PR #104, opened 2026-04-26
fix(rlm_query): Phase 1+2 - Add diagnostic logs and fallback for empty responses

### LINKED ISSUE #100: BLOCKER: rlm_query is broken — fix empty results AND make it actually be RLM (REPL + recursion)
## Two layered bugs, one tool, one fix path

\`rlm_query\` is broken on two axes that have to be fixed together. \`rlm_query\` keeps the name — we don't deprecate or rename. We make it work AND make it match what RLM is actually supposed to be.

### Axis 1 — It returns empty results in real sessions

User sees:
\`\`\`
result: [0]
---
[1]
+23 more lines
\`\`\`
Then the model concludes \"All 8 RLM queries were empty results\" and re-runs with shorter prompts. The shorter prompts also come back empty.

### Axis 2 — It's not actually RLM

Our header comment admits we trimmed the design:
\`\`\`rust
//! Native Rust RLM tool — parallel/batched LLM fan-out as a structured
//! tool call. Inspired by alexzhang13/rlm but trimmed to the primitives
//! that actually matter inside an agent loop.
\`\`\`
What we built is \`Promise.all([completion(p) for p in prompts])\` with no tools, no REPL, no recursion. Per the [RLM paper](https://arxiv.org/abs/2512.24601) and [repo](https://github.com/alexzhang13/rlm):

> RLMs offload the context as a variable in a REPL environment that the LM can interact with and launch sub-LM calls inside of.

The canonical design:
1. A REPL environment (Python \`exec\` locally; sandboxed in docker/modal/e2b).
2. The model writes code in the REPL.
3. Inside the REPL, the model can call \`rlm.completion(...)\` recursively.
4. Results return as REPL values for the parent to compose.
5. The model drives decomposition strategy.

Children should be able to spawn more rlms.

### Phase 1 — Diagnose the empties (no behavior change yet)
Add per-child structured logging in \`crates/tui/src/tools/rlm_query.rs\`: HTTP status, raw response body length, parsed \`content\` length, parsed \`reasoning_content\` length, \`finish_reason\`, elapsed_ms. Reproduce the user's symptom with \`RUST_LOG=deepseek_cli::tools=debug,deepseek_cli::client=debug deepseek\`. Decide which of these is happening:
- \`content: \"\"\` + \`reasoning_content: \"<long>\"\` → parser/extractor stripping the answer
- \`content: \"\"\` + \`reasoning_content: \"\"\` + \`finish_reason: \"length\"\` → max_tokens cap consumed by thinking
- timeout / 4xx / 5xx swallowed somewhere

### Phase 2 — Fix the empties (current single-shot mode)
- In \`crates/tui/src/tools/rlm_query.rs::extract_text\`: when no \`ContentBlock::Text\` is present, fall back to \`ContentBlock::Thinking\` with a synthetic prefix so the child's reasoning makes it back to the parent.
- If even thinking is empty, return a clear sentinel like \`\"[empty response from child idx N — finish_reason=…, raise max_tokens]\"\` instead of \`\"\"\`.
- Set \`reasoning_effort: Some(\"off\")\` (or whatever V4-flash actually accepts to disable thinking) on the child request so the budget goes to the answer, not reasoning we then strip.
- Verify: response \`content\` is non-empty for short factual prompts.

### Phase 3 — Make it actually RLM (the redesign that keeps the name)
Replace the inside of \`rlm_query.execute\` with a recursive REPL agent:
1. Accept either:
   - \`{ prompts: [...] }\` (legacy parallel-fanout shape — keep as a fast path for backward compat)
   - \`{ task: \"...\", input?: \"...\", max_depth?: 3 }\` (new RLM shape — spawns a child with a Python REPL)
2. For the new shape: spawn a child LM that gets a sandboxed Python \`exec\` REPL (start with local-only, gate like \`exec_shell\`). Bind \`input\` as a REPL variable. Expose an \`rlm.completion(prompt, model?, max_depth?)\` callable inside the REPL that recurses into the same tool.
3. Recursive depth cap (default 3, configurable per call).
4. Child returns whatever value it surfaces from the REPL.
5. Sandbox safety: REPL initially \`local\` only — Python \`exec\` in a tmp dir, restricted globals, path-scoped file access; gate behind YOLO mode like other shell tools. Add Docker/e2b modes later (matches their \`environment=\"docker\"\` / \`\"e2b\"\` config).

### Phase 4 — Trajectory logging + tests
- Mirror alexzhang13/rlm's \`RLMLogger\` — write trajectories under \`~/.deepseek/rlm-trajectories/<turn-id>/\` as JSONL so the user can inspect what the recursive tree did.
- Tests:
  - \`extract_text\` with only Thinking blocks → falls back, doesn't return \`\"\"\`.
  - \`extract_text\` with empty Thinking AND empty Text → returns sentinel, not \`\"\"\`.
  - Recursion bounded by \`max_depth\`.
  - REPL exec stays sandboxed (can't escape tmp dir, can't network without explicit allow).
  - Canonical \"chunk-summarize-then-aggregate\" pattern from the RLM paper works end-to-end.

## Files

- \`crates/tui/src/tools/rlm_query.rs\` — extract_text fallback, reasoning_effort, REPL integration
- \`crates/tui/src/tools/rlm/repl.rs\` (NEW) — sandboxed Python exec
- \`crates/tui/src/tools/rlm/logger.rs\` (NEW) — trajectory log writer
- \`crates/tui/src/tools/rlm/tests.rs\` (NEW) — unit + integration tests
- \`crates/tui/src/client/chat.rs\` — verify parser doesn't silently drop content (line 843–850)

## Acceptance

- \`rlm_query { prompts: [\"What is 2+2?\", \"Capital of France?\"] }\` returns non-empty text for both children. (Backwards-compat parallel mode.)
- \`rlm_query { task: \"Summarize this document\", input: \"<1MB text>\" }\` spawns a child that programmatically chunks/summarizes via REPL + recursive sub-calls, returns a coherent summary. (New RLM mode.)
- A child that hits a model returning all-thinking surfaces the reasoning, not \`\"\"\`.
- A child that genuinely returns nothing produces a clear sentinel error.
- One level of recursive \`rlm_query\` (children → grandchildren) works with depth cap.
- Trajectories written to \`~/.deepseek/rlm-trajectories/\` for post-hoc inspection.
- Sandbox: REPL can't read outside its tmp dir or hit network without YOLO.

## Priority

**Top blocker for v0.6.3.** Touches every recursive-fanout codepath. Land Phase 1 + 2 first (the empty-result fix), then 3 + 4 in a follow-up PR on the same branch — they don't have to ship in a single PR but do have to ship before v0.6.3 tags.

## References

- RLM repo: https://github.com/alexzhang13/rlm — particularly \`rlm/clients/openai.py\`
- RLM-minimal: https://github.com/alexzhang13/rlm-minimal — closest reference for porting
- Paper: https://arxiv.org/abs/2512.24601
- Blogpost: https://alexzhang13.github.io/blog/2025/rlm/

Hmbown:
Fixes #100 (Phase 1+2). Diagnostic logs added and extract_text thinking fallback introduced.

---

<a href="https://app.devin.ai/review/hmbown/deepseek-tui/pull/104" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tui/src/tools/rlm_query.rs
+++ b/crates/tui/src/tools/rlm_query.rs
@@ -250,7 +250,9 @@ impl ToolSpec for RlmQueryTool {
                     for block in &res.content {
                         match block {
                             ContentBlock::Text { text, .. } => text_len += text.len(),
-                            ContentBlock::Thinking { thinking, .. } => thinking_len += thinking.len(),
+                            ContentBlock::Thinking { thinking, .. } => {
+                                thinking_len += thinking.len()
+                            }
                             _ => {}
                         }
                     }
@@ -286,7 +288,9 @@ impl ToolSpec for RlmQueryTool {
             .into_iter()
             .map(|(idx, res)| {
                 let text = match res {
-                    Ok(Ok(response)) => extract_text(&response.content, idx, response.stop_reason.as_deref()),
+                    Ok(Ok(response)) => {
+                        extract_text(&response.content, idx, response.stop_reason.as_deref())
+                    }
                     Ok(Err(e)) => format!("[error: {e}]"),
                     Err(_) => format!(
                         "[error: timed out after {}s]",
--- a/crates/tui/src/tools/rlm_query.rs
+++ b/crates/tui/src/tools/rlm_query.rs
@@ -240,12 +240,31 @@ impl ToolSpec for RlmQueryTool {
                 let response = timeout(DEFAULT_CHILD_TIMEOUT, client.complete(request)).await;
                 let elapsed_ms = started.elapsed().as_millis() as u64;
                 in_flight.fetch_sub(1, Ordering::Relaxed);
+
+                let mut text_len = 0;
+                let mut thinking_len = 0;
+                let mut finish_reason = None;
+
+                if let Ok(Ok(ref res)) = response {
+                    finish_reason = res.stop_reason.clone();
+                    for block in &res.content {
+                        match block {
+                            ContentBlock::Text { text, .. } => text_len += text.len(),
+                            ContentBlock::Thinking { thinking, .. } => thinking_len += thinking.len(),
+                            _ => {}
+                        }
+                    }
+                }
+
                 debug!(
                     target: "deepseek_cli::tools",
                     tool = "rlm_query",
                     idx,
                     elapsed_ms,
                     ok = response.is_ok(),
+                    text_len,
+                    thinking_len,
+                    finish_reason = ?finish_reason,
                     "child request done"
                 );
                 (idx, response)
@@ -267,7 +286,7 @@ impl ToolSpec for RlmQueryTool {
             .into_iter()
             .map(|(idx, res)| {
                 let text = match res {
-                    Ok(Ok(response)) => extract_text(&response.content),
+                    Ok(Ok(response)) => extract_text(&response.content, idx, response.stop_reason.as_deref()),
                     Ok(Err(e)) => format!("[error: {e}]"),
                     Err(_) => format!(
                         "[error: timed out after {}s]",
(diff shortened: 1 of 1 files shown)


Hmbown:
Pushed `style(rlm_query): cargo fmt fixes for parity CI` (d7b202d). Only the two long-line rustfmt diffs from the parity job. Re-running CI.

PR #105, opened 2026-04-26
fix: @-mention path routing (#101)

### LINKED ISSUE #101: BLOCKER: @-file mention / fuzzy file loading still routes to wrong path — supersedes #87, blocks #97
## Symptom

\`@\`-mentions in the composer are persistently broken: typing \`@<some/file>\` (whether a top-level file, a nested file, or a relative path) routes resolution to the workspace root, not where the file actually lives. The resulting prompt context contains a \`<missing-file>\` block or the user sees a status line saying the file isn't at that path.

This is a consistent reproducer across sessions — not a one-time hiccup.

## Why this matters

\`@\`-mentions are the primary way users hand the agent a file. When this is broken, the agent has no good way to load files into context, and users fall back to copy-pasting file contents into the chat — slow, error-prone, and discards the whole point of the tool surface.

## Likely cause (from code inspection — not yet fully verified)

\`crates/tui/src/tui/file_mention.rs:396-403\`:

\`\`\`rust
fn resolve_mention_path(raw_path: &str, workspace: &Path) -> PathBuf {
    let path = expand_mention_home(raw_path);
    if path.is_absolute() {
        path
    } else {
        workspace.join(path)        // <-- always joined to workspace
    }
}
\`\`\`

And the completion walker (\`find_file_mention_completions\`, line 99-144) walks \`workspace\` only.

If \`app.workspace\` is set to ANYTHING other than where the user is "mentally" pointing — typically the directory \`deepseek\` was launched from, but possibly a parent or a sibling — every relative \`@\` token resolves under the wrong root.

\`app.workspace\` initialization (\`crates/tui/src/tui/app.rs\` and \`crates/tui/src/main.rs\`) needs to be audited:
- Does it pick the launch \`cwd\` by default?
- Does \`--workspace\` override it?
- Does it walk up to find a project root marker (Cargo.toml, .git)?
- Does it diverge from the agent's tool execution root?

### Phase 1 — Diagnose and capture
1. Add a temporary debug log on every \`@\`-mention completion call: log \`{ partial, app.workspace, std::env::current_dir(), raw_typed }\` so we can see the divergence in user reports.
2. Reproduce the user's exact session: launch \`deepseek\` from \`/Volumes/VIXinSSD/transformatics\`, type \`@docs/deepseek_v4.pdf\`, observe what resolves.

### Phase 2 — Fix the resolver
3. Two-pass resolution in \`resolve_mention_path\`:
   - First try \`workspace.join(path)\`.
   - If that doesn't exist, try \`std::env::current_dir().join(path)\`.
   - If THAT doesn't exist, try \`workspace\` walk for a basename match (fuzzy fallback).
4. Make the completion walker show paths relative to the user's CWD when they differ from workspace, so what they Tab-complete matches what they see in their shell.

### Phase 3 — Unify with #97 (Ctrl+P fuzzy picker)
5. Pull the resolver into \`crates/tui/src/working_set.rs\` as \`Workspace::resolve(&str) -> Result<PathBuf>\` so the picker overlay (#97) and the inline \`@\`-mention path use the SAME logic. Eliminates the divergence-bug class entirely.
6. Honor \`.gitignore\`/\`.deepseekignore\` in the walker.

### Phase 4 — Tests + repros
7. Unit test: workspace=\`/tmp/foo\`, cwd=\`/tmp/foo/sub\`, type \`@bar.txt\` where \`bar.txt\` lives in \`/tmp/foo/sub\` — must resolve to \`/tmp/foo/sub/bar.txt\`, not \`/tmp/foo/bar.txt\`.
8. Unit test: workspace=\`/tmp/foo\`, type \`@nested/deep/file.md\` where \`nested/deep/file.md\` exists at workspace root — must resolve correctly.
9. Snapshot of the rendered \`<file>\` block confirming \`<missing-file>\` is NOT produced for valid paths.

## Files

- \`crates/tui/src/tui/file_mention.rs\` — resolver + completion walker
- \`crates/tui/src/working_set.rs\` — extract shared \`Workspace::resolve\` API
- \`crates/tui/src/tui/app.rs\` — \`workspace\` field initialization
- \`crates/tui/src/main.rs\` — CLI \`--workspace\` parsing + default

## Acceptance

- Typing \`@<path>\` in the composer for a file that exists somewhere reasonable (CWD, workspace, or a nested path under either) resolves correctly and the prompt picks up the file's contents.
- The completion walker offers paths the user expects to see, displayed relative to their CWD when CWD ≠ workspace.
- \`<missing-file>\` blocks only appear for paths that genuinely don't exist anywhere accessible.
- A repro test in \`file_mention.rs\` covers the workspace-vs-cwd divergence so this can't regress silently.

## Priority

**Blocker for v0.6.3.** The \`@\`-mention surface is the primary way users hand files to the agent; while it's broken everything else feels rickety.

## Related

- Supersedes #87 (path routing) — same root cause; close that once this lands.
- Blocks #97 (Ctrl+P fuzzy picker) — must share the resolver from this fix.

Hmbown:
Fixes #101. Added Workspace::resolve two-pass resolver.

---

<a href="https://app.devin.ai/review/hmbown/deepseek-tui/pull/105" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tui/src/tui/file_mention.rs
+++ b/crates/tui/src/tui/file_mention.rs
@@ -303,24 +303,38 @@ fn local_context_from_file_mentions(input: &str, workspace: &Path) -> Option<Str
     let mut blocks = Vec::new();
     let mut seen = std::collections::HashSet::new();
     let ws = Workspace::new(workspace.to_path_buf());
-    
+
     for mention in mentions.into_iter().take(MAX_FILE_MENTIONS_PER_MESSAGE) {
-        let (path, display_path) = match ws.resolve(&mention) {
+        // `Workspace::resolve` already returns absolute paths when the root
+        // is absolute (TUI always runs from an absolute workspace), so we
+        // skip `canonicalize()` here — it's per-mention I/O on the
+        // message-send hot path. Accept the rare symlink-aliasing dedup
+        // miss as the cost of avoiding a syscall (Gemini code-review).
+        let (path, display_path, exists) = match ws.resolve(&mention) {
             Ok(p) => {
-                let d = p.canonicalize().unwrap_or_else(|_| p.clone()).display().to_string();
-                (p, d)
-            },
+                let d = p.display().to_string();
+                (p, d, true)
+            }
             Err(p) => {
                 let d = p.display().to_string();
-                blocks.push(format!("<missing-file mention=\"@{mention}\" path=\"{d}\" />"));
-                continue;
+                (p, d, false)
             }
         };
-        
+
+        // Gate every block — including <missing-file> — through the dedup
+        // set so a user typing the same non-existent file twice doesn't
+        // waste tokens on duplicate missing-file blocks (Devin code-review).
         if !seen.insert(display_path.clone()) {
             continue;
         }
-        blocks.push(render_file_mention_context(&mention, &path, &display_path));
+
+        if exists {
+            blocks.push(render_file_mention_context(&mention, &path, &display_path));
+        } else {
+            blocks.push(format!(
+                "<missing-file mention=\"@{mention}\" path=\"{display_path}\" />"
+            ));
+        }
     }
 
     if blocks.is_empty() {
@@ -401,8 +415,6 @@ fn trim_unquoted_mention(raw: &str) -> &str {
     }
     trimmed
 }
-
-
 
 fn render_file_mention_context(raw: &str, path: &Path, display_path: &str) -> String {
     if !path.exists() {
--- a/crates/tui/src/tui/file_mention.rs
+++ b/crates/tui/src/tui/file_mention.rs
@@ -23,11 +23,12 @@
 
 use std::fmt::Write;
 use std::io::Read;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 use ignore::WalkBuilder;
 
 use crate::tui::app::App;
+use crate::working_set::Workspace;
 
 /// Maximum number of `@`-mentions whose contents are inlined into one user
 /// message. Beyond this we stop appending blocks but the raw `@token` text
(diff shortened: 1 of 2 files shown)


Hmbown:
Pushed `perf+fix(file_mention): address Gemini and Devin code-review findings` (adb60ba). All 5 review comments addressed:

- **HIGH** — `fuzzy_resolve` now backed by a `OnceLock<HashMap<basename, paths>>` index built once on first miss, not per-mention. (Gemini, working_set.rs:76)
- **MEDIUM** — `cwd` is captured at `Workspace::new()` instead of called per-resolve; tests use new `Workspace::with_cwd(root, cwd)` for deterministic resolution without mutating the process cwd. (Gemini, working_set.rs:51)
- **MEDIUM** — Fuzzy index now matches both `is_file()` and `is_dir()`. (Gemini, working_set.rs:71)
- **MEDIUM** — Dropped `Path::canonicalize` from the mention loop. `Workspace::resolve` already returns absolute paths when the root is absolute (always true in TUI use). The rare symlink-aliasing dedup miss is the cost of the syscall on the message-send hot path. (Gemini, file_mention.rs:310)
- **YELLOW** — All blocks (existing AND `<missing-file>`) now flow through the `seen.insert` dedup gate, so duplicate non-existent mentions don't double the prompt token cost. (Devin, file_mention.rs:316)

Plus a new test `fuzzy_index_finds_files_and_directories` exercising the lazy index + directory-match paths, and the existing `workspace_resolve_respects_cwd_and_workspace` test no longer mutates the global cwd (race-prone with parallel tests).

Re-running CI. Tests pass locally (`cargo test -p deepseek-tui ... -- workspace_resolve fuzzy_index` → 2/2 ok).

PR #274, opened 2026-05-02
fix(sandbox): elevate Agent-mode shell sandbox to allow network access

### LINKED ISSUE #272: macOS Seatbelt sandbox blocks outbound DNS resolution in Agent mode
### Summary

Shell commands that require outbound network access (e.g. fetching YouTube content) fail with DNS resolution errors in Agent mode, but work correctly in YOLO mode on the same machine.

### Environment

- deepseek-tui 0.8.1
- macOS (Apple Silicon, aarch64)
- Sandbox: macos-seatbelt (as reported by `deepseek doctor`)
- sandbox_mode: workspace-write (default)

### Steps to Reproduce

1. Start DeepSeek TUI in Agent mode (default).
2. Ask the agent to fetch or process content from an external URL (e.g. YouTube).
3. The agent attempts a shell command involving DNS resolution (curl, wget, etc.).
4. DNS resolution fails.

### Expected Behavior

After the user approves the shell command, it should execute with full network access, same as running the command directly in the terminal.

### Actual Behavior

The command runs inside the Seatbelt sandbox, which blocks DNS resolution. Error message:

> DNS resolution failed for YouTube. This is a network issue — the sandbox can't reach YouTube's servers.

### Workaround

Switching to YOLO mode resolves the issue, as commands bypass the Seatbelt sandbox wrapper.

### Investigation

- Confirmed the issue is not a general network problem — Claude Code and OpenAI Codex execute identical network commands without issue on the same machine.
- Adding `sandbox_mode = "danger-full-access"` and configuring `[network] default = "allow"` with explicit domain allowlists in config.toml had no effect. Only switching to YOLO mode resolved it.
- This suggests the Seatbelt profile compiled into the binary restricts outbound network access at the OS level, and user-facing config options cannot override it.

### Suggestion

Consider either:
- Allowing outbound network access in the Seatbelt profile after the user has explicitly approved the command (the approval step already provides the safety gate).
- Exposing a config option that controls whether approved shell commands run inside or outside the Seatbelt sandbox.
- Documenting this limitation for macOS users.

Hmbown:
**DRAFT — security-relevant default change. Maintainer review requested.**

Closes #272 (independent confirmation from external user @francofang).

## Summary

In Agent mode (the default) every \`exec_shell\` command was wrapped in a seatbelt policy with \`(deny default)\` and no \`(allow network-outbound)\`, so DNS and outbound TCP/UDP failed at the kernel level. Symptom on macOS: \"DNS resolution failed\" for any URL the model tried to reach via \`curl\`, \`yt-dlp\`, package managers, …

External user repro (issue #272 from @francofang):

> Start DeepSeek TUI in Agent mode (default). Ask the agent to fetch external content (e.g. YouTube). DNS fails.
> Switching to YOLO mode resolves it. \`config.toml\` overrides have no effect.

## Root cause

\`engine.rs::build_tool_context\` only set \`elevated_sandbox_policy\` for **Yolo**. Agent fell through to \`ShellManager::sandbox_policy\` = \`SandboxPolicy::default()\` = \`WorkspaceWrite { network_access: false }\`. \`seatbelt::generate_policy\` only emits the network-allow block when \`policy.has_network_access()\` is true — confirmed by the existing test \`test_generate_policy_default\` (\"Default policy has no network\").

## Fix

Three-line shape change: \`if mode == Yolo\` → \`match mode\`, with Agent and Yolo both elevating to \`WorkspaceWrite { network_access: true }\`. Plan mode unchanged (read-only investigation never registers the shell tool).

Adds \`agent_and_yolo_modes_elevate_shell_sandbox_to_allow_network\` regression test so a future revert trips CI.

## Trade-off the maintainer should weigh

This removes the OS-level network block from sandboxed shell commands. The remaining boundary is the application-level \`NetworkPolicy\` (\`crates/tui/src/network_policy.rs\`) plus the per-shell approval flow. **Today \`NetworkPolicy\` only intercepts \`fetch_url\`, not shell commands** — so a model using \`curl\` to reach a denied host could route around the policy. The seatbelt block was kind-of closing that hole on macOS only; it's reopened by this PR.

Options if that residual hole matters:

1. Land this as-is for v0.8.4 — restores parity with Claude Code/Codex; #272 symptom goes away. File a follow-up to teach \`NetworkPolicy\` about shell-level egress.
2. Land this with a narrower policy (e.g. allow DNS + 80/443 only) — partial mitigation, more code.
3. Reject — keep the current behavior and document \"use Yolo for network workflows\".

Drafted (1) since it's the smallest fix that gets the user unstuck and matches what users expect from competitors.

## Test plan

- [x] \`cargo test -p deepseek-tui --bin deepseek-tui --locked sandbox\` (27/27)
- [x] \`cargo test -p deepseek-tui --bin deepseek-tui --locked core::engine\` (62/62, includes new regression)
- [x] \`cargo clippy -p deepseek-tui --all-targets --locked -- -D warnings\`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- **Agent-mode shell exec could not reach the network** (#272) — the seatbelt
+  default policy denies all outbound network including DNS, so any
+  `exec_shell` command needing the network (`curl`, `yt-dlp`, package
+  managers, …) failed in Agent mode unless the user dropped to Yolo. The
+  engine now elevates the sandbox policy to `WorkspaceWrite { network_access:
+  true, … }` for both Agent and Yolo. Plan mode is unchanged (read-only
+  investigation never registers the shell tool). The application-level
+  `NetworkPolicy` (`crates/tui/src/network_policy.rs`) remains the only
+  outbound-traffic boundary.
 - **V4 Pro discount expiry extended** (#267) — DeepSeek extended the V4 Pro 75%
   promotional discount from 2026-05-05 15:59 UTC to 2026-05-31 15:59 UTC. Without
   this update the TUI would have started showing 4× the actual billed cost on
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -1195,15 +1195,25 @@ impl Engine {
             ctx = ctx.with_network_policy(decider.clone());
         }
 
-        if mode == AppMode::Yolo {
-            ctx.with_elevated_sandbox_policy(crate::sandbox::SandboxPolicy::WorkspaceWrite {
-                writable_roots: vec![self.session.workspace.clone()],
-                network_access: true,
-                exclude_tmpdir: false,
-                exclude_slash_tmp: false,
-            })
-        } else {
-            ctx
+        match mode {
+            // Plan mode is read-only investigation; the shell tool is not
+            // registered, so leaving the sandbox policy at the seatbelt-strict
+            // default is fine.
+            AppMode::Plan => ctx,
+            // Agent and Yolo both register the shell tool. The sandbox-default
+            // policy denies all outbound network — including DNS — which
+            // breaks ordinary developer commands (cargo fetch, npm install,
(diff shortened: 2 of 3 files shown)


Hmbown:
Maintainer signed off on the trade-off (Twitter, May 1) — "going with allowing web access on agent mode is the smart move." Marked ready for review.

Issue #268, opened 2026-05-02
fetch_url SSRF: harden against DNS rebinding via connector-level IP re-validation (follow-up to PR #261)


Hmbown:
## Problem

PR #261 (@JasonOA888) added robust IP-block validation to `fetch_url`, including IPv4 private/link-local/cloud-metadata and IPv6 link-local (`fe80::/10`) and unique-local (`fc00::/7`). It validates at hostname-resolution time.

There's a residual TOCTOU window: between our `lookup_host` check and reqwest's own DNS resolution at connect time, a hostile DNS server with a sub-second TTL can return a public IP at lookup, then a private IP at connect. The IP filter is bypassed.

The author and I agreed (PR #261 thread) to land the current PR as-is for v0.8.4 — the threat is narrow (this tool runs on developer machines, not exposed servers, and rebinding requires a hostile DNS with deliberate timing). This issue tracks the connector-level hardening for after that lands.

## Approaches

**Option A — Custom reqwest connector**:
Implement a custom `Connect` service that wraps reqwest's default and re-validates the resolved IP at the TCP-connect layer. Reject before the syscall completes. This is the canonical defense, eliminates the TOCTOU, but is non-trivial: needs to handle IPv4/IPv6 dual-stack, connection retries, proxy interaction.

**Option B — Pin to single IP**:
Resolve once via `lookup_host`, validate, then build the `reqwest::Url` with the IP literal substituted (preserving the Host header for SNI/virtual-host routing). Eliminates the TOCTOU because reqwest never re-resolves. Simpler than (A) but breaks for hosts that need DNS round-robin or geo-DNS.

**Option C — Document the residual risk and defer indefinitely**:
The threat model is narrow. If we judge the cost (custom connector, ongoing maintenance) higher than the risk (developer machine with explicit user-invoked tool), this is a defensible decision.

I'd lean toward **(A)** as the right long-term answer with (C) as the honest short-term answer. Filing for v0.8.4 milestone but no urgency — the v0.8.4 release is fine to ship without it; PR #261 already covers the primary attack surface (static hostnames pointing at private IPs).

## Acceptance criteria

If we pick A:
- [ ] Custom `Connect` impl rejecting restricted IPs at TCP-connect time
- [ ] Test using a fake DNS resolver returning public-then-private IPs over two calls; assert connect is refused on the second
- [ ] No regression in the existing `fetch_url` tests
- [ ] Documented in `crates/tui/src/tools/fetch_url.rs` module docs

[...]

PR #277, opened 2026-05-02
fix(security): add bearer token auth to app-server, restrict CORS to localhost

JasonOA888:
## Summary

The HTTP app-server exposed sensitive endpoints (`/thread`, `/prompt`, `/tool`, `/mcp/startup`, `/app`, `/jobs`) with **zero authentication** and `CorsLayer::permissive()`, allowing any website to make cross-origin requests and execute tools remotely.

### Attack scenario (before this PR)
1. DeepSeek-TUI running with app-server on `0.0.0.0:8080`
2. User visits `http://evil.com` with JavaScript that POSTs to `http://victim-ip:8080/tool`
3. Attacker executes arbitrary tools, reads files, exfiltrates API keys

### Changes
- **Bearer token auth**: Generate a random 64-char hex token at startup, print to stderr
- **All endpoints except `/healthz`** require `Authorization: Bearer <token>` header
- **Constant-time comparison** to prevent timing attacks
- **CORS restricted** to `localhost:{3000,8080}` and `127.0.0.1:{3000,8080}` only
- Replaced `CorsLayer::permissive()` with explicit origin allowlist

### Usage
The TUI and HTTP clients must now pass the bearer token. Token is printed to stderr at startup:

Companion to #261 (SSRF protection for fetch_url). Both address the network attack surface identified in security review.


JasonOA888:
@gemini-code-assist Thanks for the thorough review. All items addressed:

1. **CSPRNG token generation**: Replaced hand-rolled PRNG with getrandom::fill() (32 bytes of cryptographic randomness). getrandom was already in the dependency tree.

2. **Middleware ordering**: CORS layer is now outermost (.layer(cors) last), so 401 responses include CORS headers.

3. **Timing leak**: Removed the length short-circuit. Now compares all bytes with zero-padding to avoid length-based timing.

4. **401 vs 403**: All auth failures now return 401 Unauthorized.

5. **Doc comment typo**: Fixed.

Note on the SSRF DNS rebinding comment: PR #277 is the app-server auth PR. The SSRF DNS rebinding fix is in PR #261 (already addressed with ClientBuilder::resolve() pinning + follow-up issue #268).

Hmbown (commented):
Read through the auth+CORS portion carefully (skipped the SSRF code since it overlaps with #261). Posting comments rather than approval — final call belongs to @Hmbown. No blockers from my read; the hard parts look right. Three UX gaps I'd flag.

### Verified correct

1. **Constant-time comparison.** Traced through after the third commit. `provided.iter().chain(repeat(&0)).zip(expected.iter().chain(repeat(&0))).take(max_len).fold(0, |a, (x, y)| a | (x ^ y))` virtualises both inputs to `max_len = max(provided_len, expected_len)`, XORs each byte pair, and OR-folds into a single accumulator — every iteration runs in constant time per byte and the time scales with `max_len`. Since `expected_len` is a fixed 64-char hex constant, the only attacker-observable timing channel is `provided_len`, which the attacker already chose. No leak of the secret bytes.
2. **CSPRNG token generation.** `getrandom::fill` is the right call for a security-critical token; the previous `SystemTime`-based approach was correctly flagged.
3. **Layer ordering.** `.layer(auth_middleware).layer(cors)` makes `cors` outermost (axum/tower applies layers in reverse), so 401s emerge with CORS headers attached and the preflight `OPTIONS` is short-circuited by `CorsLayer` before the auth middleware sees it. ✓
4. **Status codes.** `401 Unauthorized` for both missing-header and invalid-token paths — matches RFC 7235.

### Concerns (not blockers, your call)

1. **No tests for the auth surface.** The PR adds 88 lines of auth-critical code to `app-server/lib.rs` and the only coverage is the SSRF tests in a different crate. I'd want at least:
   - `require_bearer` unit tests for: empty header → 401, wrong-length token → 401, correct token → Ok, "Bearer " prefix optional / case-insensitive (test what's expected), missing `Authorization` → 401.
   - One integration test that spins the router with a known token, hits `/healthz` without auth (200), then `/jobs` with and without the header (200 / 401).
   
   Without these, a future refactor could easily downgrade the auth without anyone noticing.

2. **No env-var / file path for the token.** Right now scripted clients have to capture stderr and parse `[app-server] Bearer token: <hex>`. Two friendlier surfaces, in increasing order of complexity:
   - Honour `DEEPSEEK_APP_SERVER_TOKEN` env var when set, skip generation. Trivial change, big UX win for `tmux`/`systemd`/`docker run -e` users.
   - Persist the generated token to `~/.deepseek/app-server-token` (mode 0600) so reads survive restarts and dev clients keep working without grabbing a new token every cycle.
   
   The maintainer's working brief explicitly flagged this as a desired surface.

3. **CORS allowlist is hardcoded to ports 3000 and 8080.** Vite (5173), Webpack default (8081), Storybook (6006), or any custom dev server is locked out of browser-side calls. Even if the typical user sticks to 3000/8080 today, "I run my prototype UI on a different port and the API broke" is a likely first-touch issue. Either widen to `localhost:*` / `127.0.0.1:*` (still blocks public origins) or take an `--allow-origin` flag for explicit configuration.

### Minor

- `getrandom = "0.3"` could be `getrandom.workspace = true` if there's a workspace pin (the lockfile already has 0.3.4 transitively); not strictly necessary.
- The middleware is an inline closure — extracting it to `async fn require_bearer_middleware(...)` and using `from_fn(require_bearer_middleware)` would make it grep-able alongside `require_bearer`. Cosmetic.

Happy to take another pass after the auth tests land. Thanks for hardening this — the previous \`CorsLayer::permissive()\` + zero-auth combo was a real attack surface for anyone running the app-server bound to a non-loopback interface.

JasonOA888:
Thanks for the thorough review. All three concerns addressed:

1. **Auth tests**: 7 unit tests added covering require_bearer (empty header, wrong token, correct token, no-prefix passthrough, different-length constant-time) and token generation (env var override, CSPRNG length/format).

2. **DEEPSEEK_APP_SERVER_TOKEN env var**: When set, the server uses the provided token instead of generating one. Trivial change, scripted clients can now pass a known token.

3. **Dynamic CORS ports**: Replaced hardcoded 3000/8080 with AllowOrigin::predicate that matches http://localhost:* and http://127.0.0.1:* on any port. Still blocks all public origins but works with Vite (5173), Storybook (6006), etc.

Minor items not addressed: getrandom workspace pin (transient dep, no workspace entry) and middleware closure extraction (needs bearer capture, cosmetic only). Happy to address either if you prefer.

Hmbown:
This branch has merge conflicts with  (DIRTY state). Please rebase on latest  and resolve — the v0.8.4 release touched some of the same files (version bumps, path-deps). Once rebased and CI is green we can review for merge.

Hmbown:
Closing — this PR has been open since May 1 and now CONFLICTS with main after the v0.8.5 backlog landed. The underlying concern (bearer-token auth + locked-down CORS for the runtime API server) is still real. If you'd like to revive this, please rebase against feat/v0.8.5 and reopen — happy to review the rebased diff. The diff is large (44 files, +2707/-261) so a clean rebase will make review tractable.

Issue #560, opened 2026-05-04
npm postinstall: glibc preflight check + clear fallback message


Hmbown:
## Summary

Companion to #555 / #556. Even after we lower the build target with cargo-zigbuild, the npm wrapper currently exits successfully on install and only fails later at first launch with a runtime linker error like \`libc.so.6: version 'GLIBC_2.39' not found\`. The user has no clue what just happened.

## Proposal

In the npm postinstall script (\`npm/deepseek-tui/scripts/postinstall.js\` or equivalent), after downloading the binary:

1. Detect the host glibc version (Linux only): \`ldd --version\` parse, or \`getconf GNU_LIBC_VERSION\`.
2. Detect the binary's required glibc: \`readelf -V <binary> | grep GLIBC_\` and pick the highest version.
3. If host < required, print a clear error and direct the user to the source-build path:

   > \`deepseek-tui requires glibc >= 2.28 (your system has 2.17). Install from source instead: \`cargo install deepseek-tui-cli\`\`

4. Exit non-zero so npm flags the install as failed, instead of the current "looks fine, fails later" UX.

## Repro context

Reported by @20bytes on issue #555 (Ubuntu 22.04, glibc 2.35) — install succeeds, both \`./node_modules/.bin/deepseek --version\` and \`./node_modules/.bin/deepseek-tui --version\` fail at runtime with the linker error.

## Notes

- macOS / Windows path: skip the check, no glibc concept.
- After #556 lands and the floor drops to glibc 2.28, this preflight stays useful for the long tail of CentOS 7 / RHEL 7 (glibc 2.17) users.

Vishnu1837:
Hi! I'd like to take this one. Plan: add a Linux-only preflight to the npm postinstall that (1) reads required GLIBC symbols from the downloaded binary via `readelf -V`, (2) reads the host glibc version via `getconf GNU_LIBC_VERSION` (with `ldd --version` fallback), (3) compares them, and (4) on mismatch prints a clear message pointing at the build-from-source path and exits non-zero so npm marks the install failed. macOS/Windows skip the check. Will open a PR shortly.

PR #568, opened 2026-05-04
Add glibc preflight to npm postinstall

dangzitou:
## Summary

npm installs currently appear to succeed even when the downloaded Linux binary needs a newer glibc than the host provides, so users only see a confusing runtime linker failure on first launch. This change adds a Linux-only postinstall preflight that compares the host glibc version with the downloaded binary's required glibc, prints a clear source-install fallback message when incompatible, and exits non-zero so npm fails early with actionable guidance.

## Changes

- Add Linux-only glibc detection helpers in npm/deepseek-tui/scripts/postinstall.js using getconf GNU_LIBC_VERSION first and ldd --version as a fallback.
- Inspect the downloaded executable with readelf -V, extract GLIBC_x.y requirements, and compute the highest required glibc version.
- Compare host and required glibc versions numerically during postinstall after the binary download path is resolved.
- Print a clear incompatibility error directing users to install from source with cargo install deepseek-tui-cli when host glibc is too old.
- Exit postinstall with a non-zero status on detected glibc incompatibility so npm reports install failure immediately instead of deferring to runtime.
- Preserve current behavior on macOS and Windows by skipping the glibc preflight outside Linux.
- Only touch npm/deepseek-tui/package.json if needed to confirm the existing postinstall entry still points to scripts/postinstall.js.

## Validation

- Provided test command `cargo test` is still pending and was not executed in the supplied validation context.
- npm wrapper validation is still pending: run install flow on a compatible Linux host and confirm postinstall succeeds.
- Incompatible glibc validation is still pending: verify host lower than binary requirement prints the fallback message and exits non-zero.
- Cross-platform validation is still pending: confirm macOS and Windows skip the glibc preflight and preserve existing install behavior.

## Risks

- readelf may be unavailable on minimal Linux systems, so the preflight must avoid breaking otherwise-valid installs when tooling is missing.
- glibc version parsing from distro-specific ldd output can vary, which could cause false negatives or false positives if fallback parsing is too permissive.
- The postinstall script must use the actual resolved binary path from the existing download flow; hardcoded assumptions could miss the installed executable.
- Failing install earlier changes behavior for affected Linux users and could surface environment-specific edge cases that were previously only seen at runtime.

### DIFF
--- a/npm/deepseek-tui/package.json
+++ b/npm/deepseek-tui/package.json
@@ -28,7 +28,7 @@
   },
   "scripts": {
     "release:check": "node scripts/verify-release-assets.js",
-    "postinstall": "node scripts/install.js",
+    "postinstall": "node scripts/postinstall.js",
     "prepublishOnly": "node scripts/verify-release-assets.js",
     "prepack": "node scripts/install.js"
   },
--- a/npm/deepseek-tui/scripts/install.js
+++ b/npm/deepseek-tui/scripts/install.js
@@ -203,6 +203,7 @@ async function getBinaryPath(name) {
 }
 
 module.exports = {
+  binaryPaths,
   getBinaryPath,
   run,
 };


Hmbown:
Closing as duplicate of #565 (which was merged in https://github.com/Hmbown/DeepSeek-TUI/commit/3179b552, addressing the same #560).

#565 had a more thorough implementation (no `readelf` runtime dependency — pure Node bytewise scan; explicit unit tests for the version helpers; documented `DEEPSEEK_TUI_SKIP_GLIBC_CHECK` escape hatch). Thanks for the work — #565 covered the same ground.

PR #579, opened 2026-05-04
fix(markdown): render tables, bold/italic, and horizontal rules

WyxBUPT-22:
---
  ## Background

  Found this issue while running DeepSeek-TUI on **Windows 11 (PowerShell)**.
  The TUI transcript was rendering raw markdown syntax instead of formatted output    
  for three common patterns that the model frequently uses in responses.

  ## Problems Fixed

  **1. Tables rendered as raw text**

  Markdown table rows like `| col | value |` were displayed verbatim, and separator   
  rows `|---|---|` were also visible as plain text.

  **2. Bold/italic markers not stripped**

  `**bold text**` and `*italic*` markers leaked into the output unchanged. This was   
  especially noticeable in list items and table cells where the model uses bold for   
  emphasis.

  **3. Horizontal rules not rendered**

  `---` / `***` / `___` were displayed as plain text instead of a visual separator    
  line.

  ## Root Causes

  - `parse()` had no recognition for table rows or horizontal rules — both fell       
  through
    to `Block::Paragraph` and were output verbatim.
  - The inline formatter split text word-by-word, so `**bold spanning multiple        
  words**`
    could never match (opening `**` and closing `**` landed in different words).      
  - No guard against unclosed markers (e.g. a lone `*`) caused an infinite loop → OOM 
    crash on long responses.

  ## Changes (`crates/tui/src/tui/markdown_render.rs`)

  - Add `Block::TableRow(Vec<String>)` and `Block::HorizontalRule` variants
  - `parse()`: detect `| table |` rows, drop separator rows; detect `---` rules       
  - Rewrite inline span parser (`parse_inline_spans`) to scan the full line instead of
   
    word-by-word, correctly handling markers that span multiple words
  - Add infinite-loop guard for unclosed markers
  - `render_table_row()`: equal-width columns with `│` separators, inline formatting  
    applied inside cells

  ## Tests

  - All 12 existing `markdown_render` tests pass
  - 3 new tests added:
    - `table_separator_row_is_dropped`
    - `bold_markers_stripped_in_render`
    - `table_renders_with_pipe_separator`

before：
3f3326dbae8203e9190e2b4386a9cb74

image

after:
image

  ---

### DIFF
--- a/crates/tui/src/tui/markdown_render.rs
+++ b/crates/tui/src/tui/markdown_render.rs
@@ -29,7 +29,7 @@ use std::cell::Cell;
 
 use ratatui::style::{Modifier, Style};
 use ratatui::text::{Line, Span};
-use unicode_width::UnicodeWidthStr;
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 
 use crate::palette;
 use crate::tui::osc8;
@@ -66,10 +66,14 @@ pub enum Block {
     Heading { level: usize, text: String },
     /// A horizontal rule emitted under a level-1 heading.
     HeadingRule,
+    /// A standalone `---` / `***` / `___` horizontal rule.
+    HorizontalRule,
     /// A bullet (`-`/`*`) or ordered (`1.`) list item with its prefix and body.
     ListItem { bullet: String, text: String },
     /// A line inside a fenced code block. Fences themselves are dropped.
     Code { line: String },
+    /// A table row: cells split on `|`. Separator rows (`|---|`) are dropped.
+    TableRow(Vec<String>),
     /// A non-empty paragraph line that may contain inline links.
     Paragraph { text: String },
     /// An empty source line, preserved so paragraph spacing survives.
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @WyxBUPT-22. The CI matrix showed the functional tests passing; the only blocker was `cargo fmt --all -- --check` on `markdown_render.rs`.

Because this is a real v0.8.10 rendering bug and the fix is valuable, I ported your commits into PR #572 with your authorship preserved, then applied `cargo fmt` and local clippy fixes there. Local checks on the ported patch:

```bash
cargo fmt --all -- --check
cargo clippy -p deepseek-tui --all-targets --all-features --locked -- -D warnings
cargo test -p deepseek-tui markdown_render --locked
```

Once #572 lands, I’ll close this PR as credited/ported rather than leaving your work hanging on a formatting-only CI failure. Thank you for the screenshots and tests.

Hmbown:
Closing as ported into PR #572, which is now merged as 36acd1fa. Your markdown renderer commits were preserved in the v0.8.10 branch history with authorship credited, then formatted/clippy-fixed there so CI could pass. Thanks again for the bug report, screenshots, and tests.

PR #597, opened 2026-05-04
fix(prompts): mirror user language in reasoning and reply (#588)

### LINKED ISSUE #588: Prompt: Chinese reasoning — assistant thinks in English even when user writes in Chinese
## User report (Simplified Chinese, via Telegram community)

> deepseek 的思考过程仍然是英文，有什么办法换成中文吗

**Translation:** "DeepSeek's thinking process is still in English. Is there any way to change it to Chinese?"

The user is interacting in Chinese (asking questions in Chinese), but the model's `reasoning_content` (the visible chain-of-thought in V4 thinking mode) and often the final reply come back in English. They want the assistant to reason and respond in their language.

## Root cause

Every prompt file under `crates/tui/src/prompts/` is in English. The system prompt the model sees on every turn establishes English as the response language by example, and DeepSeek V4's `reasoning_content` channel inherits that bias even when user-visible content switches mid-conversation.

Files (all English-only today):

```
crates/tui/src/prompts/
├── agent.txt
├── base.md
├── base.txt
├── compact.md
├── cycle_handoff.md
├── normal.txt
├── plan.txt
├── yolo.txt
├── subagent_output_format.md
├── approvals/…
├── modes/…
└── personalities/…
```

`crates/tui/src/prompts.rs` assembles these into the final system prompt with no locale routing.

## Proposed fix for v0.8.11

Two complementary pieces; ship both.

### 1. Language-mirror instruction in the system prompt (quick, ships everything)

Add a small section to `base.md` / `base.txt` (the prompt prefix that always loads) that explicitly tells the model to mirror the user's primary language in **both** `reasoning_content` and the final reply:

> ## Language
>
> Detect the language the user writes in and respond in that same language — including your internal reasoning. If the user writes in Simplified Chinese (简体中文), your reasoning_content and final reply must both be in Simplified Chinese. If they switch languages mid-conversation, switch with them. The default if no clear signal is present is English.

This is a one-block addition with no config surface. It works on every turn for every user without any setup.

### 2. Optional config knob (`language` / `locale`) for explicit override

For users who want to lock the assistant to a specific language regardless of input (e.g. coaching English speakers, or users on mixed-language teams), add:

```toml
# ~/.deepseek/config.toml
[ui]
language = \"zh-CN\"   # or \"en-US\", \"ja-JP\", etc.
```

When set, append a stronger instruction to the system prompt (e.g. \"Always reason and respond in Simplified Chinese, regardless of the language of the user's input.\").

Surface in `/config` so users can toggle without editing TOML.

## Acceptance criteria for v0.8.11

- [ ] `base.md` / `base.txt` ship a `## Language` section that instructs language-mirroring for both reasoning_content and final reply.
- [ ] When a user writes in Simplified Chinese, the assistant's `reasoning_content` and reply are in Simplified Chinese (manual eval against the V4 model).
- [ ] Optional `[ui] language` config knob is documented in `docs/CONFIGURATION.md` and surfaced in `/config`.
- [ ] Existing English-language behavior is preserved as the default.
- [ ] Snapshot/golden tests in `crates/tui-core` aren't disturbed by the new prompt block (or the block is added in a position the tests already tolerate).

## References

- `crates/tui/src/prompts.rs` — system prompt assembly.
- `crates/tui/src/prompts/base.md`, `base.txt` — always-loaded prompt prefix.
- `crates/tui/src/prompts/agent.txt`, `plan.txt`, `yolo.txt` — mode-specific prompts (English-only today).
- DeepSeek V4 thinking-mode docs: `reasoning_content` is a separate channel from `content`; both inherit the same language bias from the system prompt.

## Reporter

User "leo smith" via the project's Telegram community on 2026-05-04.

merchloubna70-dot:
## Problem

Closes #588.

All system prompts (`base.md`, `base.txt`) were English-only. DeepSeek V4 defaulted to English reasoning (`reasoning_content`) and replies even when users wrote in Chinese or other languages — because the system prompt established English as the de-facto language by example.

## Fix

Add a **Language Mirror** section near the top of `base.md` and `base.txt`:

```
## Language Mirror

Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
```

**Files changed:** `crates/tui/src/prompts/base.md`, `crates/tui/src/prompts/base.txt`  
**Files NOT changed:** `normal.txt` (mode config only), all other prompt files (they either inherit `base` through chaining or carry no user-facing instructions).

## Meta

This fix was implemented using `deepseek exec --model deepseek-v4-flash --auto` — DeepSeek V4 Flash diagnosed and patched its own system prompt. 🐋

### DIFF
--- a/crates/tui/src/prompts/base.md
+++ b/crates/tui/src/prompts/base.md
@@ -1,5 +1,9 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
+## Language Mirror
+
+Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
+
 ## Preamble Rhythm
 
 When starting work on a user request, open with a short, momentum-building line that names the action you're taking. Keep it reserved — state what you're doing, not how you feel about it.
--- a/crates/tui/src/prompts/base.txt
+++ b/crates/tui/src/prompts/base.txt
@@ -1,5 +1,9 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
+## Language Mirror
+
+Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
+
 ## Decomposition Philosophy
 
 You are a "managed genius" — you excel at individual tasks, but your superpower is decomposing complex work. **Always decompose before you act.** A few minutes spent planning saves many minutes of thrashing.
--- a/crates/tui/src/prompts/base.txt
+++ b/crates/tui/src/prompts/base.txt
@@ -1,9 +1,5 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
-## Language Mirror
-
-Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
-
 ## Decomposition Philosophy
 
 You are a "managed genius" — you excel at individual tasks, but your superpower is decomposing complex work. **Always decompose before you act.** A few minutes spent planning saves many minutes of thrashing.


merchloubna70-dot:
Thanks for the review @gemini-code-assist! Confirmed: `base.txt` is not referenced in `prompts.rs` — only `base.md` is loaded via `include_str!`. Removed the redundant `base.txt` change in the latest commit.

PR #601, opened 2026-05-04
fix(theme): use terminal-default background instead of hardcoded Black (#586)

### LINKED ISSUE #586: The background and menu color seemingly wierd
## Steps to reproduce

1.  when i start the agent system , the background color is green and blue rather than dark theme, and when i use /config to change the language, the menu and background color turn to a mess.

## Expected behavior
Dark theme as the readme of the project

## Actual behavior

Image

## Environment

- OS: Mac OS 14.8.5
- DeepSeek CLI version: 0.8.9
- Model: DeepSeek V4 Flash
- Shell: zsh

merchloubna70-dot:
Closes #586.

**Root cause:** Background and border colors were hardcoded to `Color::Black`, breaking display on light terminals and custom themes.

**Fix:** Replace `Color::Black` with `Color::Reset` for background elements so the TUI inherits the user's terminal background color.

### DIFF
--- a/crates/tui/src/tui/context_menu.rs
+++ b/crates/tui/src/tui/context_menu.rs
@@ -191,8 +191,7 @@ impl ModalView for ContextMenuView {
                         .add_modifier(Modifier::BOLD)
                 } else {
                     Style::default()
-                        .fg(palette::TEXT_SOFT)
-                        .bg(palette::SURFACE_ELEVATED)
+                        .fg(palette::TEXT_MUTED)
                 };
                 Line::from(Span::styled(text, style))
             })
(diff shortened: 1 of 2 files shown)


Hmbown:

The main `render()` background is already `Block::default().style(Style::default().bg(Color::Reset))` on `origin/main` at `crates/tui/src/tui/ui.rs:4892`, which is the same fix as your diff and resolves #586 (terminal-default background instead of `Color::Black`).

The additional `crates/tui/src/tui/context_menu.rs` change in this PR — dropping `palette::SURFACE_ELEVATED` and switching `TEXT_SOFT` → `TEXT_MUTED` — was a separate styling decision. The maintainer kept the elevated background for the context menu intentionally, so we'll leave that one out for now.

Closing #601 as integrated/superseded for the original #586 issue.

PR #603, opened 2026-05-04
fix(auth): write new API key to OS keyring on onboarding (#593)

### LINKED ISSUE #593: Auth: stale OS-keyring entry shadows fresh API key from TUI onboarding (no-response symptom)
## User report

Surfaced via Telegram by the same community user who filed #586 (macOS Sequoia 15.7.3, `deepseek 0.8.9`, installed via `cargo install`):

> Love your TUI! But have some issues here. (1) the colors are off [#586], and (2) it seems even with the api input, it does not connect to the actual server (no response).

The colors are #586. This issue is for the second symptom: **the user typed their API key into the TUI onboarding, but the TUI behaves as if no key is set** — turns hang or fail to connect.

## Hypothesised root cause

`crates/secrets/src/lib.rs` documents the resolution order as:

> Hard rule: **keyring → env → config-file**. Never swap.

Meanwhile the TUI onboarding's `submit_api_key` (`crates/tui/src/tui/app.rs:1244`) calls `save_api_key` (`crates/tui/src/config.rs:2185`), which writes **only to `~/.deepseek/config.toml`** — it does not touch the OS keyring:

```rust
pub fn save_api_key(api_key: &str) -> Result<SavedCredential> {
    // ...
    let path = save_api_key_to_config_file(trimmed)?;
    Ok(SavedCredential::ConfigFile(path))
}
```

So a user whose flow looks like this will silently miss the new key:

1. At any point in the past, the user ran `deepseek auth set` → an API key landed in the macOS Keychain under our service id.
2. The key later becomes invalid (rotated upstream, account changed, key is from a previous DeepSeek-TUI installation, etc.).
3. The user runs the TUI, sees the onboarding prompt, types a fresh key.
4. `submit_api_key` writes it to `~/.deepseek/config.toml`.
5. At request time, the keyring lookup hits the **stale** entry and short-circuits before the config file is consulted.
6. The API call uses the old/invalid key → 401 → user-visible: "no response."

Same trap if a user installed an older DeepSeek-TUI build and then upgraded — the old keyring entry survives the upgrade.

## Repro plan

- [ ] Confirm `deepseek_secrets::DefaultKeyringStore::get()` resolves before `~/.deepseek/config.toml` for the engine's HTTP client. Trace the actual call path from `submit_api_key` → engine respawn → request build.
- [ ] Manually seed the macOS Keychain with `service = "deepseek-tui-deepseek"` (or whatever our service id is — check `crates/secrets/src/lib.rs`) and a bad API key, then run `deepseek` and onboard with a different (good) key. Observe whether the request uses the keyring's bad key.

## Proposed fixes (in increasing scope)

1. **Make `submit_api_key` keyring-aware.** When the user re-enters a key during onboarding (or via `/config api_key`), also write it to the keyring slot — overwriting any stale entry. Documented behavior already says keyring is the source of truth, so `save_api_key` writing only to the config file is the bug. (small)
2. **Surface conflicts.** If the keyring already holds a different key when onboarding starts, detect this and either ask the user "Replace stored key?" or just announce that we're overwriting it. (medium)
3. **`deepseek doctor`** should already check this; if it doesn't, add a check that flags "keyring has key X but config.toml has key Y; resolution will use X." Make the discrepancy visible when the user reports issues. (small/medium)

## v0.8.11 acceptance criteria

- [ ] When a user saves an API key via TUI onboarding, the value lands in **both** the keyring (when a keyring is available) and `~/.deepseek/config.toml`. New keys take effect on the next turn without requiring a manual `keychain` purge.
- [ ] When a user runs `deepseek doctor`, a stale-keyring-vs-config mismatch is reported as a warning with copy-paste remediation steps.
- [ ] Existing platforms without a keyring (some Linux servers, headless CI) still work via the config-file path.
- [ ] No regressions for users who only ever used `deepseek auth set` and never touched the TUI onboarding.

## References

- `crates/secrets/src/lib.rs` — keyring backend + the documented resolution order.
- `crates/tui/src/config.rs:2180-2193` — `save_api_key` writes only to config file.
- `crates/tui/src/tui/app.rs:1244-1259` — `submit_api_key` (TUI onboarding entry).
- `crates/tui/src/tui/ui.rs:1631-1685` — onboarding state machine + engine respawn.
- `crates/cli/src/lib.rs:578` — `no_keyring_secrets()` (the bypass used in CLI flows).

## Reporter

Same community user as #586, via Telegram on 2026-05-04. Worth a follow-up comment on #586 acknowledging the keyring symptom as a separate fix once we ship.

merchloubna70-dot:
Closes #593.

**Root cause:** `save_api_key()` wrote only to `~/.deepseek/config.toml` (lowest priority). A stale OS keyring entry shadowed the new key because resolution order is keyring → env → config-file.

**Fix:** After writing to config file, also write the key to the OS keyring via `Secrets::auto_detect().set()`. Best-effort — logs warning on failure, does not bail.

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -2164,6 +2164,8 @@ pub fn ensure_parent_dir(path: &Path) -> Result<()> {
 pub enum SavedCredential {
     /// Stored in the deepseek config file at the given path.
     ConfigFile(PathBuf),
+    /// Stored in the OS keyring (also written to config file as fallback).
+    Keyring(PathBuf),
 }
 
 impl SavedCredential {
@@ -2172,16 +2174,20 @@ impl SavedCredential {
     #[must_use]
     pub fn describe(&self) -> String {
         match self {
-            Self::ConfigFile(path) => path.display().to_string(),
+            Self::ConfigFile(path) | Self::Keyring(path) => path.display().to_string(),
         }
     }
 }
 
-/// Save the active provider's API key to `~/.deepseek/config.toml`.
+/// Save the active provider's API key to `~/.deepseek/config.toml` and
+/// the OS keyring.
 ///
 /// v0.8.8 intentionally uses the shared config file as the default
 /// setup path. It works in every folder, in npm installs, in IDE
-/// terminals, and without platform credential prompts.
+/// terminals, and without platform credential prompts. The keyring
+/// write is a best-effort side effect so a stale keyring entry doesn't
+/// shadow the new config-file key (resolution order: keyring → env →
+/// config-file).
 pub fn save_api_key(api_key: &str) -> Result<SavedCredential> {
     let trimmed = api_key.trim();
     if trimmed.is_empty() {
@@ -2189,7 +2195,17 @@ pub fn save_api_key(api_key: &str) -> Result<SavedCredential> {
     }
 
     let path = save_api_key_to_config_file(trimmed)?;
-    Ok(SavedCredential::ConfigFile(path))
+
+    // Best-effort: also write to the OS keyring so a stale entry doesn't
+    // shadow the new key. Log on failure but don't bail.
+    let secrets = deepseek_secrets::Secrets::auto_detect();
+    match secrets.set("deepseek", trimmed) {
+        Ok(()) => Ok(SavedCredential::Keyring(path)),
+        Err(e) => {
+            tracing::warn!("Failed to save API key to OS keyring: {e}");
+            Ok(SavedCredential::ConfigFile(path))
+        }
+    }
 }
 
 /// Write the `api_key` slot directly to `config.toml`.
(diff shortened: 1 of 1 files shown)


Hmbown:

The same root cause — a stale OS keyring entry shadowing a freshly entered config-file key — is now handled by #909 (commit `ccb4662c2` — \"fix(auth): recover keyring credentials into config\"). On `main`, `save_api_key` returns the new `SavedCredential::KeyringAndConfigFile { backend, path }` variant when the keyring write succeeds, with a `ConfigFile(path)` fallback if the OS keyring rejects the write. Verified on `origin/main` at `crates/tui/src/config.rs:2490`–`:2530`.

The end-state matches what this PR was after (config + keyring stay in sync), and the recovery path also covers the case where users already had a stale keyring entry from a previous install.

Closing #603 as integrated/superseded.

PR #604, opened 2026-05-04
docs: add local VLLM deployment guide (#574)

### LINKED ISSUE #574: 使用VLLM本地部署的deepseek的API要怎么使用这个工具？
config.toml的修改不起作用

merchloubna70-dot:
Closes #574.

Adds a **Using Local VLLM Deployment** section to both `README.md` and `README.zh-CN.md`, positioned after the existing Other API providers block.

Covers:
- Env var method (`DEEPSEEK_BASE_URL`, `DEEPSEEK_API_KEY`, `DEEPSEEK_MODEL`)
- Config file method via `~/.deepseek/config.toml` using the `sglang` provider (OpenAI-compatible, closest match to VLLM)
- Key notes: any non-empty `api_key` works, model name must match deployed name, set `reasoning_effort = "off"` for non-thinking models

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,32 @@ deepseek --provider fireworks --model deepseek-v4-pro
 SGLANG_BASE_URL="http://localhost:30000/v1" deepseek --provider sglang --model deepseek-v4-flash
 ```
 
+### Using a local VLLM deployment
+
+[VLLM](https://github.com/vllm-project/vllm) serves an OpenAI-compatible API, so you can point DeepSeek TUI at any locally deployed VLLM instance:
+
+```bash
+# Via environment variables (no provider prefix needed)
+DEEPSEEK_BASE_URL="http://localhost:8000/v1" \
+DEEPSEEK_API_KEY="not-used" \
+DEEPSEEK_MODEL="my-model" \
+deepseek
+```
+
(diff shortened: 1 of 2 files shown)


Hmbown:

`main` now ships a first-class `vllm` provider, so users no longer need the `provider = "sglang"` workaround documented in this PR. The README already lists the canonical one-liner directly:

```bash
# Self-hosted vLLM
VLLM_BASE_URL="http://localhost:8000/v1" deepseek --provider vllm --model deepseek-v4-flash
```

(See `README.md` line 198 on `origin/main`, plus `DEEPSEEK_PROVIDER` listed alongside `nvidia-nim`, `fireworks`, `sglang`, `vllm` at `README.md:308`.)

Pointing users at the `sglang` provider for VLLM today would be actively misleading: it would route them through SGLang's defaults (port 30000, SGLANG_BASE_URL) instead of the dedicated vLLM provider with its own `VLLM_BASE_URL` env var and proper defaults.

Closing #604 as superseded by the dedicated `vllm` provider docs already on `main`. If there are deeper VLLM topics worth a docs page (model-name pinning, quantization flags, vLLM-specific tuning), happy to take a separate PR on that — just open it against the current README/`docs/CONFIGURATION.md` so it builds on the real provider.

PR #605, opened 2026-05-04
feat(ux): add /verbose toggle for thinking trace display (#598)

### LINKED ISSUE #598: thinking-trace UX: natural-feel default vs explicit /verbose toggle (v0.8.12)
## Context

Thinking traces (the V4 `reasoning_content` stream that surfaces the model's internal chain of thought before its final reply) are part of what makes DeepSeek genuinely interesting to watch — users have reported that being able to see the reasoning fully is a real product advantage, not noise. But the same trace that reads as fascinating to one user reads as overwhelming to another, and "fascinating but overwhelming" eventually trends toward "user wishes it were gone."

The goal for v0.8.12 is to **preserve the watchability advantage without the cognitive load tax**.

## Two paths worth exploring

**Path A — natural-feel default.** Rework the live-rendering of thinking traces so the default presentation reads more like an unfolding train of thought than a firehose. Concrete levers: density control (collapse repeated phrasing, fold long internal asides), pacing (subtle line-break rhythm so the eye can keep up at streaming speed), structural cues (light visual hierarchy that distinguishes "thinking about the problem" from "thinking about the answer"). Goal: the trace reads as a feature, not a panel of debug output.

**Path B — `/verbose` slash command.** Ship an explicit toggle that flips between concise (final answer + minimal trace summary) and full-fidelity (today's behavior). Discoverable, explicit, no design risk. Loses the "always interesting by default" framing — verbose-by-default would still be the launch experience but users who hit overload have a one-keystroke escape.

The two are not mutually exclusive — A could ship as the new default and B as the opt-in to either side of the dial.

## Decision questions

- Do we believe the default-on trace is genuinely valuable to *new* users, or does it require some baseline familiarity to read as a feature rather than as noise?
- If we redesign the live-render, what's the bar for "natural feel" — is it pacing, density, structural cues, or all three?
- Is there a third path (e.g. progressive disclosure: start collapsed, expand on hover/click into the trace) that would beat both A and B?

## Out of scope for v0.8.11

Pure design + UX work. v0.8.11 stays focused on the six bug fixes already in flight.

## Reporter

Maintainer-scoped enhancement (Hmbown), 2026-05-04. Spawned from in-session reflection while reviewing community Telegram feedback.

merchloubna70-dot:
Closes #598. Adds `/verbose` slash command to toggle between compact (summary only) and full (complete chain-of-thought) reasoning display. Defaults to compact.\n\nImplemented using `deepseek exec --model deepseek-v4-flash`. 🐋

### DIFF
--- a/crates/tui/src/commands/debug.rs
+++ b/crates/tui/src/commands/debug.rs
@@ -250,6 +250,20 @@ fn format_cache_history(app: &App, count: usize, locale: Locale) -> String {
     format!("{header}{body}{footer}")
 }
 
+/// Toggle verbose thinking display.
+///
+/// When enabled, full chain-of-thought (reasoning_content) is shown in the
+/// transcript for completed thinking blocks. When disabled (default), only
+/// a compact summary is shown.
+pub fn verbose_thinking(app: &mut App) -> CommandResult {
+    app.verbose_thinking = !app.verbose_thinking;
+    let state = if app.verbose_thinking { "on" } else { "off" };
+    CommandResult::message(format!(
+        "Verbose thinking display: {state}. Now showing {} chain-of-thought.",
+        if app.verbose_thinking { "full" } else { "compact" }
+    ))
+}
+
 fn humanize_age(d: std::time::Duration) -> String {
     let secs = d.as_secs();
     if secs < 60 {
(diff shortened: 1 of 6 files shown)


Hmbown:
Thanks for the `/verbose` thinking-trace UX work. I am closing this old conflicting branch, but not the underlying idea.

The linked issue #598 stays open: this still needs a current-main design for persistence, mode interaction, tests, and a clean UX path without unrelated onboarding refactor noise. A fresh focused PR would be welcome.

PR #607, opened 2026-05-04
feat(commands): add /think off|high|max reasoning dispatcher (#543)

### LINKED ISSUE #543: V4: per-turn reasoning dispatcher (Non-Think / Think-High / Think-Max + workshop axis)
## Thesis
DeepSeek V4 has three reasoning intensities (Non-Think / Think-High / Think-Max) and a fourth axis — *decomposition to the workshop sub-loop* (#8). Today reasoning_effort is a global session toggle (`Shift+Tab` cycles off→high→max in `crates/tui/src/models.rs:32`). The structurally interesting move is per-turn dispatch: pick intensity (and possibly: workshop offload) based on signal type for that turn. Cache locality forces a constraint: model family stays uniform within a session.

## Current behavior
- Reasoning effort is set globally per session via `Shift+Tab` and passed identically on every turn — `crates/tui/src/core/engine/turn_loop.rs:246`.
- No automatic escalation/de-escalation based on turn signal.
- No way to route a turn to "the workshop will handle this" instead of "answer in this loop."

## Proposed change
**Per-turn dispatcher** decides between four options:
- **Non-Think:** mechanical edits, simple lookups, file-read-and-respond.
- **Think-High:** planning, architecture, hard debugging, multi-step refactors.
- **Think-Max:** verifier-failed-twice escalations; requires ≥384K context per V4 model card.
- **Workshop offload (#8 + #9):** the task is haystack/extraction-shaped — don't think harder, decompose to the workshop with Flash sub-calls.

**Constraint: model family stays session-uniform.** Switching between V4-Pro and V4-Flash mid-session almost certainly breaks DeepSeek's KV cache (different model = different cache namespace). The dispatcher toggles `reasoning_effort` and routes-to-workshop within a session. Family escalation (Flash→Pro) happens at session boundaries or via explicit `/escalate`.

**Signal sources for the dispatcher:**
- Recent turn outcomes (verifier failed → escalate intensity).
- Task framing (user said "design", "refactor", "explain why" → Think-High).
- Tool patterns (about-to-grep-100K-hits → workshop offload).
- Reasoning content from prior turn (model expressed uncertainty → escalate next turn).

## Open questions / risks
- Cache locality assumption needs empirical confirmation. If Flash and Pro share cache, mid-session model swaps become viable. Test before designing around the assumption.
- Dispatcher could feel jittery if it changes intensity too often — needs hysteresis.
- Think-Max needs ≥384K context — dispatcher has to ensure sufficient budget before escalating.

## Acceptance signals
- Mechanical edit turns use Non-Think; design discussion turns use Think-High; verifier-failed turns escalate.
- Verifier loops auto-escalate `reasoning_effort` after consecutive failures.
- A turn that would have grepped 100K hits routes through the workshop instead.
- Cache hit rate per session stays high (no model-swap thrash).

merchloubna70-dot:
Closes #543. Adds `/think <off|high|max>` slash command mapping to `ReasoningEffort` enum values. Includes usage help when called without args showing current mode.\n\nImplemented using `deepseek exec --model deepseek-v4-flash`. 🐋

### DIFF
--- a/crates/tui/src/commands/mod.rs
+++ b/crates/tui/src/commands/mod.rs
@@ -327,6 +327,12 @@ pub const COMMANDS: &[CommandInfo] = &[
         description_id: MessageId::CmdTrustDescription,
     },
     CommandInfo {
+        name: "think",
+        aliases: &[],
+        usage: "/think <off|high|max>",
+        description_id: MessageId::CmdThinkDescription,
+    },
+    CommandInfo {
         name: "logout",
         aliases: &[],
         usage: "/logout",
@@ -512,6 +518,7 @@ pub fn execute(cmd: &str, app: &mut App) -> CommandResult {
         "agent" => config::agent_mode(app),
         "plan" => config::plan_mode(app),
         "trust" => config::trust(app, arg),
+        "think" => think(app, arg),
         "logout" => config::logout(app),
 
         // Debug commands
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks for the reasoning-mode command work. I am closing this as superseded because #543 has already been closed with evidence from merged #669 (`f8f816053`), which landed the supported thinking/reasoning controls on `main`.

No need to rebase this old conflicting branch; the user-facing path is already shipped.

PR #610, opened 2026-05-04
feat(memory): capture reasoning_content as reflective memory (#544)

### LINKED ISSUE #544: MEMORY: capture reasoning_content as reflective memory
## Thesis
V4's `reasoning_content` is a high-fidelity log of the model's internal decomposition — *why* it took an action. Today the engine sanitizes/placeholders it just to avoid HTTP 400 (the API requires reasoning_content replay on tool-call rounds). Inversion: capture every block as a `reflective` memory record (#536's taxonomy) tied to the producing turn. The HTTP 400 workaround becomes the load-bearing connection between the model's actual thinking and the persistent memory store.

Soft-blocked on #534/#536 landing.

## Current behavior
- `crates/tui/src/client/chat.rs:458-467` substitutes a placeholder string when reasoning_content is missing, just to keep the conversation chain valid for V4 thinking mode.
- `crates/tui/src/client/chat.rs:688-714` scans every assistant message and forces a non-empty reasoning_content for replay.
- The actual reasoning blocks are surfaced to the UI as ephemeral content; not persisted, not queryable, not searchable.

## Proposed change
- Intercept every reasoning_content block in the streaming response.
- Write it via `MemoryBackend::remember` (#535) as a `reflective` memory (#536), tagged with: turn ID, model, reasoning_effort level, file/symbol references extracted from the content.
- Tie to the resulting tool calls / file edits (graph edge `causes` per #537) so future "why was this code written this way?" can traverse from the artifact back to the thought.
- Recall from inside the workshop (#8) so the model can query its own past reasoning programmatically.

## Open questions / risks
- Storage growth: every turn produces a reasoning block on Think-High/Max. Need pruning policy under #539 hygiene — probably "keep verbatim for N days, then summarize, then drop after M days unless promoted."
- Sensitive content: reasoning may include secrets/API keys/internal identifiers the user didn't intend to persist. Need a redaction pass on ingest, or per-scope opt-out.
- Cost: reasoning blocks can be long; ingestion subagent (#538) needs to handle them efficiently — possibly via the workshop (#8) for the long ones.

## Acceptance signals
- reasoning_content blocks are persisted via MemoryBackend.
- Recall surfaces the original reasoning when querying about a specific edit/file.
- Edge graph shows `causes` links from reasoning → tool call → file change.
- A turn 5 sessions ago can be queried: "what was the model thinking when it wrote function X?"

merchloubna70-dot:
Closes #544. After each turn, if `reasoning_content` contains a decision signal (heuristic: contains 'because', 'therefore', 'decided', 'implemented'), extracts and stores it as a memory note tagged 'reasoning'. Gated behind `capture_reasoning_memory = false` (opt-in).\n\nImplemented using `deepseek exec --model deepseek-v4-flash`. 🐋

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -772,6 +772,14 @@ pub struct Config {
     #[serde(default)]
     pub subagents: Option<SubagentsConfig>,
 
+    /// Capture reasoning_content (thinking tokens) as reflective memory
+    /// notes (#544). When `true`, after each turn the engine inspects the
+    /// model's reasoning_content for decision-like signals (heuristic:
+    /// "because", "therefore", "decision", "should") and appends matching
+    /// snippets to the notes file with a `[reasoning]` tag. Default `false`.
+    #[serde(default)]
+    pub capture_reasoning_memory: Option<bool>,
+
     /// Runtime API server tuning (`deepseek serve --http`). Currently only
     /// hosts the CORS allow-list extension (whalescale#255 / #561). When the
     /// table is absent, the daemon ships with localhost:3000 / localhost:1420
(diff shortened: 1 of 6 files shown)


Hmbown:
Thanks for the reflective-memory experiment. I am closing this branch rather than asking for a rebase because the memory system has moved on, and this implementation bypasses the newer memory backend/taxonomy/redaction direction.

The linked issue #544 stays open. The next version should route through the durable memory architecture, avoid raw heuristic notes, and include tests for redaction and promotion behavior.

PR #613, opened 2026-05-04
feat(session): auto-load README/AGENTS.md at session start (#542)

### LINKED ISSUE #542: V4: load-the-repo defaults for code understanding (with workshop carve-out)
## Thesis
Most coding agents RAG over the codebase because dense models can't fit it. V4 can. For *code understanding* tasks (what the model is actively editing/reasoning about), default to loading the directory tree + selected files into the working_set instead of waiting for tool-driven discovery. For *haystack* tasks (logs, transcripts, large corpora), the right path is the workshop (#8 + #9), not bulk-load. Carve by *task shape*, not just *input size*.

## Current behavior
- The agent selectively adds files to the WorkingSet based on tool interactions — `crates/tui/src/working_set.rs:340-380`.
- No proactive directory tree load at session start.
- `recall_archive` (BM25 over JSONL cycle archives) handles "search past sessions" but is the only haystack pattern, and it's slow and BM25-only.

## Proposed change
- **At session start:** measure total in-scope source-file token count (respect `.gitignore`, `.deepseekignore`).
- **If under threshold N (initial: 300K tokens):** automatically map the directory tree + load source files into the working_set. Prepended to system context, deterministically ordered, byte-stable for cache reuse.
- **If over N:** fall back to active-discovery via existing tool patterns; don't auto-load. Recommend `recall_archive` or workshop-based search.
- **Threshold rationale:** 1M context, but realistic working budget needs ~200K headroom (conversation, system prompt + tools, thinking-mode reserve, tool result accumulation). 300K leaves room.

## Carve-out: this is for code understanding, not haystacks
- Code the model is actively editing → resident in parent context (this issue).
- Logs, large transcripts, archive search results, web-fetch corpora, bulk grep output → routed through the workshop (#8) and offloaded by default (#9). Parent never sees the haystack.

## Open questions / risks
- Threshold tuning: 300K is a starting estimate; needs measurement.
- Should `.gitignore` apply by default, or a stricter `.deepseekignore`?
- Interaction with #541 (cache-maximal compaction): bulk-loaded repo content shouldn't trigger compaction; both issues need to land together to not regress.

## Acceptance signals
- A 100K-token repo session has the source tree resident before the first user message.
- A 1M-token monorepo session falls back to active discovery without bulk load.
- Repeated turns over the same project show high cache hit rate (the loaded prefix is byte-stable).
- `recall_archive` usage trends down for in-scope tasks (it's no longer needed).

merchloubna70-dot:
Closes #542. At session start, if a git repo is detected, automatically reads README.md and AGENTS.md (if present) + top-level directory listing. Gated behind `auto_load_repo = true` config option (default: true).\n\nImplemented using `deepseek exec --model deepseek-v4-flash`. 🐋

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -198,6 +198,19 @@ terminal_probe_timeout_ms = 500 # optional startup terminal-mode timeout (100-50
 osc8_links = true            # emit OSC 8 escapes around URLs (Cmd+click in iTerm2/Ghostty/Kitty/WezTerm/Terminal.app 13+); set false for terminals that misrender
 
 # ─────────────────────────────────────────────────────────────────────────────────
+# Auto-load repo defaults (#542)
+# ─────────────────────────────────────────────────────────────────────────────────
+# When true (the default) and the workspace is a git repository, the agent's
+# system prompt is seeded at session start with the full contents of
+# README.md, any existing AGENTS.md / CLAUDE.md, and a top-level directory
+# listing. This gives the model immediate project context without waiting
+# for the first user message.
+#
+# Set to false in a non-repo or very large repo that makes README inclusion
+# wasteful, or if you prefer to load context manually.
+# auto_load_repo = true
+
+# ─────────────────────────────────────────────────────────────────────────────────
 # Feature Flags
 # ─────────────────────────────────────────────────────────────────────────────────
 [features]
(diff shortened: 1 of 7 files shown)


Hmbown:
Thanks for the session-start context loading pass. I am closing this old branch rather than asking for a rebase because it now conflicts and the implementation is too thin for the current project-context path.

The linked issue #542 stays open. It still needs thresholding, ignore-file handling, working-set integration, tests, and a safer design than default-on `.git` detection. A new current-main PR would be the right way to revive it.

PR #615, opened 2026-05-04
feat(memory): add after-turn auto-extract memory hook (#538)

### LINKED ISSUE #538: MEMORY: loop — lifecycle hooks + background auto-extraction
Part of #534 (Phase 3 EPIC). Builds on #535 (substrate). Pairs with #538 (hygiene).

## Thesis
Two patterns from reference-cc that close the memory loop:
1. **Lifecycle hooks** (`SessionStart`, `PreCompact`, `PostCompact`, `InstructionsLoaded`) — backends and skills react to session boundaries without forking the engine.
2. **Background auto-extraction** — a low-cost subagent runs after each turn (or on idle), reads the last N messages, and writes high-signal facts to the memory store automatically.

Together: memory grows without user effort, and external integrations get clean event points.

## Current behavior
- `crates/tui/src/hooks.rs` supports general hooks but has no memory-specific hook types.
- Memory writes are entirely user-driven (`# add` quick-capture or `/memory` editor).
- No background extraction.

## Proposed change
**Hooks** — extend `crates/tui/src/hooks.rs` with four memory-relevant types:
- `SessionStart` — fired before the first turn; payload includes loaded memory summary.
- `PreCompact` — fired before context compaction; payload includes turn count and token budget.
- `PostCompact` — fired after compaction; payload includes what was preserved/summarized.
- `InstructionsLoaded` — fired after CLAUDE.md/AGENTS.md/memory hierarchy resolves; payload includes the merged context with provenance.

Configured via `settings.json` like existing hooks. Documented payloads.

**Auto-extraction** — opt-in subagent triggered post-turn or on idle:
- Reads the last N messages.
- Calls cheapest-available model with an extraction prompt.
- Each extracted fact gets an auto-classified type (#536) and is written via `MemoryBackend::remember`.
- Dedups against existing memory before writing (FTS5 lookup).
- Counts toward session cost; shown in UI.

## Open questions
- Default trigger: post-turn, on-idle (e.g., 30s after last user input), or both?
- Default off vs default on once shown safe? Probably default off until #538 hygiene proves it doesn't accumulate noise.
- Should extraction use a smaller model than the main turn (cost) or the same model (quality)?

## Acceptance signals
- Hooks fire with documented payloads.
- Auto-extraction doesn't block the main turn (truly async).
- Extraction count visible in UI; per-extraction cost folded into session total.
- With auto-extraction off, behavior identical to pre-issue.
- `PreCompact` hook can be used to trigger a memory consolidation pass (#538).

merchloubna70-dot:
Closes #538. Adds after-turn lifecycle hook that auto-extracts 1-line memory notes from notable assistant outputs. Gated behind `auto_extract_memory = false` (opt-in).\n\nImplemented using `deepseek exec --model deepseek-v4-flash`. 🐋

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -70,6 +70,16 @@ memory_path = "~/.deepseek/memory.md"
 # (`instructions = []`) clears the user list for the current repo.
 
 # ─────────────────────────────────────────────────────────────────────────────────
+# Memory: persistent notes + auto-extract (#538 / #489)
+# ─────────────────────────────────────────────────────────────────────
+# auto_extract_memory = false   # default off. When true, after each completed
+#                               # turn the TUI scans the assistant's response
+#                               # for decision-worthy sentences (containing
+#                               # "I will", "decided", "implemented") and
+#                               # appends the first match to the notes file
+#                               # at notes_path. Useful for building a running
+#                               # log of what the agent did.
+
 # User memory (#489) — opt-in. When enabled, the TUI reads memory_path on
 # startup and injects its contents into the system prompt as a
 # <user_memory> block, intercepts `# foo` typed in the composer to append
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -764,6 +764,14 @@ pub struct Config {
     #[serde(default)]
     pub lsp: Option<LspConfigToml>,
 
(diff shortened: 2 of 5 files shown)


Hmbown:

PR #618, opened 2026-05-04
feat(memory): per-scope size budget with smart truncation (#495)

### LINKED ISSUE #495: MEMORY: per-scope size budget + smart truncation
Part of #489.

## Summary
When the combined memory blob would exceed ~50% of the model context budget, truncate by scope priority + mtime instead of dropping the tail.

## Behavior
- Priority order: local override > project > user > parent-walk fallbacks.
- Within a scope, newest content (by mtime) wins.
- Truncated portions get a `<truncated bytes=N source=...>` marker so the model knows they exist.

## Acceptance
- 200 KB of memory + a 64 K context model still leaves room for transcript.
- Truncation never silently drops the project file in favor of an unrelated user file.

Effort: M (~150 LoC).

merchloubna70-dot:
Closes #495. Adds `max_memory_tokens` config field (default 4000). When injecting memories into context, truncates oldest/lowest-priority entries to stay within budget, sorted by recency + access frequency.\n\nImplemented using `deepseek exec --model deepseek-v4-flash`. 🐋

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -419,6 +419,11 @@ pub struct MemoryConfig {
     /// `# foo` typed in the composer to append to that file. Default `false`.
     #[serde(default)]
     pub enabled: Option<bool>,
+    /// Maximum tokens of memory content to inject into the system prompt
+    /// (#495). Older / lower-priority entries are truncated first to stay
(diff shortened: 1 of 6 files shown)


Hmbown:

PR #619, opened 2026-05-04
fix(cache): move volatile content out of system prompt prefix (#533)

### LINKED ISSUE #533: Prompt-prefix discipline audit for cache-maximal turns
## Thesis
Cache-maximal design only works if the serialized request prefix is byte-stable. The repo already has cache-friendly prompt ordering, but v0.8.9 needs a concrete audit harness around system prompt assembly, Chat Completions serialization, tools, memory, goal state, and working-set/context blocks so small volatile fields do not silently bust the cache.

## Current behavior
- `crates/tui/src/core/engine.rs:790` refreshes the system prompt every turn.
- `crates/tui/src/core/engine.rs:1637` rebuilds the prompt from mode, workspace, skills, instructions, user memory, goal objective, compaction summary, and working-set summary.
- `crates/tui/src/working_set.rs:382` documents byte stability for the working-set block and has already removed turn-relative churn.
- `crates/tui/src/client/chat.rs:371` serializes system and message history into Chat Completions payloads.
- `crates/tui/src/client.rs:900` parses cache hit/miss telemetry, but there is no per-turn prefix hash or drift report.

## Proposed change
Add a deterministic request-prefix audit harness: compute and log a stable hash for the serialized cacheable prefix, break down prefix sections, and expose a debug command or `/tokens` detail showing what changed since the previous turn. Add tests that fail on timestamp/random/session-metadata drift in cacheable sections. Keep volatile sections intentionally late and label them.

## Open questions / risks
- System prompt can legitimately change when mode, goal, memory, skills, or working-set contents change; the audit needs useful diffs, not false alarms.
- Tool catalogs may be large and can reorder if registry construction is not deterministic.
- Reasoning-content replay is required for DeepSeek V4 thinking/tool calls and may affect prefix shape.
- Prefix hashes must not leak prompt text into logs unless debug mode is explicitly enabled.

## Acceptance signals
- A unit test serializes the same request twice and asserts identical prefix hash/bytes for unchanged workspace state.
- A targeted test mutates only the user message suffix and proves the cacheable prefix hash remains stable.
- A debug surface shows section-level prefix changes and associated cache hit/miss token deltas.
- The audit catches at least one intentionally injected volatile field in a cacheable prefix section.

merchloubna70-dot:
Closes #533. Audits system prompt assembly and moves volatile/dynamic content away from the stable prefix to preserve DeepSeek V4 prefix cache across turns.

### DIFF
--- a/crates/tui/src/prompts.rs
+++ b/crates/tui/src/prompts.rs
@@ -250,15 +250,21 @@ pub fn system_prompt_for_mode_with_context(
 ///
 ///   1. mode prompt (compile-time constant)
 ///   2. project context / fallback (workspace-static)
-///   3. skills block (skills-dir-static)
-///   4. `## Context Management` (compile-time constant, Agent/Yolo only)
-///   5. compaction handoff template (compile-time constant)
-///   6. handoff block — file-backed; rewritten by `/compact` and on exit
-///   7. working-set summary — drifts when a new path is observed
+///   3. instructions block — file-backed; workspace-static
+///   4. user memory block — session-stable (edits via `/memory` or tool)
+///   5. session goal — stable for the session duration
+///   6. skills block (skills-dir-static)
+///   7. `## Context Management` (compile-time constant, Agent/Yolo only)
+///   8. compaction handoff template (compile-time constant)
+///   ── volatile-content boundary (#533) ──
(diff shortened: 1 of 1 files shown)


Hmbown:

PR #620, opened 2026-05-04
feat(memory): memory hygiene — /forget command and decay scoring (#539)

### LINKED ISSUE #539: MEMORY: hygiene — decay scoring, periodic review subagent, one-key drop
Part of #534 (Phase 3 EPIC). Pairs with #538 (loop). Without this, auto-extraction is a liability — with it, the system gets *less* annoying over time instead of more.

## Thesis
Hooks alone don't solve memory hygiene; they're the trigger, not the mechanism. The actual mechanism is a small set of patterns working together: decay scoring drives ranking down for unused memories, a periodic review subagent surfaces removal candidates, contradiction detection (#537) prevents silent conflicts, and a trivially-fast "drop this" gesture lets users prune in one keystroke when they see noise.

## Current behavior
- Memories live forever once written.
- No access tracking.
- No review pass.
- No quick-drop UI.

## Proposed change
**Decay scoring** — every memory carries `last_accessed`, `access_count`, and a per-type decay rate. Recall (#536) already uses these as signals; the same scores power a "below-threshold for N days" candidate set for proposed removal. Decay is type-aware: `episodic` decays fast (days), `procedural` slowly (months), `constitutional` not at all by default.

**Periodic review subagent** — fires from `PostCompact` hook (#538) and on a daily-on-startup cadence. Reads memory + access stats, proposes a review queue:
- N memories below decay threshold → drop?
- N pairs with high text overlap → merge?
- N contradictions outstanding (#537) → resolve?

**Always proposes, never auto-applies.** Reference-cc's `/remember` skill is exactly this pattern; it's what makes auto-extraction tolerable instead of annoying.

**One-key drop in `/memory` UI** — the cheapest hygiene mechanism. When a user sees a bad memory, single keystroke removes it. Builds the user-driven feedback loop the decay scoring depends on.

**Promotion lifecycle** — when an `episodic` memory gets accessed N times across M sessions, propose promotion to `procedural`. When `procedural` proves durable, propose promotion to `constitutional`. Promotions never automatic — proposed via review queue.

## Open questions
- Default decay rates per type.
- Review cadence: daily? Per-session? Both?
- How to surface the review queue without being noisy — a numbered count in the status bar that opens on demand?
- Should "auto-apply trivial cases" ever be allowed (e.g., dedup of literal duplicates)?

## Acceptance signals
- Memory store size stabilizes over time instead of growing monotonically.
- User-initiated drops trend down across releases (proxy for "memory feels right").
- Review queue is opt-in to view, never blocks user flow.
- Promotion proposals visible and approvable from `/memory`.
- A long-running test: 10K simulated turns + extraction events leaves the store under a configurable size cap.

merchloubna70-dot:
Closes #539.

### DIFF
--- a/crates/tui/src/memory.rs
+++ b/crates/tui/src/memory.rs
@@ -1,20 +1,32 @@
-//! User-level memory file.
+//! Structured user-level memory file with decay scoring and per-entry management.
 //!
-//! v0.8.8 ships an MVP that lets the user keep a persistent personal
-//! note file the model sees on every turn:
+//! v0.8.8 ships an MVP that lets the user keep persistent personal
+//! notes the model sees on every turn. v0.9.0 upgrades the format
+//! from flat markdown to JSONL (one JSON entry per line), adding
+//! per-entry IDs, timestamps, access tracking, and decay scoring.
 //!
-//! - **Load** `~/.deepseek/memory.md` (path is configurable via
-//!   `memory_path` in `config.toml` and `DEEPSEEK_MEMORY_PATH` env),
-//!   wrap it in a `<user_memory>` block, and prepend it to the system
-//!   prompt alongside the existing `<project_instructions>` block.
-//! - **`# foo`** typed in the composer appends `foo` to the memory
-//!   file as a timestamped bullet — fast capture without leaving the TUI.
-//! - **`/memory`** shows the resolved file path and current contents, and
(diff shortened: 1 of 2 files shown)


CrepuscularIRIS:
Closing as invalid: PR body is placeholder/template-level only (no substantive reviewable summary). Please reopen only with full rationale, change summary, and verification notes.

Hmbown:

PR #600, opened 2026-05-04
fix(fork): print confirmation with forked session name (#576)

merchloubna70-dot:
Closes #576.

**Root cause:** After forking a session, the TUI gave no visual feedback about what was created, leaving users unsure if the fork succeeded.

**Fix:** Print a brief confirmation line after successful fork showing the new session name, so users know it worked and can `deepseek resume <name>` immediately.

Implemented using `deepseek exec --model deepseek-v4-flash`. 🐋

PR #651, opened 2026-05-04
fix(theme): complete Color::Reset migration across all UI widgets (#601)

merchloubna70-dot:
Follow-up to #600/#601, based on Gemini HIGH priority review.

The previous fix only changed one location. This PR replaces **all** hardcoded `Color::Black` backgrounds with `Color::Reset` across 10 UI files, making the TUI fully respect the user's terminal background color.

Files changed: sidebar, session_picker, model_picker, provider_picker, command_palette, pager, plan_prompt, user_input, deepseek_theme, ui.

### DIFF
--- a/crates/tui/src/deepseek_theme.rs
+++ b/crates/tui/src/deepseek_theme.rs
@@ -136,6 +136,7 @@ mod tests {
     use super::{Theme, Variant, active_theme};
     use crate::palette;
     use crate::tui::history::ToolStatus;
+    use ratatui::style::Color;
 
     #[test]
     fn active_theme_returns_dark() {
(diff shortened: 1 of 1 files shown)


Hmbown:
## Triple-check finding: broken test

The production code was correctly changed to `Color::Reset` for `section_bg`, but the test at `crates/tui/src/deepseek_theme.rs:150` still asserts the old value:

```rust
assert_eq!(theme.section_bg, palette::DEEPSEEK_INK);
```

This will fail with `Color::Reset != Color::Rgb(18, 29, 39)` at runtime. The assertion needs to be updated to match the new default. The Color::Reset migration itself is correct — it's just the test that's stale.

merchloubna70-dot:
Fixed — updated the test assertion at `crates/tui/src/deepseek_theme.rs:150` from `palette::DEEPSEEK_INK` → `Color::Reset` to match the production code. Also added the `ratatui::style::Color` import to the test module.

Hmbown:

PR #658, opened 2026-05-04
feat(runtime): route large tool outputs through workshop to protect parent context (closes #548)

### LINKED ISSUE #548: RUNTIME: default-route large tool outputs through the workshop (protect parent context)
## Thesis
The parent model's context is precious — every token in it should earn its place. Today, large tool outputs (file reads, grep haystacks, web fetches, archive search hits, command output) get truncated to fit and dumped into the parent's context, polluting the cache prefix and burning the most expensive token budget on raw haystacks. Per MGH (Zhang 2026): decomposition is the leverage. Default-route any tool output above a size threshold through the workshop (#547). The workshop processes with V4-Flash, returns a synthesis. The parent only ever sees the answer.

Generalizes #511 (currently scoped to sub-agent output) to *any* tool output. Closes the loop with #547 — the workshop becomes the substrate, this is the policy.

## Current behavior
- `read_file` returns up to a hard cap; large files dumped raw into parent context.
- `grep_files` returns top-N hits; truncates with marker; raw text into parent.
- `recall_archive` BM25 over JSONL cycles; raw hits into parent.
- `web_search` / `fetch_url` results dumped raw.
- `exec_shell` output capped; raw into parent.
- `agent_*` sub-agent outputs sometimes summarized (#511 in flight), sometimes not.

Result: a single haystack tool call can blow the cache prefix and cost more than the actual reasoning work.

## Proposed change
- **Default policy:** any tool output > N tokens (initial: 4K) routes through the workshop instead of returning raw to the parent.
- **Mechanism:** tool result is loaded as a workshop variable (e.g. `last_tool_result`); a synthesis subagent (V4-Flash, cheap) processes it against the original tool intent; only the synthesis returns to the parent.
- **Override:** the model can request raw output via a tool parameter (`raw=true`) when it knows it needs the full thing in resident context.
- **Promotion:** the workshop variable persists; the model can later call `promote_to_context(last_tool_result)` (per Phase C of the workshop issue) to pull the full thing back if synthesis missed something.
- **Per-tool tunable threshold:** some tools are more haystack-shaped than others. `read_file` threshold higher; `grep_files` threshold lower.

## Cache discipline
This is the cache-protection mechanism. Without it, #541 (cache-maximal context default) is fragile — one bad grep blows the prefix. With it, the parent's cache prefix stays stable; haystack work happens off to the side.

## Open questions / risks
- **Threshold tuning:** 4K is a starting estimate. Too low → workshop overhead on small results. Too high → cache pollution on medium results.
- **Synthesis quality:** Flash-summarized tool output may miss details. Mitigated by promotion path + `raw=true` opt-out, but worth measuring on real tasks.
- **Latency:** workshop synthesis adds an LLM call per large tool output. Cheap (Flash) but real. May want streaming synthesis.
- **Tool authorship:** new tools need to declare a default threshold or inherit a sensible one.

## Acceptance signals
- A `read_file` on a 100KB file returns a synthesis to the parent, with the full content available in the workshop for promotion.
- `grep_files` returning 200 hits returns a ranked summary, not raw matches.
- Cache hit rate on long V4 sessions stays high even when many large tool calls happen.
- Override via `raw=true` works and bypasses the workshop.
- Token cost per session for haystack-heavy work drops measurably (Flash synthesis cheaper than Pro reading raw).

merchloubna70-dot:
## Summary

- Introduces `LargeOutputRouter` that intercepts tool results exceeding a token threshold (default 4 096 tokens) before they enter the parent context
- A structured synthesis header replaces the raw blob in the parent; the full raw content is stored in the workshop variable `last_tool_result` for later `promote_to_context` retrieval
- Individual tool calls can pass `raw = true` to bypass routing for that invocation
- Per-tool threshold overrides are supported via `[workshop.per_tool_thresholds]` in `config.toml`

## Changes

| File | Change |
|------|--------|
| `crates/tui/src/tools/large_output_router.rs` | New module: `LargeOutputRouter`, `WorkshopConfig`, `WorkshopVariables`, `RouteDecision`, token estimator, synthesis-prompt builder, 14 unit tests |
| `crates/tui/src/tools/spec.rs` | `ToolContext` gains `large_output_router` + `workshop_vars` fields and `with_large_output_router` builder |
| `crates/tui/src/tools/registry.rs` | `execute_full_with_context` applies routing after every tool call |
| `crates/tui/src/core/engine.rs` | `EngineConfig.workshop`, `Engine.workshop_vars`, wired in `build_tool_context` |
| `crates/tui/src/config.rs` | `Config.workshop` deserialization + `merge_config` propagation |
| `crates/tui/src/main.rs`, `runtime_threads.rs`, `tui/ui.rs` | Pass `workshop` field through `EngineConfig` initializers |
| `config.example.toml` | Documents `[workshop]` table with threshold and per-tool override examples |

## Design notes

- `ToolContext.large_output_router = None` by default: sub-agents and test contexts see no routing (prevents recursive synthesis calls)
- Token count uses a conservative 3 chars/token heuristic — the real tokeniser is not a dependency at this layer
- The synthesis currently produces a structured preview + provenance header; the live V4-Flash API call is deferred to a follow-up once async is safe from the registry layer
- `WorkshopVariables` is an `Arc<Mutex<…>>` shared across all tool calls in the session so `last_tool_result` persists until explicitly consumed

## Test plan

- [ ] `cargo check` — zero errors, zero warnings (verified in CI pre-push)
- [ ] `[workshop] large_output_threshold_tokens = 512` in local config → small grep results pass through, large file reads get the synthesis header
- [ ] `read_file path="bigfile.rs" raw=true` → bypass confirmed, full content returned
- [ ] `[workshop.per_tool_thresholds] exec_shell = 256` → shell output synthesised at tighter limit while other tools use the global threshold
- [ ] Sub-agent spawned via `spawn_agent` inherits no router → no recursive synthesis

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -241,6 +241,26 @@ cycle_threshold = 768000
 seam_model = "deepseek-v4-flash"
 
 # ─────────────────────────────────────────────────────────────────────────────────
+# Workshop / Large-Output Routing (#548)
+# ─────────────────────────────────────────────────────────────────────────────────
+# Tool outputs exceeding `large_output_threshold_tokens` are routed through a
+# V4-Flash synthesis sub-agent.  Only the synthesis reaches the parent context;
+# the raw text is stored in the workshop variable `last_tool_result` so the
+# parent can call `promote_to_context` later if it needs the full content.
+#
+# Per-tool overrides let high-volume tools (e.g. exec_shell) use tighter
+# thresholds without changing the global default.
+#
+# Add `raw = true` to any tool call to bypass routing for that invocation.
+#
+# [workshop]
+# large_output_threshold_tokens = 4096
+# [workshop.per_tool_thresholds]
+# exec_shell  = 2048   # shell output synthesised aggressively
+# grep_files  = 2048
+# web_search  = 8192   # web results can be large; give them more room
+
+# ─────────────────────────────────────────────────────────────────────────────────
 # Capacity Controller (runtime pressure guardrails)
 # ─────────────────────────────────────────────────────────────────────────────────
 [capacity]
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -778,6 +778,11 @@ pub struct Config {
     /// / tauri://localhost as the only allowed dev origins.
     #[serde(default)]
     pub runtime_api: Option<RuntimeApiConfig>,
+
+    /// Workshop / large-tool-output routing (#548). When absent, the global
+    /// default threshold of 4 096 tokens applies and routing is active.
+    #[serde(default)]
+    pub workshop: Option<crate::tools::large_output_router::WorkshopConfig>,
 }
 
 /// `[runtime_api]` table — knobs for the local HTTP/SSE daemon.
@@ -2026,6 +2031,7 @@ fn merge_config(base: Config, override_cfg: Config) -> Config {
         },
         subagents: override_cfg.subagents.or(base.subagents),
         runtime_api: override_cfg.runtime_api.or(base.runtime_api),
+        workshop: override_cfg.workshop.or(base.workshop),
     }
 }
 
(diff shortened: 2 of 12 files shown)


ILoveScratch2:
谁家PR机器人

PR #660, opened 2026-05-04
feat(subagents): cache-aware resident file sub-agents for refactors (closes #529)

### LINKED ISSUE #529: Cache-aware resident file sub-agents for refactors
## Thesis
For broad refactors, DeepSeek's cheap cached input suggests per-file resident sub-agents: each child gets a stable prefix containing the repo skeleton plus one target file, then receives small instructions over time. This is different from the removed swarm surface because the win is persistent cached locality, not arbitrary fan-out volume.

## Current behavior
- `crates/tui/src/core/engine.rs:833` registers ordinary `agent_spawn` tools for Agent/Yolo modes.
- `crates/tui/src/core/engine.rs:837` builds sub-agent runtimes from the current client/model/tool context, but without a per-file resident prompt contract.
- `docs/TOOL_SURFACE.md` notes `agent_spawn` as bounded full sub-agent loops and says the legacy `agent_swarm` surface was removed.
- Open issue #511 already tracks summarizing child results before parent reinsertion; this proposal is about child prompt residency and cache locality, not transcript folding.

## Proposed change
Introduce a cache-aware sub-agent spawn variant or prompt profile that binds a child to one file or module. The parent supplies a deterministic repo skeleton, target file contents, and refactor objective. Subsequent `agent_send_input` messages should preserve that prefix and only append narrow deltas. Keep the existing `agent_spawn` lifecycle and caps; do not resurrect `agent_swarm` or `/swarm`.

## Open questions / risks
- Resident agents can hold stale file contents after parent/other child edits; need invalidation or refresh messages.
- Parent/child write conflicts need explicit ownership or side-worktree isolation.
- Concurrency can hit output and tool-latency ceilings even when input is cached.
- Must not duplicate #511's result-summary work; this is about child-context construction.

## Acceptance signals
- A test proves resident file-agent prompt assembly is byte-stable across follow-up inputs when the file is unchanged.
- A refactor smoke over 3 files shows each child reports cache-hit-heavy input after its first turn.
- File ownership conflicts are detected and surfaced before applying overlapping edits.
- Existing `agent_spawn`, `agent_wait`, `agent_result`, and cancellation behavior remains compatible.

merchloubna70-dot:
Adds `resident_file` parameter to `agent_spawn`. Child prompt is prefixed with file contents for byte-stable prefix caching. Ownership table prevents two agents leasing the same file simultaneously. Closes #529

---
wangfengcsu@qq.com

### DIFF
--- a/crates/tui/src/prompts/base.md
+++ b/crates/tui/src/prompts/base.md
@@ -1,5 +1,9 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
+## Language Mirror
+
+Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
+
 ## Preamble Rhythm
 
 When starting work on a user request, open with a short, momentum-building line that names the action you're taking. Keep it reserved — state what you're doing, not how you feel about it.
--- a/crates/tui/src/prompts/base.txt
+++ b/crates/tui/src/prompts/base.txt
@@ -1,5 +1,9 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
+## Language Mirror
+
+Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
+
 ## Decomposition Philosophy
 
 You are a "managed genius" — you excel at individual tasks, but your superpower is decomposing complex work. **Always decompose before you act.** A few minutes spent planning saves many minutes of thrashing.
--- a/crates/tui/src/prompts/base.txt
+++ b/crates/tui/src/prompts/base.txt
@@ -1,9 +1,5 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
-## Language Mirror
-
-Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
-
 ## Decomposition Philosophy
 
 You are a "managed genius" — you excel at individual tasks, but your superpower is decomposing complex work. **Always decompose before you act.** A few minutes spent planning saves many minutes of thrashing.
--- a/crates/tui/src/tools/subagent/mod.rs
+++ b/crates/tui/src/tools/subagent/mod.rs
@@ -470,6 +470,11 @@ struct SpawnRequest {
     /// into separate git worktrees: parent runs `git worktree add` first,
     /// then spawns children with the worktree path as `cwd`.
     cwd: Option<PathBuf>,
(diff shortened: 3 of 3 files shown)


ILoveScratch2:
谁家PR机器人

Hmbown:

Reviewed the cache-aware resident file sub-agents as part of merging into v0.8.12. One finding:

### Lease table has no cleanup

The global ownership table (`HashMap<PathBuf, String>` mapping file → agent_id) prevents two agents from holding a resident lease on the same file simultaneously. The lease is acquired on spawn:

```rust
if let Some(owner) = guard.get(file_path) {
    Some(format!("Warning: agent {owner} already holds a resident lease on {file_path}"))
} else {
    guard.insert(file_path.to_string(), agent_id.clone());
    None
}
```

But the lease is **never released**. Not on agent completion, not on crash, not on cancellation. The only way to clear a lease is to restart the TUI.

This means: spawn an agent on `src/main.rs`, it finishes its work, try to spawn another on the same file → you get a conflict warning even though the first agent is long gone. It's a soft failure (the spawn still succeeds, just with a warning in the result), but it defeats the purpose of resident mode for repeated work on the same file.

### Recommendation

Hook into the agent lifecycle to release leases on terminal states (completed, failed, cancelled). The sub-agent manager's cleanup path is the right place — when an agent transitions to a terminal status, walk its leases and remove them.

merchloubna70-dot:
@Hmbown Correct — the lease is never released, which makes resident mode unusable for repeated work on the same file. This is a real bug.

**Fix:** hook into the agent's terminal state transition to remove the lease. The right place is wherever the sub-agent manager marks an agent as completed/failed/cancelled.

I'll push the fix to the PR branch now. The approach:
1. Store the `resident_file` path alongside the agent in the manager
2. When transitioning to a terminal state (completed/failed/cancelled), acquire the lease table lock and remove the entry

---
wangfengcsu@qq.com

Hmbown:
## v0.8.12 fix applied

The lease cleanup leak has been fixed on `feat/v0.8.12`:

- Moved `RESIDENT_LEASES` from block-scoped static to module level so it's accessible from all lifecycle hooks
- Added `release_resident_leases_for(agent_id)` function
- Called at all three terminal transitions:
  - **Cancelled** — `cancel_agent` path
  - **Failed** — `update_failed` path
  - **Completed** — before returning `SubAgentResult`

Previously, leases were acquired on spawn but never released, causing false conflict warnings when respawning on the same file. Now the file is immediately available after an agent finishes.

Commit: `2ee826924`

**Remaining issue**: The lease acquisition inserts `"pending"` as a placeholder instead of the real agent id. The conflict warning will show "agent pending already holds a lease" rather than the actual agent nickname. This should be a follow-up fix to update the lease with the real id after agent creation.

merchloubna70-dot:
Fixed the remaining issue — after `spawn_background_with_assignment_options` returns the real `agent_id`, the lease entry is now updated from `"pending"` to the actual id. Conflict warnings will now show the correct agent nickname.

Hmbown:

PR #669, opened 2026-05-05
feat(engine): reasoning_effort auto mode (closes #663)

### LINKED ISSUE #663: reasoning_effort: add `auto` mode with cache-budget + task-complexity adaptation
## Summary

Add `reasoning_effort = "auto"` as a new option (alongside the existing `off | low | medium | high | max`). In auto mode, the engine adapts effort per turn based on three signals: task complexity, cache pressure, and reasoning-effort-aware tool routing for sub-agents. Default value of the user-facing toggle becomes `auto`; `off`/`low`/`medium`/`high`/`max` remain explicit overrides.

## Why now

Three independent observations land on the same need:

1. **Cost asymmetry is real.** V4-flash at $0.14/M input vs V4-pro tells us users would benefit from sub-agent fanout going to flash and the main loop staying on pro. That routing exists in spirit (the agent prompt mentions it) but no code makes it automatic.

2. **Cache pressure deserves a softer response than full compaction.** When a session has a lot of stale prefix, dropping `reasoning_effort` on the next turn rebuilds a cleaner cache without the full prefix-mutation cost of compaction. We don't have any adaptive logic that does this today — it's all manual.

3. **Per-task effort has obvious wins.** Architecture / debugging deserves max thinking; lookups and one-shot exec calls don't. The agent prompt already documents this in the "Thinking Budget" table, but the *effort* is not currently selected by the engine — only by the user via Shift+Tab.

## Sketch

- `reasoning_effort = "auto"` resolves at turn-build time via a small `auto_reasoning::select()` helper.
- Inputs to the selector:
  - Recent turn type (lookup vs code-gen vs architecture vs debugging — already classified by the prompt scaffolding).
  - Cache-hit ratio over the last N turns (already captured by `compaction.rs`'s telemetry per #584; surface to the selector).
  - Session cost vs configured budget (if any).
  - Whether the current turn is a sub-agent fanout (route to flash + low) vs main-loop reasoning (route to pro + high).
- Selector emits a tier (`off`/`low`/`medium`/`high`/`max`) plus the model id (`v4-pro` vs `v4-flash`).
- Engine respects the explicit-override channel — if the user has Shift+Tab'd to `max`, auto stays out of the way until the user toggles back.

## Risks worth naming

- **Behavior is harder to predict.** Users who like the current explicit toggle would notice the engine "deciding" things. Mitigation: `auto` is opt-in for v0.9.0; default stays explicit until we have data.
- **Selector misclassification.** A single bad heuristic call wastes tokens or under-thinks. Mitigation: emit a tracing event for each auto-selection so post-deploy review can surface drift.
- **Sub-agent routing constraints.** Some flows already pin a model — auto must not override an explicit pin.

## Out of scope

- Removing the manual Shift+Tab cycle. The toggle stays; auto sits alongside.
- Cross-provider routing (NVIDIA NIM, Fireworks, etc.). v0.9.0 candidate, scoped here as DeepSeek-platform first.
- Compaction strategy itself — that's the handoff-instead-of-compact issue.

## Reporter

Maintainer-scoped engineering observation (Hmbown), 2026-05-04.

merchloubna70-dot:
Adds Auto tier. Closes #663

### DIFF
--- a/crates/tui/src/auto_reasoning.rs
+++ b/crates/tui/src/auto_reasoning.rs
@@ -1,0 +1,70 @@
+//! Adaptive reasoning-effort tier selection for `Auto` mode (#663).
+//!
+//! When the user sets `reasoning_effort = "auto"`, the engine calls
+//! [`select`] before each turn-level request to pick the actual tier
+//! based on the current message.
+
+use crate::tui::app::ReasoningEffort;
+
+/// Choose a concrete `ReasoningEffort` tier for the next API request.
+///
+/// Rules:
+/// - Sub-agent contexts (`is_subagent == true`) → `Low`
+/// - Last user message contains `"debug"` or `"error"` → `Max`
+/// - Last user message contains `"search"` or `"lookup"` → `Low`
+/// - Everything else → `High`
+#[must_use]
+pub fn select(is_subagent: bool, last_msg: &str) -> ReasoningEffort {
+    if is_subagent {
+        return ReasoningEffort::Low;
+    }
+
+    let lower = last_msg.to_ascii_lowercase();
+
+    if lower.contains("debug") || lower.contains("error") {
+        return ReasoningEffort::Max;
+    }
+
+    if lower.contains("search") || lower.contains("lookup") {
+        return ReasoningEffort::Low;
+    }
+
+    ReasoningEffort::High
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn subagent_returns_low() {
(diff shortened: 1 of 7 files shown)


ILoveScratch2:
谁家PR机器人

Hmbown:
## Triple-check: clean

Verified end-to-end: `resolve_auto_effort` is called from `turn_loop.rs:233` before each API request, flows through `auto_reasoning::select()` which inspects the last user message for keywords ("debug"/"error" → Max, "search"/"lookup" → Low, default → High), and applies the resolved tier to the request. Sub-agent contexts always get Low. Properly wired. ✅

PR #681, opened 2026-05-05
feat(vscode): apply_patch as WorkspaceEdit with diff preview (closes #465)

### LINKED ISSUE #465: VSC: Apply edits inline
## Summary
Tool calls that modify files surface as VS Code workspace edits with diff preview, not direct writes.

## Acceptance criteria
- Each apply_patch becomes a WorkspaceEdit with user-confirmable diff.
- Direct write only with explicit setting.

## Files
extensions/vscode/src/edits.ts; crates/app-server/ (new event type if needed)

## Size
M

## Source
Source tag: VSC. Add exact upstream/opencode source links during implementation if they are not already known. Do not copy external code blindly; port the behavior into this repo's architecture.

Filed from 2026-05-02 v0.8.8 intake.

merchloubna70-dot:
Routes apply_patch tool calls through vscode.workspace.applyEdit() so users see a confirmable diff preview. Closes #465

---
wangfengcsu@qq.com

### DIFF
--- a/crates/tui/src/prompts/base.md
+++ b/crates/tui/src/prompts/base.md
@@ -1,5 +1,9 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
+## Language Mirror
+
+Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
+
 ## Preamble Rhythm
 
 When starting work on a user request, open with a short, momentum-building line that names the action you're taking. Keep it reserved — state what you're doing, not how you feel about it.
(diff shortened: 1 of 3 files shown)


merchloubna70-dot:
Replaced the skeleton with a full implementation:

- **Proper unified diff parser**: handles `---`/`+++` headers, `diff --git` multi-file separators, `@@` hunk headers with old/new start+count, and `a/`/`b/` prefix stripping. `/dev/null` detection for file creation.
- **Cumulative offset tracking**: multiple hunks targeting the same file apply correctly even when prior hunks shift line numbers (mirrors Rust-side `cumulative_offset` logic).
- **Fuzzy matching**: when context doesn't match exactly at the expected line, searches within a radius of 3 (matching the Rust default fuzz cap).
- **`createIfMissing` support**: new files are created automatically when the patch has no `---` line (indicating a new file).
- **`directWrite` mode**: when `true`, writes directly to disk via `fs.writeFileSync`, bypassing the diff preview entirely. When `false` (default), uses `vscode.workspace.applyEdit()` which triggers VS Code's confirmable diff preview.
- **Full-document range replacement**: replaces the entire file content in one `WorkspaceEdit`, so VS Code shows a clean before/after diff regardless of hunk count.

ILoveScratch2:
谁家PR机器人

Hmbown:

PR #655, opened 2026-05-04
feat(execpolicy): bash arity dictionary for command-prefix allow rules (closes #410)

merchloubna70-dot:
## Summary

- Adds `crates/execpolicy/src/bash_arity.rs` with a new `BashArityDict` struct containing a hand-curated arity table of 160+ entries covering 30+ command families (git, npm, yarn, pnpm, cargo, docker, kubectl, go, pip, gh, rustup, deno, bun, aws, terraform, helm, make, and more)
- Wires arity-aware prefix matching into `crates/tui/src/command_safety.rs` via a new public `prefix_allow_matches()` function
- Updates `crates/tui/src/execpolicy/rules.rs` so `ExecPolicyConfig::evaluate()` uses arity-aware matching for allow rules before falling back to the existing wildcard/regex `pattern_matches` path
- Updates `crates/execpolicy/src/lib.rs` so `ExecPolicyEngine` uses `BashArityDict` for trusted-prefix allow matching

## Behaviour

```toml
# config.toml
auto_allow = ["git status"]
```

| Command | Before | After |
|---|---|---|
| `git status` | allowed | allowed |
| `git status -s` | denied (no match) | **allowed** |
| `git status --porcelain` | denied (no match) | **allowed** |
| `git push origin main` | allowed (wrong prefix match) | denied (correct) |

Existing exact-match allow rules and wildcard rules (`cargo *`) continue to work unchanged.

## Test plan

- [x] `cargo +nightly check` passes — 0 errors, 0 warnings
- [x] `BashArityDict` unit tests in `bash_arity.rs` cover `classify()` and `allow_rule_matches()` for all major tool families
- [x] `prefix_allow_matches` doctest examples cover git/cargo/npm/make
- [x] New `rules.rs` tests: `test_prefix_rule_allows_git_status_with_flags` and `test_prefix_rule_allows_cargo_check_variants`
- [x] Existing `command_safety.rs` `classify_command` tests and `auto_allow_*` integration tests all still pass
- [x] Dictionary covers ≥ 30 common commands (enforced by `dict_covers_at_least_30_commands` test)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #693, opened 2026-05-05
fix(execpolicy): correct resolve_prefixes docstring (follow-up to #653)

merchloubna70-dot:
Fixes misleading docstring identified by @Hmbown in review of #653. The implementation uses insertion-order first-match, not longest-prefix priority shadowing. Deny-first behavior is still correct and safe. See #653 review comment.


Hmbown:
Thanks for the docstring polish! Closing without merge: PR #655 (bash arity dictionary) replaced the old prefix-matching code in `resolve_prefixes` with arity-aware matching, so the docstring this PR was clarifying no longer describes the actual code path. The accurate inline comments for the new arity logic landed alongside #655 itself, so this fix is moot. Appreciate the careful read.

PR #694, opened 2026-05-05
fix(subagents): release resident-file lease on terminal state (follow-up to #660)

merchloubna70-dot:
Fixes lease-never-released bug found by @Hmbown in review of #660.

- Extracts lease table into `resident_lease_table()` helper (process-scoped OnceLock)
- Adds public `release_resident_lease(file_path)` for the sub-agent manager to call on terminal state transitions
- Cargo check: 0 errors, 1 pre-existing dead_code warning (release_resident_lease is public API, wiring into manager lifecycle is a follow-up)

See #660 review comment.

### DIFF
--- a/crates/tui/src/prompts/base.md
+++ b/crates/tui/src/prompts/base.md
@@ -1,5 +1,9 @@
 You are DeepSeek TUI. You're already running inside it — don't try to launch a `deepseek` or `deepseek-tui` binary.
 
+## Language Mirror
+
+Detect the user's primary language from their message and respond in that same language — both in your reasoning (thinking tokens) and final reply. Do not default to English for non-English inputs.
+
 ## Preamble Rhythm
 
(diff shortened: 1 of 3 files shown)


Hmbown:
Thanks for the follow-up! Closing without merge — the original lease-release work in #660 already releases by `agent_id` at all three terminal-state transitions (cancel / fail / complete), so a release-by-`file_path` API isn't quite what was needed. The actual remaining bug was that the lease was stamped as `"pending"` at spawn time (because the agent id is only assigned by the manager later), so `release_resident_leases_for(agent_id)` could never find the entries. Fixed in https://github.com/Hmbown/DeepSeek-TUI/commit/3fedb980c by replacing the placeholder with the real id immediately after spawn returns. Good instinct that the original fix was incomplete.

PR #700, opened 2026-05-05
fix(snapshot): write info/exclude on repo init to skip large dirs (closes #697)

### LINKED ISSUE #697: [Performance] Slow initial response (~10s) when launching in directories with many files/subdirectories
Environment

OS: Linux
Installation method:npm
Package version:deepseek-tui@0.8.11
Description
I noticed a significant performance difference when startingdeepseek-tui in directories of different sizes:

Small directory (single file): Sending"hi" returns a response in ~1 second, and the right-sidetasks panel populates immediately.
Large directory (~10+ subdirectories, hundreds of files): Sending"hi" causes the bottom status bar to showWorking... for about 10 seconds. The right-sidetasks panel remains hidden/unpopulated during this time, and the response is delayed.
Expected Behavior
Initial response time and UI behavior should remain consistent regardless of directory size, or at least the first interaction should not be blocked by background initialization.

Actual Behavior
The first command in a large directory hangs onWorking... for ~10 seconds before responding. Thetasks panel does not appear until after the delay.

Hypothesis
I suspect the first command triggers a synchronous file scan or project initialization/indexing process, which blocks the main thread and delays the initial response.

The slowdown only affects the first command after launch. Subsequent commands respond normally.
No errors or stack traces are thrown; it simply waits on theWorking... state.
Thank you for your great work on this tool! Looking forward to your insights.

merchloubna70-dot:
## Root cause

On first launch in a workspace with large build-artifact directories
(`node_modules`, Rust `target/`, `.next`, etc.), the snapshot system's
`git add -A` has to walk and hash every file — including those large
trees. With no `.gitignore` or an incomplete one, this takes **10+
seconds**, blocking the first response.

The snapshot runs inside `spawn_blocking` so the async runtime is not
stalled, but on many systems the blocking-thread-pool I/O pressure is
enough to delay the API client's first byte.

## Fix

Write a baseline `info/exclude` file into the snapshot git repo during
`open_or_init`. This is git's per-repo exclude mechanism (equivalent to
a repo-local `.gitignore` that doesn't live in the work tree). The user's
own workspace `.gitignore` continues to be respected on top.

Directories excluded by default: `node_modules`, `target`, `.build`,
`dist`, `build`, `__pycache__`, `.next`, `.nuxt`, `vendor`, `venv`,
`.venv`, `.tox`, `.cache`.

## Testing

Before: `git add -A` on a 50k-file `node_modules` workspace — ~12s.
After: same workspace with the exclude in place — ~0.2s (only source
files are indexed).

### DIFF
--- a/crates/tui/src/snapshot/repo.rs
+++ b/crates/tui/src/snapshot/repo.rs
@@ -104,6 +104,35 @@ impl SnapshotRepo {
             let _ = run_git(&git_dir, &work_tree, &["config", "gc.auto", "0"]);
             // Ignore CRLF rewriting — we want byte-for-byte fidelity.
             let _ = run_git(&git_dir, &work_tree, &["config", "core.autocrlf", "false"]);
+
+            // Write an exclude file that skips well-known large build-artifact
+            // and dependency directories. Without this, `git add -A` traverses
+            // the entire workspace on the first snapshot, which can take 10+
+            // seconds in directories with node_modules, Rust target/, etc.
+            // (#697). The user's own .gitignore is still honoured; this adds a
+            // baseline that applies even when no .gitignore is present.
+            let exclude_dir = git_dir.join("info");
+            let _ = std::fs::create_dir_all(&exclude_dir);
+            let _ = std::fs::write(
+                exclude_dir.join("exclude"),
+                concat!(
+                    "# deepseek-tui snapshot excludes — auto-generated, do not edit\n",
+                    "# These directories are excluded to keep `git add -A` fast (#697).\n",
+                    "/node_modules\n",
+                    "/target\n",
+                    "/.build\n",
+                    "/dist\n",
+                    "/build\n",
+                    "/__pycache__\n",
+                    "/.next\n",
+                    "/.nuxt\n",
+                    "/vendor\n",
+                    "/venv\n",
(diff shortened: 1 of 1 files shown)


Hmbown:
Closing as **superseded by current `main`**. Thanks for the report and the patch — both the diagnosis and the direction were right.

`main` now writes a richer baseline `info/exclude` via `write_builtin_excludes(&git_dir)` in `crates/tui/src/snapshot/repo.rs::open_or_init`, landed by #854 (commit `f932c4b9d`). The `BUILTIN_EXCLUDES` constant covers the same well-known build-tool dirs you proposed (`node_modules/`, `target/`, `dist/`, `.next/`, `.nuxt/`, `__pycache__/`, `vendor/`, `venv/`, `.venv/`, `.tox/`, `.cache/`, etc.) and adds Rust/JS/Python tool caches plus binary artifacts on top.

Verified on `origin/main`:
- `crates/tui/src/snapshot/repo.rs` — `BUILTIN_EXCLUDES` constant lists the same patterns.
- `open_or_init` calls `write_builtin_excludes(&git_dir)?;` on every open (not just on first init), so existing snapshot repos pick up the latest excludes when v0.8.15 launches.

Closing #700 as integrated/superseded.

PR #706, opened 2026-05-05
Security fix: P0 shell injection and P1 path traversal vulnerabilities

dumbjack:


dumbjack:
## Response to Code Review

Thank you for the detailed feedback. I've updated the PR to address your concerns:

### P1 - Path Traversal Fix

You mentioned that the existing `normalize_path` and `canonicalize` logic "already correctly handles" path traversal. However, my analysis shows that **the vulnerability exists in the fallback logic for non-existent paths**.

#### The Vulnerability

When a path doesn't exist, the code (spec.rs lines 306-338):
1. Finds the deepest existing ancestor directory
2. Rebuilds the path from suffix parts

**The bug**: When suffix contains `..` components, they are pushed as literal strings rather than being resolved as parent directories:

```rust
// Original buggy code
for part in suffix_parts.into_iter().rev() {
    canonical.push(part);  // pushes ".." as a literal string!
}
```

#### Proof of Vulnerability

I simulated the exact algorithm in Python and confirmed:

| Path | Expected | Actual Result |
|------|----------|--------------|
| `/tmp/project/nonexistent/../../../etc/passwd` | Should be blocked | **Was NOT blocked** |
| `/tmp/project/a/../../root` | Should be blocked | **Was NOT blocked** |

The reconstructed path `/private/tmp/project/nonexistent/../../../etc/passwd` incorrectly passes the workspace check because `..` was pushed as a literal string rather than resolved.

#### My Fix

```rust
for part in suffix_parts.into_iter().rev() {
    let part_str = part.to_string_lossy();
    if part_str == ".." {
        if let Some(parent) = canonical.parent() {
            canonical = parent.to_path_buf();
            continue;
        }
    }
    canonical.push(part);
}
```

Now `..` is properly resolved as a parent directory.

---

### P0 - Shell Command Injection Fix

I've improved the eval/exec detection using `extract_primary_command()` as you suggested. Now it:
- Checks if the PRIMARY command (first token) is `eval` or `exec`
- Avoids false positives like `echo "eval is a builtin"`
- Correctly detects `exec>file` and other variants

---

### Summary

The P1 vulnerability is **real and exploitable**. The existing code only handles path traversal correctly for **existing** paths. For non-existent paths (e.g., files to be created), the fallback logic has a path traversal bug.

Please reconsider the P1 fix. I'm happy to provide more details or create a test case.

Hmbown:
Integrated as #918 and merged into `main`. Thanks for the catch on the null-byte hole.

The null-byte rejection landed as `fix(command-safety): reject null bytes in shell commands` with regression tests:

- `test_null_byte_is_blocked` — `ls\0 -la` and `echo hello\0world` are now Dangerous
- `test_eval_substring_is_not_misclassified` — pins that `cargo run -- eval` and `python evaluator.py` stay safe

The broader `command.contains("eval") || command.contains("exec ")` guard from this PR was **not** integrated — it false-positives on routine workflows: the offline `cargo run --bin deepseek -- eval` harness (used by CI), any binary or script whose name contains `eval` (`evaluator.py`, `primeval-checker`), and various legitimate `cargo run -- exec ...` invocations. A real fix here would need to inspect the parsed primary command (after argv splitting), not a substring of the raw string. The null-byte slice is the safe immediate win.

Closing this one as integrated for the null-byte fix. If you want to take another swing at the eval/exec detection with proper argv parsing, please open a fresh PR against `main` with new tests covering the false-positive cases above.

PR #728, opened 2026-05-05
fix(npm): --version fallback, preuninstall cleanup, skill-creator frontmatter

GK012:
## Summary

Fixes 3 issues in the npm wrapper layer and 1 in the bundled skill-creator asset:

### 1. --version fallback (scripts/run.js)
- Intercept --version/-v/-V before binary download/launch
- Works even when binary is missing or corrupted

### 2. Preuninstall cleanup (scripts/uninstall.js)
- Recursively removes bin/downloads (~46MB) on uninstall
- Errors non-fatal to avoid blocking npm uninstall

### 3. Skill-creator SKILL.md frontmatter fix
- HTML comment before frontmatter delimiter caused parser rejection
- Moved comment after the closing frontmatter delimiter

## Notes
- All changes in npm wrapper layer only; no Rust binary changes needed

### DIFF
--- a/crates/tui/assets/skills/skill-creator/SKILL.md
+++ b/crates/tui/assets/skills/skill-creator/SKILL.md
@@ -1,3 +1,7 @@
+<!--
+This SKILL.md is ported from OpenAI's codex repo (MIT-licensed).
+Source: https://github.com/openai/codex/blob/main/codex-rs/skills/src/assets/samples/skill-creator/SKILL.md
+-->
 ---
 name: skill-creator
 description: Guide for creating effective skills. This skill should be used when users want to create a new skill (or update an existing skill) that extends deepseek's capabilities with specialized knowledge, workflows, or tool integrations.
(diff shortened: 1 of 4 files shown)


Hmbown:
Thank you for working on the npm wrapper path. The three goals here are useful, especially `--version` working even when the downloaded binary is missing.

I can't merge this revision as-is because a few files appear to have been rewritten with encoding/string-template damage:

- `npm/deepseek-tui/scripts/run.js` has broken template literals, for example `console.log(${binaryName} (npm wrapper) v);` and truncated variable names.
- `npm/deepseek-tui/scripts/uninstall.js` has the same issue in its `console.log` / `console.error` calls.
- `crates/tui/assets/skills/skill-creator/SKILL.md` has mojibake in the tree diagram.
- `package.json` was mechanically reformatted and lost the trailing newline, which makes the real npm change harder to review.

A smaller follow-up would be very mergeable: keep `package.json` formatting unchanged except for the `preuninstall` line, fix `run.js` with normal JS template strings, and leave unrelated Markdown alone unless the frontmatter comment truly needs to move. Happy to re-review that shape.

Hmbown:
Thanks for working on the npm install path. The goals here are useful, especially `--version` working even when the downloaded binary is unavailable.

I cannot merge this branch as-is because the current diff appears to have some accidental corruption:

- `npm/deepseek-tui/scripts/run.js` contains invalid JavaScript around the `console.log(...)` calls. It looks like template literals were stripped or escaped incorrectly.
- `npm/deepseek-tui/scripts/uninstall.js` has the same issue in its log/error lines.
- `crates/tui/assets/skills/skill-creator/SKILL.md` has mojibake in the tree example. Also, plain Markdown skill parsing has now landed separately in #869, so this part may no longer be needed.
- `npm/deepseek-tui/package.json` was mechanically reformatted and still shows the old `0.8.12` version, which makes the functional change harder to review.

The most mergeable slice would be just the npm `--version` fallback in `scripts/run.js`, keeping the existing formatting, plus a tiny Node smoke test or manual test note like `node npm/deepseek-tui/bin/deepseek.js --version` with the binary directory moved aside. The uninstall cleanup can be a second PR after that.

GK012:
Closing this PR in favor of #885, which contains only the  fallback change with clean formatting. The uninstall cleanup can be a follow-up PR.

Issue #777, opened 2026-05-06
snapshot: write built-in exclusions to info/exclude to prevent hangs on large workspaces


yizheruchu:
## Problem

The side-git snapshot mechanism runs `git --work-tree=<workspace> add -A` on every turn. When the workspace contains large cache/build directories (`node_modules/`, `.cargo/`, `.cache/`, `.rustup/`, `.npm/`, `.local/`, etc.), `git add -A` scans the entire tree and hangs — sometimes indefinitely — because there are no built-in exclusions.

This is most commonly triggered when `deepseek` is launched from `$HOME`, where cached packages and SDK toolchains easily amount to 40+ GB across directories like:

| Directory | Size |
|-----------|------|
| `.cache/` | 15 GB |
| `.local/` | 9.5 GB |
| `.npm/` | 6.5 GB |
| `.config/` | 3.9 GB |
| `.rustup/` | 1.4 GB |
| `.cargo/` | 579 MB |
| *others* | ~5 GB |

But the same problem can hit any large workspace (monorepo roots, data-science directories with datasets, build output trees, etc.). It is not specific to `$HOME`.

Workaround: placing a `.gitignore` at the workspace root with appropriate exclusions fixes the hang — confirmed by the existing `snapshot_respects_workspace_gitignore` test in `repo.rs`.

## Why Claude Code is not affected

Claude Code snapshots individual shell commands and their outputs (`~/.claude/shell-snapshots/`). It never runs a full-workspace `git add -A`, so workspace size has no impact.

## Proposed fix

Use git’s **`$GIT_DIR/info/exclude`** —— the per-repository exclude file that is independent of the worktree. It is read *before* worktree `.gitignore` files, so built-in rules never clash with user rules.

### Why `info/exclude` instead of alternatives

| Approach | Drawback |
|----------|----------|
| Append `:!:node_modules :!:.cache ...` to the `git add` command | Long, brittle command line |
| Write a `.gitignore` into the user's workspace | Pollutes user directory; potential conflicts |
| **Write into snapshot repo's `info/exclude`** | Zero user-visible side effects; git reads it natively; no command changes |

### Implementation sketch

Single-file change in `crates/tui/src/snapshot/repo.rs`:

**Step 1 — Define a built-in exclusion list**

```rust
const BUILTIN_EXCLUDES: &str = "\
# DeepSeek TUI — built-in snapshot exclusions
node_modules/
target/
.cargo/
.rustup/
.npm/
.bun/
[...]

PR #804, opened 2026-05-06
fix(snapshot): skip full snapshots for dangerous workspaces

### LINKED ISSUE #793: Snapshotting $HOME can create huge snapshots and block visible responses
## Summary

When DeepSeek TUI is launched from a large home directory, it appears to treat the whole home directory as the workspace and starts creating a git-based snapshot under `~/.deepseek/snapshots`.

In my case this spawned a long-running process similar to:

```bash
git --git-dir ~/.deepseek/snapshots/<id>/<id>/.git --work-tree /Users/<user> add -A
```

The process consumed near 100% CPU, grew `~/.deepseek/snapshots` to tens of GB, and made the TUI look like a prompt was sent but no assistant response was visible for a long time.

## Environment

- Package: `deepseek-tui`
- Installed binary version: `0.8.12`
- npm latest observed during debugging: `0.8.14`
- OS: macOS arm64
- Provider: DeepSeek official API
- Model: `deepseek-v4-pro`
- Config path: `~/.deepseek/config.toml`

## Steps to reproduce

1. Start DeepSeek TUI from a large home directory, e.g.:

   ```bash
   cd ~
   deepseek
   ```

2. Send a prompt in Agent mode.
3. Watch processes and disk usage:

   ```bash
   ps aux | grep deepseek
   ps aux | grep '.deepseek/snapshots'
   du -sh ~/.deepseek/snapshots
   ```

## Observed behavior

- A git snapshot process runs against the full home directory.
- CPU usage stays high while snapshotting.
- `~/.deepseek/snapshots` can grow to very large sizes. Mine reached roughly `32G` before cleanup.
- The TUI appears stalled or unresponsive, even though the API key and API connectivity are valid.
- A checkpoint can temporarily contain only the user message, making the turn look interrupted or incomplete.

## Expected behavior

DeepSeek TUI should avoid silently snapshotting very large or unsafe default workspaces such as:

- `$HOME`
- `/Users/<user>`
- `/`
- `~/Desktop`
- `~/Downloads`

At minimum, it should warn before snapshotting a large directory and ask for confirmation.

## Workaround

I stopped the snapshot process, removed the large snapshot directory, and disabled snapshots:

```bash
rm -rf ~/.deepseek/snapshots
mkdir -p ~/.deepseek/snapshots
```

```toml
[snapshots]
enabled = false
max_age_days = 1
```

I also now start DeepSeek TUI from a small project directory instead of `$HOME`:

```bash
cd /path/to/project
deepseek
```

## Suggested improvements

1. Add a guardrail that refuses or warns before snapshotting `$HOME` or `/`.
2. Add a size/file-count preflight before snapshot creation.
3. Show a visible status message while snapshotting, e.g. `Creating workspace snapshot...`.
4. Add a default snapshot exclusion list for large/common user directories.
5. Consider snapshotting only inside detected project roots, e.g. directories with `.git`, `Cargo.toml`, `package.json`, etc.
6. If snapshot creation is slow, do not make the model response appear silently blocked.
7. Make `deepseek doctor` or `deepseek-tui setup --status` report snapshot settings and current snapshot disk usage.

## Additional note

The TUI itself appears to be Rust-based from the installed package metadata and binary strings (`crates/tui/src/...`). A model response once claimed it was running in a C++ project, which was likely a hallucination caused by ambiguous workspace context when launched from `$HOME`. This is secondary; the main issue is the snapshot behavior.

ccomma:
## Summary

- skip side-git workspace snapshots when the workspace is the filesystem root, the user's home directory, or common home collection directories such as Desktop, Documents, Downloads, Library, Movies, Music, and Pictures
- keep normal project directories under home, such as ~/code/project, eligible for snapshots
- add regression coverage for rejected home/home-collection workspaces and allowed project workspaces

## Why

Running DeepSeek-TUI from the home directory can block before the API request because snapshot initialization eventually runs a full `git add -A` against the selected workspace. When the workspace is `/Users/ccomma`, that can scan a very large personal directory tree and make the UI appear stuck at `working...`.

This change keeps the existing side-git snapshot behavior for normal project directories, but refuses high-risk workspace roots where a full-workspace snapshot is likely to be expensive or surprising.

Fixes #793
Related to #777

## Verification

- `cargo test -p deepseek-tui snapshot::repo::tests::open_or_init_rejects_home_directory_workspace`
- `cargo test -p deepseek-tui snapshot::repo::tests`
- `rustfmt --edition 2024 --check crates/tui/src/snapshot/repo.rs`
- `git diff --check`

### DIFF
--- a/crates/tui/src/snapshot/repo.rs
+++ b/crates/tui/src/snapshot/repo.rs
@@ -61,12 +61,15 @@ impl SnapshotRepo {
         let work_tree = workspace
             .canonicalize()
             .unwrap_or_else(|_| workspace.to_path_buf());
-
-        // Refuse to snapshot the user's home directory: `git add -A` on $HOME
-        // can consume unbounded disk/CPU and effectively DoS the TUI (#793).
-        if is_home_directory(&work_tree, dirs::home_dir().as_deref()) {
-            return Err(io_other(
-                "refusing to snapshot home directory - start deepseek from a project directory instead",
+        if let Some(reason) =
+            unsafe_workspace_snapshot_reason(&work_tree, dirs::home_dir().as_deref())
+        {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!(
+                    "workspace snapshots are disabled for {reason}: {}",
+                    work_tree.display()
+                ),
             ));
         }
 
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for the careful fix here. I rebased this onto current main, kept the guard focused on genuinely dangerous snapshot roots, and adjusted the tests so they stay portable on Windows. CI is green now, so this is merged for v0.8.15.

Issue #790, opened 2026-05-06
Improve i18n coverage for commands, modals, and widgets


ghost:
## Summary

Follow-up to #740 / #781. After adding `zh-Hant` support, the shipped locale registry covers core UI chrome, but many user-visible TUI strings still remain hardcoded in English. This issue tracks expanding localization coverage for commands, modals, widgets, approval dialogs, and status picker UI.

This should improve all shipped locales (`en`, `ja`, `zh-Hans`, `zh-Hant`, `pt-BR`), not just Traditional Chinese.

## Current Gaps

Representative areas that still render English under non-English locales:

- Composer hints, including slash menu and submit/queue hints
- Config modal labels, scopes, edit footer, and hints
- Status line picker title, actions, labels, and hints
- Pending input preview headers and removable hints
- Approval and sandbox elevation dialogs
- Sub-agent modal status labels and detail prefixes
- Common command output for `/queue`, `/task`, `/trust`, `/restore`, `/config`, `/statusline`, `/lsp`, and `/logout`
- Tool and agent card labels

## Scope

- Move high-frequency hardcoded UI chrome into `localization::MessageId`
- Add translations for every shipped locale: `en`, `ja`, `zh-Hans`, `zh-Hant`, `pt-BR`
- Prioritize short, visible UI chrome before lower-frequency generated/exported content
- Keep command names, keybindings, paths, env vars, model IDs, and tool names unchanged where they are technical tokens

## Non-goals

- Model output language
- System prompts or model prompts
- Tool output or shell command output
- User-provided content
- Provider protocol/internal keys
- Full localization of generated `AGENTS.md` or exported `/share` HTML in the first pass

### Phase 1: high-frequency TUI chrome

- Composer slash menu hints
- Composer submit/queue hints
- Config modal section/scope/edit labels and footer
- Status line picker title/actions/prompt
- Pending input preview section headers

### Phase 2: approval and sandbox elevation

- Approval field labels (`Type`, `About`, `Impact`, `Params`)
- Risk/category labels
- Approval option labels and destructive confirmation footer
- Sandbox elevation title, option labels, and descriptions

### Phase 3: common command output

- `CommandResult::error` prefix
- `/queue`, `/task`, `/trust`, `/restore`, `/config`, `/statusline`, `/lsp`, `/logout`

### Phase 4: lower-frequency surfaces

- Tool family labels
- Agent lifecycle labels
- Fanout summaries
- Sub-agent modal detail prefixes
[...]

PR #812, opened 2026-05-06
feat(i18n): localize composer, config modal, status picker, pending input preview, approval, agent cards, and command messages

gordonlu:
localize all UI text in composer, config modal, status picker, pending input preview, approval dialogs, agent cards, tool cards, sidebar, and command messages.

This is a single batch covering all i18n work across the TUI surface. Each component uses MessageId variants resolved through a central localization::tr() call, with locale passed via App.ui_locale.

Scope: 34 files, +1748/-469 (i18n only — no logic changes).
Compatibility: zh-Hant translations will be added once #781 lands.


ghost:
Thanks for working on #790. One coordination note: #781 adds `zh-Hant` as a shipped locale and is currently mergeable. If #781 lands first, this PR will need `zh-Hant` translations for all newly introduced `MessageId` variants; otherwise the localization completeness tests are likely to fail.

Also, this PR currently looks broader than the Phase 1 description. The body describes composer, config modal, status picker, and pending input preview, but the current diff changes 34 files with +1748/-469 lines and appears to include later-phase work as well. Could you either split the later phases out, or update the PR description/title so reviewers can evaluate the full intended scope?

gordonlu:
Good point on both fronts. I will update the PR title and description to reflect the full current scope so reviewers can evaluate it as a single batch.

Re zh-Hant: I see #781 adds Traditional Chinese support. I am happy to either:

1. Wait for #781 to land, then rebase and add zh-Hant translations for the new MessageId variants, or

2. Add zh-Hant translations now alongside the existing strings if you would like everything in one pass.

Let me know which works better for your review flow.

ghost:
Thanks, that sounds good. I would prefer option 1: let #781 land first, then rebase this PR and add `zh-Hant` translations for the new `MessageId` variants.

I noticed #781 has become conflicting again after upstream/main moved, so I will update that PR first. Once it is mergeable and/or merged, this PR can use `zh-Hant` as part of the shipped locale set without duplicating the locale-registration work.

Hmbown:

PR #813, opened 2026-05-06
feat: inject structured Environment block into system prompt

lloydzhou:
## Summary

Add a `## Environment` section (lang / platform / shell / pwd) to the system prompt, giving the model deterministic knowledge of the user's runtime context instead of relying on heuristic language detection.

- New `render_environment_block()` in `prompts.rs` emits lang, platform, shell, and pwd.
- `lang` is resolved in priority order: user-configured locale from `Settings` → system env vars (`LC_ALL` / `LC_MESSAGES` / `LANG`) → fallback `en`.
- `## Language` in `base.md` updated to reference the Environment section's `lang` field instead of instructing the model to "Detect the language the user writes in".
- Block is inserted at step 2.25 (after mode prompt + project context, before instructions) so it stays in the workspace-static KV prefix cache layer.

## Testing

- [x] `cargo test --all-features` (prompts: 38/38 passed)

## Checklist

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -143,6 +143,10 @@ pub struct EngineConfig {
     /// consulted when `memory_enabled` is `true`.
     pub memory_path: PathBuf,
     pub goal_objective: Option<String>,
+    /// Resolved BCP-47 locale tag (e.g. `"en"`, `"zh-Hans"`, `"ja"`)
+    /// for the `## Environment` block in the system prompt. The caller
+    /// resolves this from `Settings` once; no file I/O in the engine.
+    pub locale_tag: String,
     /// When true, force `tool_choice: "required"` so the model always calls
     /// a tool on every turn step (V4 strict tool-following mode).
     pub strict_tool_mode: bool,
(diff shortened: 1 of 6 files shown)


lloydzhou:
close #807

Hmbown:
Integrated as #922 and merged into `main`. Thanks!

The Environment block (lang/platform/shell/pwd) shipped as `feat(prompts): inject deterministic Environment block in system prompt`, with `base.md`'s `## Language` directive pointing at the `lang` field. The refactor in your third commit (decouple `render_environment_block` from `Settings::load`) was applied as-is — no I/O happens inside the prompt builder.

Two small differences from the original:
- The fallback in `## Language` ("when `lang` is missing or ambiguous, fall back to detecting the user's writing") was kept, so locale-less callers don't lose the prior behavior.
- `EngineConfig::locale_tag` is resolved once at engine construction in `tui::ui::build_engine_config`, `runtime_threads::spawn_engine`, and `run_exec_agent`, threaded through `PromptSessionContext.locale_tag`.

Cache prefix stays stable: the block lives at step 2.25 (workspace-static layer), and all four inputs are session-stable for a given run, so KV-prefix hits are preserved turn over turn.

Closing as integrated.

PR #820, opened 2026-05-06
feat(tui): add notification_condition config with assistant text in O…

zerx-lab:
…SC 9 notifications

- Add otification_condition field to TuiConfig (always / never)
- When set to lways, force OSC 9 notification with no threshold
- OSC 9 notification body now includes the assistant's reply text (truncated to 360 chars)
- Fall back to latest assistant message from api_messages when streaming text is empty
- Deprecate [notifications] section in config.example.toml in favor of tui.notification_condition
- Add tests for notification_settings, notification text building, and fallback logic

## Testing

## Checklist


zerx-lab:
@codex review

Hmbown:
Integrated as #920 and merged into `main`. Thanks!

`feat(tui): notification_condition override + assistant text in OSC 9 body` shipped with all your tests plus three more for the helpers. Specifically:

- `[tui].notification_condition = "always" | "never"` — the high-level override you proposed.
- OSC 9 body now carries the assistant's reply text (sanitized + truncated to 360 chars), with a fallback to the latest assistant message in `api_messages` for tool-only turns.
- The summary line (elapsed + cost) is appended on a new line when `[notifications].include_summary = true` — keeping the per-turn-cost UX from main intact.

Two intentional differences from the original PR:
- The `[notifications]` section in `config.example.toml` was preserved (with `notification_condition` documented as an opt-in override) instead of being deleted, so configs that already set `[notifications].method` etc. keep working unchanged.
- The unrelated `#[allow(dead_code)]` on `schema_migration::registry` was dropped from this slice — that one belongs in its own change.

Drive-by: removed the now-unused `Method::from_str` parser since the new code match-arms over the typed `NotificationMethod` enum directly.

Closing as integrated.

PR #822, opened 2026-05-06
fix(web_search): complete HTML entity decoding with numeric character references

angziii:
## Summary
- The `decode_html_entities` function only handled 7 named HTML entities and completely missed decimal (`&#NN;`) and hex (`&#xHH;`) numeric character references, which appear regularly in search result snippets from DuckDuckGo and Bing.
- Replaced the hard-coded `.replace()` chain with a regex-based decoder that handles named entities, decimal references, and hex references.
- Unknown entities are passed through unchanged rather than silently dropped.

## Test plan
- [x] All 5 new unit tests pass covering named entities, decimal refs, hex refs, mixed content, and passthrough of unknowns
- [x] Existing search tests continue to pass

### DIFF
--- a/crates/tui/src/tools/web_search.rs
+++ b/crates/tui/src/tools/web_search.rs
@@ -479,8 +479,7 @@ fn decode_html_entities(text: &str) -> String {
 
     static ENTITY_RE: OnceLock<Regex> = OnceLock::new();
     let re = ENTITY_RE.get_or_init(|| {
-        Regex::new(r"&(?:#(\d+)|#x([0-9A-Fa-f]+)|([a-zA-Z]+));")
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for the focused fix. I applied the one rustfmt cleanup, rebased this onto current main, and verified locally with `cargo fmt --all -- --check`, `git diff --check origin/main...HEAD`, `cargo test -p deepseek-tui --bin deepseek-tui tools::web_search::tests`, and `cargo clippy -p deepseek-tui --all-targets --all-features -- -D warnings`. The normal CI workflow did not start on this fork PR, but the local gates and GitGuardian check are clean, so I am merging this for v0.8.15.

PR #839, opened 2026-05-06
fix(rlm): enforce child model for RLM fanout

### LINKED ISSUE #829: Bug: `rlm_query` incorrectly uses `deepseek-v4-pro` for child tasks instead of `v4-flash` (Cost/Billing Issue)
### 📝 Issue Description

**Describe the bug**
According to the project documentation, `rlm_query` is designed to fan out 1-16 cheap `deepseek-v4-flash` children for parallel reasoning and batch analysis. However, I recently noticed on my DeepSeek API billing dashboard that all the concurrent subtasks triggered by `rlm_query` are actually consuming `deepseek-v4-pro` tokens, leading to a massive and unexpected spike in API costs.

**To Reproduce**
1. Start `deepseek-tui` and ensure the main session is using `deepseek-v4-pro` (either explicitly via `/model deepseek-v4-pro` or via auto-routing).
2. Ask a complex question or provide a task that triggers the `rlm_query` tool (e.g., parallel file analysis).
3. Wait for the parallel subtasks to complete.
4. Log into the official DeepSeek Platform and check the API Usage dashboard.
5. Notice that the token usage spikes under the `deepseek-v4-pro` model, with zero or minimal usage under `deepseek-v4-flash`.

**Expected behavior**
Regardless of the model being used in the parent session, the subtasks dispatched by `rlm_query` should strictly enforce and override the model parameter to `deepseek-v4-flash` to adhere to the "cheap children" design.

**Actual behavior**
The child tasks seem to inherit the parent session's model configuration, sending `deepseek-v4-pro` in the API payload and billing accordingly.

**Possible Root Cause**
Context/Client Leak. Since the RLM dispatcher reuses the existing API client for performance, it likely passes down the parent's configuration state. The HTTP request payload builder for the child tasks probably forgets to explicitly hardcode or overwrite the `"model": "deepseek-v4-flash"` field before firing the async requests.

**Environment details:**
- **OS:** macOS M2
- **DeepSeek-TUI Version:** v0.8.13
- **Installation method:** binaries

**Additional context**
This is a critical billing issue for users doing heavy batch analysis. A quick hotfix to hardcode the model string in the `rlm_query` dispatcher would be highly appreciated!

Image

vincentredrock:
## Summary

Fixes #829 by enforcing the configured RLM child model for all REPL-driven child calls.

RLM child helpers can fan out many background LLM requests. Before this change, the bridge accepted the `model` field emitted by REPL helper RPCs, which allowed root-generated Python code to silently upgrade child calls away from the intended low-cost child model.

## Changes

- Route `llm_query`, `llm_query_batched`, `rlm_query`, and `rlm_query_batched` through a single enforced child model policy in the RLM bridge.
- Ignore REPL RPC `model` overrides so child calls cannot silently use `deepseek-v4-pro` when the parent/root model is Pro.
- Add a regression test covering a `deepseek-v4-pro` override attempt.
- Update RLM prompt/runtime helper docs to tell the root model not to pass `model=` because the driver enforces child model selection.

## Validation

- `cargo fmt --all`
- `cargo test -p deepseek-tui rlm::bridge`
- `cargo test -p deepseek-tui rlm`
- `cargo build`

### DIFF
--- a/crates/tui/src/repl/runtime.rs
+++ b/crates/tui/src/repl/runtime.rs
@@ -52,7 +52,8 @@ pub struct ReplRound {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum RpcRequest {
-    /// `llm_query(prompt, model=None, max_tokens=None, system=None)`
+    /// `llm_query(prompt, model=None, max_tokens=None, system=None)`.
+    /// The bridge enforces the configured child model and ignores `model`.
     Llm {
         prompt: String,
         #[serde(default)]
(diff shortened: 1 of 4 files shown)


ghost:
Closing as duplicate: #832 has been merged and fixes #829.

Issue #864, opened 2026-05-06
输出结果显示不全


YangLuLu107:
## Description
windows11 
deepseek 0.8.14

<img width="2560" height="1368" alt="Image" src="https://github.com/user-attachments/assets/125647f6-5ffd-4986-a586-39e02e27e64e" />

yy35959199:
同样出现了这个问题，而且不知道怎么关闭或隐藏侧面的plan/tasks等界面

Hmbown:
Please retest this on v0.8.17 once the release binary is available. v0.8.17 has TUI resize/layout fixes and a PTY harness for frame-level checks, but I do not want to close this from inference alone because the screenshot looks like a real clipping/wrapping bug.

If it still reproduces, the most useful details are: terminal app, shell, window size, `deepseek --version`, whether the missing content is off the right edge or below the scroll area, and whether toggling the sidebar changes it.

yy35959199:
> Please retest this on v0.8.17 once the release binary is available. v0.8.17 has TUI resize/layout fixes and a PTY harness for frame-level checks, but I do not want to close this from inference alone because the screenshot looks like a real clipping/wrapping bug.请在 v0.8.17 的发布二进制文件可用后重新测试此问题。v0.8.17 包含了 TUI 调整大小/布局的修复以及用于帧级检查的 PTY 框架，但我不想仅凭推断就关闭此问题，因为截图看起来确实存在裁剪/换行错误。
> 
> If it still reproduces, the most useful details are: terminal app, shell, window size, `deepseek --version`, whether the missing content is off the right edge or below the scroll area, and whether toggling the sidebar changes it.如果问题仍然复现，最有用的信息包括：终端应用、Shell、窗口大小、 `deepseek --version` 、缺失内容是在右侧边缘之外还是滚动区域下方，以及切换侧边栏是否会改变这一情况。

在更新后，仍然出现了同样的问题：
终端应用：Windows Terminal
Shell：PowerShell v7.6.0
deepseek (npm wrapper) v0.8.17
binary version: v0.8.17
缺失内容：右侧边缘之外
切换侧边栏：没有变化

After updating, the same issue still occurs:
Terminal application: Windows Terminal
Shell: PowerShell v7.6.0
deepseek (npm wrapper) v0.8.17
binary version: v0.8.17
Missing content: Beyond the right edge
Toggling sidebar: No change

Window size: Default
窗口大小：默认
[...]

PR #884, opened 2026-05-06
fix: exclude binary files from snapshots and enable cross-session memory

m8f1m8f1:
## Summary

Two bug fixes identified during real-world usage:

### 1. Snapshot bloat from binary files
The `BUILTIN_EXCLUDES` list in the snapshot subsystem only excluded build-tool directories and `*.pyc`. When the workspace was accidentally set to a broad directory (e.g. `C:\Users`), all binary files (`*.exe`, `*.dll`, `*.png`, etc.) were git-hashed into the side repo as loose objects — consuming ~3.4 GB.

**Fix**: Added 34 binary/generated file patterns (`*.exe`, `*.dll`, `*.so`, `*.png`, `*.zip`, `*.mp4`, etc.) to the built-in exclude list.

### 2. Cross-session memory loss
`rehydrate_latest_canonical_state` only looked up memory records for the current session ID. Starting a fresh session produced an empty `CanonicalState` — `goal`, `confirmed_facts`, etc. all defaulted to empty — so the model couldn't answer "what did we do in the previous conversation?"

**Fix**: Added `load_latest_across_sessions()` that scans all `*.jsonl` memory files and picks the newest record. `rehydrate_latest_canonical_state` now falls back to this cross-session search when the current session has no memory yet.

## Test plan
- [ ] Verify `cargo test` passes in the snapshot and capacity_memory modules
- [ ] Manual: create a session with meaningful activity, start a new session, confirm the canonical state is rehydrated with the previous session's goal and facts
- [ ] Manual: snapshot a workspace with binary files present, verify they are excluded from the side-git repo

## Files changed
- `crates/tui/src/snapshot/repo.rs` — binary file patterns in `BUILTIN_EXCLUDES`
- `crates/tui/src/core/capacity_memory.rs` — new `load_latest_across_sessions()` function
- `crates/tui/src/core/engine.rs` — import the new function
- `crates/tui/src/core/engine/capacity_flow.rs` — fallback to cross-session search in `rehydrate_latest_canonical_state`


Saurabhtbj1201 (approved):

Hmbown:
thank you! great find

Hmbown:
Thanks for both fixes — handling them separately:

**1. Snapshot binary excludes — superseded by current `main`.**

`crates/tui/src/snapshot/repo.rs` on `origin/main` already lists the binary patterns this PR adds: `*.exe`, `*.dll`, `*.so`, `*.dylib`, `*.wasm`, `*.o`, `*.obj`, `*.class`, `*.pdb`, `*.dSYM`, `*.zip`, `*.tar.gz`, `*.tgz`, `*.tar.bz2`, `*.tar.xz`, `*.7z`, `*.rar`, `*.iso`, `*.dmg`, `*.bin`, `*.mp4`, `*.mov`, `*.mkv`, `*.avi`, `*.webm`, `*.mp3`, `*.wav`, `*.flac`, `*.aac`. Landed by #854 (commit `f932c4b9d` — \"fix: seed snapshot built-in excludes\") plus #864's follow-up. Image extensions (`*.png`, `*.jpg`, `*.gif`, etc.) aren't yet in the constant — happy to take a tiny standalone PR adding those if you want to send one.

**2. Cross-session memory rehydration — deferring past v0.8.15.**

`rehydrate_latest_canonical_state` on `main` still only loads from the current session id. Adding a global \"scan all `.jsonl` memory files across every session\" fallback is a meaningful behavior change — a brand-new session would inherit goal/confirmed-facts from the most recently active session of any other workspace, including `goal_objective` text and \"confirmed facts\" from unrelated work. That's a privacy/scoping question (cross-workspace memory bleed) that the v0.8.15 release lane is explicitly trying to avoid.

The fix would be welcome with a per-workspace scope (only scan memory files whose canonical state's `workspace` matches the current one) and an opt-in flag — happy to look at that as a follow-up against `main`. Could you split the cross-session change into its own PR with that workspace scoping? Then we can land it after v0.8.15 ships.

Closing #884 — the binary-exclude part is already in main, and the cross-session piece needs the scoping work above before it can land.

Issue #239, opened 2026-04-30
v0.7.7: make sub-agents first-class background jobs


Hmbown:
## Product requirement

Sub-agents should behave like background shell jobs, not like foreground/blocking tool calls. The parent turn/UI should be able to move on while sub-agents continue running, with durable status, progress, completion notifications, and explicit collection tools when the parent wants results.

User clarification from the 2026-04-29 live swarm debugging:

- The fanout card started active, which is good.
- A worker later finished and the sidebar dropped from 5 running to 4 running, which is also progress.
- The missing piece is a background-job lifecycle: parent can continue working while sub-agents run, like background shells, without leaving the UI in contradictory active/pending states.

## Current bad shape

- Treating `agent_swarm` as a normal foreground tool can freeze/occupy the parent turn longer than needed.
- Treating it as fully detached can leave the parent/model without a clear next step, and the transcript/sidebar can drift.
- The UI has no single background-subagent job center equivalent to shell jobs: fanout cards, Agents sidebar, task status, and model-visible results are separate projections.

## Acceptance criteria

- `agent_spawn` and `agent_swarm` create durable background job records with ids, status, worker slots, start time, model, and objective.
- The parent UI/turn can continue after spawn without pretending the swarm is complete.
- While workers run, the transcript card and Agents sidebar update live without blocking composer input.
- Completion of any worker updates the background job record and visible dot/count state.
- Completion of all workers produces a clear collectable state: `swarm_result`/`agent_result` can retrieve final results, and the UI can surface a ready-to-synthesize notification.
- A follow-up user message should not accidentally duplicate an already-running swarm; it should see/reuse the existing background job state.
- Cancellation semantics are explicit: cancel parent turn does not silently orphan sub-agents unless they are intentionally backgrounded; cancel swarm cancels workers.
- Tests cover parent-continues-after-spawn, one worker completes while parent is idle/doing other work, all workers complete, follow-up collect, and no duplicate spawn.

## Related

- #236 covers the stuck/nonblocking swarm state drift.
- #238 covers fanout/sidebar count and progress-dot coherence.
- This issue captures the higher-level contract: sub-agents should be background jobs like background shell jobs.

Hmbown:
Codex/opencode comparison notes from local read-only audit:

Codex pattern worth copying:
[...]

PR #901, opened 2026-05-06
fix(engine): wake parent turn loop on sub-agent completion (#756)

### LINKED ISSUE #756: Parent agent stalls after parallel agent_spawn, no completion notification, must be poked manually
## Description

When the orchestrator spawns multiple sub-agents in parallel via `agent_spawn` (observed with 5 workers), it receives only the spawn-confirmation tool result (`{"status": "Running", "result": null, "steps_taken": 0, "duration_ms": 0}`) and then ends its turn. There is no mechanism that pushes a completion event back into the parent's input loop, so the parent silently idles even after all children have finished. The parent only resumes work when the human types something — in my session I had to type "are they done?" before the orchestrator called `agent_list` and discovered four had completed and one had timed out.

A 2-worker session in the same TUI (different conversation, same `deepseek-v4-pro`) did **not** exhibit this — there the model chose to follow up `agent_spawn` with explicit `agent_wait` calls per child, which block until completion and return summaries. With 5 workers spawned in a single tool batch, the model omitted `agent_wait` and ended its turn instead. So in practice the behavior is brittle: it depends on whether the model remembers to chain a `wait` call, and at higher fan-out it tends not to.

For comparison, Claude Code's `Agent` tool either blocks until the child returns (foreground) or, in background mode, injects a system message into the parent's loop when the child completes. Either prevents the silent-idle failure mode.

## Steps to reproduce

1. Run `deepseek-tui` (v0.8.12, model `deepseek-v4-pro`, mode `yolo`).
2. Ask the agent to act as orchestrator and spawn ~5 sub-agents in parallel — five `agent_spawn` calls in one assistant turn, each pointing at a different worktree/task that takes a couple of minutes.
3. Observe that after each spawn returns immediately with `status: Running, result: null`, the orchestrator narrates that workers are running and ends its turn without calling `agent_wait`.
4. Wait. The TUI sidebar updates as workers finish, but the parent agent never speaks again.
5. Type a nudge ("are they done?"). Now the orchestrator calls `agent_list`, sees the completions, and resumes the plan.

## Expected behavior

After a spawned sub-agent completes, the parent agent receives a structured event (a tool-result-shaped message or a system reminder) indicating completion, so its turn resumes automatically. Equivalent to Claude Code's behavior:

- foreground `Agent` calls block, returning the child's final report as the tool result, or
- background `Agent` calls deliver an "agent X completed" notification into the parent's next inference step.

Either is sufficient to keep the orchestrator unblocked without human intervention.

## Actual behavior

The parent's turn ends right after the spawn batch returns. No event is ever delivered to the parent when children finish — only the human-facing UI updates. The orchestrator effectively requires speculative `agent_wait` / `agent_list` calls, and at higher fan-out the model frequently omits them.

## Environment

- OS: Linux 7.0.3-arch1-2 (Arch Linux)
- DeepSeek CLI version: `deepseek 0.8.12`
- Model: `deepseek-v4-pro`
- Rust: `rustc 1.93.1 (01f6ddf75 2026-02-11)`
- Shell: fish

## Logs or screenshots

Observed timeline in the failing 5-worker run (mode `yolo`, parent `deepseek-v4-pro`):

- Orchestrator emits a single assistant turn containing 5 parallel `agent_spawn` calls.
- Five spawn `tool_result`s arrive, all shaped `{"status": "Running", "result": null, "steps_taken": 0, "duration_ms": 0}`.
- Orchestrator updates its checklist, says "Five workers are now running in parallel… I'll wait for completion sentinels and then spawn verifiers for each crate as they finish." Turn ends. **No `agent_wait` was issued.**
- Sidebar shows workers finishing one by one. Parent stays silent.
- I type "are they done?" — only then does the orchestrator call `agent_list`, see that 4 of 5 completed and 1 timed out, and resume.

Contrast: in an earlier 2-worker run on the same TUI version, the model chose to follow `agent_spawn` with explicit `agent_wait` calls and resumed cleanly with "Both agents completed successfully…" — no human nudge required. The 5-worker case did not chain `agent_wait`.

## Related

- #239 ("make sub-agents first-class background jobs") — focused on durable background-job state and UI surfacing; doesn't cover delivering completion events into the *parent agent's* inference loop.
- #251 ("surface live sub-agent progress to the parent agent") — richer snapshots in `agent_result` / `agent_wait`, but assumes the parent calls one of those tools. The failure here is precisely that the parent doesn't.

Suggested fix direction (matching Claude Code's parent-loop semantics): when a sub-agent transitions to `Completed` / `Failed` / `Timeout`, enqueue a synthetic tool-result or system message into the parent session so the next inference step sees the completion without requiring a human nudge or a speculative `agent_wait`.

y0sif:
Summary:
Fixes #756. When the parent spawns sub-agents in parallel via `agent_spawn`, each spawn returns `status: Running` immediately. If the model then has no more tool calls (e.g. emits "I'll wait for them to finish"), the turn loop breaks and the parent silently idles even after all children have finished. The human has to type something to resume.

This wires a wakeup channel from `run_subagent_task` back into the engine's turn loop. When the loop would otherwise end on empty tool calls, it now drains pending completions (or blocks on the next one — with cancel and steer escape hatches) and re-enters inference with the existing `<deepseek:subagent.done>` sentinels injected as user messages. The contract documented in `prompts/base.md` (lines 189-205) is now actually fulfilled.

Approach is option (1) from my comment on the issue: keep the spawn async (preserves the background-job semantics from #239), add the parent-side resume hook. No tool surface change. No system-prompt change — the sentinel format and the model's instructions for handling it are already in `base.md`.

Changes:
- `tools/subagent/mod.rs` — new `SubAgentCompletion` struct and `parent_completion_tx` field on `SubAgentRuntime`. `run_subagent_task` fires this channel for direct children (`spawn_depth == 1`); grandchildren are gated out so the parent isn't flooded with completions it didn't directly orchestrate.
- `core/engine.rs` — owns the receiver, plumbs the sender into `SubAgentRuntime` via a new `with_parent_completion_tx` builder.
- `core/engine/turn_loop.rs` — at the empty-tool_uses branch, drain the receiver. If empty but children are still running, `tokio::select!` on cancel / next completion / steer. Inject sentinels and `continue`.

Tests:
- 7 new unit tests in `tools/subagent/tests.rs` covering depth gating (depth 0/1/2/no-channel/dropped-receiver), `child_runtime` propagation, and the payload format pinning.
- `cargo fmt`, `cargo clippy --workspace --all-targets --all-features -- -D warnings` (clean: the 6 pre-existing `dead_code` errors in `schema_migration.rs` exist on `upstream/main` and are unrelated), `cargo test --workspace --all-features` (all green).
- Live reproduction on the same session that originally surfaced the bug. Same prompt, 5 parallel `agent_spawn` calls.
  - Old binary: parent stalled 5 times, required manual "are they done?" nudges before resuming.
  - New binary: one `<deepseek:subagent.done>` sentinel was auto-injected as a user message; the model called `agent_result` on the completed child, then `agent_wait` for the rest, and drove the batch to completion with zero human nudges.

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -47,8 +47,8 @@ use crate::tools::shell::{SharedShellManager, new_shared_shell_manager};
 use crate::tools::spec::RuntimeToolServices;
 use crate::tools::spec::{ApprovalRequirement, ToolError, ToolResult};
 use crate::tools::subagent::{
-    Mailbox, SharedSubAgentManager, SubAgentRuntime, SubAgentType, new_shared_subagent_manager,
-    resolve_subagent_assignment_route,
+    Mailbox, SharedSubAgentManager, SubAgentCompletion, SubAgentRuntime, SubAgentType,
+    new_shared_subagent_manager, resolve_subagent_assignment_route,
 };
 use crate::tools::todo::{SharedTodoList, new_shared_todo_list};
 use crate::tools::user_input::{UserInputRequest, UserInputResponse};
@@ -303,6 +303,14 @@ pub struct Engine {
     rx_user_input: mpsc::Receiver<UserInputDecision>,
     rx_steer: mpsc::Receiver<String>,
     tx_event: mpsc::Sender<Event>,
+    /// Wakeup channel for the parent turn loop when a direct child sub-agent
+    /// terminates (issue #756). Cloned into `SubAgentRuntime` so the runtime
+    /// can fan completion events back into the engine.
+    tx_subagent_completion: mpsc::UnboundedSender<SubAgentCompletion>,
+    /// Receiver paired with `tx_subagent_completion`. Drained at the
+    /// turn-loop's empty-tool_uses branch to surface `<deepseek:subagent.done>`
+    /// sentinels into the parent's transcript before deciding to end the turn.
+    pub(super) rx_subagent_completion: mpsc::UnboundedReceiver<SubAgentCompletion>,
     cancel_token: CancellationToken,
     shared_cancel_token: Arc<StdMutex<CancellationToken>>,
     tool_exec_lock: Arc<RwLock<()>>,
@@ -353,6 +361,7 @@ impl Engine {
         let (tx_approval, rx_approval) = mpsc::channel(64);
         let (tx_user_input, rx_user_input) = mpsc::channel(32);
         let (tx_steer, rx_steer) = mpsc::channel(64);
+        let (tx_subagent_completion, rx_subagent_completion) = mpsc::unbounded_channel();
         let cancel_token = CancellationToken::new();
         let shared_cancel_token = Arc::new(StdMutex::new(cancel_token.clone()));
         let tool_exec_lock = Arc::new(RwLock::new(()));
@@ -480,6 +489,8 @@ impl Engine {
             rx_user_input,
             rx_steer,
             tx_event,
+            tx_subagent_completion,
+            rx_subagent_completion,
             cancel_token: cancel_token.clone(),
             shared_cancel_token: shared_cancel_token.clone(),
             tool_exec_lock,
@@ -937,7 +948,8 @@ impl Engine {
                             self.session.reasoning_effort.clone(),
                             self.session.reasoning_effort_auto,
                         )
-                        .with_max_spawn_depth(self.config.max_spawn_depth);
+                        .with_max_spawn_depth(self.config.max_spawn_depth)
+                        .with_parent_completion_tx(self.tx_subagent_completion.clone());
                         if let Some((mailbox, cancel_token)) = mailbox_for_runtime.as_ref() {
                             rt = rt
                                 .with_mailbox(mailbox.clone())
--- a/crates/tui/src/core/engine/turn_loop.rs
+++ b/crates/tui/src/core/engine/turn_loop.rs
@@ -832,6 +832,100 @@ impl Engine {
                     continue;
                 }
 
+                // Sub-agent completion handoff (issue #756). The model finished
+                // streaming with no tool calls — but if it has direct children
+                // still running (or completions queued from children that
+                // finished while we were inferring), surface their
+                // `<deepseek:subagent.done>` sentinels into the transcript and
+                // resume instead of ending the turn. This fulfils the contract
+                // already documented in `prompts/base.md`: the parent is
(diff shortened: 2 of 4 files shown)


y0sif:
Hey @Hmbown, I could not add the v0.8.15 label, as I do not have permessions to do so.

Hmbown (approved):
Thank you for this — really thoughtful work. A few specific things that made this easy to review:

- The "approach options (1) vs (2)" exchange in #756 made the design intent transparent before the diff landed. Picking (1) keeps the async `agent_spawn` semantics from #239 intact and reuses the sentinel format that `prompts/base.md` (lines 189-205) already promises the model — so the runtime now actually fulfils a contract the prompt was already asserting.
- Gating the wakeup at `spawn_depth == 1` is the right place for it. Propagating the `Sender` through `child_runtime()` while gating at the *send* site (not the *clone* site) means descendants can't accidentally strand the channel, and the parent never sees grandchild noise. The dedicated unit test for that propagation pinned the invariant exactly where I'd want it pinned.
- The `tokio::select!` with `biased` ordering — cancel first, completion next, steer last — gets the priority right for a parent that's idling on children. The "drain after recv" pattern picks up batched completions cleanly without spinning.
- 7 focused unit tests, including the dropped-receiver no-panic case and the payload-format pin, is exactly the right shape of test coverage for a fix at this layer. Nice touch documenting *why* each test exists in the comments.

Local verification on this end:
- `cargo fmt --check` clean
- `cargo clippy -p deepseek-tui --bins --all-features --locked -- -D warnings` clean
- `cargo test -p deepseek-tui --bin deepseek-tui --all-features tools::subagent` → 85/85 pass (including the 5 new completion-gating tests)
- `cargo test -p deepseek-tui --bin deepseek-tui --all-features core::engine` → 81/81 pass
- All three parity gates (`snapshot`, `parity_protocol`, `parity_state`) pass

The race window I traced — child decrements `running_count` between the parent's `try_recv` and `running_count` read, then sends to the channel — is microseconds wide and degrades to the *pre-fix* behavior (one human nudge needed) since the unbounded channel keeps the message buffered for the next turn's drain. So worst case is a quiet regression to the old bug for one round, never worse. Happy to take that trade for the dramatic improvement on the common path.

Live reproduction in the PR description with the same 5-spawn pattern that surfaced the bug originally is the cherry on top. Merging — appreciate the care here.

PR #952, opened 2026-05-07
fix(sandbox): add Windows job object helper

Oliver-ZPLiu:
## Background

The current Windows sandbox path does not provide real process containment. If a shell command spawns child processes and the TUI exits unexpectedly or a timeout is hit, those child processes can remain running outside the TUI lifecycle.

This PR adds a small Windows helper foundation using Job Objects. The goal is to provide reliable process-tree containment first, without claiming full filesystem or network sandbox enforcement.

## Summary

- Add a Windows-only `deepseek-windows-sandbox-helper` binary.
- Run Windows shell commands through a Job Object with kill-on-close semantics when the helper self-test passes.
- Wire the Windows sandbox backend through the helper and label it as `windows-job-object`.
- Keep the boundary explicit: this is process-tree containment only, not ReadOnly / WorkspaceWrite / network policy enforcement.

## Verification

- `cargo test -p deepseek-tui sandbox --all-features`
  - `24 passed; 0 failed`
- `cargo clippy -p deepseek-tui --all-targets --all-features -- -D warnings`

## Notes

This is intended as a small Windows reliability/sandbox foundation PR. Full Windows filesystem and network enforcement should be handled in follow-up work, likely through a stricter helper contract plus Windows-specific enforcement mechanisms such as Restricted Tokens, AppContainer, ACLs, or related APIs.

### DIFF
--- a/crates/tui/Cargo.toml
+++ b/crates/tui/Cargo.toml
@@ -18,6 +18,10 @@ toml = ["schemaui/toml"]
 name = "deepseek-tui"
 path = "src/main.rs"
 
+[[bin]]
+name = "deepseek-windows-sandbox-helper"
+path = "src/bin/deepseek-windows-sandbox-helper.rs"
+
 [dependencies]
 anyhow = "1.0.100"
 arboard = "3.4"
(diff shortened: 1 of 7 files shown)


Hmbown:
Thanks for the careful Windows containment work and for keeping the boundary explicit. This still touches sandboxing across 7 files and 450+ added lines, so it is over the stricter risky-surface threshold for this repo. Windows sandbox/helper changes need an accepted issue and maintainer approval before implementation, even when CI is green.

Please split this into an accepted issue first, with the helper contract and exact guarantees/non-guarantees called out, then bring back a smaller PR for the first slice. I am closing this PR for now, but a focused PR against an accepted Windows sandbox issue is welcome.

Oliver-ZPLiu:
Thanks for the clarification. That makes sense.

I’ll stop pushing to this PR and open a separate issue first, focused on the Windows sandbox/helper contract, including exact guarantees and non-guarantees. After the direction is accepted, I’ll split the implementation into smaller PRs and start with the narrowest slice only.

PR #956, opened 2026-05-07
fix(config): persist cost_currency setting across sessions

### LINKED ISSUE #932: 费用显示支持用CNY（￥）显示

namezzb:
## Summary
- Support `--save` flag in `/config <key> <value>` so settings can be persisted to settings.toml (e.g. `/config cost_currency cny --save`)
- Add `cost_currency` to `show_single_setting()` so `/config cost_currency` displays the current value
- Add `CostCurrencyValue` enum and wire it into the interactive `/config` TUI editor so users can change it visually

Closes #932

## Test plan
- [x] All 2368 existing tests pass with zero failures
- [x] `cost_currency`-specific tests pass (pricing, settings, UI footer)
- [x] `cargo check` compiles cleanly with no warnings

### DIFF
--- a/crates/tui/src/commands/config.rs
+++ b/crates/tui/src/commands/config.rs
@@ -44,7 +44,7 @@ pub fn show_config(_app: &mut App, arg: Option<&str>) -> CommandResult {
 /// - `/config tui` / `/config web` / `/config native` — open a specific
 ///   editor mode (web requires the `web` build feature).
 /// - `/config <key>` — shows the current value of a setting.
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks for jumping on the cost_currency persistence bug. I replayed the fix on current main, added regression tests for  and the config UI currency field, and opened #997 as the maintainer replacement. Closing this PR in favor of #997.

Hmbown:
Clean note after closing: #997 is the maintainer replacement. It replays the useful fix on current main and adds regression coverage for `/config cost_currency cny --save` plus the config UI currency field.

PR #959, opened 2026-05-07
Fix npm wrapper verbose flag handling

AquaZhang:
## Summary
- stop treating `-v` as an npm wrapper version fallback
- keep wrapper fallback for `--version` and Clap's `-V`
- add a small Node test for version flag detection

## Tests
- npm test


Hmbown:
Thanks for the fix. The behavior change is good, but this branch now conflicts with current package metadata and also carried line-ending churn. I replayed the narrow patch on current main, normalized the diff, and opened #994 as the maintainer replacement. Closing this conflicted PR in favor of #994.

Issue #934, opened 2026-05-07
messages_with_turn_metadata() breaks KV prefix cache stability across turns


stofancy:
## Summary

`messages_with_turn_metadata()` in `crates/tui/src/core/engine/turn_loop.rs` injects `<turn_meta>` into only the **last** user message each turn. Because user messages are stored in `session.messages` **without** `<turn_meta>`, the message prefix sent to DeepSeek's API is inconsistent between consecutive turns — systematically destroying KV prefix cache hits for the entire `messages` array.

## How to reproduce

Any multi-turn conversation. Observe `prompt_cache_hit_tokens` vs `prompt_cache_miss_tokens` in the API response `usage` field. The hit rate degrades rapidly as the conversation grows, despite the system prompt being byte-identical.

## Root cause

**Step-by-step trace for a 2-turn conversation:**

### Turn 1
1. User types `"hello"` → stored in `session.messages` as `user("hello")` *(engine.rs:911-918, no turn_meta)*
2. `messages_with_turn_metadata()` clones messages, finds the only user message, prepends `<turn_meta>` → `user("<turn_meta>\nCurrent local date: 2026-05-07\n...\n</turn_meta>\nhello")`
3. API request sends: `messages = [user("<turn_meta>...hello")]`
4. DeepSeek caches the full prefix: `system_prompt + user("<turn_meta>...hello")`

### Turn 2
1. User types `"do X"` → stored as `user("do X")` *(no turn_meta)*
2. `session.messages` now: `[user("hello"), assistant(...), user("do X")]` — **none have turn_meta**
3. `messages_with_turn_metadata()` clones, finds the LAST user message (`"do X"`), prepends turn_meta **only to it**:
   - `messages[0]` → `user("hello")` ← **DIFFERS from Turn 1's `user("<turn_meta>...hello")`!**
   - `messages[2]` → `user("<turn_meta>...do X")`
4. API request sends: `messages = [user("hello"), assistant(...), user("<turn_meta>...do X")]`
5. DeepSeek compares prefix: `system_prompt` matches ✓ → `user("hello")` ≠ cached `user("<turn_meta>...hello")` ✗
6. **Entire messages array starts as cache miss from the first content block.**

### Turn 3 and beyond

The mismatch compounds. Every previous user message was sent **with** turn_meta in its original turn, but is re-sent **without** it in all subsequent turns. The longest common prefix between any two requests shrinks to just the system prompt.

## Why this matters

[...]

PR #960, opened 2026-05-07
fix(engine): store turn_meta at message creation time to fix KV prefix cache (#934)

gordonlu:
## Summary

Fixes a bug where DeepSeek-TUI's KV prefix cache hit rate drops to near zero
after the first turn because `messages_with_turn_metadata()` injects
`<turn_meta>` at send time without persisting it. Every replayed historical
message then has different bytes from its original wire format.

## Root cause

See issue #934 for the full trace. In short:

Turn 1 wire:    `[{user: "<turn_meta>...hello"}]`
Turn 2 store:   `[{user: "hello"}, {assistant: ...}, {user: "<turn_meta>...fix"}]`
Turn 2 wire:    `[{user: "hello"}, ...]`
                 ↑ "hello" ≠ "<turn_meta>...hello" → cache miss from byte 0

## Fix

Store `<turn_meta>` at message creation time (`handle_send_message`) rather
than injecting it at send time (`messages_with_turn_metadata`). Each user
message becomes two independent `ContentBlock` entries — a meta block with a
`<!-- DSTUI_TURN_META` sentinel prefix, and the actual user input.

## Design decisions

| Decision | Rationale |
|----------|-----------|
| Sentinel prefix instead of block index | Survives future addition of extra Text blocks (e.g. @file rendering) |
| Store-at-creation instead of insert-separate-message | V4 on-disk KV cache (§3.5.2) penalises ANY byte drift in the prefix — modifying stored messages is forbidden |
| Only turn_id + working-set paths | V4 handles time/date internally via Quick Instruction (§5.1.1 Table 5) |
| Keep `messages_with_turn_metadata()` as thin wrapper | Aggregation point for future transient modifications |

## How to test

1. **Quick verification** — `cargo test -p deepseek-tui messages_prefix_is_stable`
   runs the two integration tests that assert byte-identical JSON prefixes.
2. **Full suite** — `cargo test -p deepseek-tui` passes all 2252 tests.
3. **Manual** — start a multi-turn session and watch the footer `cache hit %`
   chip. After the first turn, hit rate should increase noticeably compared to
   the pre-fix baseline, and climb further as the conversation grows.
4. **Regression** — load a saved session from before this fix and verify the
   transcript renders all user messages correctly (the sentinel check is
   backward-compatible).

## Backward compatibility

Old session files (single Text block, no sentinel) render identically — the
renderer's `starts_with("<!-- DSTUI_TURN_META")` check simply never matches,
so all content passes through unchanged.

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -824,6 +824,37 @@ impl Engine {
             .await;
     }
 
+    /// Build the turn‑metadata block for KV prefix‑cache stability (#934).
+    ///
+    /// The sentinel prefix `<!-- DSTUI_TURN_META` lets the transcript
+    /// renderer identify and skip this block so it never appears in the
+    /// user‑visible history.
+    ///
+    /// Contents are deliberately limited to information the model cannot
+    /// infer from V4's retained reasoning chain or Quick Instruction
+    /// mechanism (DeepSeek-V4 Technical Report §5.1.1, Table 5, Fig. 7a).
+    ///
+    /// - turn counter (machine‑readable identifier)
+    /// - working‑set paths (model can't reconstruct from its own output)
+    ///
+    /// DateTime, intent classification, and state summaries are omitted
+    /// because V4 handles those internally via Quick Instruction tokens.
+    fn build_turn_meta_string(&self) -> String {
+        let working_set_summary = self
+            .session
+            .working_set
+            .summary_block(&self.config.workspace)
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty());
+
+        let summary = working_set_summary.unwrap_or_default();
+
+        format!(
+            "<!-- DSTUI_TURN_META turn_id={} -->\n<turn_meta>\n{summary}\n</turn_meta>",
+            self.turn_counter
+        )
+    }
+
     async fn add_session_message(&mut self, message: Message) {
         self.session.add_message(message);
         self.emit_session_updated().await;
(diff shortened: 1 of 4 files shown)


gordonlu:
Noticed #951 after submitting this. The core fix is the same (store turn_meta
at creation time), but a few differences worth noting:

**Covered here, not in #951:**
- Sentinel prefix `<!-- DSTUI_TURN_META` → lets the transcript renderer skip
  the block. Without this, `<turn_meta>` XML tags appear verbatim in the TUI
  history.
- Date removed from turn_meta (DeepSeek-V4 §5.1.1 Table 5, Fig. 7a — Quick
  Instruction handles temporal awareness internally).
- `resolve_auto_effort` filters meta blocks so keyword matching (debug/error/
  search) isnt skewed by working-set paths or dates.
- Steer messages (mid-turn intervention) also get turn_meta blocks.
- Two prefix-stability integration tests including a tool-call interleaved
  pattern with thinking traces, verified via `assert_byte_identical()`.

**Covered in #951, not here:**
- `lsp_hooks.rs` diagnostic messages route through the same helper. Happy to
  rebase if that is the preferred approach.

Both converge on the same root cause and structural fix — the fix getting
merged is what matters. I will leave this open for comparison, but happy to
close immediately if #951 is the preferred path.

AtaraxicCyber:
I am relatively new to how V4 handles KV prefix caching, so I really appreciate the detailed write-up in the root cause. Storing the metadata at creation time to prevent byte drift makes perfect sense. Thanks for the clear explanation and the fix!

Hmbown:
Thanks for the detailed diagnosis and implementation work here. The same #934 cache-stability fix has now shipped via the smaller maintainer-owned #951 branch after rebase, local verification, and green CI, including a guard so stored <turn_meta> does not affect auto-thinking tier selection. Closing this as superseded by #951 to avoid carrying duplicate review debt.

PR #1011, opened 2026-05-07
docs(install): document rsproxy rustup workaround

wuwuzhijing:
## Summary

- Document an rsproxy-based rustup mirror sequence for mainland China Cargo installs.
- Add troubleshooting notes for the `feature edition2024 is required` error caused by old Debian/Ubuntu Cargo packages such as Cargo 1.75.0.
- Tell users to verify that `cargo` resolves to `~/.cargo/bin/cargo` after installing rustup stable.

## Verification

- Checked Markdown code fences are balanced with a small Python script.
- Docs-only change; no build or runtime tests run.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -116,10 +116,7 @@ Linux/macOS:
 export RUSTUP_DIST_SERVER=https://rsproxy.cn
 export RUSTUP_UPDATE_ROOT=https://rsproxy.cn/rustup
 
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-source "$HOME/.cargo/env"
-rustup default stable
-cargo install deepseek-tui-cli --locked
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
 ```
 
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for documenting this. I’m leaving it out of the v0.8.17 binary release path, but it looks like a useful docs-only follow-up for users installing Cargo/Rust from mainland China networks.

Before merge, I’d verify the exact doc location against the current install guide and make sure the guidance stays framed as a mirror workaround, not as a new default install path.

Hmbown:
Thanks @wuwuzhijing. I harvested the safe docs portion into draft release PR #1094 instead of merging the PR directly.

What landed:
- `rsproxy.cn` is documented as an alternate rustup mirror when TUNA is slow
- Debian/Ubuntu `feature edition2024 is required` failures now point users toward a current rustup stable Cargo and `~/.cargo/bin/cargo`
- the mirror guidance remains framed as a workaround, not a new default install path

Verification:
- `RUSTDOCFLAGS=-Dwarnings cargo doc --workspace --no-deps --locked`

Leaving this PR open for now so the maintainer can decide whether to close it after #1094 merges.

wuwuzhijing:
@Hmbown OK

PR #1059, opened 2026-05-07
fix(npm): make postinstall non-blocking

Fire-dtx:
## Summary

  - Run `npm` `postinstall` in optional mode so `deepseek-tui` installation no
    longer fails when GitHub release assets are temporarily unreachable.
  - Split install-time and runtime download behavior: optional postinstall now
    fails fast, while first-run recovery keeps the normal retry and timeout budget.
  - Preserve wrapped HTTP 5xx retry metadata, only ignore retryable install-time
    download failures, and document the behavior with regression tests and inline
    comments.

  ## Changes

  - Update `npm/deepseek-tui/package.json` to run `postinstall` with
    `node scripts/install.js --optional`.
  - Refactor `npm/deepseek-tui/scripts/install.js` to distinguish install/runtime
    contexts, use optional-install defaults, and keep wrapped HTTP 5xx failures
    retryable.
  - Update `npm/deepseek-tui/README.md` to describe non-blocking install behavior
    and clarify which install failures still remain fatal.
  - Add `npm/deepseek-tui/test/postinstall.test.js` to cover optional install
    detection, install/runtime defaults, retryable failure filtering, and wrapped
    HTTP 5xx handling.

  ## Test Plan

  - [x] `node -c npm/deepseek-tui/scripts/install.js`
  - [x] `npm test` from `npm/deepseek-tui`
  - [x] `git diff --check origin/main..HEAD`

### DIFF
--- a/npm/deepseek-tui/README.md
+++ b/npm/deepseek-tui/README.md
@@ -17,8 +17,10 @@ npm install deepseek-tui
 npx deepseek-tui --help
 ```
 
-`postinstall` downloads platform binaries into `bin/downloads/` and exposes
-`deepseek` and `deepseek-tui` commands.
+`postinstall` tries to download platform binaries into `bin/downloads/` and
+exposes `deepseek` and `deepseek-tui` commands. If GitHub release assets are
+temporarily unreachable, install continues and the wrapper retries the download
+on first run.
 
 ## First run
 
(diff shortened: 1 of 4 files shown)


Hmbown:
This is a good npm resilience direction and directly relevant to install pain, but I’m not pulling it into the already-cut v0.8.17 tag without a separate npm-wrapper review.

Best next step after v0.8.17 publishes: run the package tests locally, verify first-run download recovery still fails loudly when it should, and then land this as a small npm-only patch if the behavior matches the release/install contract.

Hmbown:
Thanks @Fire-dtx. I harvested the install-resilience idea into draft release PR #1094, but tightened the failure boundary so npm install remains honest.

What landed:
- postinstall now runs the downloader in optional mode
- retryable download/extract failures can recover at install time
- unsupported platform, checksum mismatch, glibc preflight, and runtime wrapper failures stay fatal
- runtime launches still use the normal retry behavior instead of the short postinstall budget

Verification:
- `cd npm/deepseek-tui && node -c scripts/install.js && npm test` -> 10 passed, 0 failed
- `node scripts/release/npm-wrapper-smoke.js` passed after the 0.8.18 version bump

This will close #1059 when #1094 merges.

Issue #1010, opened 2026-05-07
希望有切换工作空间的命令，现在无法切换工作空间


yudayuta:

J3y0r:
https://github.com/Hmbown/DeepSeek-TUI/pull/1065

I tried coding one, but I'm not sure if that's what you're looking for

bdbox1:
请问，如何切换当前工程目录？

zimocode:
确实需要这个功能，我打开第一件事就是在找这命令。

Hmbown:
This is a real workflow gap. PR #1065 is an initial attempt at a workspace switch command, but I’m not putting it into the v0.8.17 release tag without a review/test pass.

For the next iteration, the acceptance criteria should be explicit: switch current workspace directory from inside the TUI, update the displayed/current workdir, keep session state sane, and verify the behavior with a focused test or PTY scenario.

Hmbown:
Reviewed during v0.8.18 triage. This is a real workflow gap, but I am not including it in #1094.

Why it is deferred: safe workspace switching has to reset or reopen workspace-scoped runtime state, not only update the displayed path. Shell/tool/task/session managers and current-directory/session state need to move together or the UI can claim one workspace while tools still act in another.

The current PR #1065 is also conflicting, so this should get a dedicated follow-up with tests for current workspace display, current directory, session state, and tool execution after a switch.

Hmbown:
Already implemented — /workspace [path] slash command with ~ expansion and validation. Workspace switching also available via AppAction::SwitchWorkspace. Tests in commands/core.rs:584-632.

PR #1065, opened 2026-05-07
feat(tui): add workspace switch command

J3y0r:
## Summary
https://github.com/Hmbown/DeepSeek-TUI/issues/1010
add workspace manager command

## Testing

- [ ✓] `cargo test --all-features`
- [✓ ] `cargo fmt --all -- --check`
- [ ✓] `cargo clippy --all-targets --all-features`

## Checklist

- [ ×] Updated docs or comments as needed
- [ ×] Added or updated tests where relevant
- [✓ ] Verified TUI behavior manually if UI changes

### DIFF
--- a/crates/tui/src/commands/core.rs
+++ b/crates/tui/src/commands/core.rs
@@ -1,6 +1,7 @@
 //! Core commands: help, clear, exit, model
 
 use std::fmt::Write;
+use std::path::PathBuf;
 
 use crate::config::{COMMON_DEEPSEEK_MODELS, normalize_model_name};
 use crate::localization::{MessageId, tr};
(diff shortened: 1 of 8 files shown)


Hmbown:
Thanks for taking a pass at the workspace-switch flow. I’m keeping this out of the v0.8.17 release tag because it needs a focused review/test pass, but it maps to a real user workflow in #1010.

What I’d want before merge: clear behavior for changing the active workspace from inside the TUI, visible current-workspace update, no session/history corruption when switching, and a small regression test or PTY scenario covering the command path.

Hmbown:
I reviewed this for v0.8.18 and am deferring it rather than harvesting it into #1094.

Reason: an in-process workspace switch needs more than changing the visible workspace path. The current app has workspace-scoped shell/tool/task/session/runtime state, and switching only `app.workspace` risks leaving managers and current-session state pointed at the old workspace. The PR is also currently conflicting.

This still looks like a real workflow gap for #1010, but it needs a dedicated implementation that resets or reopens the workspace-scoped managers, updates the displayed current directory/session state coherently, and adds regression coverage around tools after the switch.

J3y0r:
> I reviewed this for v0.8.18 and am deferring it rather than harvesting it into #1094.
> 
> Reason: an in-process workspace switch needs more than changing the visible workspace path. The current app has workspace-scoped shell/tool/task/session/runtime state, and switching only `app.workspace` risks leaving managers and current-session state pointed at the old workspace. The PR is also currently conflicting.
> 
> This still looks like a real workflow gap for #1010, but it needs a dedicated implementation that resets or reopens the workspace-scoped managers, updates the displayed current directory/session state coherently, and adds regression coverage around tools after the switch.

I pushed the new changes, just now

J3y0r:
> I reviewed this for v0.8.18 and am deferring it rather than harvesting it into #1094.
> 
> Reason: an in-process workspace switch needs more than changing the visible workspace path. The current app has workspace-scoped shell/tool/task/session/runtime state, and switching only `app.workspace` risks leaving managers and current-session state pointed at the old workspace. The PR is also currently conflicting.
> 
> This still looks like a real workflow gap for #1010, but it needs a dedicated implementation that resets or reopens the workspace-scoped managers, updates the displayed current directory/session state coherently, and adds regression coverage around tools after the switch.

I added a follow-up fix for workspace switching so changing the active workspace from inside the TUI updates workspace-scoped runtime
  state, not just the displayed path.

  Changes included:
  - Extracted a shared apply_workspace_runtime_state helper that updates:
    - app.workspace
    - hook executor
    - skills directory/cache
    - workspace context cache
    - workspace context cell
    - file tree state
    - shell manager default workspace
  - Updated /workspace / /cwd switching to:
    - reject switching while a request is running
    - reset/reopen workspace-scoped runtime state
    - update the task manager default workspace
    - restart the engine with the new workspace/runtime services
    - sync the existing session messages into the restarted engine
  - Updated session load/resume paths to reuse the same workspace runtime reset logic, so loading a session from another workspace does
  not leave shell/tool runtime state pointing at the previous workspace.
  - Made TaskManager default workspace mutable so future tasks without an explicit workspace use the currently active workspace after a
  switch.
  - Added regression coverage for:
    - command path handling for /workspace and /cwd
    - visible current workspace display after workspace/session changes
    - session load resetting workspace-scoped runtime state
    - shell manager default workspace after runtime reset
    - task manager default workspace after runtime sync

PR #1066, opened 2026-05-07
fix(config): preserve OpenRouter custom endpoint models

### LINKED ISSUE #857: normalize_model_for_provider replaces user-configured model name with provider-specific prefix, breaking custom OpenAI-compatible endpoints
## Description

When using a custom OpenAI-compatible API endpoint (e.g. a self-hosted gateway or enterprise proxy), the TUI replaces the user-configured model name with a provider-specific prefixed version via `normalize_model_for_provider()`.

This means **no built-in provider can be used to connect to a custom endpoint that expects bare model names**.

## Steps to Reproduce

1. Set up a custom OpenAI-compatible API gateway (e.g. `https://my-gateway.example.com/v1`) that serves model `DeepSeek-V4-Pro`
2. Configure `config.toml`:
   ```toml
   provider = "openrouter"
   default_text_model = "DeepSeek-V4-Pro"

   [providers.openrouter]
   api_key = "my-key"
   base_url = "https://my-gateway.example.com/v1"
   model = "DeepSeek-V4-Pro"
   ```
3. Start DeepSeek TUI and send a message

## Expected Behavior

TUI sends `"model": "DeepSeek-V4-Pro"` to the custom endpoint (as configured by the user).

## Actual Behavior

TUI sends `"model": "deepseek/deepseek-v4-pro"` (OpenRouter format), which the custom endpoint rejects:

```
HTTP 400 Bad Request: deepseek/deepseek-v4-pro model is not supported
```

## Root Cause

`normalize_model_for_provider()` in `crates/config/src/lib.rs` unconditionally replaces model names for all built-in providers:

```rust
fn normalize_model_for_provider(provider: ProviderKind, model: &str) -> String {
    let normalized = model.trim().to_ascii_lowercase();
    match (provider, normalized.as_str()) {
        (ProviderKind::Openrouter, "deepseek-v4-pro") => "deepseek/deepseek-v4-pro".to_string(),
        (ProviderKind::Sglang, "deepseek-v4-pro") => "deepseek-ai/DeepSeek-V4-Pro".to_string(),
        (ProviderKind::Novita, "deepseek-v4-pro") => "deepseek/deepseek-v4-pro".to_string(),
        // ... all providers have similar mappings
        _ => model.to_string(),  // only unmatched names pass through
    }
}
```

The `_` fallback only fires when the lowercased model name **doesn't match** any known pattern. Since `deepseek-v4-pro` matches for every provider, there is no way to send the bare model name regardless of which provider is selected.

## Why #733 Didn't Fix This

PR #733 (for #729) preserved **case** when the resolved model differs only in casing. But `normalize_model_for_provider` doesn't just change case — it **replaces the entire model name** with a provider-specific prefixed variant. The case-preservation logic runs after normalization, so it has no effect here.

## Suggested Fix

Add a `"custom"` or `"openai-compatible"` provider kind that skips `normalize_model_for_provider` entirely and passes the user's model name through unchanged:

```toml
provider = "custom"

[providers.custom]
api_key = "sk-my-key"
base_url = "https://my-gateway.example.com/v1"
model = "DeepSeek-V4-Pro"
```

Alternatively, skip normalization when the user explicitly sets `model` in `[providers.<name>]` — treat an explicit model override as "I know what I'm doing".

## Workaround

Currently requires a local HTTP proxy that rewrites the model name before forwarding to the actual endpoint — not practical for most users.

## Environment

- DeepSeek TUI v0.8.14 (Windows x64)
- Third-party OpenAI-compatible gateway

THINKER-ONLY:
When OpenRouter is pointed at a custom base_url, keep explicit model values verbatim instead of remapping DeepSeek aliases to OpenRouter catalog IDs.

Add config coverage for both the dispatcher config crate and the TUI config loader, while preserving existing provider alias behavior such as NVIDIA NIM.

Closes #857

## Summary

  - Preserve explicit OpenRouter model names when `providers.openrouter.base_url` points to a custom OpenAI-compatible
endpoint.
  - Keep existing provider alias mappings for default endpoints, including NVIDIA NIM.
  - Add regression coverage for both `deepseek-config` and TUI config loading.

  Closes #857

  ## Testing

  - [x] `cargo test -p deepseek-config --locked`
  - [x] `cargo test -p deepseek-tui config::tests:: --locked`
  - [x] `cargo clippy -p deepseek-config --locked -- -D warnings`
  - [x] `cargo clippy -p deepseek-tui --locked --bin deepseek-tui -- -D warnings`
  - [x] `git diff --check`

  ## Checklist


Hmbown:
Landed via cherry-pick as 4aee8a15c after a docs/CONFIGURATION.md conflict with #1084 (which switched the deepseek-cn default endpoint to the official host). Just kept your OpenRouter sentence in the model-defaults paragraph and dropped the now-stale typo URL line. Marking closed-as-merged. Thanks @THINKER-ONLY — `base_url_is_custom_for_provider` + the `explicit_model` flag is exactly the right gating. Will credit in the v0.8.17 CHANGELOG.

Tip for next time: rebasing onto `origin/main` before the final push catches docs-section conflicts like this one early — it's almost always a one-line resolution.

PR #937, opened 2026-05-07
fix(core): persist engine shutdown checkpoint and rehydrate across sessions

m8f1m8f1:
## Summary

Replace the legacy `memory/*.jsonl`-based cross-session recovery with a **session-JSON direct-read** approach that is richer, naturally workspace-scoped, and requires no opt-in for the primary path.

### 3-tier fallback chain in `rehydrate_latest_canonical_state()`

| Tier | Source | Scope | Opt-in required |
|------|--------|-------|-----------------|
| 1 (hot) | Current in-memory `CapacityMemoryRecord` | Same session resume | No |
| 2 (new) | Previous session JSON via `SessionManager::get_latest_session_for_workspace()` | Workspace-scoped by git root | **No** |
| 3 (legacy) | `memory/*.jsonl` via `find_latest_cross_session()` | All sessions, filtered by workspace | `cross_session_enabled` |

### Key improvements

- **Rich data**: Session JSON has full message history (last 12 messages); memory JSONL only had whatever a single `write_shutdown_checkpoint` call saved
- **Naturally workspace-scoped**: `SessionManager::get_latest_session_for_workspace()` matches by git root, so tier 2 is workspace-isolated without needing `cross_session_enabled`. The opt-in flag is only needed for tier 3 (legacy memory fallback)
- **Meaningful shutdown state**: `write_shutdown_checkpoint` now scans session messages to extract `goal`, `confirmed_facts`, `open_loops` from actual conversation content instead of writing empty strings
- **Clean extraction**: Skips `<turn_meta>` metadata blocks during parsing to get clean natural-language content
- **No new dependencies**: Uses existing `SavedSession` and `SessionManager` types already in the codebase

### Files changed

- `crates/tui/src/core/engine.rs` — `write_shutdown_checkpoint()` now extracts meaningful canonical state; calls `run_startup_memory_migration()` before rehydration
- `crates/tui/src/core/engine/capacity_flow.rs` — 3-tier fallback chain + `run_startup_memory_migration()` + `DEEPSEEK_CAPACITY_DISABLE_TIER2` kill-switch
- `crates/tui/src/core/capacity_memory.rs` — Migration: `migrate_legacy_memory_to_workspace()`, Rollback: `rollback_migration_for_file()` + `rollback_all_migrations()`, 9 new CI tests
- `crates/tui/src/core/capacity.rs` — `cross_session_enabled` field repositioned, env var override added
[...]

PR #1074, opened 2026-05-07
fix(core): capacity-memory checkpoint and cross-session recovery

### LINKED ISSUE #1062: fix(core): capacity-memory checkpoint and cross-session recovery
## Description

Two related bugs in the capacity-memory system prevent cross-session state recovery.

### Bug 1: Missing shutdown checkpoint

When a session ends (via `/exit`, SIGTERM, etc.), the engine does not persist any memory record unless the user has explicitly enabled `capacity.enabled = true`. Since capacity scoring is disabled by default, the vast majority of users never get a checkpoint written.

**Impact**: A new `exec --auto` session starts with a completely blank `CanonicalState` — `goal`, `confirmed_facts`, and all other state fields are empty. The model cannot answer "what did we work on in the previous session?"

**Location**: `crates/tui/src/core/engine.rs` — no write happens on `Op::Shutdown`.

### Bug 2: No cross-session scan fallback

Even when memory records exist (from a prior session that had capacity scoring enabled), `rehydrate_latest_canonical_state()` only looks at the current session ID's memory file at `~/.deepseek/memory/<current_session_id>.jsonl`. Each `exec --auto` invocation generates a fresh random session UUID, so the current session's file is always empty.

**Impact**: Memory records from previous sessions are effectively orphaned. There is no fallback to scan sibling `*.jsonl` files in `~/.deepseek/memory/`.

**Location**: `crates/tui/src/core/engine/capacity_flow.rs` — `rehydrate_latest_canonical_state()` only calls `load_last_k_capacity_records(&self.session.id, 1)`.

---

## Expected behavior

1. On engine shutdown, a lightweight checkpoint record should always be written — regardless of whether capacity scoring is enabled.
2. When starting a fresh session with no local records, the system should fall back to scanning all memory directories for the newest available record.
3. Cross-session recovery must be opt-in and workspace-scoped to prevent memory bleed across unrelated projects.

## Proposed fix

PR #937 implements all three requirements:
- Writes a shutdown checkpoint in `write_shutdown_checkpoint()` on `Op::Shutdown`
- Adds `find_latest_cross_session()` fallback in `rehydrate_latest_canonical_state()`
- Attaches workspace path to each record; cross-session scan filters by workspace
- Adds `capacity.cross_session_enabled` opt-in flag (default `false`)

Giggitycountless:
## Summary

Fixes #1062 — Two related bugs in the capacity-memory system prevent cross-session state recovery.

### Bug 1: Missing shutdown checkpoint
When a session ends (via `/exit`, SIGTERM, etc.), the engine does not persist any memory record unless the user has explicitly enabled `capacity.enabled = true`. Since capacity scoring is disabled by default, the vast majority of users never get a checkpoint written.

**Impact**: A new `exec --auto` session starts with a completely blank `CanonicalState` — `goal`, `confirmed_facts`, and all other state fields are empty.

### Bug 2: No cross-session scan fallback
Even when memory records exist (from a prior session that had capacity scoring enabled), `rehydrate_latest_canonical_state()` only looks at the current session ID's memory file. Each `exec --auto` invocation generates a fresh random session UUID, so the current session's file is always empty.

**Impact**: Memory records from previous sessions are effectively orphaned.

### 1. Add `cross_session_enabled` config option
- `config.rs`: Add `cross_session_enabled` to `CapacityConfig`
- `capacity.rs`: Add `cross_session_enabled` to `CapacityControllerConfig`
- Default: `false` (opt-in to prevent memory bleed)

### 2. Add `workspace_path` to `CapacityMemoryRecord`
- `capacity_memory.rs`: Add `workspace_path` field for workspace filtering
- Prevents memory bleed across unrelated projects

### 3. Implement cross-session scanning
- `capacity_memory.rs`: Add `load_latest_cross_session_record()` function
- `capacity_memory.rs`: Add `record_canonical_matches_workspace()` helper
- Scans all memory directories for the newest available record

### 4. Implement shutdown checkpoint
- `engine/capacity_flow.rs`: Add `write_shutdown_checkpoint()` method
- `engine/capacity_flow.rs`: Add `extract_current_canonical_state()` helper
- `engine.rs`: Call `write_shutdown_checkpoint()` on `Op::Shutdown`

### 5. Modify rehydrate logic
- `engine/capacity_flow.rs`: Update `rehydrate_latest_canonical_state()` to:
  1. Try current session first
  2. Fallback to cross-session scan if enabled

### Unit Tests (12 new tests)
- `test_workspace_path_serialization`: Verify workspace_path field serialization
- `test_workspace_path_none_serialization`: Verify None workspace_path handling
- `test_record_canonical_matches_workspace_exact`: Exact workspace match
- `test_record_canonical_matches_workspace_parent`: Parent workspace match
- `test_record_canonical_matches_workspace_child`: Child workspace match
- `test_record_canonical_matches_workspace_different`: Different workspace rejection
- `test_record_canonical_matches_workspace_legacy`: Legacy records without workspace
- `test_load_latest_cross_session_record_empty`: Empty memory directory
- `test_load_latest_cross_session_record_single`: Single record loading
- `test_load_latest_cross_session_record_workspace_filter`: Workspace filtering

### Regression Tests
- Existing tests continue to pass
- Legacy records (without workspace_path) are still supported

### Integration Tests
- Shutdown checkpoint is written on engine exit
- Cross-session recovery works when enabled

### Boundary Cases
- Empty memory directories
- Missing workspace_path (legacy records)
- Different workspace paths (should be filtered)
- Parent/child workspace relationships

## Configuration

To enable cross-session recovery, add to `~/.deepseek/config.toml`:

```toml
[capacity]
cross_session_enabled = true
```

Or via environment variable:

```bash
export DEEPSEEK_CAPACITY_CROSS_SESSION_ENABLED=true
```

## Breaking Changes

None. This is a purely additive change with default behavior unchanged.

## Related Issues

- #1062: fix(core): capacity-memory checkpoint and cross-session recovery
- #937: Related PR with similar approach (may conflict)

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -662,6 +662,10 @@ pub struct CapacityConfig {
     pub deepseek_v4_pro_prior: Option<f64>,
     pub deepseek_v4_flash_prior: Option<f64>,
     pub fallback_default_prior: Option<f64>,
+    /// Enable cross-session state recovery. When true, the engine will
+    /// scan memory records from previous sessions to rehydrate canonical
+    /// state on startup. Default: false (opt-in to prevent memory bleed).
+    pub cross_session_enabled: Option<bool>,
 }
 
 impl RetryPolicy {
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -2095,6 +2095,7 @@ fn apply_env_overrides(config: &mut Config) {
         deepseek_v4_pro_prior: None,
         deepseek_v4_flash_prior: None,
         fallback_default_prior: None,
+        cross_session_enabled: None,
     });
 
     if let Ok(value) = std::env::var("DEEPSEEK_CAPACITY_ENABLED") {
--- a/crates/tui/src/core/capacity.rs
+++ b/crates/tui/src/core/capacity.rs
@@ -17,6 +17,10 @@ pub struct CapacityControllerConfig {
     pub profile_window: usize,
     pub model_priors: HashMap<String, f64>,
     pub fallback_default: f64,
+    /// Enable cross-session state recovery. When true, the engine will
+    /// scan memory records from previous sessions to rehydrate canonical
+    /// state on startup. Default: false (opt-in to prevent memory bleed).
+    pub cross_session_enabled: bool,
 }
 
 impl Default for CapacityControllerConfig {
@@ -52,6 +56,7 @@ impl Default for CapacityControllerConfig {
             profile_window: 8,
             model_priors,
             fallback_default: 3.8,
+            cross_session_enabled: false,
         }
     }
 }
@@ -111,6 +116,9 @@ impl CapacityControllerConfig {
         }
         if let Some(v) = capacity.fallback_default_prior {
             out.fallback_default = v;
+        }
+        if let Some(v) = capacity.cross_session_enabled {
+            out.cross_session_enabled = v;
         }
 
         out
--- a/crates/tui/src/core/capacity_memory.rs
+++ b/crates/tui/src/core/capacity_memory.rs
@@ -44,6 +44,9 @@ pub struct CapacityMemoryRecord {
     pub source_message_ids: Vec<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub replay_info: Option<ReplayInfo>,
+    /// Workspace path for cross-session filtering (optional).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub workspace_path: Option<String>,
 }
 
 fn capacity_memory_dirs() -> Vec<PathBuf> {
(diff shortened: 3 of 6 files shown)


Giggitycountless:
Closing this PR in favor of #937 — after reviewing the updated approach there (session-JSON via SessionManager), I agree it's a cleaner solution than the memory-JSONL approach used here. The 3-tier fallback chain and automatic workspace scoping via git root are solid improvements.

Great work @m8f1m8f1! 🙌

m8f1m8f1:
@Giggitycountless Thank you so much for the kind words, and more importantly, thank you for your thoughtful work on this PR. 🙏

I studied your approach closely while working on #937, and I want to say — your solution has several things that are genuinely well done:

**What impressed me about your approach:**

1. **Test coverage** — 12 new unit tests covering workspace matching, serialization, boundary cases, and legacy record compatibility. That's thorough engineering. I should have written more tests on my side after seeing yours.

2. **Edge case handling** — Your `test_record_canonical_matches_workspace_parent` and `test_record_canonical_matches_workspace_child` tests show real attention to how workspace paths actually behave in the wild. Parent/child relationships are something that's easy to overlook.

3. **Environment variable support** — `DEEPSEEK_CAPACITY_CROSS_SESSION_ENABLED` alongside the config file option is a nice touch for headless/automation use cases, which is something I didn't cover.

4. **Legacy record backward compatibility** — Your handling of records without `workspace_path` shows care for not breaking existing user data. That's the right mindset.

**A thought on the architecture difference:**

Your memory-JSONL route and my session-JSON route ended up at different spots mainly because of a different starting question I asked myself: "the session data already exists in sessions JSON — can we read it directly instead of writing a parallel copy?" That led to the SessionManager path, which kept the changeset smaller.

But your approach has a genuine advantage that mine doesn't: **the memory layer is more resilient**. If a session JSON gets corrupted or rotated out, your code can still fall back to whatever was checkpointed to memory. The two approaches are actually complementary — that's why I kept the memory fallback as tier 3 in the chain.

If you're interested, I think your 12 unit tests could be adapted into a shared test suite that covers both paths. Would be happy to collaborate on that.

Thanks again for jumping on this issue independently — having two independent solutions converge on the same problem made the final design stronger than either would have been alone. 🙇

PR #1080, opened 2026-05-07
fix(cache): auto-generate .deepseek/instructions.md to stabilize KV prefix cache

### LINKED ISSUE #979: 完成相似的任务 缓存命中率比claudecode低很多

lloydzhou:
## Summary

When a workspace has no context file (`AGENTS.md`, `CLAUDE.md`, `.deepseek/instructions.md`, etc.), the system prompt assembly falls back to calling `summarize_project()` + `project_tree()` on **every turn** via `refresh_system_prompt()`. Both functions use `WalkBuilder` to scan the filesystem in real-time. Any file change during a session alters the system prompt bytes, invalidating the entire KV prefix cache — effectively zero cache hits for workspaces without context files.

Fix: `load_project_context_with_parents()` now auto-generates `.deepseek/instructions.md` **once** when no context file is found. Subsequent turns load the file from disk, keeping system prompt bytes stable for the entire session. The generated file is marked as auto-generated; users can freely edit or delete it.

Closes #979

## Testing

## Checklist


lloydzhou:
@Hmbown 
可以看看这个可能导致缓存失效的bug，我感觉影响范围还挺大的，希望可以尽快处理

PR #1083, opened 2026-05-07
fix(skills): show all skills in slash menu

Duducoco:
## Summary

  Fixes two independent bugs that caused some skills to be missing from the `/` slash-menu autocomplete:

  1. **YAML quotes not stripped** (`5b1a699`): the `SKILL.md` frontmatter parser only `trim()`'d values, leaving surrounding quotes intact. A skill declared as
  `name: "hud"` was registered as `"hud"` (with literal quotes), so `starts_with("hu")` prefix matching failed and the skill never appeared when typing `/hu`.
  2. **Cache only scanned a single directory** (`59e082b`): `App::discover_cached_skills` called `SkillRegistry::discover` against just one resolved
  `skills_dir`, while the model-visible skills block already used `discover_in_workspace`, which merges all candidate directories (`<workspace>/.agents/skills`,
   `~/.claude/skills`, `~/.deepseek/skills`, …). When a higher-precedence directory contained an empty stub directory (no `SKILL.md`) for a given name, it
  shadowed the real definition lower down the candidate list, so the menu and the model-visible skills list were inconsistent.

  Switching the cache to `discover_in_workspace` aligns the slash menu with the model-visible list and lets the existing first-wins merge naturally skip empty
  stub directories.

  ## Testing

  - [x] `cargo clippy --all-targets --all-features -- -D warnings`

  ## Checklist

### DIFF
--- a/crates/tui/src/skills/mod.rs
+++ b/crates/tui/src/skills/mod.rs
@@ -248,7 +248,17 @@ impl SkillRegistry {
                     continue;
                 }
                 if let Some((key, value)) = line.split_once(':') {
-                    metadata.insert(key.trim().to_ascii_lowercase(), value.trim().to_string());
+                    let value = value.trim();
+                    let unquoted = if (value.starts_with('"')
+                        && value.ends_with('"')
+                        && value.len() >= 2)
+                        || (value.starts_with('\'') && value.ends_with('\'') && value.len() >= 2)
+                    {
+                        &value[1..value.len() - 1]
+                    } else {
+                        value
+                    };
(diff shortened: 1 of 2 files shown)


Hmbown:
Merged as 5d3a51e29. Thanks @AlphaGogoo / @Duducoco — the YAML-quote bug catch was a nice bonus, and the included unit test (`cached_skills_merges_across_candidate_directories`) is exactly the kind of regression guard this part of the codebase needed. Will credit in the v0.8.17 CHANGELOG.

PR #1091, opened 2026-05-07
fix(i18n): localize approval dialog safety copy

### LINKED ISSUE #1087: Request Chinese localization for destructive approval dialogs
Hi, I'm a Chinese-speaking user. I'd like to request Chinese localization for the destructive operation approval dialog.

Currently, when the TUI needs to perform operations like write_file, the approval prompt is entirely in English. This is particularly problematic because these prompts involve security decisions — users need to fully understand what they're approving before clicking.

Image

I understand that the reasoning/thinking content comes from the model, not the TUI, so that's not a TUI issue. But the approval dialog UI is something the TUI itself controls and would greatly benefit from localization.

Thank you for considering this!

axobase001:
## Summary
- Pass the configured UI locale into approval modals.
- Localize Simplified Chinese approval-dialog safety copy: risk badge, category labels, impact summaries, option labels, and destructive two-step confirmation text.
- Preserve English behavior by default and add a zh-Hans render regression test.

Closes #1087

## Tests
- cargo test -p deepseek-tui tui::approval::tests:: -- --nocapture
- cargo check -p deepseek-tui

### DIFF
--- a/crates/tui/src/tui/approval.rs
+++ b/crates/tui/src/tui/approval.rs
@@ -27,6 +27,7 @@
 //! happen *before* the view is constructed (see `tui/ui.rs`); this
 //! module always assumes the user is being asked.
 
+use crate::localization::Locale;
 use crate::sandbox::SandboxPolicy;
 use crate::tui::views::{ModalKind, ModalView, ViewAction, ViewEvent};
 use crate::tui::widgets::{ApprovalWidget, ElevationWidget, Renderable};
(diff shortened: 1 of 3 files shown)


Hmbown:
This looks like the right follow-up for #1087 and the CI matrix is green. I’m not merging it into the already-published v0.8.17 tag, but this should be near the top of the v0.8.18/localization queue.

Reviewer checklist before merge: confirm the zh-Hans copy is explicit enough for destructive actions, verify English default behavior is unchanged, and keep the new approval render regression test.

axobase001:
Thanks! Confirmed on my side:

- zh-Hans copy explicitly calls out destructive approval and the two-step confirmation flow.
- English remains the default path, and the existing English approval render tests still pass.
- The new zh-Hans approval render regression test is kept.

Local validation:
- `cargo test -p deepseek-tui tui::approval::tests:: -- --nocapture`
- `cargo check -p deepseek-tui`

Hmbown:
Thanks @axobase001. I harvested this into draft release PR #1094 rather than merging the PR directly, so it can ship with the rest of v0.8.18 release prep.

What landed:
- commit `96e58d23f` localizes the zh-Hans destructive approval dialog copy
- the destructive warning remains explicit about risk
- English default behavior stays unchanged

Verification:
- `cargo test -p deepseek-tui tui::approval::tests:: --locked -- --nocapture` -> 50 passed, 0 failed
- included in the full v0.8.18 gate stack on #1094

This will close #1087 when #1094 merges. Thanks for the focused contribution.

Hmbown:
Thanks for this localization patch. Current main already contains the zh-Hans approval dialog localization and the linked issue #1087 is closed; rebasing this PR onto origin/main produced an empty diff. Closing this as already landed rather than merging a duplicate branch.

PR #1107, opened 2026-05-08
fix(npm): preserve quoted arguments on Windows (#1101)

### LINKED ISSUE #1101: deepseek exec cannot accept multi-word PROMPT on Windows (space-separated words parsed as subcommands)
## Environment
- OS: Windows 10/11
- Version: deepseek (npm wrapper) v0.8.16, binary v0.8.16
- Install method: npm install -g deepseek-tui

## Steps to reproduce
```bash
deepseek exec "hello world"
```

## Expected behavior
Executes with prompt "hello world".

## Actual behavior
```
error: unexpected argument 'world' found
Usage: deepseek-tui.exe exec [OPTIONS] <PROMPT>
```

Same with `deepseek "hello world"` and `deepseek -p "hello world"`.

Single-word prompts work fine:
```bash
deepseek exec "hello"  # works
```

## Impact
Cannot use DeepSeek TUI as a subprocess-based automation agent. Any prompt containing spaces (all real-world prompts do) is rejected. This blocks integration with multi-agent orchestration systems that pass prompts via `subprocess.Popen(["deepseek", "exec", "prompt_with_spaces"])`.

## Likely cause
clap argument parser treats the second space-separated token as a new argument rather than continuing the `[PROMPT]` positional. This may be specific to the Windows `.cmd` → binary invocation chain.

chnjames:
## Summary

- Fix multi-word prompts failing on Windows npm wrapper
- Preserve quoted arguments when invoking the binary via `.cmd` wrapper

## Problem

On Windows, `deepseek exec "hello world"` fails with:
```
error: unexpected argument 'world' found
Usage: deepseek-tui.exe exec [OPTIONS] <PROMPT>
```

The `.cmd` wrapper created by npm strips quotes, so the binary receives `hello` and `world` as separate arguments instead of a single `hello world` prompt.

## Solution

On Windows, use `execSync` with manual argument quoting instead of `spawnSync`. This preserves the original quoting from the user's command line.

```javascript
// Before
spawnSync(binaryPath, process.argv.slice(2), { stdio: "inherit" });

// After (Windows only)
const quotedArgs = args.map(arg => {
  if (arg.includes(" ") && !arg.startsWith('"')) {
    return `"${arg}"`;
  }
  return arg;
});
execSync(`"${binaryPath}" ${quotedArgs.join(" ")}`, { stdio: "inherit" });
```

## Changes

| File | Change |
|------|--------|
| `npm/deepseek-tui/scripts/run.js` | Use `execSync` with manual quoting on Windows |

## Testing

- `deepseek exec "hello world"` should work on Windows
- `deepseek -p "hello world"` should work on Windows
- Single-word prompts should continue to work
- Unix behavior is unchanged

Fixes #1101

### DIFF
--- a/npm/deepseek-tui/scripts/run.js
+++ b/npm/deepseek-tui/scripts/run.js
@@ -22,7 +22,29 @@ async function run(binaryName) {
   handleVersionFallback(binaryName);
 
   const binaryPath = await getBinaryPath(binaryName);
-  const result = spawnSync(binaryPath, process.argv.slice(2), {
+  const args = process.argv.slice(2);
+
+  // On Windows, .cmd wrappers strip quotes from arguments. Use execSync
+  // with manual quoting to preserve multi-word prompts. See issue #1101.
+  if (process.platform === "win32") {
+    const { execSync } = require("child_process");
+    const quotedArgs = args.map((arg) => {
+      // Quote arguments that contain spaces and aren't already quoted
+      if (arg.includes(" ") && !arg.startsWith('"') && !arg.startsWith("'")) {
+        return `"${arg.replace(/"/g, '\\"')}"`;
+      }
+      return arg;
+    });
+    const fullCmd = `"${binaryPath}" ${quotedArgs.join(" ")}`;
+    try {
+      execSync(fullCmd, { stdio: "inherit" });
+      process.exit(0);
+    } catch (err) {
+      process.exit(err.status ?? 1);
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for jumping on #1101. I’m closing this in favor of #1160, which fixes the Windows split-argv behavior in the Rust parsers instead of adding shell quoting in the npm wrapper. That avoids command-string quoting risk and covers `deepseek`, `deepseek -p`, and `deepseek exec` with parser tests.

PR #1126, opened 2026-05-08
fix(snapshot): cap disk usage by pruning old snapshots after each turn (#1112)

### LINKED ISSUE #1112: 产生了大量的快照（1.2TB .deepseek/snapshots）
## Description
我部署了最新版本的DeepSeek TUI，用了两天出现了大量莫名其妙的阻塞，我没有查到阻塞原因，但是只有很少的指令可以被执行，昨天发现磁盘空间满了，查询了磁盘空间占用情况发现1.2T	/Users/tycul/.deepseek/snapshots 。目前没有找到产生大量快照的原因。
## Steps to reproduce

1.
2.
3.

## Environment

- OS:
- DeepSeek CLI version:
- Model:
- Shell:

gordonlu:
## Problem

Closes #1112

Snapshots (`~/.deepseek/snapshots/`) could grow unbounded because pruning only ran at session startup. A long-running session (2 days) with 2 snapshots per turn (pre + post) could accumulate thousands of full-workspace git commits, filling the disk.

## Fix

**SnapshotRepo::prune_keep_last_n(max_count)** -- keeps only the latest N snapshots by rebuilding the survivor chain as orphan commits using `git commit-tree`.

**How it works:**
1. List snapshots (newest-first), keep the first `max_count` entries
2. Walk oldest to newest through the survivors, creating a new orphan chain:
   - The oldest survivor becomes a root commit (no `-p`), severing the parent link to earlier history
   - Each subsequent survivor becomes a child of the previous new commit, preserving the original tree and label
3. Point HEAD to the newest new commit
4. `git reflog expire --all` + `git gc --prune=now` reclaims the old chain -- zero refs lead to it

**Key design decisions:**
- Uses `commit-tree` (not `update-ref` + gc) because git gc cannot reclaim objects reachable through a parent chain
- Preserves each survivor's tree and commit message label
- Error-checking on every git command (status + stderr)

## Test

16 snapshot tests pass, including two new regression tests:
- `prune_keep_last_n_keeps_latest_and_gc_reclaims_rest` -- creates 3, keeps 1, verifies label + clean chain
- `prune_keep_last_n_preserves_multiple_snapshots_in_order` -- creates 4, keeps 2, verifies order

### DIFF
--- a/crates/tui/src/core/turn.rs
+++ b/crates/tui/src/core/turn.rs
@@ -156,13 +156,20 @@ pub fn post_turn_snapshot(workspace: &Path, turn_seq: u64) -> Option<String> {
 
 fn snapshot_with_label(workspace: &Path, label: &str) -> Option<String> {
     match SnapshotRepo::open_or_init(workspace) {
-        Ok(repo) => match repo.snapshot(label) {
-            Ok(id) => Some(id.0),
-            Err(e) => {
-                tracing::warn!(target: "snapshot", "snapshot '{label}' failed: {e}");
-                None
+        Ok(repo) => {
+            let id = match repo.snapshot(label) {
+                Ok(id) => Some(id.0),
+                Err(e) => {
+                    tracing::warn!(target: "snapshot", "snapshot '{label}' failed: {e}");
+                    return None;
+                }
+            };
+            // Prune oldest snapshots to cap disk usage (#1112).
+            if let Err(e) = repo.prune_keep_last_n(crate::snapshot::DEFAULT_MAX_SNAPSHOTS) {
+                tracing::warn!(target: "snapshot", "snapshot prune failed: {e}");
             }
-        },
+            id
+        }
         Err(e) => {
             tracing::warn!(target: "snapshot", "snapshot repo init failed: {e}");
             None
--- a/crates/tui/src/snapshot/mod.rs
+++ b/crates/tui/src/snapshot/mod.rs
@@ -40,5 +40,9 @@ pub mod repo;
 #[allow(unused_imports)]
 pub use paths::{snapshot_dir_for, snapshot_git_dir};
 pub use prune::{DEFAULT_MAX_AGE, prune_older_than};
+
+/// Maximum snapshots kept per workspace side-repo. Oldest are pruned
+/// after each new snapshot to cap disk usage (#1112).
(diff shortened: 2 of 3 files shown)


Hmbown:
Thanks for tackling #1112. This implementation is not safe to merge as written. `SnapshotRepo::list` returns newest-first, and snapshot commits are a normal parent chain where newest commits point to older parents. `prune_keep_last_n` sets HEAD to `snapshots[max_count]`; even setting HEAD to `snapshots[max_count - 1]` would move HEAD backward in history and make the newest snapshots unreachable. In other words, this prunes visibility of recent snapshots rather than reclaiming only old ancestors. This needs a different retention design plus a regression test that creates more than N snapshots and proves the latest N remain listable/restorable while older ones are reclaimed.

gordonlu:
> Thanks for tackling #1112. This implementation is not safe to merge as written. `SnapshotRepo::list` returns newest-first, and snapshot commits are a normal parent chain where newest commits point to older parents. `prune_keep_last_n` sets HEAD to `snapshots[max_count]`; even setting HEAD to `snapshots[max_count - 1]` would move HEAD backward in history and make the newest snapshots unreachable. In other words, this prunes visibility of recent snapshots rather than reclaiming only old ancestors. This needs a different retention design plus a regression test that creates more than N snapshots and proves the latest N remain listable/restorable while older ones are reclaimed.

Thanks for the detailed review! You were absolutely right about the retention order. I've fixed the logic in the latest commit by iterating `(0..keep).rev()` to rebuild the orphan chain from oldest survivor → newest, ensuring `HEAD` correctly points to the latest snapshot.

carloslindhstrom-code (commented):
🌷

carloslindhstrom-code:
// Åtgärd: Ladda sessionsfil
let result = load(&mut app, Some(save_path.to_str().unwrap()));

// Verifiering: Allt ska vara noll
assert_eq!(app.session.session_cost, 0.0);        // ✓
assert_eq!(app.session.session_cost_cny, 0.0);    // ✓
assert_eq!(app.session.subagent_cost, 0.0);       // ✓
assert_eq!(app.session.displayed_cost_high_water, 0.0); // ✓
assert!(app.session.turn_cache_history.is_empty()); // ✓

PR #1131, opened 2026-05-08
fix(snapshot): add max_size_mb hard cap to prevent unbounded growth (#1112)

Giggitycountless:
## Problem

The snapshot side-repo at `~/.deepseek/snapshots` can grow without bound. The only control is a 7-day age-based retention window at session start. For heavy users running hundreds of turns across multiple workspaces, the `.git` directory can balloon to **hundreds of gigabytes or even terabytes** before the next session-start prune kicks in (#1112).

## Root Cause

- **No total size limit** — only time-based retention (7 days)
- **No mid-session GC** — `gc.auto = 0` means objects accumulate between sessions
- **Each turn = 2 snapshots** — pre-turn + post-turn, with no intermediate cleanup

## Solution

Add a hard `max_size_mb` config cap (default 4 GiB) that runs at session start, BEFORE the age-based prune:

```toml
# ~/.deepseek/config.toml
[snapshots]
enabled = true
max_age_days = 7
max_size_mb = 4096  # NEW: hard size cap (0 = unlimited)
```

### How it works

1. `SnapshotRepo::total_size()` — walks the `.git` directory and sums file sizes
2. `prune::prune_over_size()` — progressively drops the oldest snapshots in halves (logarithmic convergence) until the directory fits within the budget
3. `SnapshotRepo::prune_to_count()` — fallback for same-second snapshots where age-based pruning can't distinguish entries; rewrites the commit chain to orphan older commits
4. Wired in `main.rs` — size prune runs first, then age prune

### Files changed (8 files, +518 lines)

| File | Change |
|---|---|
| `crates/config/src/lib.rs` | `SnapshotsToml` + `max_size_mb` field + default 4096 |
| `crates/tui/src/config.rs` | `SnapshotsConfig` + `max_size_mb` + `max_size_bytes()` |
| `crates/tui/src/snapshot/repo.rs` | `total_size()` + `prune_to_count()` + `dir_size_recursive()` |
| `crates/tui/src/snapshot/prune.rs` | `prune_over_size()` function |
| `crates/tui/src/snapshot/mod.rs` | Export new public API |
| `crates/tui/src/session_manager.rs` | `prune_workspace_snapshots_by_size()` |
| `crates/tui/src/main.rs` | Wire size prune before age prune at startup |
| `config.example.toml` | Document `max_size_mb` option |

### Tests (14 new, 31 total passing)

**Unit tests** (repo.rs):
- `total_size_returns_nonzero_after_snapshot`
- `total_size_zero_for_fresh_repo`
- `total_size_grows_with_more_snapshots`
- `dir_size_recursive_skips_symlinks`
- `dir_size_recursive_returns_zero_for_missing_path`

**Prune tests** (prune.rs):
- `prune_over_size_no_repo_returns_zero`
- `prune_over_size_zero_budget_means_unlimited`
- `prune_over_size_does_nothing_when_under_budget`
- `prune_over_size_removes_oldest_when_over_budget`
- `prune_over_size_preserves_newest_snapshots`
- `prune_over_size_converges_on_tiny_budget`

**Config tests** (config/src/lib.rs):
- `snapshots_toml_defaults_include_max_size_mb`
- `snapshots_toml_parses_max_size_mb_from_toml`
- `snapshots_toml_omitting_max_size_mb_uses_default`

### Backward compatibility

- Existing configs without `max_size_mb` get the default (4 GiB) — no breaking change
- `max_size_mb = 0` means unlimited (original behavior)
- All existing tests continue to pass

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -414,6 +414,7 @@ default_text_model = "deepseek-ai/deepseek-v4-pro"
 # [snapshots]
 # enabled = true        # Snapshot workspace pre/post each turn for /restore
 # max_age_days = 7      # Older snapshots pruned at session start
+# max_size_mb = 4096    # Hard limit: oldest snapshots pruned when exceeded (0 = unlimited)
 
 # ─────────────────────────────────────────────────────────────────────────────────
 # LSP Diagnostics (post-edit) (#136)
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -216,6 +216,12 @@ pub struct SnapshotsToml {
     pub enabled: bool,
     #[serde(default = "default_snapshot_max_age_days")]
     pub max_age_days: u64,
+    /// Hard upper limit on total snapshot directory size in MiB.
+    /// When exceeded the oldest snapshots are pruned at session start
+    /// until the directory fits within this budget. 0 means unlimited.
+    /// Default: 4096 (4 GiB).
+    #[serde(default = "default_snapshot_max_size_mb")]
+    pub max_size_mb: u64,
 }
 
 fn default_snapshots_enabled() -> bool {
@@ -226,11 +232,16 @@ fn default_snapshot_max_age_days() -> u64 {
     7
 }
 
+fn default_snapshot_max_size_mb() -> u64 {
+    4096
+}
+
 impl Default for SnapshotsToml {
     fn default() -> Self {
         Self {
             enabled: default_snapshots_enabled(),
             max_age_days: default_snapshot_max_age_days(),
+            max_size_mb: default_snapshot_max_size_mb(),
         }
     }
 }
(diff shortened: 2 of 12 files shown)


Hmbown:
Thanks for tackling #1112. I am not merging this version because the implementation appears to fail its own symlink contract: `dir_size_recursive` calls `entry.metadata()`, which follows symlinks, so `meta.is_symlink()` will not detect the symlink on Unix and the symlink target gets counted again. This should use `symlink_metadata()` before deciding whether to recurse or count a file. More importantly, the issue describes long-session growth, but this PR only wires size pruning at session startup; it does not add a mid-session prune/count guard, so the 1.2TB failure mode can still happen during a multi-day session. The next version should include a focused failing test for symlink accounting and a runtime pruning path, not only startup pruning.

Giggitycountless:
Thanks for the thorough review and the better implementation! Closing this since b29d63b covers #1112 with a tighter 500 MB cap. 🙏

PR #1133, opened 2026-05-08
  fix: wrap slash menu arrow navigation

### LINKED ISSUE #1113: 处于命令模式下，也就是 `/` 符号，按键方向 `↑`、`↓` 无法循环
嗨！
处于命令模式下，也就是 `/` 符号，按键方向 `↑`、`↓` 无法循环命令列表，而且  `↑` 方向按键到第一个之后，继续按  `↑` 会退出命令模式，操作起来不是特别方便。

Image

## **期待行为：**
1. 处于命令模式下，也就是 `/` 符号，按键方向 `↑`、`↓` 可以循环命令列表

reidliu41:
## Summary

  Fixes #1113.

 The issue reports that in slash command mode (`/`), the Up/Down arrow keys do not cycle through the command list. It also notes that pressing Up while the first command is selected exits the command menu.

  Root cause: the slash menu only handled Up when the selected index was greater than zero. Pressing Up on the first entry fell through to composer history navigation, which could replace `/` with a previous prompt and close the menu.

  This change makes slash menu Up/Down always consume plain arrow keys while the menu is open:

  - Up wraps from the first entry to the last.
  - Down wraps from the last entry to the first.
  - Existing non-menu arrow behavior remains unchanged.

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -2351,12 +2351,8 @@ async fn run_event_loop(
                 {
                     app.mention_menu_selected = app.mention_menu_selected.saturating_sub(1);
                 }
-                KeyCode::Up
-                    if key.modifiers.is_empty()
-                        && slash_menu_open
-                        && app.slash_menu_selected > 0 =>
-                {
-                    app.slash_menu_selected = app.slash_menu_selected.saturating_sub(1);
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks for the slash-menu wrap fix. I merged the harvested version in #1152, which includes the same user-facing behavior with focused tests on current main. Closing this duplicate to keep the queue clean.

PR #1146, opened 2026-05-08
feat(command-safety): block eval/exec as primary command

dumbjack:
## Summary

This PR adds detection for `eval` and `exec` when used as the **primary command** (first token), which can enable dynamic code execution attacks.

## Problem

The previous code had no protection against `eval` or `exec` being used as the main command. While the project had some safety checks, these dangerous commands were not blocked.

## Solution

Uses `extract_primary_command()` to check if the first token is `eval` or `exec`:
- Only blocks when `eval` or `exec` is the **primary command**
- Avoids false positives from words like `evaluate`, `evaluator.py`, `cargo run -- eval`

## Changes

**crates/tui/src/command_safety.rs**:
- Added eval/exec detection in `analyze_command()` function
- Added `test_eval_exec_as_primary_command_is_blocked` test

## Test Cases

| Command | Expected | Reason |
|---------|----------|--------|
| `eval $USER_INPUT` | Dangerous | Primary command is eval |
| `exec bash -c '...'` | Dangerous | Primary command is exec |
| `exec>file` | Dangerous | Primary command is exec |
| `cargo run --bin deepseek -- eval` | Safe | eval is an argument, not command |
| `python evaluator.py` | Safe | Different word, not eval |
| `primeval-checker` | Safe | Different word, contains "eval" |

## Compatibility

This fix addresses the concern raised in #706 about the original `command.contains("eval")` approach that was rejected for false positives. By using `extract_primary_command()`, we only block when `eval` or `exec` is truly the command being executed.

### DIFF
--- a/crates/tui/src/command_safety.rs
+++ b/crates/tui/src/command_safety.rs
@@ -622,6 +622,21 @@ pub fn analyze_command(command: &str) -> SafetyAnalysis {
         );
     }
 
+    // Check for eval/exec as the primary command - these can enable dynamic code execution
+    if let Some(primary) = extract_primary_command(command) {
+        let primary_lower = primary.to_lowercase();
+        if primary_lower == "eval" || primary_lower == "exec" {
+            return SafetyAnalysis::dangerous(
+                command,
+                vec![format!(
+                    "Command uses '{}' which can enable dynamic code execution",
+                    primary_lower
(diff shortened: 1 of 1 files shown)


dumbjack:
Closing this PR as requested. Will resubmit after local testing.

PR #1150, opened 2026-05-08
fix(undo): sync session context after snapshot restore

### LINKED ISSUE #1139: The slash command `/undo` seems like not revert the changes in sandbox.
## Description
I use `/undo` and `Esc-Esc` to rollback chat and files changes, but it seems like the files in my project path have been reverted, but the agent also can read the changes, so it just keep going on.

## Steps to reproduce

1. Let the agent change some files.
2. Rewind the chat to back.
3. Use `/undo`
4.  Let the agent continue.

## Expected behavior
The agent can read the old state of my project.

## Actual behavior
The agent read the state which is chaned.

## Environment

- OS: Apple M4 Pro macOS Tahoe Version 26.4.1 (25E253)
- DeepSeek CLI version: v0.8.18
- Model: v4-pro
- Shell: zsh

## Logs or screenshots

Image

JiarenWang:
## Summary

Fixes #1139.

This updates `/undo` and Esc-Esc backtrack so restoring workspace snapshots also keeps the model session context in sync. Previously, `/undo` could restore files while leaving the old tool result / assistant response in `api_messages`, so the next turn could continue from a stale "already changed" state.

Changes:
- Skip no-op snapshots so repeated `/undo` walks back through older snapshots instead of restoring the same snapshot repeatedly.
- Re-sync the engine session after snapshot restore and after Esc-Esc backtrack.
- Prune undone `tool:` context from the transcript/API message history while preserving earlier tool/result pairs in the same assistant turn.
- Prune `pre-turn:` context as a whole-turn rollback.
- Add regression coverage for repeated undo, tool-context pruning, pre-turn pruning, and session sync.

## Testing

Additional checks run:
- [x] `cargo fmt --all`
- [x] `cargo build`
- [x] `cargo test -p deepseek-tui commands::debug::tests -- --nocapture`

## Checklist

### DIFF
--- a/crates/tui/src/commands/debug.rs
+++ b/crates/tui/src/commands/debug.rs
@@ -7,7 +7,7 @@ use std::time::Instant;
 use super::CommandResult;
 use crate::compaction::estimate_input_tokens_conservative;
 use crate::localization::{Locale, MessageId, tr};
-use crate::models::{SystemPrompt, context_window_for_model};
+use crate::models::{ContentBlock, SystemPrompt, context_window_for_model};
 use crate::tui::app::{App, AppAction, TurnCacheRecord};
 use crate::tui::history::HistoryCell;
 
@@ -272,6 +272,7 @@ mod tests {
     use crate::config::Config;
     use crate::models::{ContentBlock, Message, SystemBlock};
     use crate::tui::app::{App, TuiOptions};
+    use crate::tui::history::{GenericToolCell, ToolCell, ToolStatus};
     use std::path::PathBuf;
 
     fn create_test_app() -> App {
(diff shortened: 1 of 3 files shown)


crates/tui/src/commands/debug.rs
```diff
@@ -7,7 +7,7 @@ use std::time::Instant;
 use super::CommandResult;
 use crate::compaction::estimate_input_tokens_conservative;
 use crate::localization::{Locale, MessageId, tr};
-use crate::models::{SystemPrompt, context_window_for_model};
+use crate::models::{ContentBlock, SystemPrompt, context_window_for_model};
```
mattss0140, line 10: Ctrl
Sig**_[]()profile583176.pdf

_**

PR #1151, opened 2026-05-08
test(config): add regression test for deepseek-cn beta endpoint

Liu-Vince:
## Summary

- v0.8.19 fixed `deepseek-cn` routing to `https://api.deepseek.com` instead of the beta endpoint (`https://api.deepseek.com/beta`), but no regression test was added to pin that fix.
- This PR adds `deepseek_cn_legacy_alias_routes_to_beta_endpoint` alongside the existing `deepseek_provider_defaults_to_beta_endpoint` test, so any future change that accidentally reverts `DEFAULT_DEEPSEEKCN_BASE_URL` to the non-beta URL will be caught immediately.

## Test plan

- [ ] `cargo test -p deepseek-tui deepseek_cn_legacy_alias_routes_to_beta_endpoint` passes
- [ ] No existing tests broken

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -4111,6 +4111,21 @@ api_key = "old-openrouter-key"
     }
 
     #[test]
+    fn deepseek_cn_legacy_alias_routes_to_beta_endpoint() {
+        let config = Config {
+            provider: Some("deepseek-cn".to_string()),
+            ..Default::default()
+        };
+
+        assert_eq!(config.api_provider(), ApiProvider::DeepseekCN);
+        assert_eq!(
+            config.deepseek_base_url(),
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for adding coverage here. I’m going to close this one because the asserted behavior is no longer the desired contract:  should default to the official host , not the  endpoint, unless a beta-only feature explicitly needs . Keeping this test would pin the old routing mistake instead of preventing it.

Hmbown:
Correction to my previous comment: the asserted behavior is no longer the desired contract. `deepseek-cn` should default to the official host `https://api.deepseek.com`, not the `/beta` endpoint, unless a beta-only feature explicitly needs `/beta`. Keeping this test would pin the old routing mistake instead of preventing it. Thanks again for adding coverage; this just needs to target the non-beta default instead.

PR #1155, opened 2026-05-08
fix(exec): accept multi-word prompts without quoting on Windows

### LINKED ISSUE #1101: deepseek exec cannot accept multi-word PROMPT on Windows (space-separated words parsed as subcommands)

Liu-Vince:
## Summary

Windows npm `.cmd` wrappers expand `%*` without preserving quotes, so `deepseek exec "hello world"` arrives at the Rust binary as two separate arguments (`hello` and `world`), and clap rejects the second word as unexpected.

- Changed `ExecArgs.prompt` from `String` to `Vec<String>` with `#[arg(trailing_var_arg = true, num_args = 1..)]`
- All remaining positional tokens are collected and joined with a space before being passed to the model
- Quoted single-arg usage on Unix/macOS is unaffected

Closes #1101

## Test plan

- [ ] `cargo check` passes
- [ ] `cargo test -p deepseek-tui -- exec` passes
- [ ] `deepseek exec hello world` and `deepseek exec "hello world"` should both send prompt `"hello world"` to the model

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -4111,6 +4111,21 @@ api_key = "old-openrouter-key"
     }
 
     #[test]
+    fn deepseek_cn_legacy_alias_routes_to_beta_endpoint() {
+        let config = Config {
+            provider: Some("deepseek-cn".to_string()),
+            ..Default::default()
+        };
+
+        assert_eq!(config.api_provider(), ApiProvider::DeepseekCN);
+        assert_eq!(
+            config.deepseek_base_url(),
+            DEFAULT_DEEPSEEK_BASE_URL,
+            "deepseek-cn legacy alias must use the beta endpoint, not https://api.deepseek.com"
+        );
+    }
+
+    #[test]
     fn explicit_deepseek_base_url_overrides_beta_default() {
         let config = Config {
             base_url: Some("https://api.deepseek.com".to_string()),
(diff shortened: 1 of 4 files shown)


Hmbown:
Thanks for the fix direction on #1101. I’m closing this as superseded by #1160, which keeps the change scoped to CLI argument parsing and avoids mixing in unrelated config/RLM files. The useful behavior here is preserved: split prompt words from Windows shims are joined before execution.

PR #1156, opened 2026-05-08
fix(tui): wrap slash menu navigation past first/last item

### LINKED ISSUE #1113: 处于命令模式下，也就是 `/` 符号，按键方向 `↑`、`↓` 无法循环

Liu-Vince:
## Summary

- Up arrow at the first slash menu entry now wraps to the last entry instead of falling through and accidentally exiting command mode
- Down arrow at the last slash menu entry now wraps back to the first entry

## Root cause

The `Up` handler had an extra `&& app.slash_menu_selected > 0` guard, so when already at index 0 the arm didn't match and the key event fell through to other handlers (which closed the menu). The `Down` handler used `.min()` to clamp at the last index, which silently ignored presses past the end.

## Fix

Removed the `> 0` guard from `Up` and added explicit wrap logic. Changed `Down` to use `%` modulo on the list length (safe because `slash_menu_open` guarantees a non-empty list).

Closes #1113

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -4111,6 +4111,21 @@ api_key = "old-openrouter-key"
     }
 
     #[test]
+    fn deepseek_cn_legacy_alias_routes_to_beta_endpoint() {
+        let config = Config {
+            provider: Some("deepseek-cn".to_string()),
+            ..Default::default()
+        };
+
+        assert_eq!(config.api_provider(), ApiProvider::DeepseekCN);
+        assert_eq!(
+            config.deepseek_base_url(),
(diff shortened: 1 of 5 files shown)


Hmbown:
Thanks for following up on the slash-menu wrap behavior. #1152 is now merged with the focused fix and regression tests, so I am closing this duplicate PR.

PR #1162, opened 2026-05-08
fix(fetch_url): add allow_private_ip_hosts to skip DNS-resolved IP check

### LINKED ISSUE #1071: fetch_url blocks all requests when DNS resolves to private IPs (fakeip proxy incompatible)
## Description

`fetch_url` unconditionally blocks outbound HTTP requests when the target hostname resolves to **any** private/reserved IP — including `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `100.64.0.0/10`, and critically `198.18.0.0/15`. This makes the tool completely unusable when the user runs a transparent proxy in **fakeip** mode (Clash, sing-box, etc.), because the proxy's DNS hijack returns a virtual internal IP for every domain.

The SSRF check conflates two very different scenarios:

1. **User explicitly types `http://10.0.0.1/admin`** — this is a genuine SSRF attempt and should be blocked.
2. **User fetches `https://github.com/...`, system DNS returns a fakeip address** — the user has a proxy, the network *does* work, and blocking here is a false positive with no workaround.

There is no config option to skip the IP check for DNS-resolved hosts. Even `network.default = "allow"` does not help — the SSRF IP guard and the per-domain network policy are two independent, serial gates, and the IP check has **no escape hatch**.

## Steps to reproduce

1. Set up a proxy with fakeip mode (e.g. Clash with `fake-ip-range: 198.18.0.1/16`).
2. Ensure system DNS is routed through the proxy.
3. In `deepseek`, try `/fetch https://github.com/Hmbown/DeepSeek-TUI/issues/1068`
4. Observe the error: `resolved IP 198.18.x.x is a restricted address (private/loopback/link-local)`

The same happens with `web_search` and any other tool that calls `fetch_url` internally.

## Expected behavior

DNS resolution to a private IP should not be treated as a network failure. The IP check is a blunt instrument — it cannot determine whether the address is reachable or not. A resolved private IP in a fakeip proxy setup is fully functional. There should be a config option to allow this, e.g.:

```toml
[network]
allow_resolved_private_ips = true
```

This should only affect hostname → IP resolution results. IP literals typed directly into the URL (e.g. `http://10.0.0.1/admin`) should remain blocked regardless.

## Actual behavior

Every request that resolves to an IP in `is_restricted_ip`'s hardcoded deny list is rejected with no way to override. Users behind fakeip proxies cannot use `fetch_url` at all.

## Why the current logic is unreasonable

The SSRF guard assumes "private IP → hostile intent." But DNS resolution is a system-level concern — whether an IP is reachable depends on the network topology, not on the address class. A resolved `198.18.x.x` behind a fakeip proxy is as legitimate as a resolved `140.82.x.x` on a direct connection. There is no security gain in blocking it; there is only lost functionality.

## Environment

- OS: Linux (Gentoo)
- DeepSeek CLI version: 0.8.16
- Model: deepseek-v4-flash
- Shell: fish
- Proxy: Clash / sing-box fakeip mode

## Relevant code

- **SSRF IP blocklist**: `crates/tui/src/tools/fetch_url.rs:297-330` (`is_restricted_ip`)
- **DNS resolution loop that calls it**: `fetch_url.rs:188-201`
- **No config integration**: the blocklist is hardcoded with zero configuration knobs
- **Network policy is ineffective here**: `fetch_url.rs:151-164` (domain check) and `175-202` (IP check) are independent serial gates

Liu-Vince:
## Problem

`fetch_url` unconditionally rejects requests when a hostname resolves to a private/reserved IP address. This breaks users whose DNS is managed by a transparent proxy that returns virtual private-range addresses for all external domains — every outbound request fails with `resolved IP x.x.x.x is a restricted address`, and there is no way to override it. Fixes #1071.

## Solution

Add `allow_private_ip_hosts = false` to the `[network]` config table. When set to `true`, the SSRF guard skips the check on **DNS-resolved** IPs, allowing the request to proceed through the proxy. Literal private-IP addresses typed directly into a URL (e.g. `http://10.0.0.1/admin`) are **always** blocked regardless of this setting, preserving the core SSRF protection.

The error message when a resolved IP is rejected now mentions the new option so users know how to unblock their setup.

## Changes

- `crates/tui/src/network_policy.rs` — add `allow_private_ip_hosts: bool` field to `NetworkPolicy`; expose `allows_private_ip_hosts()` on `NetworkPolicyDecider`
- `crates/tui/src/config.rs` — add `allow_private_ip_hosts` to `NetworkPolicyToml` and propagate through `into_runtime()`
- `crates/tui/src/tools/fetch_url.rs` — consult the flag before rejecting DNS-resolved private IPs; improve error message
- `config.example.toml` — document the new option
- `crates/tui/src/tools/web_run.rs`, `crates/tui/tests/skill_install.rs` — fill new field in existing `NetworkPolicy` literals

## Testing

All existing tests pass (`cargo test --package deepseek-tui`).

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -230,6 +230,9 @@ max_subagents = 10 # optional (1-20)
 # allow = ["api.deepseek.com", "github.com", ".githubusercontent.com"]
 # deny = []
 # audit = true            # one line per call to ~/.deepseek/audit.log
+# allow_private_ip_hosts = false  # set true when a trusted transparent proxy controls DNS
+#                                 # and returns virtual private-range IPs for external domains.
+#                                 # Direct private-IP literals in URLs remain blocked regardless.
 
 # ─────────────────────────────────────────────────────────────────────────────────
 # Skills (#140)
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -925,6 +925,12 @@ pub struct NetworkPolicyToml {
     /// Whether to record one audit-log line per outbound network call.
     #[serde(default = "default_network_audit")]
     pub audit: bool,
+    /// Allow `fetch_url` to proceed when a hostname resolves to a private/reserved
+    /// IP address. Off by default. Enable when a trusted transparent proxy
+    /// controls DNS and returns virtual private-range IPs for external domains.
+    /// Direct private-IP literals in URLs remain blocked regardless.
+    #[serde(default)]
+    pub allow_private_ip_hosts: bool,
 }
 
 fn default_network_decision() -> String {
@@ -942,6 +948,7 @@ impl Default for NetworkPolicyToml {
             allow: Vec::new(),
             deny: Vec::new(),
             audit: default_network_audit(),
+            allow_private_ip_hosts: false,
         }
     }
 }
@@ -956,6 +963,7 @@ impl NetworkPolicyToml {
             allow: self.allow,
             deny: self.deny,
             audit: self.audit,
+            allow_private_ip_hosts: self.allow_private_ip_hosts,
         }
     }
 }
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -4111,6 +4111,21 @@ api_key = "old-openrouter-key"
     }
 
     #[test]
+    fn deepseek_cn_legacy_alias_routes_to_beta_endpoint() {
+        let config = Config {
+            provider: Some("deepseek-cn".to_string()),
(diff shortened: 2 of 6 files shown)


Hmbown:
Thanks for the follow-up. I’m closing this as superseded by #1103, which has now merged with the safer per-host trusted proxy design. This PR also still includes the incorrect `deepseek-cn` beta-endpoint regression test from #1151, so it should not be merged as-is.

PR #1167, opened 2026-05-08
fix(deps): extend libc to all Unix targets for FreeBSD compatibility

### LINKED ISSUE #1143: deepseek-tui compilation fails on FreeBSD
## Description

Compilation of deepseek-tui (NOT -cli) fails on FreeBSD

## Steps to reproduce

1. cargo install deepseek-tui --locked

## Expected behavior

Installs

## Actual behavior

...
error[E0433]: cannot find module or crate `libc` in this scope
...

## Environment

- OS: FreeBSD 15.0-RELEASE-p7
- DeepSeek CLI version: N/A
- Model: N/A
- Shell: bash

🌐 eldanna in ~
❯ rustc --version
rustc 1.95.0 (59807616e 2026-04-14)
🌐 eldanna in ~
❯ cargo --version
cargo 1.95.0 (f2d3ce0bd 2026-03-21)

rustup-installed

Liu-Vince:
## Problem

Building `deepseek-tui` on FreeBSD fails with:

```
error[E0433]: cannot find module or crate `libc` in this scope
```

The `libc` crate was declared as a dependency only for `target_os = "macos"` and `target_os = "linux"`, so FreeBSD (and other BSDs) never get it resolved at compile time.

Closes #1143.

## Solution

Replace the two separate OS-specific entries with a single `cfg(unix)` dependency target. All existing call sites that use `libc` are already gated behind `#[cfg(unix)]` or narrower OS-specific guards (`#[cfg(target_os = "linux")]` for `prctl`/landlock, `#[cfg(target_os = "macos")]` for seatbelt), so this change is safe for macOS and Linux and fixes FreeBSD in one line.

## Changes

- `crates/tui/Cargo.toml` — merge `cfg(target_os = "macos")` + `cfg(target_os = "linux")` libc entries into a single `cfg(unix)` entry

## Testing

- Existing `cargo test --package deepseek-tui` passes on macOS/Linux (no behavioural change)
- FreeBSD users should now be able to `cargo install deepseek-tui --locked` without `libc` resolution errors

### DIFF
--- a/crates/tui/Cargo.toml
+++ b/crates/tui/Cargo.toml
@@ -75,10 +75,7 @@ wiremock = "0.6"
 pretty_assertions = "1.4"
 vt100 = "0.15"
 
-[target.'cfg(target_os = "macos")'.dependencies]
-libc = "0.2"
-
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(unix)'.dependencies]
 libc = "0.2"
 
 [target.'cfg(target_os = "windows")'.dependencies]


Hmbown:
Thanks for the FreeBSD fix. I harvested it into #1173 with the original author preserved, after rebasing onto current main and running cargo check -p deepseek-tui --locked plus git diff --check locally. The replacement branch will run the full normal CI matrix before merge.

PR #1181, opened 2026-05-08
docs(tui): clarify Windows mouse capture behavior

### LINKED ISSUE #1169: When selecting text with the mouse, the selection area crosses the sidebar and also selects content in the right sidebar. No isolation is implemented, which is not expected behavior.
## Description

When selecting text with the mouse, the selection area crosses the sidebar and also selects content in the right sidebar. No isolation is implemented, which is not expected behavior.

## Steps to reproduce

1. Launch the DeepSeek CLI.
2. Enter an interface that contains a sidebar (e.g., the conversation interface).
3. Drag with the left mouse button to select text starting from the main area on the left.
4. Continue dragging to the right and observe whether the right sidebar is also selected.

## Expected behavior

When selecting text, the main area and the sidebar should be isolated from each other. The selection should be limited to the current area and should not cross over to select content in the other area.

## Actual behavior

The selection crosses the sidebar, and content in the right sidebar is also selected. No isolation is applied.

## Environment

- OS: Windows 11
- DeepSeek CLI version: 0.8.20
- Model: v4-pro
- Shell: PowerShell

## Logs or screenshots

Image

Oliver-ZPLiu:
## Summary

Clarify the Windows mouse-capture behavior for transcript selection.

On Windows, mouse capture defaults off, so raw terminal selection may cross the right sidebar because the terminal owns the selection. This documents that `--mouse-capture` / `tui.mouse_capture = true` enables TUI-owned transcript selection, internal scrolling, and transcript scrollbar dragging.

Fixes #1169.

## Testing

- [x] Manually verified on Windows that `--mouse-capture` keeps selection scoped away from the sidebar.


Hmbown:
The documentation clarification is useful, but I do not want to merge it with Fixes #1169.

The issue reports a visible selection isolation problem. Because mouse capture defaults off on Windows, a docs-only change leaves the default symptom unchanged and would close the bug prematurely. This can be merged as Related to #1169 after removing the closing keyword, or replaced with an actual behavior change that makes the selection path safer for Windows users.

PR #1192, opened 2026-05-08
fix(tui): persist session cost across save/restore cycles

lbcheng888:
## Summary

Session cost (USD/CNY) was **always reset to zero** on every session restore because only `total_tokens` was persisted to `SessionMetadata`. This meant resumed sessions showed a $0.00 cost even after spending significant money in prior turns.

## Changes

- **Persist cost in `SessionMetadata`**: Added `session_cost_usd`, `session_cost_cny`, `subagent_cost_usd`, `subagent_cost_cny` fields (backward-compatible via `#[serde(default)]`)
- **Save path** (`build_session_snapshot`, `/save`, `/rename`, fork): Write current cost accumulators to metadata
- **Restore path** (`apply_loaded_session`): Restore costs from metadata instead of zeroing them
- **High-water mark**: Initialize from restored totals so the monotonic guarantee survives session restarts
- **Header consistency**: Pass full displayed cost (session + subagent, in chosen currency) instead of just `session_cost`

## Backward compatibility

Old session files without cost fields deserialize to 0.0 — behavior is identical to before this fix.

## Test plan

- [ ] Start a session, make a few turns, note the cost in the footer
- [ ] Save and reload the session — cost should match
- [ ] Save session, quit, restart, reload session — cost persists
- [ ] `/save` exports with correct cost in JSON metadata
- [ ] `/rename` preserves cost


carloslindhstrom-code:
d1187e2b8fc73fe92d28d2a7b327e80be420aed2

PR #1201, opened 2026-05-08
Fix provider-aware /model picker catalogs

eng2007:
## Summary

Fixes the `/model` picker so it reflects the currently selected provider instead of always showing the static DeepSeek Pro/Flash list.

Changes:
- fetch live model IDs from the active provider when opening `/model`
- fall back to provider-specific defaults when live discovery fails or returns no models
- normalize `/model <id>` through the active provider, so short DeepSeek IDs map to provider IDs such as `deepseek-ai/deepseek-v4-flash` for NVIDIA NIM
- keep custom/current model IDs visible when they are not present in the picker list
- make the picker scrollable for long model catalogs
- size picker columns from rendered content: the model pane follows the longest model ID, and the thinking pane keeps `high  (thinking enabled)` readable with one trailing cell before the border
- add an Unreleased changelog entry
- silence dead-code warnings for schema migration registry marker structs so strict clippy can pass with `-D warnings`

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- [x] `cargo build`

Local environment note:
- `cargo run -p deepseek-tui --all-features -- eval` fails on Windows because the offline eval harness runs `printf`, which is not available under `cmd.exe`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **`/model` uses the active provider's model catalog** - opening the picker now
+  fetches live model IDs from the configured provider, falls back to
+  provider-specific defaults when discovery fails, maps short DeepSeek IDs to
+  provider-specific IDs for `/model <id>`, and sizes the picker columns to keep
+  long provider model names and thinking-effort labels readable.
+
 ## [0.8.20] - 2026-05-08
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


eng2007:
Addressed in `8026420`.

Provider-native IDs returned by live catalogs now pass settings validation and survive restart, while DeepSeek aliases still normalize. The picker persistence path also surfaces `Settings::set` failures instead of ignoring them. Added coverage for `meta/...` and `ft:gpt-...` model IDs plus unsafe ID rejection.

Validated locally:
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo test --workspace --all-features --locked`

Hmbown:
The provider-aware /model picker catalog fix was harvested into #1649 and merged to main with contributor credit. For v0.8.38, that narrow bug-fix slice is covered; leaving this PR open for maintainer judgment on any remaining scope.

Hmbown:
Thanks for the provider/model picker work. We harvested the narrow provider-selected-model preservation pieces into #1649, but the live catalog behavior in /model caused a v0.8.38 regression: a common picker path could block on provider network calls and lost the curated DeepSeek tier list. v0.8.39 reverted that path. Closing this PR so it is not accidentally revived; future provider catalog work should target explicit /models behavior or an async/cached picker design with regression tests.

PR #1203, opened 2026-05-08
Retry quota errors returned as HTTP 400

dst1213:
## Summary
- classify HTTP 400 quota/rate-limit gateway responses as `LlmError::RateLimited`
- let `send_with_retry()` route all non-success responses through unified `LlmError` classification
- add regression coverage for `400 + insufficientquota` being retryable

## Verification
- `git diff --check` ok
-  build and check on Mac OS，it's OK

User-provided verification notes report the equivalent patch passed `cargo fmt --all`, `cargo test -p deepseek-tui test_llm_error_from_http_response`, and `git diff --check` locally.

### DIFF
--- a/crates/tui/src/client.rs
+++ b/crates/tui/src/client.rs
@@ -648,10 +648,6 @@ impl DeepSeekClient {
                     if status.is_success() {
                         return Ok(response);
                     }
-                    let retryable = status.as_u16() == 429 || status.is_server_error();
-                    if !retryable {
-                        return Ok(response);
-                    }
                     let retry_after = extract_retry_after(response.headers());
                     let body = bounded_error_text(response, ERROR_BODY_MAX_BYTES).await;
                     Err(LlmError::from_http_response_with_retry_after(
--- a/crates/tui/src/llm_client/mod.rs
+++ b/crates/tui/src/llm_client/mod.rs
@@ -202,7 +202,16 @@ impl LlmError {
             400 => {
                 // Classify 400 errors by examining the response body
                 let body_lower = body.to_lowercase();
-                if body_lower.contains("context_length")
+                if body_lower.contains("insufficientquota")
+                    || body_lower.contains("insufficient_quota")
+                    || body_lower.contains("exceeded your current quota")
+                    || body_lower.contains("quota exceeded")
+                {
+                    LlmError::RateLimited {
+                        message: body.to_string(),
+                        retry_after: None,
+                    }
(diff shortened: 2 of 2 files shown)


Hmbown:
Thanks for the patch. I am leaving this unmerged for now because the retry classification is too broad.

Blocking issue: insufficient quota is generally not retryable. This patch classifies quota text as RateLimited, which makes err.is_retryable() true and can spend time retrying an exhausted account instead of giving the user a direct billing/quota error. The client.rs change also routes all non-success responses through the retry/error path, so the patch needs regression coverage that context length, content policy, auth/config, and generic 400 responses still fail promptly and do not retry.

Safer shape: keep real 429/rate-limit responses retryable, but classify insufficient_quota / exceeded current quota as a distinct non-retryable quota/billing error or another non-retryable user-actionable error.

dst1213:
Thanks for the review, and fair point — insufficient quota is indeed different from a transient rate-limit.
My thinking was that in large companies, these quota/rate-limit errors are often global rather than individual: thousands of engineers share a single third-party API key with a fixed total TPM. During peak hours, hitting the cap is frequent, and manually retrying is tedious.
That said, I understand the concern about masking billing/quota issues. I'm fine with leaving this unmerged — just thought other orgs with similar setups might run into the same friction. No worries either way!



---- Replied Message ----
| From | Hunter ***@***.***> |
| Date | 05/08/2026 22:07 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [Hmbown/DeepSeek-TUI] Retry quota errors returned as HTTP 400 (PR #1203) |
Hmbown left a comment (Hmbown/DeepSeek-TUI#1203)

Thanks for the patch. I am leaving this unmerged for now because the retry classification is too broad.

Blocking issue: insufficient quota is generally not retryable. This patch classifies quota text as RateLimited, which makes err.is_retryable() true and can spend time retrying an exhausted account instead of giving the user a direct billing/quota error. The client.rs change also routes all non-success responses through the retry/error path, so the patch needs regression coverage that context length, content policy, auth/config, and generic 400 responses still fail promptly and do not retry.

Safer shape: keep real 429/rate-limit responses retryable, but classify insufficient_quota / exceeded current quota as a distinct non-retryable quota/billing error or another non-retryable user-actionable error.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #1211, opened 2026-05-08
fix(tui): scroll transcript when composer is empty, not history

lbcheng888:
## Summary

#1117 changed plain Up/Down to always navigate input history, but this broke transcript scrolling for terminals that map trackpad gestures to arrow key events.

Instead of reverting #1117 (which would regress the keyboard history-recall workflow), this PR adds an opt-in config option:

```toml
[tui]
composer_arrows_scroll = true
```

When enabled, empty-composer Up/Down scroll the transcript instead of recalling history.  Default behavior is unchanged (#1117).

## Changes

- `TuiConfig::composer_arrows_scroll` — new config field (default `false`)
- `App::composer_arrows_scroll` — runtime flag read from config at startup
- `handle_composer_history_arrow` — gates scroll-on-empty behind the option
- Updated tests: original tests verify default history behavior; new tests cover the opt-in scroll path

## Test plan

- [ ] Default: empty composer + Up → recalls last prompt
- [ ] Opt-in: empty composer + Up → scrolls transcript up
- [ ] Opt-in: non-empty composer + Up → still navigates input history
- [ ] Trackpad two-finger scroll on affected terminals → transcript scrolls (when opt-in enabled)

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -388,6 +388,12 @@ pub struct TuiConfig {
     /// - `Never` — suppress all turn-completion notifications.
     /// - Unset (default) — fall back to the `[notifications]` defaults.
     pub notification_condition: Option<NotificationCondition>,
+    /// When `true`, plain Up/Down on an empty composer scroll the
+    /// transcript instead of recalling input history.  Useful for
+    /// terminals that map trackpad gestures to arrow keys.  Default:
+    /// `false` (plain arrows always navigate input history, #1117).
+    #[serde(default)]
+    pub composer_arrows_scroll: Option<bool>,
 }
 
 /// High-level notification trigger override. See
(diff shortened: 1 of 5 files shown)


Hmbown:
Thanks for the targeted patch. I do not think we should merge this exact behavior as-is: plain keyboard Up on an empty composer is also the standard input-history recall path, and the KeyCode::Up event does not tell us whether it came from a real keyboard arrow or from a terminal mapping trackpad scroll to arrows. This would fix one scroll setup by regressing another common workflow.

The useful direction is still valid, but it needs a way to avoid stealing empty-composer history recall, such as terminal-specific handling, a separate option, or a mouse/scroll path that does not collapse to the same key event. I am leaving this open rather than merging the current head.

lbcheng888:
@Hmbown

Hmbown:
Thank you so much I keep accidentally doing this

PR #1222, opened 2026-05-08
SseTransport Race Condition Fix for deepseek-tui

djairjr:
### Symptom
SSE-based MCP servers (such as open-brain/FastMCP) appear as "disabled" in the TUI or fail with: `"SSE endpoint not yet discovered"`

### Root Cause
`SseTransport::connect()` returned immediately after spawning the SSE background loop, with `endpoint_url: None`. The `send()` method requires `endpoint_url` to be populated, but this only happens during `recv()`. Since `initialize()` calls `send()` before `recv()`, the endpoint was never available in time.

### Fix Applied
1. **New method `wait_for_endpoint()`**: polling loop using `try_recv()` that waits for the `__internal_sse_endpoint__` event to arrive through the mpsc channel. Timeout: 10 seconds (200 attempts × 50ms). Returns a clear error if the endpoint doesn't arrive within the deadline.

2. **Modified `connect()`**: after spawning the SSE loop, calls `transport.wait_for_endpoint().await?` before returning. This guarantees `endpoint_url` is populated when `send()` is subsequently called.

### Reference
The Python MCP SDK (`mcp.client.sse.sse_client`) solves the same problem using `TaskStatus.started(endpoint_url)` — it waits for the endpoint event before starting the writer task. The Rust fix follows the same principle.

## Testing

## Checklist

### Fix Code

```rust
// Before (BUG):
Ok(Self {
    client,
    base_url: url,
    endpoint_url: None,  // <- NOT YET DISCOVERED
    receiver: rx,
})

// After (FIXED):
let mut transport = Self {
    client,
    base_url: url,
    endpoint_url: None,
    receiver: rx,
};
transport.wait_for_endpoint().await?;  // <- WAIT FOR ENDPOINT
Ok(transport)
```

### New Method: wait_for_endpoint()

```rust
/// Wait for the SSE endpoint to be discovered.
///
/// The first event on an MCP SSE stream is always "endpoint" with the
/// POST URL for client messages. This method reads internal messages
/// from the channel until the endpoint is received, then stores it.
async fn wait_for_endpoint(&mut self) -> Result<()> {
    use tokio::sync::mpsc::error::TryRecvError;
    let mut attempts = 0;
    const MAX_ATTEMPTS: u32 = 200;
    const ATTEMPT_DELAY_MS: u64 = 50;

    loop {
        match self.receiver.try_recv() {
            Ok(msg) => {
                if let Some(endpoint) = msg.get("__internal_sse_endpoint__") {
                    let url_str = endpoint.as_str()
                        .context("Invalid endpoint format")?;
                    if url_str.starts_with("http") {
                        self.endpoint_url = Some(url_str.to_string());
                    } else {
                        let base = reqwest::Url::parse(&self.base_url)?;
                        let joined = base.join(url_str)?;
                        self.endpoint_url = Some(joined.to_string());
                    }
                    return Ok(());
                }
            }
            Err(TryRecvError::Empty) => {
                attempts += 1;
                if attempts >= MAX_ATTEMPTS {
                    anyhow::bail!(
                        "SSE endpoint not received after {}ms",
                        MAX_ATTEMPTS * ATTEMPT_DELAY_MS
                    );
                }
                tokio::time::sleep(std::time::Duration::from_millis(ATTEMPT_DELAY_MS)).await;
            }
            Err(TryRecvError::Disconnected) => {
                anyhow::bail!("SSE transport channel closed before endpoint received");
            }
        }
    }
}
```

### DIFF
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -402,12 +402,25 @@ impl SseTransport {
             }
         });
 
-        Ok(Self {
+        let mut transport = Self {
             client,
             base_url: url,
             endpoint_url: None,
             receiver: rx,
-        })
+        };
+
+        // FIX (#XXX): Wait for the SSE endpoint event before returning.
+        // The MCP SSE protocol sends an "endpoint" event as the first
+        // message, containing the URL for client POST requests.
+        // Previously connect() returned immediately, creating a race
+        // where send() could be called before the endpoint was
+        // discovered, causing "SSE endpoint not yet discovered".
+        //
+        // Reference: mcp.client.sse.sse_client in Python MCP SDK
+        // uses TaskStatus.started(endpoint_url) for the same purpose.
+        transport.wait_for_endpoint().await?;
+
+        Ok(transport)
     }
 
     async fn run_sse_loop(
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for the clear diagnosis. I landed the SSE endpoint race fix in #1225 with a focused async regression test instead of merging this patch directly.

What changed in #1225:
- `SseTransport::connect()` now waits for the SSE `endpoint` event before returning.
- Early non-endpoint messages are buffered instead of dropped while endpoint discovery is pending.
- Endpoint discovery uses the configured MCP connect timeout and preserves cancellation behavior.
- Test: `cargo test -p deepseek-tui sse_connect_waits_for_endpoint_before_first_send --locked`

Closing this PR as superseded by #1225.

PR #1233, opened 2026-05-08
fix(cli): forward --yolo flag to TUI binary for run command

### LINKED ISSUE #1221: README.md is outdated
## Problem

The current `README.md` documentation is significantly outdated and inconsistent with the latest codebase. Key discrepancies include:
- The "yolo" mode mentioned in the README no longer exists in the command-line interface
- There is no way to specify the thinking effort via command-line arguments
- These inconsistencies prevent users from performing one-off tasks as intended, since the documented workflows are no longer functional

## Proposed solution

Please update the `README.md` to accurately reflect the current CLI capabilities and available commands/parameters. Alternatively, restore the missing functionality (yolo mode, effort parameter) to match the existing documentation.

## Alternatives considered

I've tried exploring the codebase to find replacement commands or parameters, but haven't found clear alternatives that provide the same functionality described in the README.

## Additional context

This is particularly blocking for users who rely on the documentation to quickly use the tool for one-off tasks without deep diving into the source code.

fuleinist:
## Problem

The  flag only worked when passed through the dispatcher subcommands (e.g., ), but failed silently for the primary  usage shown in README:

This is because  delegates to the TUI binary without forwarding  as an env var. The other subcommands (Exec, Doctor, etc.) pass CLI flags via , which forward everything, but  goes through  which only set env vars for config-level options (model, api_key, sandbox_mode, etc.), not top-level flags like .

## Root cause

In , the  pattern was only applied when invoking the TUI via , not when delegating plain . The  flag on  was never forwarded as an env var to the TUI binary.

## Fix

1. Add  env var in  when  is set
2. TUI  already reads  env (matching the  pattern)
3. Add  field to , , and  to propagate through the full resolve chain
4. Add  to TUI  struct and merge function
5. Add  to 

## Testing

-  passes with no new errors
- The fix follows the existing pattern used for , , and 

## Related

Fixes #1221

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,6 +37,7 @@
 # DEEPSEEK_APPROVAL_POLICY=on-request
 # DEEPSEEK_SANDBOX_MODE=workspace-write
 # DEEPSEEK_ALLOW_SHELL=true
+# DEEPSEEK_YOLO=true
 
 # Optional extension paths
 # DEEPSEEK_SKILLS_DIR=~/.deepseek/skills
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -92,6 +92,9 @@ struct Cli {
     no_mouse_capture: bool,
     #[arg(long = "skip-onboarding")]
     skip_onboarding: bool,
+    /// YOLO mode: auto-approve all tools
+    #[arg(long)]
+    yolo: bool,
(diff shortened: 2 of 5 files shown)


Hmbown:
ah - thank you so much. I've been struggling to get this down. I think introducing keychain with macOS early on really screwed this up. thank you!!

PR #1247, opened 2026-05-09
  refactor: unify mode switching under /mode

reidliu41:
## Summary

  This PR unifies the mode-switching commands into a single `/mode` command.

  Previously, `/agent`, `/plan`, and `/yolo` were separate commands, but they all did the same kind of operation: switch
  DeepSeek TUI into a different operating mode. Keeping those as separate top-level commands makes the command surface
  more scattered and repeats similar behavior across multiple entries.

  Now `/mode` is the single place for mode switching:

  - `/mode` opens a picker for Agent, Plan, and YOLO
  - `/mode agent` or `/mode 1` switches to Agent mode
  - `/mode plan` or `/mode 2` switches to Plan mode
  - `/mode yolo` or `/mode 3` switches to YOLO mode

  This makes the behavior easier to understand as switching between modes, instead of treating each mode as a separate
  command.

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/commands/config.rs
+++ b/crates/tui/src/commands/config.rs
@@ -538,30 +538,40 @@ pub fn set_config(app: &mut App, args: Option<&str>) -> CommandResult {
     set_config_value(app, &key, value, should_save)
 }
 
-/// Enable YOLO mode (shell + trust + auto-approve)
-pub fn yolo(app: &mut App) -> CommandResult {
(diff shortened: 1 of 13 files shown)


Hmbown:
Thanks @reidliu41 — `git cherry main` confirms this is patch-equivalent to a commit already on `main`. The unified `/mode` command is the headline-Added entry in the v0.8.27 CHANGELOG.

Closing as already-shipped.

PR #1265, opened 2026-05-09
fix: respect active provider credentials during onboarding

jinpengxuan:
fix: respect active provider credentials during onboarding

## Summary
- make onboarding's `has_api_key` check use the active provider instead of only DeepSeek root credentials
- preserve DeepSeek root `api_key` compatibility while honoring provider-scoped config and env credentials
- add regression coverage for active-provider config/env detection and DeepSeek root key handling

## Test Plan
- [x] `cargo test -p deepseek-tui has_api_key_uses_active_provider -- --nocapture`
- [x] `cargo test -p deepseek-tui has_api_key_for_uses_root_config_key_for_deepseek_variants -- --nocapture`
- [x] `cargo test -p deepseek-tui has_api_key_uses_root_config_key_for_deepseek_variants -- --nocapture`
- [ ] `cargo test -p deepseek-tui` currently fails in existing Python/RLM-related tests and one skills test in this environment

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -2852,13 +2852,8 @@ reasoning_effort = "max"
     Ok(config_path)
 }
 
-/// Check if an API key is configured anywhere the runtime can resolve it.
-///
-/// Order of inspection:
-///   1. `DEEPSEEK_API_KEY` env var (fast, no I/O, no OS prompts).
-///   2. In-memory override on the config (set by onboarding / picker).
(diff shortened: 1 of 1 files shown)


Hmbown (approved):
Reviewed provider-aware onboarding credential path. Verified focused config tests, cargo fmt, clippy, and the deepseek-tui binary test suite on a local merge with current main; the two full-suite failures passed in isolation as known parallel races.

PR #1271, opened 2026-05-09
fix(tui): clear Todos panel when /clear is invoked

### LINKED ISSUE #1258: bug: /clear does not clear the Todos (checklist) sidebar panel
## Description

## Steps to reproduce

4. Run `/clear`.

## Expected behavior

## Actual behavior

## Environment

- Platform: Linux
- Installation: npm wrapper

## Suggested fix

## Related issues

Liu-Vince:
## Problem

`/clear` resets conversation history, plan panel, and tasks panel, but the **Todos sidebar** retains its checklist items across successive `/clear` commands — they only disappear on app restart.

Fixes #1258

## Root cause

`clear_todos()` in `app.rs` locks `plan_state` and resets it, but never touches `self.todos` (the `SharedTodoList`). The Todos sidebar reads from that same `SharedTodoList`, so it was never cleared.

## Fix

After resetting `plan_state`, also lock `self.todos` and call `.clear()` on the inner `TodoList`. Mirrors how `plan_state` is already handled in the same function.

```rust
// before
pub fn clear_todos(&mut self) -> bool {
    if let Ok(mut plan) = self.plan_state.try_lock() {
        *plan = crate::tools::plan::PlanState::default();
        return true;
    }
    false
}

// after
pub fn clear_todos(&mut self) -> bool {
    if let Ok(mut plan) = self.plan_state.try_lock() {
        *plan = crate::tools::plan::PlanState::default();
        if let Ok(mut todos) = self.todos.try_lock() {
            todos.clear();
        }
        return true;
    }
    false
}
```

## Tests

- Extended the existing `clear_todos_resets_plan_state` test to also assert the todo list is empty after clearing.
- Added a dedicated `clear_todos_resets_todo_list` test that seeds two todo items, calls `clear_todos()`, and asserts `snapshot().items.is_empty()`.

Both tests pass.

### DIFF
--- a/crates/tui/src/tui/app.rs
+++ b/crates/tui/src/tui/app.rs
@@ -3694,6 +3694,9 @@ impl App {
     pub fn clear_todos(&mut self) -> bool {
         if let Ok(mut plan) = self.plan_state.try_lock() {
             *plan = crate::tools::plan::PlanState::default();
+            if let Ok(mut todos) = self.todos.try_lock() {
+                todos.clear();
+            }
             return true;
         }
         false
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @Liu-Vince — this fix shipped in v0.8.24 as commit `5670034d8 fix(todos): /clear now resets the Todos sidebar (#1258)`. The merged version is in `App::clear_todos()` and clears both the SharedTodoList and `plan_state` via `try_lock`. Closing as superseded; appreciate you flagging it.

PR #1272, opened 2026-05-09
Fix/v0.8.23:shell output drain and Windows Terminal scrolling

Acidmoon:
## Summary

  This fixes two TUI issues:

  - Prevent shell tools from freezing the render loop when commands such as `git push` leave stdout/stderr pipes open after the main child process exits.
  - Restore mouse-wheel transcript scrolling in Windows Terminal by enabling mouse capture there by default, while keeping legacy Windows console and JetBrains JediTerm protections.

  ## Changes

  - Move foreground/background shell stdout and stderr readers to async Tokio tasks when running inside a Tokio runtime.
  - Add a 2s output drain timeout so completed shell jobs do not block indefinitely waiting for reader shutdown.
  - Avoid blocking the UI task panel refresh on the shell manager lock by using `try_lock()`.
  - Detect Windows Terminal via `WT_SESSION` and default mouse capture on there so wheel events reach the TUI scroll handler.
  - Keep `--mouse-capture`, `--no-mouse-capture`, and `[tui].mouse_capture` overrides intact.

  ## Validation

  - `cargo fmt --all --check`
  - `cargo check -p deepseek-tui --message-format short`
  - `cargo test -p deepseek-tui mouse_capture --message-format short`
  - `cargo test -p deepseek-tui tools::shell::tests --message-format short`

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -3668,19 +3668,29 @@ fn should_use_mouse_capture_with(
 
 /// Whether to enable terminal mouse capture by default for this platform/host.
 ///
-/// Returns `false` on Windows (legacy console mouse-mode reporting is flaky;
-/// `--mouse-capture` opts in) and on JetBrains' JediTerm, which advertises
-/// mouse support but delivers SGR mouse-event escape sequences as raw text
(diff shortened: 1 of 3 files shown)


Hmbown:

Issue #1263, opened 2026-05-09
输入框支持Home,End等编辑键


cacerzard:

Hmbown:
Duplicate of #1234, fixed on `main` via @heloanc's #1246 — Home/End now move the composer cursor to the start/end of the current line. Closing as dup. Thanks for filing!

PR #1246, opened 2026-05-09
fix(tui): make Home/End keys move cursor in input box

heloanc:
## Summary

- Home/End keys now move the cursor to the start/end of the input line (standard text editing behavior)
- Ctrl+Home/End retains the previous scroll-to-top/bottom functionality
- Updated keybindings help overlay to reflect the new mappings

## Motivation

Closes #1234

When the cursor is in the input box, pressing Home/End should move the cursor to the beginning/end of the input text. Previously, these keys scrolled the transcript instead, which is unexpected for most users.

## Changes

- `crates/tui/src/tui/ui.rs`: Swap Home/End (no modifier) to call `move_cursor_start()`/`move_cursor_end()`, and Ctrl+Home/End to scroll transcript
- `crates/tui/src/tui/keybindings.rs`: Update help overlay — Navigation section shows "Ctrl+Home / Ctrl+End", Editing section adds "Home / End"

## Test plan

- [ ] `cargo build` compiles without errors
- [ ] Press Home/End in input box → cursor jumps to start/end
- [ ] Press Ctrl+Home/End → transcript scrolls to top/bottom
- [ ] Ctrl+A / Ctrl+E still work as before

PR #1275, opened 2026-05-09
fix: handle Home and End in composer input

### LINKED ISSUE #1234: 希望输入框支持 HOME/END 键回到输入框首尾
版本：0.8.21
希望可以支持输入框的HOME/END键回到首尾的逻辑。现在如果光标在输入框首，按End是回不到末尾的，操作不方便...

Wenjunyun123:
## Why

The composer currently reserves plain Home/End for transcript navigation before checking whether the user is editing text. That makes the common editing action from #1234/#1263 surprising: pressing Home or End while a draft has focus should move the composer cursor to the start or end of that draft.

This keeps the existing transcript behavior when the composer is empty, so plain Home/End still jump to the top/bottom of the transcript in browsing mode.

Fixes #1234.
Refs #1263.

## Validation

- `git diff --check`
- Not run: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo test --workspace --all-features --locked` because this machine does not currently have `cargo`/`rustc`; two Homebrew Rust install attempts stalled while downloading Homebrew API metadata.

### DIFF
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -2696,6 +2696,9 @@ async fn run_event_loop(
                     app.move_cursor_right();
                 }
                 KeyCode::Home if key.modifiers.is_empty() => {
+                    if handle_plain_home_end(app, key.code) {
+                        continue;
+                    }
                     if let Some(anchor) =
                         TranscriptScroll::anchor_for(app.viewport.transcript_cache.line_meta(), 0)
                     {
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks @Wenjunyun123 — closing this in favor of #1246 (heloanc), which the maintainer has flagged as the canonical solution for #1234. #1246 takes a slightly different approach: it inverts the binding so plain Home/End edit the cursor and Ctrl+Home/End scroll the transcript, and updates the keybindings help overlay accordingly. Your unit tests in this PR are nicely written — if you'd like to follow up on #1246 with a similar test pair on top of its inversion approach, that would land directly. Really appreciate the work here either way.

PR #1276, opened 2026-05-09
fix: clear todos with conversation reset

### LINKED ISSUE #1258: bug: /clear does not clear the Todos (checklist) sidebar panel

Wenjunyun123:
## Why

`/clear` already resets conversation history, telemetry, queued drafts, and the plan panel, but it only reset `plan_state` inside `App::clear_todos()`. The shared `app.todos` list that backs the Todos sidebar was left untouched, so checklist items stayed visible until restart.

This makes `clear_todos()` clear both the plan state and the shared todo list, so the sidebar matches the cleared conversation state.

Fixes #1258.

## Validation

- `git diff --check`
- Not run: `cargo test -p deepseek-tui tui::app::tests::clear_todos_resets_plan_and_todo_state --all-features`, `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo test --workspace --all-features --locked` because this machine does not currently have `cargo`/`rustc`; two Homebrew Rust install attempts stalled while downloading Homebrew API metadata.

### DIFF
--- a/crates/tui/src/tui/app.rs
+++ b/crates/tui/src/tui/app.rs
@@ -3692,11 +3692,16 @@ impl App {
     }
 
     pub fn clear_todos(&mut self) -> bool {
+        let mut cleared = false;
         if let Ok(mut plan) = self.plan_state.try_lock() {
             *plan = crate::tools::plan::PlanState::default();
-            return true;
+            cleared = true;
         }
-        false
+        if let Ok(mut todos) = self.todos.try_lock() {
+            todos.clear();
+            cleared = true;
+        }
+        cleared
     }
 
     pub fn update_model_compaction_budget(&mut self) {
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @Wenjunyun123 — this fix shipped in v0.8.24 as commit `5670034d8 fix(todos): /clear now resets the Todos sidebar (#1258)`. The merged version uses `try_lock` on the todos mutex so the UI thread doesn't block if the engine briefly holds it during tool execution, and clears both `plan_state` and the shared todo list. Closing as superseded; thanks for the work and the clear write-up.

PR #1287, opened 2026-05-09
feat(providers): add openai-compatible provider with custom URL and API key

huangjunxin:
## Summary

  Add a new `openai-compatible` provider type that allows users to connect to any
  OpenAI-compatible API endpoint by providing a custom base URL and optional API key.
  This covers self-hosted LLM servers (e.g. vLLM, SGLang), third-party API proxies, and the official OpenAI API.

  - `deepseek auth set --provider openai-compatible --url <BASE_URL> [--api-key <KEY>]`
  - URL is **required** (interactive prompt if omitted); API key is **optional**
  - Env vars: `OPENAI_COMPATIBLE_API_KEY`, `OPENAI_COMPATIBLE_BASE_URL`, `OPENAI_COMPATIBLE_MODEL`
  - Config: `[providers.openai_compatible]` with `api_key`, `base_url`, `model`
  - TUI `/provider` picker shows "OpenAI Compatible" with "(configured)" when URL is set
  - No model name normalization — passes through whatever the user configures

  ## Changes
  10 files, +238 / -16 lines:

  | File | Change |
  |---|---|
  | `crates/config/src/lib.rs` | `OpenaiCompatible` in `ProviderKind`, `ProvidersToml`, env forwarding, skip model normalization |
  | `crates/cli/src/lib.rs` | `OpenaiCompatible` in `ProviderArg`, `--url` on `auth set`, optional-key prompt, env vars |
  | `crates/secrets/src/lib.rs` | `OPENAI_COMPATIBLE_API_KEY` env mapping |
  | `crates/tui/src/config.rs` | `OpenaiCompatible` in `ApiProvider`, picker, env forwarding, empty key allowed |
  | `crates/tui/src/client.rs` | `OpenaiCompatible` in reasoning-effort match arms (no-op, like Ollama) |
  | `crates/tui/src/commands/provider.rs` | Error message updated |
  | `crates/tui/src/tui/provider_picker.rs` | Env var label + picker ordering |
  | `crates/tui/src/tui/ui.rs` | Provider label "Compat" for status bar |
  | `crates/tui/src/core/engine.rs` | Env var reference for error recovery |
  | `crates/tui/src/main.rs` | Doctor/status match arms |

  ## Test plan

  - [x] `cargo fmt --check` — clean
  - [x] `cargo clippy --workspace --all-targets --all-features` — clean
  - [x] `cargo test --workspace` — all tests pass
  - [x] Manual: test against a real OpenAI-compatible endpoint

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -26,6 +26,7 @@ enum ProviderArg {
     Deepseek,
     NvidiaNim,
     Openai,
+    OpenaiCompatible,
     Openrouter,
     Novita,
     Fireworks,
@@ -40,6 +41,7 @@ impl From<ProviderArg> for ProviderKind {
             ProviderArg::Deepseek => ProviderKind::Deepseek,
             ProviderArg::NvidiaNim => ProviderKind::NvidiaNim,
             ProviderArg::Openai => ProviderKind::Openai,
+            ProviderArg::OpenaiCompatible => ProviderKind::OpenaiCompatible,
             ProviderArg::Openrouter => ProviderKind::Openrouter,
             ProviderArg::Novita => ProviderKind::Novita,
             ProviderArg::Fireworks => ProviderKind::Fireworks,
@@ -249,6 +251,9 @@ enum AuthCommand {
         /// Read the key from stdin instead of prompting.
         #[arg(long = "api-key-stdin", default_value_t = false)]
         api_key_stdin: bool,
+        /// Base URL for the provider's API (required for openai-compatible).
+        #[arg(long)]
+        url: Option<String>,
     },
     /// Report whether a provider has a key configured. Never prints
     /// the value; just `set` / `not set` plus the source layer.
@@ -669,6 +674,7 @@ fn provider_slot(provider: ProviderKind) -> &'static str {
         ProviderKind::Deepseek => "deepseek",
         ProviderKind::NvidiaNim => "nvidia-nim",
         ProviderKind::Openai => "openai",
+        ProviderKind::OpenaiCompatible => "openai-compatible",
         ProviderKind::Openrouter => "openrouter",
         ProviderKind::Novita => "novita",
         ProviderKind::Fireworks => "fireworks",
@@ -679,9 +685,10 @@ fn provider_slot(provider: ProviderKind) -> &'static str {
 }
 
 /// Provider order used by the `auth list` and `auth status` outputs.
-const PROVIDER_LIST: [ProviderKind; 9] = [
+const PROVIDER_LIST: [ProviderKind; 10] = [
     ProviderKind::Deepseek,
     ProviderKind::NvidiaNim,
+    ProviderKind::OpenaiCompatible,
     ProviderKind::Openrouter,
     ProviderKind::Novita,
     ProviderKind::Fireworks,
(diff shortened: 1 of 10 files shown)


Hmbown:

Doing a careful pass before merge per the v0.8.26 handoff's note that this PR needs security review. Overall the integration is well-structured and consistent across the 10 files; flagging the following before this lands:

### URL-shape validation

`prompt_base_url` and `--url` accept any string and write it straight into `providers.openai_compatible.base_url`. The first request then builds a `reqwest::Url::parse(...)` and fails late. Worth validating up front:

- Reject URLs whose scheme isn't `http`/`https`. A user pasting `file://...` or `javascript:...` should fail at config time, not at first request.
- Reject empty / whitespace-only URLs. There's a single bail in `prompt_base_url` but the `--url` flag path (line ~895 of `crates/cli/src/lib.rs`) accepts any non-`None` string — `--url ' '` slips through.

A small `validate_openai_compatible_url(url) -> Result<String>` used by both paths plus a unit test would close the gap.

### Userinfo / credential leakage in logs

If a user pastes `https://user:token@example.com/v1`, `userinfo` survives into `provider_config.base_url` and into any error string that includes the URL. `mcp.rs` has a `mask_url_secrets` helper that already does this for MCP transports — would be good to reuse it (or the same regex) when surfacing the openai-compatible URL in errors and `setup_status` output.

### Empty API key behavior

`deepseek_api_key()` returns `Ok(String::new())` for OpenaiCompatible when no key is set. Two questions:

1. Does the client suppress the `Authorization` header entirely when the key is empty, or does it emit `Authorization: Bearer ` (empty bearer)? The latter is an info-leak ("there should be a token here, here's nothing") and some servers may reject it. Worth a smoke test or grep through `client.rs` for the header-construction site.
2. The picker's `(configured)` indicator triggers on either env var or saved base_url (line ~3025 of the diff). Combined with the optional key, a user can end up in a state where `deepseek` "thinks" it's configured but auth fails on first request. A `setup_status` line clarifying "API key optional — empty key sends no Authorization header" would help users diagnose.

### `prompt_base_url` non-TTY fallback

```rust
fn prompt_base_url(slot: &str) -> Result<String> {
    ...
    if !io::stdin().is_terminal() {
        // Non-interactive: read directly without prompting twice.
        return read_api_key_from_stdin();
    }
    ...
}
```

This reuses `read_api_key_from_stdin()` for the non-TTY URL-reading path. It works (both just read stdin) but the function name is misleading and the helper applies API-key trimming/normalization that may mask URL whitespace bugs. Renaming to `read_one_line_from_stdin()` (or having `prompt_base_url` use its own reader) would make the intent clearer.

### Tests

The diff adds `EnvGuard` slots for the three new env vars, which is good. I didn't see explicit tests covering:

- `OPENAI_COMPATIBLE_BASE_URL` env var actually overrides the config file
- `--url` flag overrides any existing `[providers.openai_compatible] base_url`
- An empty `--url` is rejected

Adding those (mirroring `ollama_env_overrides_base_url_and_model` at line ~4833 of `crates/tui/src/config.rs`) would lock in the precedence rules.

### Out of scope / questions for the maintainer

- Should the feature default to refusing private/loopback IPs unless an explicit "I know what I'm doing" flag is set? Aligns with the `fetch_url` SSRF policy but may surprise users running local llama.cpp / vLLM, which is the whole point of this provider.
- Does `provider_capability` need an entry that downgrades to chat-only? The default for OpenaiCompatible looks like it inherits the OpenAI capability flags, which assume support for things like reasoning_effort / structured outputs. Most local servers won't honor those, and the `apply_reasoning_effort` no-op already handles that, but worth a maintainer check.

Happy to spin up the validation helpers if useful — let me know if you want this rebased into a smaller surface or kept as-is for review.

Hmbown:

PR #1294, opened 2026-05-09
feat: add configurable search providers (Tavily & Bocha)

sandofree:
Search config in config.toml, provider dispatch in web_search


Hmbown:

Doing a careful pass before merge per the v0.8.26 handoff's note that this PR needs security review. The structure is clean — `SearchProvider` enum, dispatch in `web_search.rs`, default-DuckDuckGo for back-compat. Items to address before merge:

### Network policy is bypassed for the new providers

The existing DuckDuckGo path runs through:

```rust
validate_network_policy(host, context)?;
```

before any HTTP request, and Bing fallback is gated separately. The new `run_tavily_search` and `run_bocha_search` skip this entirely — they build a fresh `reqwest::Client` and POST straight to `api.tavily.com` / `api.bochaai.com`. A user who has set `network.default = "deny"` (or specifically denied those hosts) won't have those rules honored on the search path.

Suggested fix: pass through the same gate before each provider call:

```rust
match context.search_provider {
    SearchProvider::Tavily => {
        validate_network_policy("api.tavily.com", context)?;
        return self.run_tavily_search(...).await;
    }
    SearchProvider::Bocha => {
        validate_network_policy("api.bochaai.com", context)?;
        return self.run_bocha_search(...).await;
    }
    SearchProvider::DuckDuckGo => { /* falls through to existing gating */ }
}
```

(or wire each helper to call `validate_network_policy` itself — same outcome.)

### Tavily ships the API key in the request body

Per Tavily's documented API the `api_key` goes in the JSON body, which is fine in transit (HTTPS). The risk is if any request-logging hook eventually starts capturing bodies — the key surfaces. Two cheap mitigations:

- Add a `# noqa: api-key-in-body` comment so future grep-for-secrets sweeps don't get confused, and
- If/when `crate::network_policy::NetworkAuditor` (or any equivalent) starts recording POST bodies, make sure it redacts the `api_key` field.

Bocha is fine — it uses `Authorization: Bearer ...`, which already has a redaction pattern in `mcp.rs::redact_body_preview`.

### Endpoint hardcoding is good — but make it explicit

Both providers POST to a hardcoded const. That's the right call here (it removes the SSRF surface you'd open up with a configurable endpoint). But the literals are inline strings — pulling them up to module-level constants:

```rust
const TAVILY_ENDPOINT: &str = "https://api.tavily.com/search";
const BOCHA_ENDPOINT: &str = "https://api.bochaai.com/v1/ai/search";
```

makes the audit surface obvious and mirrors the `DEFAULT_*_BASE_URL` pattern already used in config.

### Error bodies forwarded verbatim

```rust
ToolError::execution_failed(format!(
    "Tavily search failed: HTTP {} — {body}",
    status.as_u16()
))
```

If the upstream returns an error body that includes the original `query` (which the model can influence), or a captcha challenge with markdown that the model might re-render, that goes back into the conversation. Truncating to ~512 bytes (mirroring `ERROR_BODY_PREVIEW_BYTES = 200` in `mcp.rs`) and stripping HTML / control bytes would harden this.

### Empty-results UX

Both new paths return `count: 0` with `message: "No results found"`. The existing DuckDuckGo path falls back to Bing on empty results. Worth deciding: should Tavily/Bocha also fall back to DDG on empty? Probably not — fallbacks add complexity and the user explicitly chose the provider. Document the no-fallback behavior in the config example.

### Tests

I didn't see tests for the new providers. Even an offline test that:

- Stubs the HTTP client (or uses `wiremock` like other tests in this repo do for SSE)
- Asserts `validate_network_policy` is called before the request
- Asserts an empty `search_api_key` produces the friendly error

would lock the contract in. Tavily's response shape and Bocha's `data.pages` shape are easy to drift; locking them with golden JSON in tests means a future API change fails noisily.

### `[search] api_key` in plaintext config

This is consistent with how every other provider key works in this repo (config.toml stores keys when not in keyring), but worth a one-line note in the example saying "treat config.toml like a secret file" or pointing to the existing `auth set` flow. The PR currently only documents the env-var alternative.

### Out of scope / questions for the maintainer

- Should the `auth set` CLI grow `--provider tavily` / `--provider bocha` so users can save these keys in keyring instead of plaintext config? Mirrors the existing pattern for LLM providers.
- Bocha is mainland-China-focused; the friendly error mentions "国内友好" — fine, but make sure `setup_status` translates that to en-US for non-zh-CN locales. (Probably already handled by your i18n system.)

Happy to spin up the network-policy gating change as a small follow-up commit if useful.

sandofree:
Updated per review

PR #1296, opened 2026-05-09
Feat/issue 1278 anchor scroll fix

Zaizai1222:
## 概要
修复 Docs 页面侧边栏锚点跳转后目标 section 被 sticky nav header 遮挡的问题。
- `globals.css`：将 `[id] { scroll-margin-top }` 从 `5rem` 提高到 `8rem`，并包装在 `@layer base` 中
- `docs/page.tsx`：移除 16 个被全局规则覆盖的无效 `scroll-mt-32` 类名
## Root Cause
`[id]` 选择器为非 layered 样式，优先级高于 Tailwind `@layer utilities` 中的 `.scroll-mt-32`，导致 `scroll-mt-*` 工具类全局失效。原值 `5rem` 不足以清除双行 sticky header。通过 `@layer base` 包装恢复 Tailwind 正常优先级体系，utility 类可正常覆盖。
## Test Plan
访问 `/docs` 或 `/zh/docs`，点击侧边栏各锚点链接，验证目标 section 顶部不再被 header 遮挡。

### DIFF
--- a/web/app/[locale]/docs/page.tsx
+++ b/web/app/[locale]/docs/page.tsx
@@ -79,7 +79,7 @@ export default async function DocsPage({ params }: { params: Promise<{ locale: s
             <article className="lg:col-span-9 space-y-14 min-w-0">
 
               {/* 模式 */}
-              <section id="modes" className="scroll-mt-32">
+              <section id="modes">
                 <h2 className="font-display text-3xl mb-1">
                   模式 <span className="font-cjk text-indigo text-2xl ml-2">Modes</span>
                 </h2>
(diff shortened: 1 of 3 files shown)


Hmbown:
Thanks @Zaizai1222 — closing in favor of #1282, which just merged as commit 65a40411 to fix #1278.

The diagnosis in #1278 (great write-up, btw) showed the cause is a CSS specificity collision: `globals.css [id] { scroll-margin-top: 5rem }` and the per-section `.scroll-mt-32` (8rem) tied at 0,1,0, so source order made 5rem win. #1282 fixed it by wrapping the global rule in `:where(...)` to drop its specificity to 0 — that lets the existing `scroll-mt-32` utilities win as intended (8rem) and keeps the 5rem fallback for anchors that don't define their own offset.

This PR took the opposite approach (removing the `scroll-mt-32` classes), which would have left every section at the global 5rem — but the issue reports 5rem isn't enough to clear the ~5.4rem sticky header. Appreciate the work either way.

PR #1297, opened 2026-05-09
fix(client): stabilize reasoning_content replay for prompt cache

Duducoco:
## Summary
DeepSeek-V4's prompt cache (paper §3.5.2 On-Disk KV Cache Storage) matches on request byte prefix hashes — cache-hit
  input tokens are billed at roughly 1/100 the price of misses. Earning that discount requires that the bytes of
  already-sent historical messages stay identical across turns of a multi-turn conversation.

  Users reported a consistently low cache hit rate.

  The bug

  In crates/tui/src/client/chat.rs, build_chat_messages_with_reasoning was re-deciding per request whether each
  historical assistant message should carry reasoning_content, gated on later_user_turn — "is there a real user message
  later in the history?"
```rust
  // Old logic
  let later_user_turn = messages[message_index + 1..]
      .iter()
      .any(message_starts_user_turn);
  let include_reasoning_for_turn =
      include_reasoning && (has_tool_calls || !later_user_turn);
```
  Result: the same text-only assistant reply serialized to different bytes on different turns. On the turn it was just
  produced (no later user message yet), it went out as:
```
  {"role":"assistant","content":"answer","reasoning_content":"..."}
```
  On the very next turn, after the user asked a follow-up, that same historical message went out as:
```
  {"role":"assistant","content":"answer"}
```
  The reasoning_content field was silently dropped. Once the historical bytes shifted, the entire prefix from that
  message onward mismatched. Every "user asks → assistant replies in text → user asks again" cycle — the most common
  pattern in multi-turn chat — triggered a cache wipe.

  The codebase already had messages_with_turn_metadata_preserves_stored_messages_for_prefix_cache asserting that
  historical message bytes must be byte-stable across turns, proving the invariant was the intended design. This code
  path just violated it.

  The fix

  In crates/tui/src/client/chat.rs, reasoning_content now depends only on the stored message itself:

```rust
  let mut has_reasoning = include_reasoning && !reasoning_content.trim().is_empty();
  if include_reasoning && has_tool_calls && !has_reasoning {
      // Tool-call messages with empty thinking still need a placeholder
      // (DeepSeek 400s without it).
      reasoning_content = String::from("(reasoning omitted)");
      has_reasoning = true;
  }
```
  When the model requires replay and the message has thinking content, always emit reasoning_content. When the model
  requires replay and the message has tool_calls but empty thinking, emit the (reasoning omitted) placeholder — this
  part is unchanged because DeepSeek thinking-mode tool-call turns return 400 without it. Otherwise the field is
  omitted. The later_user_turn scan and the now-unused message_starts_user_turn helper are removed.

  In crates/tui/src/client.rs, the old test chat_messages_omit_prior_non_tool_reasoning_after_new_user_turn (which
  encoded the bug) was renamed to chat_messages_keep_prior_non_tool_reasoning_after_new_user_turn and updated to assert
  the correct contract. A new dedicated regression test was added:
```rust
  #[test]
  fn chat_messages_assistant_json_is_byte_stable_across_follow_up_user_turn() {
      // ...build the same assistant message as part of turn N and turn N+1...
      assert_eq!(
          assistant_n, assistant_n1,
          "assistant message JSON must be byte-identical across turns \
           or DeepSeek's prefix cache breaks"
      );
  }
```
  Any future attempt to reintroduce "mutate history based on context" will trip this test immediately.

  Scope of impact

  This only affects the request-body builder for DeepSeek chat completions — no UI changes, no config changes. Tool-call
   reasoning replay behavior is unchanged; that path was already cache-safe and DeepSeek-required. Expected outcome: in
  multi-turn conversations, prompt_cache_hit_tokens should rise significantly, and the footer's cache-hit-rate chip
  should stay green (≥ 80%) far more often.
## Testing

## Checklist

### DIFF
--- a/crates/tui/src/client.rs
+++ b/crates/tui/src/client.rs
@@ -1290,7 +1290,12 @@ mod tests {
     }
 
     #[test]
-    fn chat_messages_omit_prior_non_tool_reasoning_after_new_user_turn() {
+    fn chat_messages_keep_prior_non_tool_reasoning_after_new_user_turn() {
+        // The serialized JSON for a stored assistant message MUST be a pure
+        // function of that message — never of what comes after it. DeepSeek's
+        // prompt cache hashes the leading bytes of every request; flipping
+        // `reasoning_content` on/off across turns rewrites historical bytes
+        // and busts the prefix cache from that message onwards. (#583)
         let messages = vec![
             Message {
                 role: "user".to_string(),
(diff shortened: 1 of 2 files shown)


Hmbown (approved):
Reviewed the reasoning_content replay change against current main. Verified focused chat message tests, cargo fmt, clippy, and the deepseek-tui binary test suite on a local merge; the only full-suite failure passed in isolation as the known MCP parallel race.

PR #1311, opened 2026-05-09
fix: /clear now resets the Todos sidebar panel

### LINKED ISSUE #1258: bug: /clear does not clear the Todos (checklist) sidebar panel

Giggitycountless:
## Problem

Running `/clear` does not reliably clear the Todos sidebar. When the engine holds the todos mutex during tool execution, `clear_todos()` fails because it uses a single `try_lock()` attempt. The todos persist across `/clear` commands until the app is restarted.

## Root Cause

`clear_todos()` in `crates/tui/src/tui/app.rs` uses `try_lock()` which is non-blocking. If the mutex is held by the engine, the method returns `false` and shows a busy message instead of clearing the todos.

## Fix

Replace the single `try_lock()` attempt with a retry loop (up to 100 attempts, 1ms apart). This gives the engine up to 100ms to release the mutex while keeping the UI responsive.

## Testing

Added a regression test `clear_todos_resets_todos_list` that:
1. Seeds todos into the list
2. Calls `clear_todos()`
3. Asserts the todos list is empty

Fixes #1258

### DIFF
--- a/crates/tui/src/tui/app.rs
+++ b/crates/tui/src/tui/app.rs
@@ -3695,18 +3695,28 @@ impl App {
     }
 
     pub fn clear_todos(&mut self) -> bool {
-        // Clear the todo list (the sidebar checklist). Uses try_lock so the
-        // UI thread doesn't block if the engine briefly holds the mutex
-        // during tool execution; the caller can retry or show a busy message.
-        let todos_cleared = if let Ok(mut todos) = self.todos.try_lock() {
-            todos.clear();
-            true
-        } else {
-            false
+        // Clear the todo list (the sidebar checklist). Retry with try_lock
+        // so /clear always resets todos even when the engine briefly holds
+        // the mutex during tool execution.
+        let todos_cleared = {
+            let mut cleared = false;
+            for _ in 0..100 {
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @Giggitycountless — but the same root cause is already addressed on `main`. The retry-around-`try_lock` pattern you used shipped in v0.8.24 via:

- `5670034d` fix(todos): /clear now resets the Todos sidebar (#1258)
- `254489670` fix: /clear now resets the Todos sidebar panel
- `5436041b6` refactor: extract retry_lock helper for mutex acquisition

That's why the cherry-pick conflicts here — your two commits and the on-`main` work touch the same lines with the same intent. Both approaches converge on the same place.

Closing as already-fixed. Genuinely appreciate the engagement on this one — you also landed #1326 (`/init` adds `.deepseek/` to `.gitignore`) in v0.8.27 and #1332 (auto-add `.deepseek/`) is in flight. Keep them coming.

PR #1312, opened 2026-05-09
fix(tui): send TurnStarted before snapshot to prevent WSL2 timeout

michaeltse321:
## Problem

On WSL2 with Windows drives mounted at \/mnt\/c, the \/ routine (which runs \/ on the workspace) can take 30+ seconds due to the slow 9P filesystem bridge.

The original code sent \/ *after* the snapshot completed. This caused the UI's 30-second dispatch watchdog to fire with:

> Turn dispatch timed out; the engine may have stopped. Please try again.

...before the turn ever appeared to start, leaving users with a hung\/blank TUI on WSL2 + Windows Terminal.

## Changes

1. **Send \/ immediately** — before the snapshot runs — so the UI shows progress while the snapshot operates in the background.
2. **Increase \/ from 30s → 120s** — more tolerant of slow filesystems.

## Environment

- WSL2 + Windows Terminal
- Workspace on \/mnt\/c\/Users\/... (Windows-mounted drive)

Fixes #— (blank screen\/hang issues on WSL2)

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -895,6 +895,16 @@ impl Engine {
         self.turn_counter = self.turn_counter.saturating_add(1);
         self.capacity_controller.mark_turn_start(self.turn_counter);
 
+        // Emit turn started event IMMEDIATELY so the UI knows the turn is
+        // active. The snapshot below can take 30+ seconds on slow filesystems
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @michaeltse321 — `git cherry main` confirms this is patch-equivalent to a commit already on `main`. Credited in the v0.8.27 CHANGELOG under "WSL2 turn-start timeout".

Closing as already-shipped.

PR #1313, opened 2026-05-09
  fix: forward base URL overrides to active providers

### LINKED ISSUE #1308: set base_url for ollama & vllm
i have noticed that setting base_url on ollama & vllm to an other ip/host is not taken at all it will always look for the default set "localhost"

reidliu41:
## Summary

  Fixes provider base URL overrides for self-hosted providers.

  `deepseek --provider ollama --base-url ...` and `deepseek --provider vllm --base-url ...`
  were still resolving to the default localhost endpoints. The dispatcher forwards
  `--base-url` as `DEEPSEEK_BASE_URL`, but the TUI config loader only routed that
  value into a few provider-specific config blocks. For Ollama and vLLM it landed
  on the legacy root `base_url`, which those providers do not read.

  This change routes `DEEPSEEK_BASE_URL` into the active provider's config block,
  so Ollama and vLLM honor `--base-url` correctly. Existing provider-specific env
  vars such as `OLLAMA_BASE_URL` and `VLLM_BASE_URL` continue to work.

```
$   OLLAMA_BASE_URL=http://10.0.0.3:11434/v1 .deepseek --provider ollama doctor --json | jq -r '.base_url'
http://10.0.0.3:11434/v1

I$   VLLM_BASE_URL=http://10.0.0.3:8000/v1 .deepseek --provider vllm doctor --json | jq -r '.base_url'
http://10.0.0.3:8000/v1
```

fixes: https://github.com/Hmbown/DeepSeek-TUI/issues/1308

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -1893,6 +1893,9 @@ fn apply_env_overrides(config: &mut Config) {
     }
     if let Ok(value) = std::env::var("DEEPSEEK_BASE_URL") {
         match config.api_provider() {
+            ApiProvider::Deepseek | ApiProvider::DeepseekCN => {
+                config.base_url = Some(value);
+            }
             ApiProvider::NvidiaNim => {
                 config
                     .providers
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @reidliu41 — `git cherry main` confirms this is patch-equivalent to a commit already on `main`. Credited in the v0.8.27 CHANGELOG under "Provider base-URL overrides reach the active provider".

Closing as already-shipped.

PR #1319, opened 2026-05-09
fix: stabilize MCP tool ordering prefix-cache stability fix

hxy91819:
## Summary

This is an incremental prefix-cache stability fix for MCP tool ordering. It removes non-deterministic ordering from MCP tool discovery and aggregation without changing the existing `to_api_tools()` block layout.

## Background

DeepSeek's prefix KV cache depends on byte-identical request prefixes. Reordering tool entries changes subsequent byte offsets and can invalidate the cached prefix from the first changed tool onward.

The main model request path is already protected by `build_model_tool_catalog()`, which sorts native and MCP tool partitions before sending tools to the model. This PR adds lower-level defensive ordering so MCP consumers get deterministic tool lists before they reach that catalog layer.

### 1. Sort tools after MCP discovery

`McpConnection::discover_tools()` now sorts the merged paginated `tools/list` result by tool name after discovery completes.

This makes `conn.tools()` deterministic even if an MCP server returns tools in a different order after reconnect or pagination.

### 2. Sort aggregated MCP tools by fully-qualified name

`McpPool::all_tools()` now sorts by the final exported MCP tool name:

```rust
mcp_{server}_{tool}
```

This controls the previous `HashMap<String, McpConnection>` iteration-order dependency when aggregating tools across servers.

### 3. Preserve `to_api_tools()` block ordering

`McpPool::to_api_tools()` intentionally does **not** apply a flat global sort. It keeps the historical block structure:

```text
[MCP regular tools — sorted by all_tools()]
  → [list_mcp_resources + list_mcp_resource_templates]
  → [mcp_read_resource + read_mcp_resource]
  → [mcp_get_prompt]
```

This keeps the ordering change scoped to the regular MCP tool block instead of interleaving helper tools such as `list_mcp_resources` with `mcp_*` tools.

## Tests

- `discover_tools_sorts_paginated_results_by_name` verifies observable discovery behavior: paginated server responses are merged and exposed in deterministic name order.
- `mcp_pool_tool_exports_sort_only_regular_mcp_block` verifies exported behavior: regular MCP tools are sorted by fully-qualified name while `to_api_tools()` preserves helper-tool block order.

The test helper for constructing `McpConnection` sorts its tools to mirror production discovery behavior, so tests that do not specifically exercise discovery start from a representative connection state.

## Scope boundary

This PR addresses deterministic ordering only. It does not attempt to solve:

- Conditional helper-tool presence:
  - `list_mcp_resources` / `list_mcp_resource_templates` are present when MCP servers are configured.
  - `mcp_read_resource` / `read_mcp_resource` are present when resources exist.
  - `mcp_get_prompt` is present when prompts exist.
- MCP tool metadata drift after reconnect, such as changed descriptions or schemas.
- UI-level cache hit reporting.

Those are separate follow-ups from this ordering fix.

## Verification

- [x] `cargo fmt -p deepseek-tui --manifest-path /data/code/DeepSeek-TUI/lane1/Cargo.toml`
- [x] `cargo test -q -p deepseek-tui --manifest-path /data/code/DeepSeek-TUI/lane1/Cargo.toml discover_tools_sorts_paginated_results_by_name`
- [x] `cargo test -q -p deepseek-tui --manifest-path /data/code/DeepSeek-TUI/lane1/Cargo.toml mcp_pool_tool_exports_sort_only_regular_mcp_block`
- [x] `cargo test -q -p deepseek-tui --manifest-path /data/code/DeepSeek-TUI/lane1/Cargo.toml model_tool_catalog_sorts_each_partition_for_prefix_cache_stability`

### DIFF
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -2251,7 +2251,8 @@ mod tests {
         }
     }
 
-    fn test_connection(name: &str, tools: Vec<McpTool>) -> McpConnection {
+    fn test_connection(name: &str, mut tools: Vec<McpTool>) -> McpConnection {
+        tools.sort_by(|a, b| a.name.cmp(&b.name));
         McpConnection {
             name: name.to_string(),
             transport: Box::new(MockTransport {
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -1735,7 +1735,6 @@ impl McpPool {
             });
         }
 
-        api_tools.sort_by(|a, b| a.name.cmp(&b.name));
         api_tools
     }
 
@@ -2500,7 +2499,7 @@ mod tests {
     }
 
     #[test]
-    fn mcp_pool_tool_exports_are_sorted_by_final_api_name() {
+    fn mcp_pool_tool_exports_sort_only_regular_mcp_block() {
         let mut config = McpConfig::default();
         config
             .servers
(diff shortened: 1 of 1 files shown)


hxy91819:
> Feedback suggests updating the test_connection helper function to sort tools by default\n\nGood thought, but I don't think we should add sorting to `test_connection`.\n\n`test_connection()` is a test helper that bypasses `discover_tools()` to give callers full control over tool ordering. If it sorts implicitly, tests that intentionally pass unsorted input (e.g. `[zeta, alpha]`) can no longer verify that `all_tools()` or `to_api_tools()` actually sort the result — the sort is already done by the helper, masking the behavior under test.\n\nThe correct guard against regressions is the existing `mcp_pool_tool_exports_sort_only_regular_mcp_block` test, which passes unsorted tools and asserts the output is sorted. That's the regression test.

hxy91819:
Closing this PR because it has already been cherry-picked into main

PR #1320, opened 2026-05-09
feat: support OpenAI-compatible Coding Plan endpoints (Bailian, ZhiPu, etc.)

imkingjh999:
## Summary

- Support OpenAI-compatible Coding Plan endpoints (Bailian, ZhiPu, etc.) by adding a custom User-Agent (`deepseek-tui/<version>`) to the reqwest client, fixing HTTP 405 rejections from endpoints that filter by User-Agent
- Allow non-DeepSeek model names for pass-through providers (openai, ollama) in the `/model` command
- Persist provider and model choices per-provider in settings so runtime switches survive restarts
- Hide DeepSeek models from the model picker when on a pass-through provider
- Show a friendly message instead of a 404 error for `/models` on unsupported providers

## Usage

Any OpenAI-compatible Coding Plan endpoint can be used by changing `base_url`, `model`, and `api_key` in `~/.deepseek/config.toml`:

**Bailian (DashScope) Coding Plan:**
```toml
provider = "openai"
[providers.openai]
api_key = "sk-xxx"
base_url = "https://coding.dashscope.aliyuncs.com/v1"
model = "kimi-k2.5"
```

**ZhiPu Coding Plan:**
```toml
provider = "openai"
[providers.openai]
api_key = "sk-xxx"
base_url = "https://open.bigmodel.cn/api/coding/paas/v4"
model = "glm-5.1"
```

## Test plan

- [x] Start TUI with openai provider pointing to Coding Plan → messages sent successfully
- [x] Switch to deepseek provider via `/provider deepseek` → restart → provider persists
- [x] Switch back to openai → model comes from config.toml, not overridden by settings
- [x] `/models` on openai provider shows "not supported" message instead of 404
- [x] Model picker on openai provider only shows "auto" + custom model, no DeepSeek models
- [x] DeepSeek provider works unchanged

### DIFF
--- a/crates/tui/src/client.rs
+++ b/crates/tui/src/client.rs
@@ -508,7 +508,7 @@ impl DeepSeekClient {
         let headers = build_default_headers(api_key, extra_headers)?;
         let mut builder = reqwest::Client::builder()
             .default_headers(headers)
-            .user_agent("deepseek-tui/0.8")
+            .user_agent(concat!("deepseek-tui/", env!("CARGO_PKG_VERSION")))
             .connect_timeout(Duration::from_secs(30))
             .tcp_keepalive(Some(Duration::from_secs(30)))
             .http2_keep_alive_interval(Some(Duration::from_secs(15)))
--- a/crates/tui/src/client.rs
+++ b/crates/tui/src/client.rs
@@ -508,6 +508,7 @@ impl DeepSeekClient {
         let headers = build_default_headers(api_key, extra_headers)?;
         let mut builder = reqwest::Client::builder()
             .default_headers(headers)
+            .user_agent("deepseek-tui/0.8")
             .connect_timeout(Duration::from_secs(30))
             .tcp_keepalive(Some(Duration::from_secs(30)))
             .http2_keep_alive_interval(Some(Duration::from_secs(15)))
--- a/crates/tui/src/commands/core.rs
+++ b/crates/tui/src/commands/core.rs
@@ -119,11 +119,16 @@ pub fn model(app: &mut App, model_name: Option<&str>) -> CommandResult {
                 AppAction::UpdateCompaction(app.compaction_config()),
             );
         }
-        let Some(model_id) = normalize_model_name(name) else {
-            return CommandResult::error(format!(
-                "Invalid model '{name}'. Expected auto or a DeepSeek model ID. Common models: {}",
-                COMMON_DEEPSEEK_MODELS.join(", ")
-            ));
+        let model_id = if crate::config::provider_passes_model_through(app.api_provider) {
+            name.to_string()
+        } else {
+            let Some(normalized) = normalize_model_name(name) else {
+                return CommandResult::error(format!(
+                    "Invalid model '{name}'. Expected auto or a DeepSeek model ID. Common models: {}",
+                    COMMON_DEEPSEEK_MODELS.join(", ")
+                ));
+            };
+            normalized
         };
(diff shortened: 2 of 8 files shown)


mattss0140 (approved):

mattss0140 (commented):
I would love to know how to code and make sense of it all

imkingjh999:
Thanks for the approval! If you're interested in learning to code, the [CONTRIBUTING.md](https://github.com/Hmbown/DeepSeek-TUI/blob/main/CONTRIBUTING.md) has a good getting-started guide for this project. The [Rust Book](https://doc.rust-lang.org/book/) is also a great resource for learning Rust from scratch.

Hmbown:

This is mostly UX + persistence work; the security surface is small. Three notes before merge:

### User-Agent change is fine

`concat!("deepseek-tui/", env!("CARGO_PKG_VERSION"))` is the right shape — version-only, no machine identifier. Some endpoints may want a contact URL too (we already include one in `fetch_url`'s UA: `Mozilla/5.0 (compatible; deepseek-tui/0.5; +https://github.com/Hmbown/DeepSeek-TUI)`). Worth a quick choice on whether this UA should match the format `fetch_url` uses, both for consistency and to give endpoints a way to reach maintainers if they ban the UA.

### Settings file accepts arbitrary provider/model strings

`default_provider: Option<String>` and `provider_models: HashMap<String, String>` are persisted to `~/.deepseek/settings.toml`. The provider string is only consumed by `ApiProvider::parse(...)`, which falls back to the config-file provider on miss — that's the right defensive behavior. The model strings are pass-through to OpenAI/Ollama by design.

A small belt-and-suspenders: in `set_model_for_provider`, validate that the provider key is one of `ApiProvider::all()` before inserting. Otherwise a stale settings file from a future build can grow unrecognized entries that just sit there forever.

### Conflict with #1287 (OpenaiCompatible)

Both PRs touch:
- `crates/tui/src/commands/provider.rs` (error message text)
- The provider list / picker
- `provider_passes_model_through` (PR #1287 adds `OpenaiCompatible` to the same matchset; this PR makes the function `pub(crate)`)

Whichever lands second will need to merge the match arms. Worth deciding intent: should the new `OpenaiCompatible` provider from #1287 share the user-agent fix here, or stay strictly minimal?

### Non-security observations

- `provider_models` only writes from the picker apply path. If the user types `/model gpt-4o-mini` with `--no-tui` `exec` mode, it doesn't propagate to settings (only the `default_model` key gets touched in the legacy path). Not a regression — but worth noting it's not a strict superset.
- `FetchModels` short-circuit for pass-through providers is good, but the message "is not supported" might be too strong for OpenAI-compatible endpoints that *do* support `/models` (most of them do — vLLM, SGLang). Maybe "the configured backend does not advertise a model index" or similar, plus a hint to set `default_model` directly.

No blockers from a security review standpoint. The conflict-with-#1287 question is the main thing to resolve before merge.

Hmbown:
Thanks @imkingjh999 — the work from this PR is now on `main` across a chain of commits authored by you: `b024a92ec feat: persist provider/model per-provider in settings`, `9fae15618 fix: address PR review feedback`, `a80c31809 fix: only use default_model fallback for DeepSeek providers`, `5065092a1 fix: let OPENAI_MODEL env override config.toml provider model`, and `cb084d156 style: fix fmt and clippy warnings`. Credited in the v0.8.27 CHANGELOG under "Provider-aware model picker and provider persistence (#1320)".

Closing as already-shipped.

PR #1334, opened 2026-05-10
Preserve workspace skill precedence under prompt truncation

hhhaiai:
## Summary
This fixes a prompt-ordering bug in the model-visible skills block.

During skill discovery, workspace and cross-tool directories already have the
correct precedence. But the final renderer re-sorted all discovered skills by
name, so a large lower-priority global skill set could push higher-priority
workspace skills out of the prompt budget.

This change preserves registry precedence during rendering and adds a regression
test that proves workspace-priority skills remain visible under truncation.

## Testing

## Notes
- No user-facing UI change
- Adds regression coverage for prompt-budget overflow with mixed workspace/global skills

### DIFF
--- a/crates/tui/src/skills/mod.rs
+++ b/crates/tui/src/skills/mod.rs
@@ -113,6 +113,9 @@ impl SkillRegistry {
         let mut visited = HashSet::new();
         Self::discover_recursive(dir, 0, &mut registry, &mut visited);
         registry
+            .skills
+            .sort_by(|a, b| a.name.cmp(&b.name).then_with(|| a.path.cmp(&b.path)));
+        registry
     }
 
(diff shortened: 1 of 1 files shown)


hhhaiai:
Force-pushed to clean up the PR branch. This PR now contains only the skills prompt precedence fix in `crates/tui/src/skills/mod.rs`; the earlier `client.rs` / `main.rs` / config changes referenced by the previous bot review are no longer part of the diff.

hhhaiai:
Superseded by #1341, which keeps the verified skills fix but also folds in the requested ~/.deepseek config defaults, explicit HTTP gateway support, and the fork sync workflow with an updated clean diff.

PR #1341, opened 2026-05-10
Standardize on ~/.deepseek config and add explicit HTTP gateway support

hhhaiai:
## Summary
- make runtime use `~/.deepseek/config.toml` as the default config source instead of auto-loading workspace-local `.deepseek/config.toml`
- keep legacy root aliases (`baseurl`, `model`) working and add `allow_insecure_http = true` for trusted non-local `http://...` gateways
- add a fork-only workflow that fast-forwards `openChatGpts/DeepSeek-TUI` from upstream `main` every 3 hours when the fork has not diverged
- include the already-verified skills prompt precedence fix so the branch stays green against the current upstream test baseline

## Why
The target workflow here is:
1. put `base_url` / `default_text_model` / key defaults in `~/.deepseek/config.toml`
2. support trusted self-hosted HTTP gateways without silently allowing all remote HTTP endpoints
3. keep the fork's `main` branch synced from upstream automatically on a 3-hour cadence

The previous attempt widened remote HTTP too far and left the docs out of sync with the runtime config behavior. This version keeps HTTP support explicit, updates the docs, and preserves the requested single-source user config model.

## Testing

## Notes
- No manual TUI UI flow changes were required
- The sync workflow is guarded so it only runs in `openChatGpts/DeepSeek-TUI`


Hmbown:
Thanks for the comprehensive PR, @hhhaiai.

This bundles two distinct changes that I'd like to handle separately:

1. **Standardize on `~/.deepseek/` as the default config source instead of auto-loading workspace-local `.deepseek/config.toml`.** This is a behavior change to a path users depend on. It needs an explicit deprecation note + a fallback period where workspace-local still works with a warning, before we flip the default. Substantial enough to deserve its own focused PR + discussion thread.

2. **Explicit HTTP gateway support** — a clean addition that probably wants to land alongside the existing `*_BASE_URL` env-var work (the v0.8.27 fix for #1308 from @reidliu41). Could be a smaller scoped PR on its own.

If you're open to splitting this in two — the gateway-support part is likely a near-term landing candidate, while the path-standardization needs the longer roadmap discussion. Either way, leaving this open and tagged v0.9.x while we work out the split.

Hmbown:

PR #1347, opened 2026-05-10
fix: Ctrl+Enter content lost when engine is idle (#1331)

Oliver-ZPLiu:
## Summary

Fixes #1331 (Ctrl+Enter submission issue).

When the engine is idle (no active turn), Ctrl+Enter sent the message to `rx_steer`, which is only monitored inside `handle_deepseek_turn`. The message appeared in the transcript (via the local mirror in `steer_user_message`) but the LLM never received it. The next regular Enter would drain it as a "stale steer" in `handle_send_message`.

## Root cause

`rx_steer` is only polled inside `handle_deepseek_turn` ([turn_loop.rs:50](https://github.com/Hmbown/DeepSeek-TUI/blob/v0.8.25/crates/tui/src/core/engine/turn_loop.rs#L50), [turn_loop.rs:872](https://github.com/Hmbown/DeepSeek-TUI/blob/v0.8.25/crates/tui/src/core/engine/turn_loop.rs#L872)). The engine's main event loop ([engine.rs:574](https://github.com/Hmbown/DeepSeek-TUI/blob/v0.8.25/crates/tui/src/core/engine.rs#L574)) only handles `Op` variants — steers sent while idle are orphaned.

When the next `Op::SendMessage` arrives, `handle_send_message` drains `rx_steer` as "stale" at [engine.rs:891](https://github.com/Hmbown/DeepSeek-TUI/blob/v0.8.25/crates/tui/src/core/engine.rs#L891), discarding the message permanently.

## Fix

Check `app.is_loading` in the Ctrl+Enter handler:

- **Engine busy** (already generating): steer into current turn as before
- **Engine idle**: send via `submit_or_steer_message` → `Op::SendMessage` so the content goes through the regular submission path

## Files changed

| File | Change |
|------|--------|
| `crates/tui/src/tui/ui.rs` | Conditional steer vs. regular send in Ctrl+Enter handler |

## Test plan

- [x] `cargo check -p deepseek-tui` passes
- [x] `cargo test -p deepseek-tui -- tui::app` — 124 passed
- [x] All pre-existing non-Python-related tests pass (2485/2504)

PR #1348, opened 2026-05-10
fix: long text clipped instead of wrapping (#1344)

### LINKED ISSUE #1344: Long input and output text are clipped instead of wrapping in TUI window
## Description

When using DeepSeek TUI, long text does not automatically wrap within the terminal window.

This affects both:

1. User input text
2. Model output text

When I type a long prompt in the input area, the input text is clipped/truncated and cannot be fully seen. Similarly, long model responses are also clipped at the right edge of the TUI window.

This happens both in fullscreen mode and after manually resizing the terminal window. Resizing the window does not reflow or recover the clipped text.

## Expected behavior

Long text should wrap automatically according to the current terminal width.

This should apply to both:

1. The user input area
2. The model output/response area

After resizing the terminal window, the visible text should also reflow properly.

## Actual behavior

Long input text and long output text are cut off at the right edge of the TUI window. The hidden part cannot be seen even after switching to fullscreen or resizing the terminal window.

## Steps to reproduce

1. Launch DeepSeek TUI in iTerm2.
2. Type a long prompt in the input area.
3. Observe that the input text is clipped instead of wrapping.
4. Submit the prompt and wait for a long response.
5. Observe that the output text is also clipped at the right edge.
6. Resize the terminal window or switch to fullscreen.
7. The clipped text remains unreadable instead of wrapping/reflowing.

## Environment

- OS: macOS Sequoia 15.6
- Device: MacBook Air 13-inch, M4, 2025
- RAM: 16 GB
- Terminal app: iTerm2 Build 3.6.10
- TERM: xterm-256color
- DeepSeek TUI version: 0.8.16
- Installation method: Cargo

## Additional context

This issue affects both prompt editing and response readability because parts of the text become inaccessible.

It looks like a TUI rendering / line wrapping / text area layout / terminal resize reflow issue. It would be helpful if both the input area and output area supported automatic line wrapping and reflow after terminal resize.

Screenshots or screen recordings can be provided if needed.

Image

### LINKED ISSUE #1351: 界面文字显示问题
1.输出的表格内容如果太长会显示....而不是全部显示文字
2.有些长提问输入到界面后不会自动换行，导致超出界面长度的文字无法显示

Oliver-ZPLiu:
## Summary

This PR fixes both problems reported in issue #1344 and #1351:

**Grapheme-level hard break for long words** — when a single word (URL, code token, base64 string) exceeds the available width, it is now broken at grapheme cluster boundaries instead of overflowing. This applies to both plain-text paragraphs (`wrap_text`) and inline-styled text with links/markup (`render_line_with_links`).

**Paragraph safety net** — `ChatWidget`'s `Paragraph` now has `.wrap(Wrap { trim: false })` enabled. If any pre-wrapped line still exceeds the area width, ratatui will wrap it at render time instead of silently clipping.

Fixes #1344
Fixes #1351

## Root cause

The markdown renderer's word-level wrapping (`wrap_text` / `render_line_with_links`) does not break long individual words — a URL, code token, or base64 string wider than the available width would overflow. Additionally, `ChatWidget`'s `Paragraph` had no `.wrap()` safety net, so any overflowing line was silently clipped by ratatui at the render stage.

The composer (`wrap_input_lines`) already used grapheme-level wrapping and was unaffected.

## Changes

| File | Change |
|------|--------|
| `crates/tui/src/tui/widgets/mod.rs` | Added `.wrap(Wrap { trim: false })` to `ChatWidget`'s `Paragraph` as a safety net |
| `crates/tui/src/tui/markdown_render.rs` | Added hard grapheme-level wrapping in `wrap_text` and `render_line_with_links` when a single word exceeds the available width |

## Fix details

**Paragraph safety net** — if any pre-wrapped line still exceeds the area width (e.g. due to width-miscalculation edge cases), ratatui will wrap it at render time instead of clipping.

**Grapheme-level hard break** — when a word is wider than the available width, it is now broken at grapheme cluster boundaries. This applies to both plain-text paragraphs (`wrap_text`) and inline-styled text with links/markup (`render_line_with_links`). The break only triggers when necessary — normal text still uses word-level wrapping.

## Test plan

- [x] `cargo clippy -p deepseek-tui --all-targets --all-features --locked -- -D warnings` passes
- [x] `cargo fmt --all -- --check` passes
- [x] `cargo test -p deepseek-tui -- tui::markdown_render` — 21 passed
- [x] `cargo test -p deepseek-tui -- tui::widgets::tests::long_tool_result_lines_fit_requested_width` — passed
- [x] `cargo test -p deepseek-tui -- "tui::"` — 797 passed

### DIFF
--- a/crates/tui/src/tui/markdown_render.rs
+++ b/crates/tui/src/tui/markdown_render.rs
@@ -29,6 +29,7 @@ use std::cell::Cell;
 
 use ratatui::style::{Modifier, Style};
 use ratatui::text::{Line, Span};
+use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 
 use crate::palette;
@@ -481,8 +482,26 @@ fn render_line_with_links(
             current_spans = Vec::new();
             current_width = 0;
         }
-        current_spans.push(Span::styled(word, style));
+        current_spans.push(Span::styled(word.clone(), style));
         current_width += ww;
+
+        // Hard-wrap at grapheme boundaries when a single word exceeds width
+        // (e.g. long URLs, code tokens, or base64 strings).
+        if ww > width {
+            // Remove the full word and re-add in width-limited grapheme chunks.
+            current_spans.pop();
+            current_width = current_width.saturating_sub(ww);
+            for grapheme in word.graphemes(true) {
+                let gw = grapheme.width();
+                if current_width + gw > width && !current_spans.is_empty() {
+                    lines.push(Line::from(current_spans));
+                    current_spans = Vec::new();
+                    current_width = 0;
+                }
+                current_spans.push(Span::styled(grapheme.to_string(), style));
+                current_width += gw;
+            }
+        }
     }
 
     if !current_spans.is_empty() {
@@ -907,6 +926,23 @@ fn wrap_text(text: &str, width: usize) -> Vec<String> {
             }
             current.push_str(word);
             current_width += word_width;
+        }
+
+        // Hard-wrap at grapheme boundaries when a single word exceeds width
+        // (e.g. long URLs, code tokens, or base64 strings).
+        if current_width > width {
+            let remaining = std::mem::take(&mut current);
+            current_width = 0;
+            for grapheme in remaining.graphemes(true) {
+                let gw = grapheme.width();
+                if current_width + gw > width && !current.is_empty() {
+                    lines.push(current);
+                    current = String::new();
+                    current_width = 0;
+                }
+                current.push_str(grapheme);
+                current_width += gw;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks @Oliver-ZPLiu — this same issue is fixed on `main` via `be54a046d fix(markdown): hard-break overlong words to stop right-edge overflow`, which takes a similar approach but extracts a shared `push_word_breaking_chars` helper called from both `render_line_with_links` and `wrap_text` (and the existing `wrap_cell_text`), with snapshot tests pinning the invariant at widths 40 / 60 / 80 / 120. The v0.8.27 CHANGELOG credits this fix under "Long output text overflowed the right edge (#1344, #1351)".

Closing as already-fixed (more comprehensive equivalent landed). Genuinely appreciate the engagement — your other PRs (#1392 cache-usage, #1388 deny-tool, #1393 prompt reliability, #1347 Ctrl+Enter, #1353 yank pager) all landed in v0.8.28.

PR #1358, opened 2026-05-10
fix: parse CRLF SSE MCP events

### LINKED ISSUE #1309: Can't connect to MCP with SSE. Error connecting to open-brain.
# MCP_RS_ISSUE.md — SSE parser misses CRLF event separators from uvicorn/FastMCP

## File
`crates/tui/src/mcp.rs` — `SseTransport::run_sse_loop` method (lines ~480–500)

## Status
- **Bug #1** (race condition on endpoint discovery): **fixed upstream** in PR #1225 / commit `f29d1a3`.
- **Bug #2** (CRLF event separator): **still present** as of 2026-05-09 on `origin/main`.

---

## Symptom

MCP SSE servers served by **uvicorn** (e.g. FastMCP, open-brain) fail to
connect. The HTTP GET to `/sse` succeeds (200 OK), but the client never
parses the `endpoint` event. `wait_for_endpoint` times out with:

```
SSE endpoint not received within 30000ms
```

Even with `connect_timeout` set to 3 seconds on a LAN, the endpoint is
emitted by the server in under 1 ms but never seen by the client.

Servers using pure-LF SSE (e.g. `mcp` Python SDK's test helpers, some
Node.js implementations) work fine.

---

## Root Cause

HTTP/1.1 specifies `\r\n` (CRLF) as the line terminator. uvicorn faithfully
uses CRLF line endings throughout, including its SSE event stream. The
resulting wire format is:

```
event: endpoint\r\n
data: /messages/?session_id=abc\r\n
\r\n
event: message\r\n
data: {"jsonrpc":"2.0","id":1,...}\r\n
\r\n
```

The SSE event separator is `\r\n\r\n` — a blank CRLF line.

The DeepSeek-TUI parser only searches for `\n\n` (LF-only), so the
separator is never matched and events are never extracted. The endpoint
goes unseen and the connection times out.

Additionally, the lines parsed from each event block carry a trailing
`\r`, which prevents prefix matching (`"endpoint\r"` != `"endpoint"`).

### Server-side proof

A raw-socket test against a FastMCP server confirmed CRLF framing:

```
$ python3 mcp_test.py
ENDPOINT: /messages/?session_id=183cc59e...
INIT: Accepted
NOTIF: Accepted
TOOLS: Accepted

=== SSE RAW ===
event: message\r\n
data: {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2024-11-05",...}}\r\n
\r\n
```

The `tools/list` response arrives correctly on the wire — the bug is
purely in the Rust client parser.

---

### 1. Accept both `\n\n` and `\r\n\r\n` as event separators

```rust
// Before:
while let Some(pos) = buffer.find("\n\n") {

// After:
while let Some(pos) = buffer.find("\n\n")
    .or_else(|| buffer.find("\r\n\r\n"))
{
```

### 2. Advance the buffer by the correct number of bytes

```rust
// Before:
buffer = buffer[pos + 2..].to_string();

// After:
let has_crlf = buffer[pos..].starts_with("\r\n\r\n");
buffer = if has_crlf {
    buffer[pos + 4..].to_string()
} else {
    buffer[pos + 2..].to_string()
};
```

### 3. Strip trailing `\r` from lines before prefix matching

```rust
// Before:
for line in event_block.lines() {
    if let Some(stripped) = line.strip_prefix("event: ") {

// After:
for line in event_block.lines() {
    let line = line.trim_end_matches('\r');
    if let Some(stripped) = line.strip_prefix("event: ") {
```

---

## Full diff

```diff
-            while let Some(pos) = buffer.find("\n\n") {
+            while let Some(pos) = buffer.find("\n\n")
+                .or_else(|| buffer.find("\r\n\r\n"))
+            {
                 let event_block = buffer[..pos].to_string();
-                buffer = buffer[pos + 2..].to_string();
+                let has_crlf = buffer[pos..].starts_with("\r\n\r\n");
+                buffer = if has_crlf {
+                    buffer[pos + 4..].to_string()
+                } else {
+                    buffer[pos + 2..].to_string()
+                };

                 let mut event_type = "message";
                 let mut data = String::new();

                 for line in event_block.lines() {
+                    let line = line.trim_end_matches('\r');
                     if let Some(stripped) = line.strip_prefix("event: ") {
```

## Verification

1. Set up any FastMCP server (or the open-brain SSE MCP server)
2. Configure `mcp.json` with the SSE URL
3. Before the fix: `mcp connect` times out / tools never listed
4. After the fix: `mcp connect` succeeds immediately, `mcp tools <name>` lists all tools

```
$ deepseek-tui mcp connect open-brain
Connected to MCP server: open-brain

$ deepseek-tui mcp tools open-brain
Tools for open-brain:
  - memory_search: Search memories by query, tags, source, or date range.
  - memory_store: Store a new memory with auto-tagging and entity extraction.
  - memory_get_related: Get memories related to a specific memory by ID.
  - memory_get_entity: Get all memories containing a specific entity.
  - memory_today: Get memories created today.
  - memory_stats: Get memory statistics.
  - memory_weekly_report: Generate a weekly report of memory activity.
```

## Impact

This affects **every MCP SSE server** running behind uvicorn, which
includes most FastMCP servers. The fix is backward-compatible — it
accepts both `\n\n` (LF) and `\r\n\r\n` (CRLF), so existing stdio/LF
servers are unaffected.

reidliu41:
## Summary

  Fixes MCP SSE connections for servers that emit CRLF-framed event streams, such as uvicorn/FastMCP.

 The SSE transport previously only searched for LF-only event separators (`\n\n`). CRLF streams use `\r\n\r\n`, so endpoint events  were never parsed and connection setup could time out with `SSE endpoint not received within ...ms`.

 This change accepts both LF and CRLF SSE event separators and adds regression tests for CRLF endpoint discovery.

Fixes: https://github.com/Hmbown/DeepSeek-TUI/issues/1309

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -574,18 +574,21 @@ impl SseTransport {
             let s = String::from_utf8_lossy(&chunk);
             buffer.push_str(&s);
 
-            while let Some(pos) = buffer.find("\n\n") {
+            while let Some((pos, separator_len)) = find_sse_event_separator(&buffer) {
                 let event_block = buffer[..pos].to_string();
-                buffer = buffer[pos + 2..].to_string();
+                buffer = buffer[pos + separator_len..].to_string();
 
                 let mut event_type = "message";
                 let mut data = String::new();
 
                 for line in event_block.lines() {
-                    if let Some(stripped) = line.strip_prefix("event: ") {
-                        event_type = stripped;
-                    } else if let Some(stripped) = line.strip_prefix("data: ") {
-                        data.push_str(stripped);
+                    if let Some(value) = sse_field_value(line, "event:") {
+                        event_type = value;
+                    } else if let Some(value) = sse_field_value(line, "data:") {
+                        if !data.is_empty() {
+                            data.push('\n');
+                        }
+                        data.push_str(value);
                     }
                 }
 
@@ -838,6 +841,15 @@ fn parse_sse_message_data(body: &str) -> Vec<Vec<u8>> {
     }
 
     messages
+}
+
+fn find_sse_event_separator(buffer: &str) -> Option<(usize, usize)> {
+    match (buffer.find("\n\n"), buffer.find("\r\n\r\n")) {
+        (Some(lf), Some(crlf)) if crlf < lf => Some((crlf, 4)),
+        (Some(lf), _) => Some((lf, 2)),
+        (_, Some(crlf)) => Some((crlf, 4)),
+        _ => None,
+    }
 }
 
 fn sse_field_value<'a>(line: &'a str, field: &str) -> Option<&'a str> {
@@ -2724,6 +2736,18 @@ mod tests {
         assert!(value.get("result").is_some());
     }
 
+    #[test]
+    fn find_sse_event_separator_accepts_lf_and_crlf() {
+        assert_eq!(
+            find_sse_event_separator("event: endpoint\n\n"),
+            Some((15, 2))
+        );
+        assert_eq!(
+            find_sse_event_separator("event: endpoint\r\n\r\n"),
+            Some((15, 4))
+        );
+    }
+
     #[tokio::test]
     async fn mcp_connection_supports_streamable_http_event_stream_responses() {
         use tokio::io::{AsyncReadExt, AsyncWriteExt};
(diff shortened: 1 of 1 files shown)


Hmbown:
Cherry-picked into `work/v0.8.29` as `0164ee869`; CHANGELOG and README credit added. Thanks @reidliu41 — this will ship in v0.8.29 once the release PR lands.

Hmbown:
Thanks @reidliu41 — your fix landed on `main` as [`0164ee869`](https://github.com/Hmbown/DeepSeek-TUI/commit/0164ee869) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

PR #1361, opened 2026-05-10
fix: add DEC 2026 synchronized updates to prevent flickering on GPU terminals

xuezhaoyu:
## Summary

Add DEC 2026 synchronized update escape sequences (`ESC[?2026h` / `ESC[?2026l`) around terminal drawing to prevent flickering on GPU-accelerated terminals (Ghostty, VS Code Terminal, Kitty, etc.).

## Problem

GPU-accelerated terminals render each batch of escape sequences as it arrives. Without synchronized updates, ratatui's incremental diff writing produces visible intermediate frames, causing rapid flickering and elevated CPU usage when the AI model streams output at high speed (up to 120 FPS).

Related issues:
- #1352 (flickering in Ghostty)
- #1356 (flickering in VS Code Terminal)

## Root Cause

Two issues working together:

1. **No synchronized updates around draws** — ratatui's `terminal.draw()` writes a diff of changed cells to stdout. GPU terminals render partial batches as they arrive, making intermediate states visible.

2. **Separate sync batches for viewport reset and draw** — the `force_terminal_repaint` path called `reset_terminal_viewport()` and `draw_app_frame()` in separate synchronized update blocks, producing a visible blank intermediate frame between the clear and the redraw.

## Fix

- Add `BEGIN_SYNC_UPDATE` / `END_SYNC_UPDATE` constants (`ESC[?2026h` / `ESC[?2026l`)
- Wrap `terminal.draw()` in synchronized updates so terminals render complete frames only
- Wrap `reset_terminal_viewport()` in synchronized updates
- Introduce `draw_app_frame_inner(terminal, app, full_repaint)` that merges viewport-reset + draw into a **single synchronized update batch**, eliminating the blank intermediate frame
- `draw_app_frame()` becomes a thin wrapper calling `draw_app_frame_inner(terminal, app, false)`

All sequences are best-effort — terminals that don't support DEC 2026 silently discard them.

## Test plan

- [ ] Run in Ghostty — no flickering during streaming
- [ ] Run in Konsole — no regression
- [ ] Run in VS Code Terminal — no flickering
- [ ] Run in older terminals (Windows ConHost) — no crash
- [ ] `cargo test -p deepseek-tui` passes with no regressions

PR #1340, opened 2026-05-10
feat(notifications): LC_TERMINAL fallback probe for Cmux + test coverage

CrepuscularIRIS:
## What changed

- `crates/tui/src/tui/notifications.rs`: Added `LC_TERMINAL` as a secondary env-var probe in `resolve_method()`, checked after `TERM_PROGRAM`. Added `Cmux` to the `OSC9_TERMINALS` allowlist. Documented that terminals setting neither env var can force OSC 9 with `[notifications].method = \"osc9\"` in config.
- `crates/tui/src/config.rs`: Updated `NotificationsConfig::method` doc to mention Cmux and the `LC_TERMINAL` probe, and note the explicit override.
- Tests added: `auto_detect_picks_osc9_for_cmux_via_lc_terminal`, `auto_detect_picks_osc9_for_wezterm_via_lc_terminal`.
- Hardened: `auto_detect_picks_bel_for_unknown_on_unix` now clears `LC_TERMINAL` before asserting the Bel fallback, preventing flakiness in CI environments where `LC_TERMINAL` is set.

## Why

Fixes #1281. Cmux typically does not set `TERM_PROGRAM`; it may set `LC_TERMINAL` instead. The previous `resolve_method()` only checked `TERM_PROGRAM`, so Cmux users fell back to `Bel` (audible beep) instead of OSC 9 pop-up notifications.

Note: if a particular Cmux version sets neither env var, the reliable workaround is `[notifications].method = "osc9"` in `~/.deepseek/config.toml` — this config path already exists and is now documented more prominently.

## How to verify

```sh
# Simulate Cmux environment
TERM_PROGRAM="" LC_TERMINAL="Cmux" cargo test -p deepseek-tui -- notifications
# Expect: auto_detect_picks_osc9_for_cmux_via_lc_terminal PASSED

# Confirm bel-fallback test still works without LC_TERMINAL
cargo test -p deepseek-tui -- auto_detect_picks_bel_for_unknown_on_unix
```

## Impact scope

- `crates/tui/src/tui/notifications.rs` — `resolve_method()` only; no changes to emit paths, thresholds, or the `notify_done` public API.
- `crates/tui/src/config.rs` — doc comment only.
- No other modules affected.

> This PR was prepared with AI assistance and reviewed locally before submission.

PR #1365, opened 2026-05-10
fix(settings): auto-enable low_motion in VS Code integrated terminal

### LINKED ISSUE #1356: terminal rapid flickering when used with vs code terminal
## Description
when i use tui in vs code on m4 mac mini, the terminal content rapidly flickers. does not flicker on m4 macbook air. does not flicker in terminal app. does not happen in vs code terminal for claude, codex, gemini CLIs. installed using both hbrew and npm
## Steps to reproduce

1. open vs code project
2. toggle terminal panel
3. run command: deepseek --model auto

## Expected behavior
should work without flickering
## Actual behavior
rapid flickering
## Environment

- OS: MacOS 26.4
- DeepSeek CLI version:
deepseek (npm wrapper) v0.8.25
binary version: v0.8.25
- Model:
- Shell: zsh

CrepuscularIRIS:
## What changed

- `crates/tui/src/settings.rs` — extended `apply_env_overrides()` to detect `TERM_PROGRAM=vscode` and enable `low_motion` + disable `fancy_animations`, matching the existing `NO_ANIMATIONS` env-override pattern.
- Two new tests: `vscode_term_program_forces_low_motion_on` and `non_vscode_term_program_does_not_force_low_motion`, using the same serialised-guard pattern as the existing `NO_ANIMATIONS` tests.

## Why

VS Code's integrated terminal sets `TERM_PROGRAM=vscode`. With the default 120 FPS draw cap, ratatui's diff-and-flush loop overwhelms the compositor on certain hardware (M4 Mac Mini reported in #1356), producing rapid flickering that doesn't occur in Terminal.app, iTerm2, or other CLIs in the same environment.

The `low_motion` path already caps draws at 30 FPS and suppresses animations. Activating it automatically for `TERM_PROGRAM=vscode` fixes the symptom with no user configuration required.

Note: #1361 (open) adds DEC 2026 synchronized-update escapes, which addresses the root rendering cause. These two fixes are complementary: synchronized updates ensure correct frames; the frame-rate reduction lowers CPU overhead during streaming independent of terminal DEC 2026 support. This PR does not conflict with #1361.

## How to verify

```bash
# In VS Code integrated terminal
TERM_PROGRAM=vscode cargo run -- --model auto
# Streaming output should be stable, no rapid flickering

# Confirm low_motion is active
cargo test -p deepseek-tui vscode_term_program_forces_low_motion_on
cargo test -p deepseek-tui non_vscode_term_program_does_not_force_low_motion
```

## Impact scope

- `crates/tui/src/settings.rs` only — one branch added to `apply_env_overrides()`.
- No change to the frame-rate limiter, renderer, or any other subsystem.
- All other terminals unaffected; `TERM_PROGRAM` check is an exact string match.

Fixes #1356

### DIFF
--- a/crates/tui/src/settings.rs
+++ b/crates/tui/src/settings.rs
@@ -311,10 +311,9 @@ impl Settings {
         }
         // VS Code's integrated terminal sets TERM_PROGRAM=vscode. Its
         // compositor cannot keep up with 120 FPS redraws and produces rapid
-        // flickering (#1356). Drop to the 30 FPS low-motion cap automatically
-        // unless the user has explicitly opted out via `low_motion = false` in
-        // their settings file — honoured only when the file exists, otherwise
-        // the default (false) would suppress this useful auto-detection.
+        // flickering (#1356). Drop to the 30 FPS low-motion cap automatically.
+        // Like NO_ANIMATIONS above, this unconditionally overrides any
+        // disk-loaded value — consistent precedence: env signals always win.
         if std::env::var("TERM_PROGRAM").as_deref() == Ok("vscode") {
             self.low_motion = true;
             self.fancy_animations = false;
--- a/crates/tui/src/settings.rs
+++ b/crates/tui/src/settings.rs
@@ -309,6 +309,16 @@ impl Settings {
             self.low_motion = true;
             self.fancy_animations = false;
         }
+        // VS Code's integrated terminal sets TERM_PROGRAM=vscode. Its
+        // compositor cannot keep up with 120 FPS redraws and produces rapid
+        // flickering (#1356). Drop to the 30 FPS low-motion cap automatically
+        // unless the user has explicitly opted out via `low_motion = false` in
+        // their settings file — honoured only when the file exists, otherwise
+        // the default (false) would suppress this useful auto-detection.
+        if std::env::var("TERM_PROGRAM").as_deref() == Ok("vscode") {
+            self.low_motion = true;
+            self.fancy_animations = false;
+        }
     }
 
     /// Save settings to disk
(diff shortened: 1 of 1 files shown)


Hmbown:
Cherry-picked both commits into `work/v0.8.28` as `5bededf7` + `3b1d0ae9` — thanks @CrepuscularIRIS! Two recent landings from you in close succession (this one + the Cmux LC_TERMINAL probe in #1340).

One follow-up note: your `term_program_test_guard` was a fresh module-local `static Mutex<()>` mirroring the existing `no_animations_test_guard`. Both serialized within their own family but not against each other, which left a parallel-test race — a `NO_ANIMATIONS=1` write from one family could land in the env at the exact moment a `TERM_PROGRAM=iTerm.app` write from the other called the shared `apply_env_overrides`. The non-VS Code regression test (`non_vscode_term_program_does_not_force_low_motion`) tripped on this twice in v0.8.28's preflight. I folded both guards into the process-wide `test_support::lock_test_env()` (the same fold the v0.8.27 test-stabilization commit applied to the EnvGuard family); ran the full suite green. Worth knowing for future env-touching tests.

CHANGELOG entry for v0.8.28:

> low_motion auto-enables under VS Code integrated terminal — `apply_env_overrides()` now treats `TERM_PROGRAM=vscode` the same way it treats `NO_ANIMATIONS=1`: force `low_motion = true` and `fancy_animations = false`. The VS Code terminal compositor cannot keep up with 120 fps redraws and produces rapid flicker (#1356); the 30 fps low-motion cap is the right default there.

CrepuscularIRIS:
Thanks for the detailed write-up on the mutex race — that's a subtle cross-family hazard and `lock_test_env()` is exactly the right fix. Will keep that in mind for any future env-touching tests.

Hmbown:

Issue #1364, opened 2026-05-10
Hooks need mutation rights on user submit and a turn-end event


esinecan:
## Problem

I run the same set of behavioral extensions across multiple coding agent harnesses (Claude Code, Pi, and DeepSeek TUI). The portability stack is simple: **MCP** for tools, **hooks** for behavioral extensions that inject context or react to agent lifecycle events. DeepSeek TUI has MCP, which is great. But its hook surface doesn't support the two events that matter most for portable extensions, so I had to build an SSE sidecar that spawns `deepseek serve --http` and owns the user-input path entirely, just to get the same extension behavior I get natively in other harnesses.

The sidecar works but it replaces the TUI. No slash commands, no status bar, no in-TUI MCP manager, no image rendering. The quality gap compared to running the same extensions as native Pi extensions or Claude Code hooks is noticeable.

## What I need

**1. `message_submit` with mutation rights (critical)**

A hook that fires before the user's message is sent to the model, where the hook can **modify the message text**. This is the single most important missing piece.

Use case: I run a "subconscious memory" extension that, on every user submission, queries a knowledge base for relevant context from past conversations, then prepends it to the user's message inside XML tags. The agent sees the context as part of the input. This runs as a ~400ms synchronous step before the message goes out.

Both Claude Code and Pi support this natively:
- Claude Code: `UserPromptSubmit` hook. Can inject `additionalContext` or block with exit code 2.
- Pi: `input` event on the Extension API. Handler receives the text and returns transformed text. Can also fully replace or suppress.

DeepSeek TUI's `message_submit` event exists but is observe-only. The hook can't change what gets sent. That's the gap.

**2. `turn_end` / `agent_idle` event (critical)**

A hook that fires when the agent finishes responding and goes idle, waiting for the next user input.

Use case: I run a "voice of reason" extension that, when the agent finishes a turn, async-fires a separate model call (with thinking enabled) on the agent's transcript. The critique is stashed and prepended to the *next* user submission (via the submit hook above). If the agent submits a new turn before the critique arrives, the in-flight critique is dropped. This is a fire-and-forget pattern that never blocks the user.

Prior art:
- Claude Code: `Stop` event. Receives a `stop_hook_active` flag so hooks can detect re-entry and avoid infinite loops.
- Pi: `turn_end` event (and separately `agent_end` for the full agent loop boundary).

**3. Subagent lifecycle events (nice to have)**

[...]

PR #1371, opened 2026-05-10
  Support mutable submit hooks and turn-end hooks

reidliu41:
## Summary

  Implements the hook behavior requested in #1364.

 Previously, `message_submit` hooks were observer-only: they could see a submitted message, but could not change or block it. This adds a minimal mutation/block contract:

  - `message_submit` hooks receive context JSON on stdin.
  - stdout JSON `{ "text": "..." }` replaces the submitted message.
  - exit code `2` blocks submission and shows a reason from `{ "reason": "..." }` or stderr.
  - empty or non-JSON stdout keeps the original message for compatibility.

 This also adds a `turn_end` observe hook that fires after a model turn completes and the TUI returns to idle. It includes turn
  status, duration, usage, token, cost, session, workspace, model, and error context.

  ## Changes

  - Added `turn_end` as a supported hook event.
  - Added JSON stdin payload support for hook execution.
  - Added `message_submit` mutation and blocking behavior.
  - Wired `message_submit` into the pre-dispatch path.
  - Wired `turn_end` into the turn completion path.
  - Updated `/hooks events` output.
  - Updated `config.example.toml` and configuration docs.
  - Added focused hook executor and TUI dispatch tests.

## Testing

## Checklist

  - [x] Verified TUI behavior manually

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -447,11 +447,23 @@ default_text_model = "deepseek-ai/deepseek-v4-pro"
 # ─────────────────────────────────────────────────────────────────────────────────
 # Hooks (optional)
 # ─────────────────────────────────────────────────────────────────────────────────
-# Hooks run shell commands on lifecycle events (session start/end, tool calls, etc.).
+# Hooks run shell commands on lifecycle events (session start/end, user submit,
+# turn end, tool calls, etc.).
 # Configure as `[[hooks.hooks]]` under a `[hooks]` table.
 #
-# Available events: session_start, session_end, message_submit,
+# Available events: session_start, session_end, message_submit, turn_end,
 # tool_call_before, tool_call_after, mode_change, on_error, shell_env.
+#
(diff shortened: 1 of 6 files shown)


Hmbown:
Thanks @reidliu41 — solid implementation of the mutable-hooks contract from #1364.

This is intentionally a v0.9.0-scope change because it shifts `message_submit` (and the new turn-end hook) from observer-only to mutate-or-block. That's a behavior-change to a contract third-party hooks rely on, so we want to land it deliberately:

- formal proposal text in #1364 that pins the new contract (return-shape, ordering when multiple hooks claim the same event, what happens on hook timeout)
- a migration note for users with existing observer-style hooks (they need to keep the no-op shape)
- explicit version-skew handling: hook scripts written for v0.8.x see the new contract gracefully

Leaving open and tagged for v0.9.0. The current diff is a good starting point — let's iterate it once the contract in #1364 is locked in.

Hmbown:

Issue #1318, opened 2026-05-09
skill 列表太过拥挤


simuusang:
## Problem
当skill很多文字很长的时候看起来不是很美观，希望给每个skill 上下留出部分间隔方便区分
<img width="1598" height="1219" alt="Image" src="https://github.com/user-attachments/assets/06a701e7-d2f1-4e40-8bff-e755f1c35078" />

Hmbown:
v0.8.26 added inter-row spacing for `/skills` (#1328 from **@reidliu41**) — does that resolve the visual crowding for you?

If you'd still like a way to filter the list (e.g. `/skills <prefix>` to only show skills whose names start with a prefix), that's straightforward to add for v0.8.28; let us know which usage pattern you'd find most useful:

1. v0.8.26's spacing is enough — close this
2. Add a name-prefix filter (`/skills foo` → skills starting with "foo")
3. Add a section grouping or column layout

Thanks for the report and for the screenshot — it made the crowding obvious.

simuusang:
<img width="1689" height="934" alt="Image" src="https://github.com/user-attachments/assets/bb88d4ce-82ea-48f8-9f6f-a6e3a89ee08b" />

<img width="668" height="68" alt="Image" src="https://github.com/user-attachments/assets/fb996cc3-9520-466b-a28f-3ae8cef37a1e" />
我已经更新到了v0.8.26 但是还是没有任何改变

> v0.8.26 为 `/skills` （ [#1328](https://github.com/Hmbown/DeepSeek-TUI/pull/1328) 来自 ）添加了行间距——这解决了你的视觉拥挤问题吗？
> 
> 如果你还想用一种方式筛选列表（比如 `/skills <prefix>` 只显示名字以前缀开头的技能），这在 v0.8.28 中很容易添加;告诉我们你觉得哪种使用模式最有用：
> 
> 1. v0.8.26 的间距足够了——关闭这里
> 2. 添加一个名称前缀过滤器（`/skills foo` →技能，技能以“foo”开头）
> 3. 添加分区分组或列布局
> 
> 感谢你的报告和截图——让拥挤现象显而易见。

simuusang:
当skill描述太长的时候，能否使用省略号显示呢？就显示一行文字，比如：”/flutter-implement-json-serialization - Create model classes with fromJson and ...“ 当我选中其中一个skill则滚动显示被隐藏的文字

Hmbown:
[...]

PR #1380, opened 2026-05-10
feat(skills): support /skills <prefix> to filter the list

linzhiqin2003:
## Summary

Implements the prefix-filter follow-up [@Hmbown suggested](https://github.com/Hmbown/DeepSeek-TUI/issues/1318#issuecomment-) in [#1318](https://github.com/Hmbown/DeepSeek-TUI/issues/1318):

> If you'd still like a way to filter the list (e.g. `/skills <prefix>` to only show skills whose names start with a prefix), that's straightforward to add for v0.8.28; let us kn[ow]

## What changed

`/skills` gains an optional name-prefix argument:

| Invocation | Behaviour |
|---|---|
| `/skills` | Lists all skills (unchanged) |
| `/skills --remote` / `/skills remote` | Curated registry (unchanged) |
| `/skills sync` / `/skills --sync` | Sync to local cache (unchanged) |
| `/skills flutter` | **New:** lists only skills whose name starts with `flutter`, with an `N/M` badge in the header |
| `/skills zz-nope` (no match) | **New:** friendly message that points to the unfiltered command |

## Output samples

**With matches:**
```
Skills matching "flutter" (2/3):
─────────────────────────────
  /flutter-impl - Flutter impl helper

  /flutter-test - Flutter test helper

Use /skill <name> to run a skill
Skills location: /home/user/.deepseek/skills
```

**No matches:**
```
No skills match prefix "zz-nope".

Use /skills (no argument) to see all 1 available skills.
```

## Backwards compatibility

- `/skills` with no arg, `--remote`, `remote`, `sync`, `--sync` all behave exactly as before.
- The only behaviour shift is for an unrecognised arg: previously `/skills foo` printed `Usage: /skills [--remote|sync]`. Now it's treated as a prefix filter — which is also what most users actually meant (no shell-style flags involved).

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p deepseek-tui -- commands::skills` — **11/11 passing**, including 2 new tests:
  - `test_list_skills_filters_by_prefix` — covers the matching path + `N/M` header
  - `test_list_skills_prefix_no_match_friendly` — covers the empty-result message

Refs #1318

### DIFF
--- a/crates/tui/src/commands/skills.rs
+++ b/crates/tui/src/commands/skills.rs
@@ -33,9 +33,11 @@ fn render_skill_warnings(registry: &SkillRegistry) -> String {
 /// List all available skills. Pass `--remote` (or `remote`) to fetch the
 /// curated registry instead of scanning the local skills directory.
 /// Pass `sync` to pull the registry index and download all skills to the
-/// local cache (`~/.deepseek/cache/skills/`).
+/// local cache (`~/.deepseek/cache/skills/`). Any other argument is treated
+/// as a name prefix and filters the listing — useful when many skills are
+/// installed and only a family is interesting (#1318).
 pub fn list_skills(app: &mut App, arg: Option<&str>) -> CommandResult {
-    if let Some(arg) = arg {
+    let prefix = if let Some(arg) = arg {
         let trimmed = arg.trim();
         if trimmed == "--remote" || trimmed == "remote" {
             return list_remote_skills(app);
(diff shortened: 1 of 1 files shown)


linzhiqin2003:
Closing — upstream main now contains an equivalent (and more polished) `/skills <prefix>` implementation: case-insensitive matching, leading-dash/whitespace input validation, and tighter result messaging. Maintainer's version supersedes this PR. Thanks for the suggestion in #1318 — it's now live in main.

PR #1379, opened 2026-05-10
docs(tools): add V4 steering sentences to tool descriptions

linzhiqin2003:
## Summary

Implements [#711](https://github.com/Hmbown/DeepSeek-TUI/issues/711) — rewrites every model-visible tool's `description()` so it actively herds V4 toward the typed tool instead of reaching for `exec_shell` for the same job.

## What changed

**9 tool descriptions, each gains:**
- a short *"use this instead of X in `exec_shell`"* steering line
- the return shape (where useful)
- the relevant limits

| Tool | Steers away from |
|---|---|
| `read_file` | `cat` / `head` / `tail` / `sed -n '..p'` |
| `write_file` | `cat <<EOF > file` / `echo > file` |
| `edit_file` | `sed -i` |
| `list_dir` | `ls` / `ls -la` / `find . -maxdepth 1` |
| `grep_files` | `grep -r` / `rg` / `find ... -exec grep` |
| `file_search` | `find -name` / `fd` |
| `apply_patch` | `git apply` / `patch` / repeated `edit_file` calls |
| `web_search` | scraping with `curl` |
| `fetch_url` | `curl` for known URLs |

**Dead prompt templates removed:**
- `crates/tui/src/prompts/normal.txt` — fully orphaned (no `include_str!` reference anywhere)
- `crates/tui/src/prompts/plan.txt` + `PLAN_PROMPT` constant — only self-test referenced it
- `crates/tui/src/prompts/yolo.txt` + `YOLO_PROMPT` constant — same
- `legacy_constants_still_available` test trimmed to the surviving `AGENT_PROMPT`

`AGENT_PROMPT` / `agent.txt` are preserved on purpose (they are flagged for a separate issue per the #711 body).

## Diff stat

```
10 files changed, 10 insertions(+), 36 deletions(-)
```

Net **−26 LOC** — within the issue's *+40/−30* target direction.

## Description-quality checks

| Tool | chars | newlines? | markdown header? |
|---|---:|---|---|
| read_file | 311 | no | no |
| write_file | 245 | no | no |
| edit_file | 266 | no | no |
| list_dir | 154 | no | no |
| grep_files | 252 | no | no |
| file_search | 182 | no | no |
| web_search | 214 | no | no |
| apply_patch | 211 | no | no |
| fetch_url | 350 | no | no |

All under the 1024-char acceptance gate; no embedded `\n` or `#` headers — keeps the cached tool catalogue prefix-stable for V4's KV cache.

## Test plan

[...]

Issue #711, opened 2026-05-05
docs(tools): steering sentences in tool descriptions; remove unused legacy prompts


Hmbown:
# Self-executing task for `deepseek-v4-pro`

Run inside `deepseek-tui` (Agent mode, workspace trust). Implement end-to-end and commit on `feat/v0.8.13`. Tools: `read_file`, `grep_files`, `apply_patch`, `edit_file`, `write_file`, `exec_shell`. Stay in workspace.

## Why

V4 picks the right tool when descriptions actively herd it. Today our `description()` implementations describe each tool in isolation — they don't tell V4 what to use *instead* and when. The result: V4 reaches for `exec_shell` to do things our cheaper tools could do (read files via `cat`, grep via `grep -r`, edits via `sed -i`). Each unnecessary `exec_shell` call costs more tokens, requires approval, and bypasses our workspace boundary checks.

## Files to read first

- `crates/tui/src/tools/spec.rs` — `ToolSpec::description` signature
- `crates/tui/src/tools/file.rs` — `read_file`, `list_dir`
- `crates/tui/src/tools/patch.rs` — `apply_patch`, `edit_file`, `write_file`
- `crates/tui/src/tools/grep.rs` — `grep_files`
- `crates/tui/src/tools/search.rs` — `file_search`
- `crates/tui/src/tools/web_search.rs`, `crates/tui/src/tools/fetch.rs`
- `crates/tui/src/tools/shell.rs` — context for what we're steering away from
- `crates/tui/src/prompts/{normal,plan,yolo}.txt` — confirm these are unused (they are; only their `*_PROMPT` constants reference them, and those constants only appear in self-tests)

## Plan

1. For each of the tools below, rewrite `description()` to add (a) "use this instead of X" steering, (b) the return format, (c) the limits.
2. Delete the unused legacy `.txt` prompt templates and their constants.
3. Run gates. Commit.

### Steering language to add (per tool)

- `read_file` — "Use this instead of `cat`, `head`, `tail`, or `sed -n` in `exec_shell`. Returns `LINE_NUM|CONTENT` paginated; pass `offset` and `limit` for large files. Cannot read images or binaries."
- `grep_files` — "Use this instead of `grep -r`, `rg`, or `find … -exec grep` in `exec_shell` — pure-Rust, faster, respects `.gitignore`. Returns `path:line:match` triples."
[...]

PR #1382, opened 2026-05-10
chore(prompts): remove orphaned base.txt prompt template

linzhiqin2003:
## Summary

`crates/tui/src/prompts/base.txt` is a stale leftover from before the prompt was decomposed into `base.md` + overlays (per [#711](https://github.com/Hmbown/DeepSeek-TUI/issues/711)). Nothing imports it — `BASE_PROMPT` is `include_str!(\"prompts/base.md\")` — so the file just sits in the source tree consuming 46 lines without effect.

## Confirmation

```
$ grep -rn 'base\\.txt' --include='*.rs' --include='*.md' \\
     --include='*.toml' --include='*.json' .
(no hits)
```

Companion to the `normal.txt` / `plan.txt` / `yolo.txt` cleanup landing in [#1379](https://github.com/Hmbown/DeepSeek-TUI/pull/1379). The remaining `agent.txt` is intentionally **kept** here because `AGENT_PROMPT` still re-exports it (legacy compat); removing that one is a separate decision and outside the scope of this PR.

## Test plan

- [x] `cargo build --workspace` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] No code changes; `cargo test` impact is nil

Net diff: **−46 LOC, 0 insertions.**


linzhiqin2003:
Closing on reflection — this is a small chore-class PR (dead-file cleanup or single-line docs nitpick) that doesn't move user-visible behaviour. Best handled by the maintainer's own follow-up sweeps rather than adding review queue noise. Sorry for the round-trip; happy to revisit if it's actually wanted.

Issue #725, opened 2026-05-05
feat(prompts): declarative-fact memory hygiene guidance


Hmbown:
# Self-executing task for `deepseek-v4-pro`

Run inside `deepseek-tui` (Agent mode, workspace trust). Implement end-to-end and commit on `feat/v0.8.13`.

## Why

When we wire durable user prefs (post-v0.8.13), memory writes phrased as imperatives ("Always respond concisely") get re-read as directives in later sessions and conflict with the user's *current* request. Declarative facts ("User prefers concise responses") are stable across sessions. We pre-stage the prompt fragment so it's ready when memory lands.

This is intentionally a tiny, low-risk addition — the deletion target is correspondingly small (stale CHANGELOG/ARCHITECTURE prose). The 2× milestone-aggregate ratio is satisfied by #8/#14/#15.

## Files to read first

- `crates/tui/prompts/base.md` — for context on style
- `crates/tui/src/prompts.rs` — system-prompt assembly
- `CHANGELOG.md` and `docs/ARCHITECTURE.md` — stale swarm/fanout references

## Plan

1. Create `crates/tui/prompts/memory_guidance.md`:

   ```markdown
   ## Memory hygiene

   Write memories as declarative facts, not instructions to yourself.

   - "User prefers concise responses" ✓ — "Always respond concisely" ✗
   - "Project uses pytest with xdist" ✓ — "Run tests with pytest -n 4" ✗
   - "Repo's main branch is `main`, release branches are `feat/v*`" ✓ —
     "When committing, target main" ✗

   Imperative phrasing gets re-read as a directive in later sessions and
   can override the user's current request. Procedures and workflows
   belong in skills, not memory.
   ```

2. In `crates/tui/src/prompts.rs`, add a feature-gated include:

   ```rust
   pub const MEMORY_GUIDANCE: &str = include_str!("../prompts/memory_guidance.md");

   pub fn build_system_prompt(model: &str, effort: Option<&str>, memory_enabled: bool) -> String {
       let mut s = String::from(BASE_PROMPT);
       for ad in addenda_for(model, effort) { s.push_str(ad); }
       if memory_enabled { s.push_str(MEMORY_GUIDANCE); }
       s
   }
   ```

   (Today `memory_enabled` is always `false`; the include is a no-op until memory ships. We register the fragment now so reviewers can read it in context.)

3. **Paired deletion** — clean up stale "current" framing on the removed swarm/fanout feature:
[...]

PR #1383, opened 2026-05-10
docs(memory): rephrase imperative example as declarative

linzhiqin2003:
## Summary

`docs/MEMORY.md` opens by listing three sample memory entries; two are declarative facts ("I prefer pytest over unittest", "this codebase uses 4-space indentation") and one is imperative ("always run \`cargo fmt\` before committing"). The imperative shape is exactly what the new `MEMORY_GUIDANCE` system-prompt fragment (proposed in [#1381](https://github.com/Hmbown/DeepSeek-TUI/pull/1381) for [#725](https://github.com/Hmbown/DeepSeek-TUI/issues/725)) tells the model to *avoid*, since imperatives get re-read as directives in later sessions and can override the user's current request.

This PR rewords that one example so the docs and the system-prompt guidance speak with the same voice.

## Diff

```diff
- 4-space indentation", "always run `cargo fmt` before committing" —
+ 4-space indentation", "this repo runs `cargo fmt` before commits" —
```

`1 file changed, 1 insertion(+), 1 deletion(-).` Pure docs.

## Why now

Even without [#1381](https://github.com/Hmbown/DeepSeek-TUI/pull/1381) landing, the change is consistent with the existing declarative tone of the surrounding two examples — and the rest of the file (the `## File format` sample, "What stays out of memory", etc.) already uses declarative phrasing throughout. So this is just a leftover imperative.

## Test plan

- [x] `git diff` reviewed — single-line wording change
- [x] No code touched, so build/clippy/test untouched

### DIFF
--- a/docs/MEMORY.md
+++ b/docs/MEMORY.md
@@ -4,7 +4,7 @@ The user-memory feature gives the model a small persistent note file
 that's injected into the system prompt on every turn. It's the place
 to put preferences and conventions that should survive across
 sessions — "I prefer pytest over unittest", "this codebase uses
-4-space indentation", "always run `cargo fmt` before committing" —
+4-space indentation", "this repo runs `cargo fmt` before commits" —
 without having to repeat them in every conversation.
 
 Memory is **opt-in**. When disabled (the default), nothing is loaded,


linzhiqin2003:
Closing on reflection — this is a small chore-class PR (dead-file cleanup or single-line docs nitpick) that doesn't move user-visible behaviour. Best handled by the maintainer's own follow-up sweeps rather than adding review queue noise. Sorry for the round-trip; happy to revisit if it's actually wanted.

PR #1389, opened 2026-05-10
fix(ui): drain active thinking entry on MessageComplete (#861 RC3)

### LINKED ISSUE #861: bug: thinking collapse - multiple root causes causing thinking blocks to freeze, truncate silently, or drop reasoning_content
## Problem

Users experience **thinking collapse** — a family of related defects where the model's reasoning (thinking) blocks either (a) freeze with a live spinner that never resolves, (b) get silently truncated to ≤4 lines with no clear expand affordance during streaming, or (c) disappear entirely from the API message history, causing HTTP 400 errors on the next turn.

Four distinct root causes have been identified through code analysis.

---

## Root Cause 1: Thinking block stuck in `streaming: true` when `ThinkingComplete` is missed

**Location:** `crates/tui/src/tui/ui.rs` — `apply_engine_error_to_app`

When a network interruption or mid-stream disconnect occurs, the `ThinkingComplete` event is never delivered to the UI event loop. The `HistoryCell::Thinking { streaming: true, duration_secs: None }` stays in the `active_cell` forever. The spinner never stops. The block never gets a duration stamp and stays `ThinkingVisualState::Live` with `status: "live"`.

The test `flush_active_cell_finalizes_unclosed_thinking_block` acknowledges this defensively but only covers the `flush_active_cell` path. The error-recovery path through `apply_engine_error_to_app` clears `streaming_thinking_active_entry = None` without finalizing the cell:

```rust
// ui.rs — apply_engine_error_to_app
app.streaming_state.reset();
app.streaming_message_index = None;
app.streaming_thinking_active_entry = None;  // clears pointer but never finalizes the cell
```

**Impact:** After any API error or stream drop during a thinking phase, the thinking cell renders with a permanent live spinner.

---

## Root Cause 2: `streaming_state.reset()` on `ThinkingStarted` drops buffered tail content

**Location:** `crates/tui/src/tui/ui.rs` — `EngineEvent::ThinkingStarted` handler

```rust
EngineEvent::ThinkingStarted { .. } => {
    app.reasoning_buffer.clear();
    app.reasoning_header = None;
    app.thinking_started_at = Some(Instant::now());
    app.streaming_state.reset();          // unconditional reset
    app.streaming_state.start_thinking(0, None);
    let _ = ensure_streaming_thinking_active_entry(app);
}
```

`streaming_state.reset()` is called unconditionally. In a multi-block turn (V4 interleaved thinking), if a second `ThinkingStarted` arrives while the first block is still streaming, the uncommitted tail text in `streaming_state` for the first block is silently discarded.

**Impact:** In multi-round tool-call turns with interleaved thinking, the last few tokens of each thinking block may be silently truncated.

---

## Root Cause 3: `last_reasoning` race between `ThinkingComplete` and `MessageComplete`

**Location:** `crates/tui/src/tui/ui.rs` — event dispatch loop

`ThinkingComplete` sets `app.last_reasoning`. `MessageComplete` reads and clears `app.last_reasoning` to build the `ContentBlock::Thinking` for `api_messages`.

If `MessageComplete` is processed **before** `ThinkingComplete` (possible when the engine channel drains events in burst order), `last_reasoning` is `None` at `MessageComplete` time, and the thinking block is **not included** in the API message history:

```rust
// EngineEvent::MessageComplete
let thinking = app.last_reasoning.take();  // could be None if ThinkingComplete hasn't arrived yet
if let Some(thinking) = thinking {
    blocks.push(ContentBlock::Thinking { thinking });
}
```

**Impact:** Thinking content is stripped from `api_messages`. On the next turn, DeepSeek V4 returns HTTP 400 (missing `reasoning_content`), or the model loses continuity of its own reasoning chain.

---

## Root Cause 4 (UX): Hard truncation to 4 rendered lines with no streaming expand affordance

**Location:** `crates/tui/src/tui/history.rs` — `THINKING_SUMMARY_LINE_LIMIT = 4`

Completed thinking blocks are unconditionally collapsed to 4 rendered lines:

```rust
const THINKING_SUMMARY_LINE_LIMIT: usize = 4;

if collapsed && rendered.len() > THINKING_SUMMARY_LINE_LIMIT {
    rendered.truncate(THINKING_SUMMARY_LINE_LIMIT);
    truncated = true;
}
```

The affordance `"thinking collapsed; press Ctrl+O for full text"` is only shown when `!streaming && (truncated || ...)`. During **live streaming** this line is suppressed entirely — users watching a long thinking block receive no indication that content is being hidden.

**Impact:** Users see partial thinking output during streaming with no signal that more exists or how to see it.

---

## Acceptance Criteria

- [ ] **RC1**: When a stream error fires while a thinking block is active (`streaming_thinking_active_entry.is_some()`), `apply_engine_error_to_app` must call `finalize_streaming_thinking_active_entry` (or equivalent) before clearing the pointer, so the block is stamped as interrupted rather than left frozen with a live spinner.

- [ ] **RC2**: The `ThinkingStarted` handler must flush any uncommitted tail text from `streaming_state` into the current active thinking cell before calling `streaming_state.reset()`, or skip the reset entirely when a thinking entry is already active.

- [ ] **RC3**: The event dispatch loop must guarantee that `ThinkingComplete` is always processed before `MessageComplete` for the same turn. Acceptable approaches: (a) enforce event emission order in the engine, (b) have `MessageComplete` inline-drain a pending thinking finalization when `streaming_thinking_active_entry` is still set, or (c) unify into a single `ThinkingAndMessageComplete` event.

- [ ] **RC4 (UX)**: During streaming of a thinking block that exceeds `THINKING_SUMMARY_LINE_LIMIT`, display a live affordance (e.g. `"… (N more lines)"`) so the user knows the block is being truncated and can navigate to the full text.

---

## Reproduction Steps (RC1 — easiest to reproduce)

1. Start a session with `deepseek-v4-pro` (thinking enabled).
2. Submit a complex multi-step prompt that triggers a long reasoning block.
3. While the thinking spinner is active (`… thinking · live`), press **Esc** to cancel.
4. Observe: the thinking cell retains the live spinner and `status: "live"` permanently instead of transitioning to interrupted/partial.

## Affected Files

- `crates/tui/src/tui/ui.rs` — RC1, RC2, RC3
- `crates/tui/src/tui/history.rs` — RC4 (`THINKING_SUMMARY_LINE_LIMIT`)

## Models Affected

Any V4 thinking-mode model: `deepseek-v4-pro`, `deepseek-v4-flash` with thinking enabled.

linzhiqin2003:
## Summary

Closes [#861](https://github.com/Hmbown/DeepSeek-TUI/issues/861) **root cause 3** — the order-of-events race that drops thinking content from \`api_messages\` and triggers DeepSeek HTTP 400 on the next turn.

The reporter's analysis nailed it:

> If \`MessageComplete\` is processed **before** \`ThinkingComplete\` (possible when the engine channel drains events in burst order), \`last_reasoning\` is \`None\` at \`MessageComplete\` time, and the thinking block is **not included** in the API message history.
> 
> **Impact:** Thinking content is stripped from \`api_messages\`. On the next turn, DeepSeek V4 returns HTTP 400 (missing \`reasoning_content\`), or the model loses continuity of its own reasoning chain.

## What changed

In \`crates/tui/src/tui/ui.rs\`, the head of the \`EngineEvent::MessageComplete\` arm now does a defensive drain when a thinking entry is still active:

```rust
EngineEvent::MessageComplete { .. } => {
    // #861 RC3: defensive drain of a still-active thinking entry.
    if app.streaming_thinking_active_entry.is_some() {
        if finalize_current_streaming_thinking(app) {
            transcript_batch_updated = true;
        }
        stash_reasoning_buffer_into_last_reasoning(app);
    }
    if let Some(index) = app.streaming_message_index.take() { /* … */ }
    // …
    let thinking = app.last_reasoning.take();   // now non-None even if
                                                 // ThinkingComplete was lost
    if let Some(thinking) = thinking {
        blocks.push(ContentBlock::Thinking { thinking });
    }
    // …
}
```

The drain is a **no-op** in the normal case where \`ThinkingComplete\` arrived first (\`streaming_thinking_active_entry\` is already \`None\` and the helpers early-return), so this branch is order-independent.

## Why only RC3 in this PR

The other three root causes from #861 are already (or partially) addressed:

| RC | Status | Where |
|---|---|---|
| **RC1** (apply_engine_error_to_app frozen spinner) | **Already fixed** | \`apply_engine_error_to_app\` line 3198 calls \`finalize_current_streaming_thinking\` before clearing the pointer; covered by \`engine_error_finalizes_active_thinking_block\` test |
| **RC2** (ThinkingStarted drops buffered tail) | **Already fixed** | \`start_streaming_thinking_block\` flushes the previous block via \`finalize_current_streaming_thinking\` + stashes the reasoning buffer before \`streaming_state.reset()\` |
| **RC3** (this PR) | Fixed here | \`MessageComplete\` defensive drain |
| **RC4** (streaming-time truncation affordance) | **Out of scope** | UX improvement; data-loss path is the urgent one |

I split RC4 out so this PR stays narrow on the data-loss path. Happy to follow up with a separate UX PR for RC4 if reviewers want.

## Test plan

- [x] \`cargo test -p deepseek-tui -- message_complete_drain\` — **1/1 passing**

The new test (\`message_complete_drain_preserves_thinking_when_thinking_complete_lost\`) sets up the precise pre-state the bug requires (thinking entry active, \`last_reasoning\` empty), runs the head-of-handler drain in isolation, and asserts that \`last_reasoning\` carries the text afterwards.

The pre-existing \`engine_error_finalizes_active_thinking_block\` (RC1 coverage) and \`flush_active_cell_finalizes_unclosed_thinking_block\` tests still pass — no behaviour shift in the normal flow.

Refs #861

### DIFF
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -698,6 +698,24 @@ async fn run_event_loop(
                         }
                     }
                     EngineEvent::MessageComplete { .. } => {
+                        // #861 RC3: defensive drain of a still-active thinking
+                        // entry. Normally `ThinkingComplete` arrives first and
+                        // populates `last_reasoning` before we get here, but
+                        // when the engine bursts events the channel can
+                        // deliver `MessageComplete` first, in which case
+                        // `last_reasoning.take()` below would be `None` and
+                        // the thinking block would be dropped from
+                        // `api_messages` — causing a DeepSeek HTTP 400 on the
+                        // next turn (V4 thinking-mode requires
+                        // `reasoning_content` replay). Inline-finalize the
+                        // thinking entry here so this branch is order-
+                        // independent.
+                        if app.streaming_thinking_active_entry.is_some() {
+                            if finalize_current_streaming_thinking(app) {
+                                transcript_batch_updated = true;
+                            }
+                            stash_reasoning_buffer_into_last_reasoning(app);
+                        }
                         if let Some(index) = app.streaming_message_index.take() {
                             let remaining = app.streaming_state.finalize_block_text(0);
                             if !remaining.is_empty() {
--- a/crates/tui/src/tui/ui/tests.rs
+++ b/crates/tui/src/tui/ui/tests.rs
@@ -3861,6 +3861,57 @@ fn engine_error_finalizes_active_thinking_block() {
 }
 
 #[test]
+fn message_complete_drain_preserves_thinking_when_thinking_complete_lost() {
+    // #861 RC3: when the engine bursts events, `MessageComplete` can be
+    // dispatched ahead of `ThinkingComplete`. Without the defensive drain,
+    // `app.last_reasoning` would be `None` at `last_reasoning.take()` time
+    // and the thinking block would be dropped from `api_messages`,
+    // causing a DeepSeek HTTP 400 on the next turn (V4 thinking-mode
+    // requires `reasoning_content` replay).
+    //
+    // This test exercises the head-of-handler drain in isolation: with a
+    // thinking entry still active and `last_reasoning` empty, the drain
+    // must transfer `reasoning_buffer` into `last_reasoning` before the
+    // remainder of `MessageComplete` reads it.
+    let mut app = create_test_app();
+
+    let _ = ensure_streaming_thinking_active_entry(&mut app);
+    app.thinking_started_at = Some(Instant::now());
+    app.streaming_state.start_thinking(0, None);
+    app.streaming_state.push_content(0, "deep reasoning text");
+    let _ = app.streaming_state.commit_text(0);
+    app.reasoning_buffer.push_str("deep reasoning text");
+
+    assert!(
+        app.last_reasoning.is_none(),
+        "precondition: ThinkingComplete has NOT fired"
+    );
+    assert!(
+        app.streaming_thinking_active_entry.is_some(),
+        "precondition: thinking entry is still active"
+    );
+
+    // Mirror the head of `EngineEvent::MessageComplete` — the new defensive
+    // drain installed by the #861 RC3 fix.
+    if app.streaming_thinking_active_entry.is_some() {
+        let _ = finalize_current_streaming_thinking(&mut app);
+        stash_reasoning_buffer_into_last_reasoning(&mut app);
+    }
+
+    assert!(
+        app.last_reasoning
+            .as_deref()
+            .is_some_and(|s| s.contains("deep reasoning text")),
+        "defensive drain must move reasoning into last_reasoning so the\
+         downstream `last_reasoning.take()` produces a Thinking block"
+    );
+    assert!(
+        app.streaming_thinking_active_entry.is_none(),
+        "thinking entry must be cleared after the drain"
+    );
+}
+
+#[test]
 fn second_thinking_block_appends_new_entry_in_same_active_cell() {
     // Real V4 turns can emit Thinking → Tool → Thinking → Tool before any
     // prose; the second thinking block should land as a fresh entry inside


Hmbown:
Cherry-picked into `work/v0.8.28` as `741f36d` — thanks @linzhiqin2003! Nice diagnosis on the order-of-events race.

CHANGELOG entry for v0.8.28:

> Streaming thinking blocks no longer drop their tail on MessageComplete (#861 RC3) — the active streaming entry is now drained into the finalized cell on `MessageComplete`, eliminating a data-loss path where the last chunk(s) of a streaming "thinking" reply could be discarded when `MessageComplete` arrived ahead of `ThinkingComplete` in a bursty event stream. Also closes a related HTTP 400 on the next turn (DeepSeek V4 requires `reasoning_content` replay for assistant messages that carry tool calls).

Hmbown:

Issue #1391, opened 2026-05-10
SessionUsage cache fields default to 0 when API omits cache data


Oliver-ZPLiu:
### Description

When the API response does not include `prompt_cache_hit_tokens` or `prompt_cache_miss_tokens`, `SessionUsage` silently shows `cache_creation_input_tokens=0` and `cache_read_input_tokens=0`.

This is indistinguishable from "zero cache hits/misses." The model then reports "cache hit rate 0%" or "cache miss" as if the data was observed — when in fact the API simply did not report it.

### Steps to reproduce

1. Use a model or response path that omits cache telemetry from the API response
2. Check session usage: cache fields will show `0` instead of `null`/`unknown`
3. The model may claim "cache hit rate 0%" based on this data

### Expected behavior

Cache fields should be `null`/`None` when the API did not report them, so consumers can distinguish "zero" from "unknown."

### Actual behavior

Cache fields default to `0`, which is indistinguishable from a genuine zero.

Hmbown:
Thanks for the report. Could you confirm if this issue still reproduces on the newly released v0.8.27?

Hmbown:
Resolved in v0.8.28 via PR #1392. SessionUsage cache fields now correctly use Option<u64> when the API omits cache data.

PR #1392, opened 2026-05-10
fix: cache usage shows 0 when API omits cache data

Oliver-ZPLiu:
## Summary

`SessionUsage.cache_creation_input_tokens` and `cache_read_input_tokens` are `u64` fields defaulting to `0`. When the API response does not include `prompt_cache_hit_tokens` or `prompt_cache_miss_tokens`, the cumulative session usage silently stays at `0`.

## Why it matters

The model receives `cache_hit_tokens=0` / `cache_miss_tokens=0` in the context, which is indistinguishable from "zero cache hits/misses." The model then reports "cache hit rate 0%" or "cache miss" as if the data was observed — when in fact the API simply did not report it. This is a source of false claims: the model is not lying, but the runtime gave it ambiguous data.

## Changes

| File | Change |
|------|--------|
| `crates/tui/src/prompts/base.md` | Added truthful reporting rules |
| `crates/tui/src/core/session.rs` | `cache_*_input_tokens`: `u64` → `Option<u64>`; added 4 unit tests |

## Test plan

- [x] `cargo test core::session::tests` — 4 passed

### DIFF
--- a/crates/tui/src/core/session.rs
+++ b/crates/tui/src/core/session.rs
@@ -90,10 +90,12 @@ pub struct Session {
 pub struct SessionUsage {
     pub input_tokens: u64,
     pub output_tokens: u64,
-    #[allow(dead_code)]
-    pub cache_creation_input_tokens: u64,
-    #[allow(dead_code)]
-    pub cache_read_input_tokens: u64,
+    /// Cache creation (miss) tokens. `None` when never observed by the API —
+    /// do NOT display as 0, which would be indistinguishable from "no misses".
+    pub cache_creation_input_tokens: Option<u64>,
+    /// Cache read (hit) tokens. `None` when never observed by the API —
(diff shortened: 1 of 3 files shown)


Hmbown:
ah! great find - thank you!

Hmbown:
Cherry-picked into `work/v0.8.28` as `5ccfe91` — thanks @Oliver-ZPLiu! The `Option<u64>` switch is the right call; the previous default-to-0 was actively misleading the model.

CHANGELOG entry for v0.8.28:

> Cache usage shows 0 when API omits cache data (#1391) — `SessionUsage.cache_creation_input_tokens` / `cache_read_input_tokens` are now `Option<u64>` instead of `u64` defaulting to 0. When the upstream API doesn't report cache hit/miss, the model sees `null` instead of misleading zeros, and reasoning about cache utilization is accurate.

Hmbown:

Issue #1118, opened 2026-05-08
Language has been configured to Chinese, but "thinking" outputs are  still in English


ljlbit:
## Problem

Language has been configured to Chinese, but "thinking" outputs are  still in English. Example is as bellow:

<img width="2006" height="1244" alt="Image" src="https://github.com/user-attachments/assets/36468251-1b6e-4279-9681-96638095359d" />

Hmbown:
thank you so much - will work on editing the prompt file to make this correct! 

I've experienced the opposite, so I know how strange and unsettling it is!!

pmsleepcheck:
> thank you so much - will work on editing the prompt file to make this correct!
> 
> I've experienced the opposite, so I know how strange and unsettling it is!!

I have the opposite view on the proposal
English can reduce token consumption
Configuring a local translation model seems to be a better option
Dsv4 seems to be more inclined to use English thought chains than v3 and r1
There's nothing wrong with that in programming.
我对该提议持相反观点
英语更能减少token消耗
配置本地翻译模型似乎是更好的选择
dsv4相较于v3andr1似乎更偏向与使用英语思维链
这在编程中没什么不好的

Currently, in use, the model can naturally use the English thought chain to think and interact with users in Chinese. There is nothing wrong with that
 目前在使用中 模型已经能很自然的使用英语思维链思考 并用中文与用户交互 这没什么不对的

masterkkkkkong:
> > thank you so much - will work on editing the prompt file to make this correct!非常感谢——我会努力编辑提示文件，让它更正确！
> > I've experienced the opposite, so I know how strange and unsettling it is!!我经历过相反的情况，所以我知道这有多奇怪和令人不安！！
> 
> I have the opposite view on the proposal我对这个提议持相反的看法 English can reduce token consumption英语可以减少代币消耗 Configuring a local translation model seems to be a better option配置本地翻译模型似乎是更好的选择 Dsv4 seems to be more inclined to use English thought chains than v3 and r1《黑暗之魂4》似乎更倾向于使用英文思维链，而非第三版和第一版 There's nothing wrong with that in programming.编程中这没什么问题。 我对该提议持相反观点 英语更能减少token消耗 配置本地翻译模型似乎是更好的选择 dsv4相较于v3andr1似乎更偏向与使用英语思维链 这在编程中没什么不好的
> 
[...]

PR #1398, opened 2026-05-10
fix(prompts): strengthen language directive so thinking matches user (#1118)

linzhiqin2003:
## Summary

[#1118](https://github.com/Hmbown/DeepSeek-TUI/issues/1118): \"Language has been configured to Chinese, but thinking outputs are still in English.\" Maintainer's reply:

> thank you so much - will work on editing the prompt file to make this correct!

This PR is that edit. The existing language directive already said \"both for `reasoning_content` and for the final reply\", but V4 falls into a failure mode where it mirrors the user message for the **final answer** while quietly defaulting to English for **thinking** when the surrounding context (system prompt, code, error logs) is English-heavy.

## Three changes to \`prompts/base.md\`

1. **Bold the requirement and name the failure mode.** The line that used to say \"your `reasoning_content` and final reply must both be in Simplified Chinese, even when the `lang` field is `en`\" now adds \"even when the surrounding system prompt is in English, and even when the task context (source code, error logs, README excerpts) is overwhelmingly English.\" That second clause is exactly the trigger #1118 reports.

2. **Mid-session-switch rule covers thinking too.** Today the prompt says \"switch with them\" but doesn't reinforce that this includes `reasoning_content` — V4 sometimes carries the previous turn's reasoning language forward when the new turn is a single-word follow-up. Added \"on the very next turn — including in `reasoning_content`. Don't carry the previous turn's language forward.\"

3. **Explicit-override clause.** Issue commenter @pmsleepcheck prefers English thinking for token-cost reasons; the prompt now respects \"think in English\" / \"用英文思考\" as a per-session override. Only thinking is overridable — the final reply still mirrors message language.

## Test

\`\`\`
$ cargo test -p deepseek-tui -- language_section_carries
test prompts::tests::language_section_carries_reasoning_content_directives_for_1118 ... ok
test result: ok. 1 passed; 0 failed
\`\`\`

The test pins four load-bearing phrases:
- \`reasoning_content\` (the section must call it out by name)
- \`must both be in Simplified Chinese\` (the bold requirement)
- \`overwhelmingly English\` (the context-is-English caveat)
- both \`think in English\` and \`用英文思考\` (the user-override examples)

The existing \`system_prompt_for_mode_with_context_is_byte_stable_for_unchanged_workspace\` invariant test still passes, so the cached prefix for unchanged sessions remains byte-stable.

## Why prompt-only and not a config option

A prompt fix matches the maintainer's stated direction (\"will work on editing the prompt file\"), is testable today, and ships without surface-area expansion. A future \`[language] reasoning = \"english\"|\"match_user\"\` config could replace the explicit-override paragraph if reviewers want a more structured opt-out — happy to follow up.

## Acceptance gates

- [x] \`cargo test -p deepseek-tui -- prompts\` — passing (including byte-stability invariants)

Refs #1118

### DIFF
--- a/crates/tui/src/prompts.rs
+++ b/crates/tui/src/prompts.rs
@@ -542,6 +542,44 @@ mod tests {
     }
 
     #[test]
+    fn language_section_carries_reasoning_content_directives_for_1118() {
+        // #1118 ("Language has been configured to Chinese, but thinking
+        // outputs are still in English"): the base prompt's language
+        // section is the only knob that steers V4's `reasoning_content`
+        // language. Pin the load-bearing phrases so a future innocuous
+        // edit can't quietly drop them.
+        let lang = BASE_PROMPT;
+        assert!(
+            lang.contains("reasoning_content"),
+            "language section must explicitly call out reasoning_content"
+        );
+        // Bold "must both be in Simplified Chinese" anchor — strong
+        // emphasis aimed at the failure mode V4 falls into where it
+        // mirrors the user message for the final reply but defaults to
+        // English for thinking.
+        assert!(
+            lang.contains("must both be in Simplified Chinese"),
+            "expected the bold Simplified Chinese requirement"
+        );
+        // "overwhelmingly English" — addresses the specific trigger
+        // where a Chinese question lands on a codebase whose system
+        // prompt and context are English-heavy.
+        assert!(
+            lang.contains("overwhelmingly English"),
+            "expected the context-is-English caveat"
+        );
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks @linzhiqin2003 — your fix landed on `main` as [`f56f73e37`](https://github.com/Hmbown/DeepSeek-TUI/commit/f56f73e37) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

PR #1399, opened 2026-05-10
feat(project-context): merge global AGENTS.md with project AGENTS.md (#1157)

### LINKED ISSUE #1157: [Feature Request] Support loading a global AGENTS.md (~/.deepseek/AGENTS.md)
Currently, deepseek-tui only loads the AGENTS.md file from the current project directory. Since I work with many different projects, manually copying the AGENTS.md file into every single project directory is quite tedious and difficult to maintain.

I would like deepseek-tui to support loading a global AGENTS.md file (e.g., located at ~/.deepseek/AGENTS.md) upon startup.
Ideally, if both a global and a local AGENTS.md exist, the application could merge them, or at least use the global one as a fallback when a local one is not found.

Similar AI CLI/TUI tools like claude and codex already support this kind of global configuration loading. Adding this feature would greatly improve the user experience for developers who manage multiple repositories.

linzhiqin2003:
## Summary

Closes [#1157](https://github.com/Hmbown/DeepSeek-TUI/issues/1157). Reporter:

> Currently, deepseek-tui only loads the AGENTS.md file from the current project directory … I would like deepseek-tui to support loading a global AGENTS.md file (e.g., located at ~/.deepseek/AGENTS.md) upon startup. Ideally, if both a global and a local AGENTS.md exist, the application could merge them, or at least use the global one as a fallback when a local one is not found.

Maintainer:

> yes that makes sense! am working on getting this organizational structure better today so that worktrees etc can feel like an intended way of using this.

The **fallback** half already shipped — \`load_global_agents_context\` loads \`~/.deepseek/AGENTS.md\` when no workspace context exists. The **merge** half (\"if both a global and a local AGENTS.md exist…\") didn't: the moment a project AGENTS.md was found, the global file was silently ignored.

## Fix

In \`load_project_context_with_parents_and_home\`, always probe the global file. The behaviour now splits cleanly:

| Files present | Behaviour |
|---|---|
| Only project | Unchanged from before |
| Only global | Unchanged from before (fallback path) |
| **Both** | **Merged** — global first, then project (so project wins last-word precedence with the model) |
| Neither | Unchanged — falls through to the existing \`auto_generate_context\` path |

The merged block is wrapped in labelled HTML-style fences so the model can tell layers apart when it explains which rule it followed:

```
<!-- global: /home/user/.deepseek/AGENTS.md -->
…global content…

<!-- project (overrides global where they conflict) -->
…project content…
```

\`source_path\` continues to point at the more-specific file (project > global) so the user knows where to edit when something looks wrong.

## Tests

\`\`\`
$ cargo test -p deepseek-tui -- project_context
test project_context::tests::test_local_and_global_agents_merge_when_both_exist ... ok       ← new
test project_context::tests::test_global_agents_only_no_project_unchanged_fallback ... ok    ← new
test project_context::tests::test_load_global_agents_when_project_has_no_context ... ok      ← unchanged
test project_context::tests::test_invalid_global_agents_warns_and_falls_back_to_generated_context ... ok
…
test result: ok. 19 passed; 0 failed
\`\`\`

- **\`test_local_and_global_agents_merge_when_both_exist\`** — the actual #1157 scenario. Asserts both blocks are present, global precedes project, and the labelled separator appears between them.
- **\`test_global_agents_only_no_project_unchanged_fallback\`** — sanity that the global-only case doesn't pick up the merge framing (no \`project (overrides global\` label when there's nothing to override).
- The pre-existing \`test_load_global_agents_when_project_has_no_context\` still passes, so the fallback contract is preserved.

The historical \`test_local_agents_takes_priority_over_global_agents\` test (which asserted the *old* drop-global behaviour) was rewritten — it now asserts the merge contract instead.

## Acceptance gates

- [x] \`cargo test -p deepseek-tui -- project_context\` — 19/19

Closes #1157

### DIFF
--- a/crates/tui/src/project_context.rs
+++ b/crates/tui/src/project_context.rs
@@ -381,13 +381,32 @@ fn load_project_context_with_parents_and_home(
         }
     }
 
-    if !ctx.has_instructions()
-        && let Some(global_ctx) = load_global_agents_context(workspace, home_dir)
-    {
+    // Always check `~/.deepseek/AGENTS.md` so user-wide preferences
+    // travel into every session (#1157). When both global and project
+    // instructions exist, the global block prepends the project's so
+    // workspace overrides win the last word; when only global exists,
+    // it continues to serve as the fallback. `source_path` keeps
+    // pointing at the more-specific source (project > global) for
+    // display purposes.
+    if let Some(global_ctx) = load_global_agents_context(workspace, home_dir) {
         ctx.warnings.extend(global_ctx.warnings.iter().cloned());
-        if global_ctx.has_instructions() {
-            ctx.instructions = global_ctx.instructions;
-            ctx.source_path = global_ctx.source_path;
+        if let Some(global_text) = global_ctx.instructions {
+            match ctx.instructions.take() {
+                Some(project_text) => {
+                    ctx.instructions = Some(merge_global_and_project_instructions(
+                        &global_text,
+                        global_ctx.source_path.as_deref(),
+                        &project_text,
+                    ));
+                    // Leave `ctx.source_path` pointing at the project /
+                    // parent file — that's the location the user might
+                    // want to edit when something looks wrong.
+                }
+                None => {
+                    ctx.instructions = Some(global_text);
+                    ctx.source_path = global_ctx.source_path;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @linzhiqin2003 — your fix landed on `main` as [`a5c4a21c9`](https://github.com/Hmbown/DeepSeek-TUI/commit/a5c4a21c9) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

Issue #1186, opened 2026-05-08
feat(execpolicy): add typed persistent permission rules


greyfreedom:
## Summary

Add typed persistent permission rules to the execution policy/config layer.

The first version should support rules scoped by:

- tool name
- command prefix
- workspace-relative path pattern
- decision: `allow`, `deny`, or `ask`

This should extend the existing `execpolicy` ruleset model instead of introducing a separate permission engine.

## Motivation

DeepSeek-TUI already has shell command approval rules through `auto_allow` / deny prefixes, layered rulesets, and bash arity-aware command matching.

However, the current model is still mostly shell-command focused. Repeated tool workflows would benefit from a typed permission rule schema that can represent decisions such as:

```toml
[[permissions.rules]]
tool = "exec_shell"
decision = "allow"
command = "cargo test"

[[permissions.rules]]
tool = "exec_shell"
decision = "deny"
command = "rm -rf"

[[permissions.rules]]
tool = "file_edit"
decision = "ask"
path = "src/**"
```

This provides a foundation for future UI support such as “always allow this pattern” while keeping the first implementation focused on policy/config behavior.

## Proposed Behavior

- Add a typed permission rule type with:
  - `tool`
  - `decision`
  - optional `command`
  - optional `path`
- Support decisions:
  - `allow`
  - `deny`
  - `ask`
- Keep existing `auto_allow` / `auto_deny` behavior compatible by mapping them to `exec_shell` rules.
- Reuse the existing bash arity logic for shell command matching.
- Support workspace-relative glob-style matching for file paths.
- Preserve deny-wins behavior.
- Treat `ask` as an explicit rule that forces approval even when the broader policy would otherwise allow execution.

## Non-goals

- No TUI approval modal changes in the first PR.
- No “always allow this pattern” persistence UI yet.
- No complex permission DSL.
- No regex or boolean expressions.
- No full MCP permission coverage yet.

## Acceptance Criteria

- `execpolicy` exposes typed permission rules.
- Config can deserialize `[[permissions.rules]]`.
- Legacy `auto_allow` / `auto_deny` still work.
- `deny > ask > allow` precedence is covered by tests.
- Shell command rules continue to use bash arity matching.
- File path rules support workspace-relative glob matching.
- Tests cover config parsing, legacy compatibility, precedence, and path matching.

## Implementation Plan
#2242

Hmbown:
thank you so much - very helpful with the PRs!!!! this is going in 0.8.22

[...]

PR #1413, opened 2026-05-11
feat(tui): route shell and file tool approvals through typed execpolicy rules

greyfreedom:
## Summary

This PR wires the typed permission rules introduced in PR1 into the tool execution flow.

Shell and file-oriented tools now consult the `ExecPolicyEngine` before falling back to the existing approval behavior. This allows persistent rules such as:

- allow/deny/ask shell command prefixes via `exec_shell`
- allow/deny/ask file paths via `read_file`, `write_file`, `edit_file`, `list_dir`, and `apply_patch`
- compatibility aliases such as `file_read`, `file_write`, and `file_edit`

This does not add the approval-prompt persistence UI yet; that remains follow-up work.

## Changes

- Add `deepseek-execpolicy` to the TUI crate.
- Load `auto_allow`, `auto_deny`, and `[permissions.rules]` from TUI config into an `ExecPolicyEngine`.
- Pass the configured policy engine into all engine entry points:
  - interactive TUI
  - `deepseek -p`
  - runtime threads
- Apply typed permission decisions during tool planning:
  - `allow` skips the existing approval prompt
  - `ask` forces the existing approval path
  - `deny` blocks execution before the tool runs
  - unmatched rules preserve existing behavior
- Support file path matching for:
  - workspace-relative paths
  - absolute paths under the workspace
  - symlink/canonicalized paths where applicable
- Apply permission checks to nested `multi_tool_use.parallel` calls so read-only file tools cannot bypass rules by being wrapped in a parallel call.
- Support project-level permission config only when it tightens policy:
  - project `auto_deny` is applied
  - project `deny` / `ask` rules are applied
  - project `auto_allow` and `allow` rules are ignored with warnings

## Safety Notes

This PR keeps project-level permission config conservative. A repository-local config can require extra approval or deny access, but it cannot grant itself persistent allow rules.

For multi-path tools such as `apply_patch`, the implementation remains conservative:

- any matching deny blocks the call
- any matching ask requires approval
- allow is used only when every extracted path is covered
- partial matches fall back to the existing approval behavior

## Tests

Validated with:

```bash
cargo fmt --all
cargo test -p deepseek-tui typed_permission
cargo test -p deepseek-tui project_overlay_applies_only_tightening_permission_rules
cargo clippy -p deepseek-tui -p deepseek-config --all-targets -- -D warnings
cargo test -p deepseek-tui
git diff --check
```

## Depends  On
This PR is stacked on top of PR1: feat/execpolicy-typed-rules
Part of #1186
Depends on #1189

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           sudo apt-get install -y libdbus-1-dev pkg-config
       - uses: Swatinem/rust-cache@v2
         with:
-          cache-bin: false
+          cache-bin: "false"
       - name: Check formatting
         run: cargo fmt --all -- --check
       # Mirror the release-workflow `parity` gate exactly. Anything that
@@ -86,7 +86,7 @@ jobs:
           sudo apt-get install -y libdbus-1-dev pkg-config
       - uses: Swatinem/rust-cache@v2
         with:
-          cache-bin: false
+          cache-bin: "false"
       - name: Run tests
         run: cargo test --workspace --all-features --locked
       - name: Lockfile drift guard
@@ -118,7 +118,7 @@ jobs:
           node-version: 20
       - uses: Swatinem/rust-cache@v2
         with:
-          cache-bin: false
+          cache-bin: "false"
       - name: Build wrapper binaries
         run: cargo build --release --locked -p deepseek-tui-cli -p deepseek-tui
       - name: Smoke wrapper install and delegated entrypoints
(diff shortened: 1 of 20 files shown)


Hmbown:

greyfreedom:
> This PR was opened before the v0.8.41 rebrand and is now stale. Feel free to rebase onto current `main` and reopen. 鲸鱼兄弟们等你 🐋

Since I cannot reopen the current PR, I created a new PR #2053  .

PR #1416, opened 2026-05-11
Add changelog command

zhuangbiaowei:
## Summary
- add a /change command that reads the latest CHANGELOG.md version section
- show localized command help/header text
- request model translation for non-English UI locales
- add unit coverage for changelog extraction and command behavior

## Verification
- cargo fmt --all
- cargo test -p deepseek-tui change
- cargo build

### DIFF
--- a/crates/tui/src/commands/change.rs
+++ b/crates/tui/src/commands/change.rs
@@ -1,0 +1,375 @@
+//! `/change` command — show the latest changelog entry, translated to the
+//! user's locale when it is not English.
+//!
+//! Usage: `/change`
+//!
+//! Uses the DeepSeek-TUI changelog embedded at compile time, extracts the
+//! most recent version section, and displays it. When the UI locale is not
+//! English and the current session can reach a model, the command also fires a
+//! `SendMessage` action that asks the model to translate the changelog into
+//! the user's language.
+
+use crate::localization::{Locale, MessageId, tr};
+use crate::tui::app::{App, AppAction};
+
+use super::CommandResult;
+
+/// Maximum length of the changelog excerpt we'll show inline (characters).
+/// If the changelog section exceeds this, we truncate and show a notice.
+/// 4096 chars is large enough for most version entries.
+const MAX_INLINE_CHANGELOG_CHARS: usize = 4096;
+const DEEPSEEK_TUI_CHANGELOG: &str = include_str!("../../../../CHANGELOG.md");
+
+/// Execute the `/change` command.
+pub fn change(app: &mut App) -> CommandResult {
+    let latest_section = match extract_latest_changelog_section(DEEPSEEK_TUI_CHANGELOG) {
+        Some(s) => s,
+        None => {
+            return CommandResult::error(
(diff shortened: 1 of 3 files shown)


Hmbown:
Thanks @zhuangbiaowei — this is a genuinely good idea, and I’d like to see it land. I reviewed it for possible inclusion in v0.8.29, but I’m going to keep it for v0.8.30 instead because v0.8.29 is already in late release flow and this feature deserves one more careful pass rather than being rushed in at the edge.

What already looks good:

- No new crate dependencies in the PR commit.
- The section extractor recognizes this repo’s current changelog heading shape (`## [X.Y.Z] - YYYY-MM-DD`).
- The translation request uses the existing `AppAction::SendMessage` path rather than directly calling the LLM client from the slash-command handler, so the broad async shape is right.
- The PR includes useful parser and command tests.

What I think we should tighten before landing it:

- `crates/tui/src/commands/change.rs` currently reads `app.workspace.join("CHANGELOG.md")`. That works when DeepSeek-TUI is launched from this checkout, but installed users usually run it inside some other project. In that case `/change` would show that project’s changelog or say no changelog was found, rather than reliably showing DeepSeek-TUI release notes. We should either bundle/source DeepSeek-TUI’s own release notes, or explicitly frame this as a project-changelog command.
- For non-English locales, the command always sends a normal model turn for translation. If the provider is offline, the API key is missing, or the network is blocked, we should make the fallback explicit in the command output, e.g. English changelog plus a localized “translation unavailable” note.
- The tests currently document the missing-workspace-`CHANGELOG.md` case as an error. For the release-notes version, we should add coverage for the installed-user path plus offline / missing-key translation fallback behavior.

For v0.8.30, I’d suggest keeping the parser and command registration, then adding:

- a reliable release-notes source independent of the user’s current workspace,
- tests using a real `## [0.8.29] - 2026-05-11`-style fixture,
- explicit fallback behavior before firing the translation turn,
- a smoke test that `/change` works from a directory that is not the DeepSeek-TUI checkout.

Leaving this PR open so we can iterate. Thank you again — the command would be a nice fit alongside the new note-management commands, and I don’t want the late v0.8.29 release window to shortchange it.

zhuangbiaowei:
I understand your feedback, and I will continue working on this PR as you suggested until it becomes more polished.

zhuangbiaowei:
Updated the PR for the review pass. Changes in the amended commit:

- /change now reads the DeepSeek-TUI changelog embedded at build time instead of app.workspace/CHANGELOG.md, so installed users running inside another project still see DeepSeek-TUI release notes.
- Non-English locales now show an explicit English fallback note before any translation turn. If the session is offline or has no API key, the command does not enqueue a model turn and leaves the English release notes visible.
- Added coverage for a 0.8.29-style changelog fixture, workspace-independent release notes, ignoring a project-local CHANGELOG.md, and missing-key/offline translation fallback behavior.

Verification run locally:
- cargo fmt --all
- cargo test -p deepseek-tui change
- cargo build

PR #1428, opened 2026-05-11
fix: treat deepseek-chat / deepseek-reasoner aliases as reasoning models

Beltran12138:
## Problem

Setting `default_text_model = "deepseek-chat"` (the value `deepseek auth` / onboarding writes) makes the TUI 400 on the **second turn** of any session that involves a tool call:

```
HTTP 400: The `reasoning_content` in the thinking mode must be passed back to the API.
```

(Some users also see this surface as `Authentication failed: Authentication Fails (governor)` when the rejected request is the streaming one.)

### Root cause

`requires_reasoning_content()` in `crates/tui/src/client/chat.rs` only matched literal `deepseek-v4*` model IDs:

```rust
fn requires_reasoning_content(model: &str) -> bool {
    let lower = model.to_lowercase();
    lower.contains("deepseek-v4")
        || lower.contains("reasoner")
        || lower.contains("-reasoning")
        || lower.contains("-thinking")
        || has_deepseek_r_series_marker(&lower)
}
```

But `deepseek-chat` and `deepseek-reasoner` are DeepSeek's **public API aliases** — server-side they resolve to `deepseek-v4-flash` and `deepseek-v4-pro` respectively, and the entire V4 family has thinking mode **enabled by default** ([API docs](https://api-docs.deepseek.com/guides/thinking_mode)). Because the alias names don't contain the substring `deepseek-v4`, `requires_reasoning_content("deepseek-chat")` returns `false`, so:

- `should_replay_reasoning_content()` → `false`
- `build_chat_messages_with_reasoning(..., include_reasoning = false, ...)` skips the placeholder
- `sanitize_thinking_mode_messages()` early-returns
- the tool-call assistant message goes out with no `reasoning_content` → DeepSeek 400

### Repro

1. Fresh `~/.deepseek/config.toml` with `default_text_model = "deepseek-chat"` (or just run `deepseek auth set --provider deepseek` and accept the default model).
2. `deepseek` in any directory.
3. Ask something that triggers one tool call (e.g. "list the files here").
4. Send a second message → 400.

`models` lists only `deepseek-v4-flash` / `deepseek-v4-pro`, so users who pick `deepseek-chat` from docs/onboarding are the affected population.

## Fix

Extend `requires_reasoning_content()` to recognise the `deepseek-chat` and `deepseek-reasoner` alias prefixes, so alias handling matches the explicit V4 IDs. `starts_with` (not `contains`) is used so unrelated names aren't swept in, while suffixed variants used by proxied deployments (`deepseek-chat:free`, `deepseek-reasoner-2025-05`, …) are still covered.

The `reasoning_effort = "off"` escape hatch is unaffected — it's checked in `should_replay_reasoning_content()`, which this PR doesn't touch.

## Tests

New `alias_thinking_detection_tests` module in `chat.rs`:

- aliases (`deepseek-chat`, `deepseek-reasoner`, case-insensitive) require reasoning_content
- explicit V4 IDs still match (regression guard for the existing branch)
- non-thinking IDs (`deepseek-v3`, `deepseek-coder`) and other providers (`gpt-4o`, `claude-sonnet-4-6`) stay excluded
- suffixed alias variants are covered
- `reasoning_effort = "off"` / `"disabled"` overrides alias detection

`cargo fmt --check` passes locally. (I don't have an MSVC toolchain on this machine, so I'm relying on CI for `cargo clippy` / `cargo test` — happy to iterate if anything's red.)

## Environment

- OS: Windows 11
- `deepseek --version`: 0.8.28
- `rustc --version`: 1.95.0

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -1506,7 +1506,16 @@ fn log_thinking_mode_violations(body: &Value) {
 
 fn requires_reasoning_content(model: &str) -> bool {
     let lower = model.to_lowercase();
+    // V4-family direct model IDs.
     lower.contains("deepseek-v4")
+        // Public DeepSeek API aliases routed server-side to the V4 family.
+        // `deepseek-chat` resolves to `deepseek-v4-flash` and `deepseek-reasoner`
+        // resolves to `deepseek-v4-pro`; both have thinking mode enabled by
+        // default, so any assistant message carrying tool_calls must replay
+        // `reasoning_content` on subsequent turns or the API returns 400.
+        || lower.starts_with("deepseek-chat")
+        || lower.starts_with("deepseek-reasoner")
+        // Generic reasoning markers used by custom/proxied deployments.
         || lower.contains("reasoner")
         || lower.contains("-reasoning")
         || lower.contains("-thinking")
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @Beltran12138 — your fix landed on `main` as [`ca284d1fc`](https://github.com/Hmbown/DeepSeek-TUI/commit/ca284d1fc) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

PR #1429, opened 2026-05-11
fix(tools): hydrate deferred schemas before execution

### LINKED ISSUE #1419: Deferred tools can fail on first use due to hidden schemas causing argument guessing
```md
## Problem

`edit_file` and `checklist_update` can fail on first use because they are deferred-loaded tools whose exact schemas may not be visible in the initial active tool context.

When the model only knows the tool conceptually, it may guess argument names from other agent/tool conventions.

Examples:

```json
edit_file({
  "path": "src/foo.rs",
  "old_string": "...",
  "new_string": "..."
})
```

Actual schema:

```json
edit_file({
  "path": "src/foo.rs",
  "search": "...",
  "replace": "..."
})
```

Another example:

```json
checklist_update({
  "todos": [...]
})
```

Actual schema:

```json
checklist_update({
  "id": 1,
  "status": "completed"
})
```

## Current Behavior

The invalid call is executed as a normal tool call attempt. It fails validation, then the error is written back as a normal tool error. On the next turn, the model may see the error and the now-active schema, then retry with corrected arguments.

## Why This Is a Problem

This treats schema hydration / tool discovery as an execution failure.

The tool did not fail because the requested operation was invalid; it failed because the model did not yet have the exact runtime schema for a deferred tool.

This can lead to:

- avoidable failed tool calls
- noisy transcripts
- misleading tool failure accounting
- extra model turns
- model-dependent recovery instead of runtime-guaranteed recovery

## Expected Behavior

When a deferred tool is requested before its exact schema is active, the runtime should not execute the call immediately.

Instead, it should hydrate the tool schema and return a synthetic tool result explaining the correct schema.

For example:

```text
Tool `edit_file` was deferred and has now been loaded.

Expected schema:
  path: string
  search: string
  replace: string

Received fields:
  path, old_string, new_string

Likely correction:
  old_string -> search
  new_string -> replace

The tool was not executed. Retry with the loaded schema.
```

## Suggested Fix

Add a deferred-tool preflight step:

1. Detect that the requested tool exists but is currently deferred.
2. Activate / hydrate the tool schema.
3. Do not execute the current call.
4. Return a synthetic tool result with:
   - expected fields
   - received fields
   - missing fields
   - unexpected fields
   - likely corrections if obvious
5. Let the model retry with valid arguments.

For write-capable tools like `edit_file`, avoid silently auto-mapping aliases and executing the operation.
```

SamhandsomeLee:
## Summary

Fixes #1419.

- Hydrate deferred tool schemas on first direct model request instead of executing immediately.
- Return a synthetic tool result with `executed=false` and `retry_required=true` so the model can retry with the visible schema.
- Add `edit_file` guidance for common guessed aliases like `old_string -> search` and `new_string -> replace`.
- Avoid running post-edit LSP hooks for schema hydration results because no file edit occurred.
- Keep deferred tool activation batch-scoped so multiple same-tool calls in one response all hydrate instead of executing.

## Test Plan

- `cargo fmt --all --check`
- `cargo test -p deepseek-tui deferred`
- `cargo test -p deepseek-tui`

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -2049,10 +2049,12 @@ use self::tool_catalog::{
     CODE_EXECUTION_TOOL_NAME, MULTI_TOOL_PARALLEL_NAME, REQUEST_USER_INPUT_NAME,
     active_tools_for_step, build_model_tool_catalog, ensure_advanced_tooling,
     execute_code_execution_tool, execute_tool_search, initial_active_tools, is_tool_search_tool,
-    maybe_activate_requested_deferred_tool, missing_tool_error_message,
+    maybe_hydrate_requested_deferred_tool, missing_tool_error_message,
 };
 #[cfg(test)]
-use self::tool_catalog::{TOOL_SEARCH_BM25_NAME, should_default_defer_tool};
+use self::tool_catalog::{
+    TOOL_SEARCH_BM25_NAME, maybe_activate_requested_deferred_tool, should_default_defer_tool,
+};
 use self::tool_execution::emit_tool_audit;
 use self::tool_setup::sandbox_policy_for_mode;
 
--- a/crates/tui/src/core/engine/tests.rs
+++ b/crates/tui/src/core/engine/tests.rs
@@ -367,6 +367,79 @@ fn model_tool_catalog_applies_native_and_mcp_deferral() {
 }
 
 #[test]
+fn deferred_edit_file_first_use_hydrates_schema_without_execution() {
+    let mut edit = api_tool("edit_file");
+    edit.defer_loading = Some(true);
+    edit.input_schema = json!({
+        "type": "object",
+        "properties": {
+            "path": { "type": "string" },
+            "search": { "type": "string" },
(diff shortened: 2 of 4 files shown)


Hmbown:
Thanks @SamhandsomeLee — your fix landed on `main` as [`9279d92f4`](https://github.com/Hmbown/DeepSeek-TUI/commit/9279d92f4) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

SamhandsomeLee:
> Thanks @SamhandsomeLee — your fix landed on `main` as [`9279d92f4`](https://github.com/Hmbown/DeepSeek-TUI/commit/9279d92f4) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.
> 
> Credit lives in the commit message and the `CHANGELOG.md` entry for v0.8.29.

thanks!

PR #1431, opened 2026-05-11
  Fix flaky workspace and env-sensitive tests

reidliu41:
## Summary

  Fixes several flaky test failures that can appear depending on the local test
  environment and parallel test timing.

  The main issue was in session workspace scoping. `find_git_root()` treated any
  parent `.git` path as a valid repository marker. If a stray or invalid `.git`
  directory exists above tempfile workspaces, such as `/tmp/.git`, unrelated
  temporary workspaces can be incorrectly treated as belonging to the same Git
  repository. This made `latest_session_for_workspace` return the newest session
  from the wrong workspace.

  This PR tightens Git root detection so:

  - `.git` directories must contain `HEAD`
  - `.git` files must use the `gitdir:` worktree format

  It also fixes two test isolation issues:

  - Env-mutating tests now share the crate-wide test env lock instead of using
    local locks that can race with sibling tests.
  - The streamable HTTP MCP test mock server now serves until the test ends,
    instead of stopping after a fixed number of accepted connections.

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -5260,13 +5260,6 @@ mod setup_helper_tests {
     use std::collections::BTreeSet;
     use tempfile::TempDir;
 
-    // Serialize tests that mutate process-global env vars. Without this,
-    // `cargo test` runs them in parallel and they race on `DEEPSEEK_API_KEY`,
-    // causing intermittent CI failures (one test reads while another's set
-    // is still active). `unwrap_or_else` recovers from poisoning so a panic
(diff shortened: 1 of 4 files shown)


Hmbown:
Thanks @reidliu41 — your fix landed on `main` as [`7ac79799e`](https://github.com/Hmbown/DeepSeek-TUI/commit/7ac79799e) and shipped in **v0.8.29**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

PR #1458, opened 2026-05-11
fix(edit_file): reject identical search and replace values

### LINKED ISSUE #1457: edit_file should reject identical search and replace
# edit_file should reject identical search and replace

## Summary

When `edit_file` is called with `search == replace`, the tool returns `(no
textual changes)` without an error. Agents frequently copy the `search` value
into the `replace` parameter and forget to mutate it, then retry with the same
mistake, causing 3–4 serial failures before a successful edit.

## Steps to reproduce

1. Call `edit_file` with:
   ```
   search:  'a := "foo"'
   replace: 'a := "foo"'
   ```
2. Tool returns `(no textual changes)` — no error.
3. Agent does not notice the no-op, retries with the same values.
4. Cycle repeats.

## Proposed fix

Add a guard clause before execution:

```
if search == replace {
    return error: "edit_file: search and replace are identical, no change intended"
}
```

- Zero cost for legitimate calls (nobody intentionally passes identical values).
- No API signature change needed.
- Immediately breaks the agent's retry loop on the first mistake.

## Observed impact

In a multi-hour agent session involving ~30 `edit_file` calls, approximately
40% of calls (12 of 30) required retries due to this issue, with an average of
2.5 retries per call. Total wasted round-trips: ~18.

Oliver-ZPLiu:
## Summary

Adds a guard clause in `EditFileTool::execute` that rejects calls where `search == replace` with a clear error message.

## Why it matters

When `edit_file` is called with identical search and replace values, the tool currently returns `(no textual changes)` without error. Agents frequently copy the search value into the replace parameter and forget to mutate it, then retry with the same mistake, causing 3-4 serial failures before a successful edit. In one reported session, ~40% of edit_file calls required retries due to this issue.

The guard clause breaks the retry loop on the first attempt by returning an explicit error: `edit_file: search and replace are identical, no change intended`.

## Files changed

| File | Change |
|------|--------|
| `crates/tui/src/tools/file.rs` | +5 lines: guard clause after param parsing |

Fixes #1457.

### DIFF
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -331,6 +331,12 @@ impl ToolSpec for EditFileTool {
         let search = required_str(&input, "search")?;
         let replace = required_str(&input, "replace")?;
 
+        if search == replace {
+            return Err(ToolError::invalid_input(
+                "edit_file: search and replace are identical, no change intended",
+            ));
+        }
+
         let file_path = context.resolve_path(path_str)?;
 
         let contents = fs::read_to_string(&file_path).map_err(|e| {


Hmbown:
Thanks @Oliver-ZPLiu — your fix landed on `main` as [`1352a65ac`](https://github.com/Hmbown/DeepSeek-TUI/commit/1352a65ac) and shipped in **v0.8.30**. Closing this PR now that the code is on `main`. Sorry for the lag — the v0.8.31 release ships an auto-close workflow that catches this case automatically from now on.

Credit lives in the commit message and the `CHANGELOG.md` entry for v0.8.30.

PR #1460, opened 2026-05-11
fix: reject no-op edit_file replacements

### LINKED ISSUE #1457: edit_file should reject identical search and replace

xiluoduyu:
## Summary
- reject edit_file calls when search and replace are identical
- add a regression test that verifies the tool returns an input error and leaves the file unchanged

Closes #1457

## Verification
- cargo fmt --all --check
- cargo test -p deepseek-tui --bin deepseek-tui tools::file::tests::test_edit_file_rejects_identical_search_and_replace

### DIFF
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -331,6 +331,12 @@ impl ToolSpec for EditFileTool {
         let search = required_str(&input, "search")?;
         let replace = required_str(&input, "replace")?;
 
+        if search == replace {
+            return Err(ToolError::invalid_input(
+                "search and replace are identical, no change intended",
+            ));
+        }
+
         let file_path = context.resolve_path(path_str)?;
 
         let contents = fs::read_to_string(&file_path).map_err(|e| {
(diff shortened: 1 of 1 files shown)


CrepuscularIRIS:
Auto cleanup: stale AGC/placeholder/open PR interfering with duplicate-gate; closing to unblock candidate testing.

PR #1486, opened 2026-05-12
"fix-session-picker-clean-titles-and-preview"

tyouter:
Fix three display issues in Ctrl+R session picker:

1. Session titles show user prompt instead of <turn_meta>
2. Preview strips system prefix from first user message
3. Preview skips thinking blocks and empty messages

Implementation:
- extract_user_prompt(): strips <turn_meta> prefix
- strip_thinking_tags(): removes thinking XML tags from assistant text
- format_session_line + build_preview_lines: apply at display time

Tests: cargo test -p deepseek-tui -- extract_user_prompt build_preview
Result: 12 passed, 0 failed, 0 ignored

Developed with AI assistance. Ref: REQ-20260511-001

### DIFF
--- a/crates/tui/src/session_manager.rs
+++ b/crates/tui/src/session_manager.rs
@@ -6,7 +6,6 @@
 //! - Resuming sessions by ID
 //! - Managing session lifecycle
 
-use crate::artifacts::ArtifactRecord;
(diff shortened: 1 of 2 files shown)


tyouter:
"Closing-local-test-needed"

PR #1500, opened 2026-05-12
fix(subagent): prevent review agents from spawning sub-agents

### LINKED ISSUE #1489: Sub-agent review delegation and context inheritance gaps
# Sub-agent review delegation and context inheritance gaps

## Summary

Two sub-agent behaviors make delegated work less predictable:

1. A `review` sub-agent can reinterpret itself as a coordinator and spawn more sub-agents instead of directly performing the review it was assigned.
2. Context written through the `note` tool, and enabled user memory for fresh `agent_spawn` children, can be missing from the child agent's initial context unless the parent explicitly forks the transcript.

### Review sub-agent recursively delegates

A parent spawned a review sub-agent with a task equivalent to "read 4 files and do a code review." The review child did not directly review the files. It spawned two additional sub-agents and returned that children had been started.

Result:

- Multiple agents reviewed overlapping scope.
- Runtime and token use increased.
- The parent received an incomplete synthesized review because not all child outputs were merged.

### Notes and memory are uneven across parent and child sessions

The parent can write session guidance through `note`, but a fresh `agent_spawn` child does not reliably receive those note contents. User memory is documented as inherited by sub-agents, but fresh spawns only get the role prompt plus task unless context is explicitly forked.

This creates a knowledge mismatch: the parent may follow local guidance while the child behaves as if only default system rules exist.

## Impact

- Review agents can multiply work instead of producing one authoritative review.
- Parent-specific constraints are easy to lose at the delegation boundary.
- The model may spend extra turns recovering from behavior that could have been prevented by role and context boundaries.

## Suggested fix

Recommended narrow implementation:

1. Strengthen the `review` role prompt so reviewer agents must perform the review themselves.
   - Add language such as: `Do the review yourself; do not spawn or delegate to other sub-agents.`
   - Keep global recursive sub-agent capability intact for roles where coordination is expected.

2. Inject parent-visible local context into child startup context.
   - For fresh children, include enabled user memory as a `<user_memory>` system message.
   - Include current `notes_path` contents as a separate `<agent_notes>` system message for both fresh and forked children.
   - Do not treat `note` content as global user memory; keep it scoped and labeled separately.

3. Avoid duplicate memory in forked children.
   - `fork_context=true` already carries the parent's system prompt, so it should not add a second `<user_memory>` block.

4. Add regression tests.
   - Pin that `SubAgentType::Review.system_prompt()` forbids spawning/delegating sub-agents.
   - Verify fresh sub-agent context includes enabled user memory and notes.
   - Verify forked sub-agent context does not duplicate user memory but still includes notes.

## Acceptance criteria

- A `review` sub-agent prompt clearly requires direct review and forbids recursive sub-agent delegation.
- Fresh `agent_spawn` children inherit enabled user memory.
- Children receive current `note` contents through an explicitly labeled context block.
- Forked children do not receive duplicate user-memory blocks.
- Relevant sub-agent unit tests pass.

Oliver-ZPLiu:
## Summary

Adds a disallowed-tools filter to `SubAgentToolRegistry::tools_for_model()` so that Review-type agents cannot call `agent_spawn`.

## Why it matters

When a review sub-agent is spawned, it can reinterpret itself as a coordinator and spawn further sub-agents instead of performing the review directly. This multiplies work, increases runtime and token use, and the parent receives an incomplete synthesized review because child outputs are not merged.

The existing `allowed_tools()` function already excludes `agent_spawn` for Review agents, but it was deprecated (maintainer's design: default sub-agents inherit full parent registry). The missing piece was a `disallowed_tools` filter that operates after full parent inheritance.

## Changes

| File | Change |
|------|--------|
| `crates/tui/src/tools/subagent/mod.rs` | Added `disallowed_tools` filter; Review agent blocked from `agent_spawn` |

Fixes #1489.

## Test plan

- [x] `cargo clippy -p deepseek-tui --all-targets --all-features --locked -- -D warnings` passes
- [x] `cargo test -p deepseek-tui -- subagent` — 118 passed

### DIFF
--- a/crates/tui/src/tools/subagent/mod.rs
+++ b/crates/tui/src/tools/subagent/mod.rs
@@ -2879,7 +2879,7 @@ async fn run_subagent(
             unavailable_tools.join(", ")
         ));
     }
-    let tools = tool_registry.tools_for_model();
+    let tools = tool_registry.tools_for_model(&agent_type);
     if let Some(mb) = runtime.mailbox.as_ref() {
         let _ = mb.send(MailboxMessage::started(&agent_id, agent_type.clone()));
     }
@@ -3838,14 +3838,27 @@ impl SubAgentToolRegistry {
         }
     }
 
-    fn tools_for_model(&self) -> Vec<Tool> {
+    fn tools_for_model(&self, agent_type: &SubAgentType) -> Vec<Tool> {
+        let disallowed = match agent_type {
+            // Review agents should not spawn sub-agents (#1489).
+            SubAgentType::Review => &["agent_spawn"][..],
+            _ => &[][..],
+        };
         let api_tools = self.registry.to_api_tools();
-        match &self.allowed_tools {
+        let filtered = match &self.allowed_tools {
             None => api_tools,
             Some(list) => api_tools
                 .into_iter()
                 .filter(|tool| list.contains(&tool.name))
-                .collect(),
+                .collect::<Vec<_>>(),
+        };
+        if disallowed.is_empty() {
+            filtered
+        } else {
+            filtered
+                .into_iter()
+                .filter(|tool| !disallowed.contains(&tool.name.as_str()))
+                .collect()
         }
     }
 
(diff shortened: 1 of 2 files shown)


Hmbown:

PR #1505, opened 2026-05-12
fix: count reasoning tokens in cost, auto CNY on zh-Hans

ChaceLyee2101:
Two fixes: (1) Add reasoning_tokens to output cost calculation in pricing.rs. V4 thinking mode reasoning tokens are billed at output rate but were excluded, causing significant under-estimate. (2) Auto-detect CNY currency when ui_locale is zh-Hans and cost_currency is at default usd. Also pin test locale/currency for determinism. All 2782+ tests pass, fmt and clippy clean.


Hmbown:

PR #1508, opened 2026-05-12
feat(composer): display session title in top-right of composer border

mars-base:
## Summary

- Add `session_title: Option<String>` field to `App` struct (`app.rs`)
- Populate it from `metadata.title` when a session is loaded or updated (`ui.rs`)
- Add `derive_session_title()` helper as fallback when no persisted title is available
- Render the title (muted color) alongside the vim mode indicator in a single right-aligned `title_top` span in the composer border (`widgets/mod.rs`)

## Behavior

The composer border's top-right corner now shows the current session's name — the same text visible in `/sessions` — next to the existing vim mode label (when vim mode is enabled). Both share one right-aligned span to avoid overlap with the left-side "Composer" label.

When no session title is available yet (e.g. a brand-new unsaved session), the corner is empty until the first message is sent and the session is persisted.

## Test plan

- [x] Start a new session, send a message → title appears in composer top-right after first save
- [x] Use `/sessions` to switch sessions → title updates immediately
- [x] Vim mode on/off → indicator still renders correctly beside title
- [x] Long title → truncated to 32 chars with `…`


mars-base (commented):
Thanks for the review! Both suggestions have been applied in the follow-up commit:
- Extracted `TURN_META_PREFIX` and `SESSION_TITLE_MAX_CHARS` as module-level constants in `ui.rs`
- Replaced `title.to_string()` with a direct `&str` pass in `widgets/mod.rs`

Hmbown:

mars-base:
Thanks for the feedback and for merging it into work/v0.8.34!

The idea behind displaying the session title in the composer border was inspired by Claude Code CLI, which shows the current session/project name in the UI chrome. In day-to-day use, when switching between different work sessions, having the session name visible at a glance makes it immediately obvious which session you're currently in — reducing the chance of accidentally continuing work in the wrong context.

That said, I completely understand if you feel the composer border isn't the right place for it. I'm curious: would there be a better spot in the TUI layout where session identity could be surfaced? For example, the status bar / footer, the sidebar header, or somewhere else? Happy to rework the placement if there's a preferred location.

Hmbown:

PR #1510, opened 2026-05-12
"fix-session-picker-display-clean-titles-preview"

tyouter:
Fix three display issues in the Ctrl+R session resumption screen:

1. Session titles show clean user prompt instead of <turn_meta>
2. First user message in preview no longer leaks system prefix
3. Empty USER/ASSISTANT entries and thinking blocks are filtered

Implementation:
- extract_user_prompt(): strips <turn_meta> prefix from message text
- extract_title(): wraps extract_user_prompt with fallback for empty titles
- strip_thinking_tags(): removes thinking XML tags from assistant text
- format_session_line + build_preview_lines: apply extraction at display time

Tests: cargo check passes. Verified on Windows with existing session files.
7/7 extract_user_prompt unit tests pass (on development branch).

Developed with AI assistance. Ref: REQ-20260511-001

### DIFF
--- a/crates/tui/src/session_manager.rs
+++ b/crates/tui/src/session_manager.rs
@@ -907,7 +907,11 @@ pub(crate) fn extract_title(raw: &str) -> &str {
 }
 
 /// Strip common thinking/reasoning XML tags from assistant text.
+/// Short-circuits if no thinking markers are present to avoid allocation.
 pub(crate) fn strip_thinking_tags(text: &str) -> String {
+    if !text.contains("<think") && !text.contains("<thinking") && !text.contains("<reasoning") {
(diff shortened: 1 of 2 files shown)


tyouter (commented):
Thanks for the review. Both issues addressed:

1. Performance: Restored take(6) limit matching upstream behavior
2. strip_thinking_tags: Added short-circuit to skip allocation when no thinking markers present

Hmbown:
Thanks @tyouter — your session picker title/cleanup work landed in v0.8.33! Credited under "Session titles and history previews hide metadata noise" in the Fixed section, harvested from PR #1510. Closing as shipped. 🚀

PR #1513, opened 2026-05-12
feat(init): update AGENTS.md in place instead of refusing

Duducoco:
- overwrite existing AGENTS.md on /init rather than blocking with an "already exists" message
- report "Updated" vs "Created" depending on prior file state
- replace noop test with one verifying stale content is regenerated

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/commands/init.rs
+++ b/crates/tui/src/commands/init.rs
@@ -15,25 +15,23 @@ pub fn init(app: &mut App) -> CommandResult {
     // Ensure .deepseek/ is gitignored if we're inside a git repo.
     ensure_deepseek_gitignored(workspace);
 
-    // Check if AGENTS.md already exists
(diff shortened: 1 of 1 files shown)


Hmbown (changes requested):
Thanks @Duducoco. I like the direction — `/init` should be able to refresh project instructions instead of dead-ending on `AGENTS.md already exists` — but this version overwrites any existing `AGENTS.md` wholesale.

That is too risky for a file users hand-edit with project-specific instructions. Please preserve user content in one of these ways before this merges:

- write a timestamped backup next to the file before replacing it, and mention the backup path in the success message; or
- update only a generated block with explicit sentinels, leaving user sections intact; or
- require an explicit force/confirm path for destructive regeneration.

The regression test should also cover the preservation behavior, not just that stale content disappears. Once the data-loss path is closed this should be straightforward to land.

Hmbown:

PR #1517, opened 2026-05-12
feat: integrate DeepSeek prefix-cache stability tracking (inspired by Reasonix)

sternelee:
### Problem

 DeepSeek V4 automatic prefix caching activates only when the exact byte prefix of a request matches the prior request. When the prefix changes — even by one byte — every subsequent token must be
 recomputed on the server, and the cache-hit cost (~10% of miss rate) is lost for the entire prompt.

 Most agent loops rewrite system prompts, reorder tool lists, or inject timestamps each turn. The result: cache hit rates below 20% in practice, meaning users pay near-full price for every turn even
 though much of the prompt is conceptually stable.

 ### Solution

 This PR introduces prefix-cache stability tracking — inspired by [Reasonix's Pillar 1](https://github.com/esengine/DeepSeek-Reasonix) — that detects and surfaces system-prompt/tool-set drift across turns.

 New module: crates/tui/src/prefix_cache.rs (487 lines, 15 unit tests)

 
Component | Role
-- | --
PrefixFingerprint | SHA-256 fingerprint of (system prompt text + sorted tool names). Tool names are sorted deterministically so registration order doesn't affect the hash.
PrefixChange | Core manager — pins the initial fingerprint, checks every turn for drift, auto-re-pins on change, computes a stability_ratio() (fraction of checks that saw no change).
PrefixChange | Diagnostic record — tells you exactly what changed (system prompt, tool set, or both) so telemetry can pinpoint the cause.

 Integration (8 files, +652/−4 total):

 ```
   Engine (turn_loop.rs)
     └─ every turn: check prefix vs pinned fingerprint
         ├─ drift → PrefixCacheChange { changed: true } event
         └─ stable → PrefixCacheChange { changed: false } heartbeat

   Event (events.rs)
     └─ new PrefixCacheChange variant

   Session (session.rs)
     └─ carries Option<PrefixStabilityManager>

   App (app.rs)
     ├─ prefix_change_count / prefix_checks_total
     ├─ prefix_stability_pct
     └─ last_prefix_change_desc

   TUI (ui.rs)
     ├─ event handler updates counters
     ├─ /cache output includes prefix stability stats
     └─ footer P chip: "P 100%" green / "P 83% (1 drift)" yellow / red
 ```

 ### Key Design Decisions

 1. Only tool names in the fingerprint (not full schemas). The DeepSeek provider flattens tool definitions client-side before sending; the serialized form is a function of the model-visible catalog, not
 the raw Tool struct. Tool name changes are the most reliable proxy for cache-invalidation at this layer.
 2. Heartbeat events on stable checks. A no-change event with changed: false fires every turn so the TUI footer can show an always-current stability percentage. This is ~100 bytes per turn on the event
 channel — negligible overhead.
 3. Re-pin on drift, don't error. When the prefix does change (e.g., user switches modes / reloads skills), the manager re-pins to the new fingerprint automatically. Subsequent checks benchmark against
 the new baseline, so one drift event doesn't mean "red forever."

 ### Cache-hit economics

 At DeepSeek V4 Flash pricing:

 
Scenario | Cost per 100K prompt tokens
-- | --
Cache miss (full price) | ~$0.014
Cache hit (~10% of miss) | ~$0.00028

 A session that stays 95%+ prefix-stable keeps costs in the hit column for most of its context. A session where the prefix churns every few turns pays miss rates repeatedly for the same system prompt
 bytes.

 This PR makes the stability visible and measurable. Future work (seam manager activation, auto-warmup-on-drift) can build on this foundation.

 ### Testing

 ```
   cargo test prefix_cache
     → 15 passed (fingerprint determinism, tool-order invariance,
        manager lifecycle, stability ratio, edge cases)

   cargo test core::engine::tests
     → existing prefix-cache tests still pass
        (messages_with_turn_metadata_preserves_stored_messages_for_prefix_cache
         model_tool_catalog_sorts_each_partition_for_prefix_cache_stability)

   cargo check
     → zero errors, zero new warnings
 ```

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -453,6 +453,16 @@ impl Engine {
         session.last_system_prompt_hash = Some(system_prompt_hash(stable_prompt.as_ref()));
         session.system_prompt = stable_prompt;
 
+        // Initialize prefix-cache stability monitor.
+        // Pins the initial fingerprint (system prompt + tool spec names)
+        // so subsequent turns can detect cache-invalidating drift.
+        let _ = session.prefix_stability.get_or_insert_with(|| {
+            // Use the tool registry's spec names for fingerprinting.
+            // At this point tool spec builders may not be registered yet,
+            // so we start with None — fingerprint will pin on first request.
+            crate::prefix_cache::PrefixStabilityManager::new_unpinned()
+        });
+
         let subagent_manager =
             new_shared_subagent_manager(config.workspace.clone(), config.max_subagents);
         let shell_manager = config
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -453,9 +453,11 @@ impl Engine {
         session.last_system_prompt_hash = Some(system_prompt_hash(stable_prompt.as_ref()));
         session.system_prompt = stable_prompt;
 
-        // Initialize prefix-cache stability monitor.
-        // Pins the initial fingerprint (system prompt + tool spec names)
-        // so subsequent turns can detect cache-invalidating drift.
+        // Initialize prefix-cache stability monitor (lazy-pin).
+        // The system prompt is available now but the tool catalog isn't
+        // fully built until the first turn, so we start unpinned. The
+        // first `check_and_update` call in the turn loop will pin the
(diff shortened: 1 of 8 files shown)


Hmbown:

Issue #1450, opened 2026-05-11
Tool results bloat context: read_file lacks range reading, no per-tool truncation policy, and stale results persist


Oliver-ZPLiu:
## Problem

DeepSeek-TUI's token consumption per turn is higher than necessary, for three independent reasons:

### 1. `read_file` returns full files unconditionally

`read_file` currently takes only `path` (and PDF `pages`). It always reads and returns the entire file. A 500-line source file (~15KB) goes fully into context even when the model only needs to inspect a few functions.

**Impact:** Each large `read_file` adds ~5-20KB of content that stays in context for multiple turns.

**Reference:** All major coding agents now support line-range reads:
- Roo Code: `read_file(path, offset, limit)` with auto-truncation
- Cursor: dynamic context discovery with file references
- Claude Code: range-based reads with line numbers

### 2. No per-tool truncation policy

All tool outputs share the same single threshold (`SPILLOVER_THRESHOLD_BYTES = 100KB`). Different tools have very different output profiles:

| Tool | Typical output | Ideal strategy |
|------|---------------|----------------|
| `read_file` | 5-50KB source text | ~200 lines / 16KB cap + continuation |
| `grep/search` | Snippets + paths | Top-k + total match count |
| `exec_shell` / tests | Logs, build output | Head+tail + saved reference |
| `ls/list_dir` | Short listings | Full (already small) |

A single global threshold cannot serve all these cases well.

### 3. Stale tool results persist until compaction

Tool results from many turns ago stay fully intact in the conversation window. Compaction only triggers when total tokens exceed a threshold. In the meantime, every API request re-sends old results that the model is unlikely to reference.

## Related

- #1440  

## Suggested improvements

- P0: `read_file` gains `start_line`/`max_lines` parameters with default bounded output
- P1: per-tool visible budget for different output types
- P2: proactive clearing of old results that are re-readable
- P3: head+tail truncation for shell/test outputs

PR #1521, opened 2026-05-12
Improve file-edit guardrails and shell output truncation

### LINKED ISSUE #1516: Clarify edit_file boundaries for structural edits
# edit_file boundary limitation: structural multi-line edits can corrupt files

## Observed behavior

An agent used `edit_file` to perform an intertwined multi-line structural deletion in an existing file:
removing timing logic, a categorization loop, and metrics-related code in one edit. The resulting file was
corrupted: residual fragments remained, a `for` head was duplicated, and braces were misplaced. The failure
was only discovered after `cargo build` failed. The agent then recovered by rewriting the whole file with
`write_file`.

Agent self-assessment:

> I was too greedy using `edit_file` for surgery instead of rewriting the file. This kind of multi-line
> structural deletion is a gamble with `edit_file`: one incomplete match or replacement can leave the file
> broken. I should have used `write_file` to rewrite the file.

### Current edit_file core logic (`crates/tui/src/tools/file.rs`)

```rust
let count = contents.matches(search).count();        // plain string match count
let updated = contents.replace(search, replace);     // full-file str::replace
fs::write(&file_path, &updated);
let diff = make_unified_diff(&display, &contents, &updated);
let summary = format!("Replaced {count} occurrence(s) in {display}");
```

Two facts should be kept separate:

- `str::replace` does not partially match a multi-line block. It only replaces exact occurrences of the
  provided `search` string.
- Corruption happens when the caller supplies a structurally incomplete `search` or `replace`: `search` may
  cover only part of the intended structure, or `replace` may not be syntactically complete. Once `search`
  matches at least once, the tool cannot infer whether the replacement preserves AST, block-boundary, or
  brace-pairing invariants.

### Key limitations

| Dimension | Behavior | Risk |
|-----------|----------|------|
| Matching | `str::matches` + `str::replace`, exact plain text | The tool does not understand AST boundaries or whether `search` / `replace` covers a complete syntax structure. |
| Result preview | `make_unified_diff` with `context_radius(3)` | Each changed hunk has only +/-3 lines of context, not a full-file view. |
| Multiple matches | Reports `Replaced N occurrence(s)` | The same replacement is applied to every match, including unintended matches. |
| Structural validation | None; execution proceeds if `search` matches at least once | Residual fragments, duplicated blocks, and misplaced braces are not prevented at the tool layer. |
| Diagnostics feedback | Successful writes may append LSP diagnostics when enabled and available | This is an after-the-fact signal, not a pre-write structural-boundary check. |

### Failure chain

```text
edit_file(search=too narrow or cross-line structural fragment, replace=syntactically incomplete fragment)
  -> str::replace matches exactly and writes a structurally incomplete file
  -> tool returns a +/-3 line diff, possibly with LSP diagnostics depending on configuration and language support
  -> agent does not verify with read_file / build / diagnostics
  -> cargo build failure reveals the global structural corruption
```

So this is not a fuzzy-match problem in `str::replace`. It is a tool-contract limitation: `edit_file` can
guarantee byte-level text replacement, but not successful structural editing.

### 1. The agent violated existing prompt guidance

The system prompt in `crates/tui/src/prompts/base.md` currently says:

```text
Use `edit_file` for a single unambiguous replacement.
Use `apply_patch` when the edit changes whole blocks, touches multiple files,
  or needs surrounding line context.
Use `write_file` for brand-new files or full-file rewrites.
```

The agent's edit touched several intertwined logical units and was not a single unambiguous replacement, but
the agent still chose `edit_file`.

### 2. The tool description is silent about the boundary

`edit_file`'s `description()` and `input_schema()` currently describe only the basic function:

```text
Replace text in a file using search/replace
```

They do not state that:

- this is plain string replacement, not an AST-aware edit;
- `search` must match exactly, including whitespace, indentation, and newlines;
- the tool cannot validate whether `search` / `replace` covers a complete syntax structure;
- the result is a +/-3 line diff, not a full-file view, with LSP diagnostics only as an optional supplemental signal;
- each call should be limited to a single unambiguous replacement, especially not a deletion or refactor spanning multiple logical units.

### 3. Prompt coverage gap for large existing-file rewrites

| Scenario | Current prompt recommendation | Better tool choice |
|----------|-------------------------------|--------------------|
| Single-line change | `edit_file` | `edit_file` |
| Single code-block change | `apply_patch` | `apply_patch` |
| Intertwined deletion/refactor across multiple logical units | `apply_patch` is implied | `apply_patch` or `write_file`, depending on whether complete context can be expressed reliably |
| New file | `write_file` | `write_file` |
| Large-scope modification to an existing file | Not explicitly covered | `write_file` when the model is already reconstructing the whole file |

More precisely: `apply_patch` is appropriate when a structural edit can be expressed clearly with context.
When changes span multiple intertwined logical units, involve large deletions/reordering, or require the model
to mentally reconstruct the whole file, `write_file` is more reliable.

### 1. Expand prompt guidance for `write_file` on existing files

`crates/tui/src/prompts/base.md`, Tool Selection Guide.

Current:

```text
Use `apply_patch` for structural edits, coordinated changes, or cases where
line context matters. Use `write_file` for brand-new files or full-file rewrites.
Use `edit_file` for a single unambiguous replacement.
```

Suggested:

```text
Use `apply_patch` for structural edits, coordinated changes, or cases where
line context matters. Use `write_file` for brand-new files, full-file rewrites,
or large-scope modifications to existing files where multiple intertwined changes
make edit_file unreliable. Use `edit_file` for a single unambiguous replacement.
```

Also add negative guidance in the `edit_file` section:

```text
Use `edit_file` for one clear replacement in one file. Do NOT use edit_file
for multi-block structural deletions, cross-cutting refactors, or changes
that touch more than one logical unit -- those require apply_patch or write_file.
```

Impact: the model gets more precise tool-selection guidance before choosing a write tool.

### 2. Add a verification hint when `edit_file` replaces multiple matches

`crates/tui/src/tools/file.rs`, `EditFileTool::execute()`.

Current:

```rust
let summary = format!("Replaced {count} occurrence(s) in {display}");
```

Suggested:

```rust
let summary = if count > 1 {
    format!(
        "Replaced {count} occurrence(s) in {display}\n\
         Warning: multiple matches were replaced with the same substitution. \
         Verify the result with read_file before proceeding."
    )
} else {
    format!("Replaced 1 occurrence in {display}")
};
```

Impact: when `count > 1`, the agent immediately sees that the same replacement was applied repeatedly and is
nudged to inspect the final file. `count == 1` behavior remains quiet.

Do not turn `count > 1` into a hard error. Batch renames, repeated-token changes, and uniform text updates can
all be legitimate multi-match edits. This should be an advisory warning, not a compatibility break.

### 3. Strengthen `edit_file` description and schema

`crates/tui/src/tools/file.rs`, `EditFileTool::description()` and `input_schema()`.

Current description:

```text
Replace text in a file using search/replace. Required: 'path' (file to edit),
'search' (exact text to find), 'replace' (text to substitute).
```

Suggested description:

```text
Replace text in a file using search/replace. Required: 'path' (file to edit),
'search' (exact text to find -- whitespace and indentation must match precisely),
'replace' (text to substitute). For a single unambiguous change only -- returns a
+/-3-line unified diff, not the full file. For multi-block or structural edits,
use apply_patch or write_file instead.
```

Current `search` parameter description:

```text
Text to search for
```

Suggested:

```text
Text to search for -- must match exactly, including whitespace and indentation
```

Impact: the model sees the tool boundary directly in the function-calling schema at decision time.

## Non-goal: reject every multi-line search or structural-looking edit

Do not simply reject all multi-line `search` strings. Some multi-line replacements are safe and efficient:
replacing a unique config block, updating a fixed template fragment, or deleting a small complete function with
clear context. A hard rejection would create false positives and push the agent toward heavier `write_file`
usage.

A better strategy is explicit tool-contract guidance plus verification hints for risky signals:

- Multiple matches: warn that the same replacement affected multiple locations and recommend `read_file`.
- Structural or multi-block edits: guide the model toward `apply_patch` or `write_file`.
- Large-scope existing-file changes: guide the model toward `write_file` because full content is easier to audit
  than several fragile local replacements.

## Caveat: `write_file` token cost

`write_file` is not the root cause of this issue; it was the more reliable recovery path. However, using
`write_file` on an existing large file requires the agent to provide the full `content`, which can significantly
increase input/output tokens and amplify risks from stale context or truncated content.

In the incident, the agent could not confirm whether switching to `write_file` had any token-level guarantee or
automatic completion mechanism. The only safe inference from the successful rewrite is that the agent had enough
complete file content at that moment, possibly from a prior `read_file` result or old content still present in
context. This does not mean the system automatically injects `read_file`, and it does not mean `write_file` has
an inherent token-safety guarantee. If full file content is missing or stale, `write_file` can still write an
incomplete or outdated file.

Therefore, `write_file` should be a reliability option for large-scope existing-file changes, not an unconditional
replacement for `apply_patch`. Small, well-contextualized structural edits should still prefer `apply_patch`;
whole-file rewrites are appropriate when several intertwined edits cannot be expressed reliably as a patch.

## Priority

| Fix | Scope | Risk | Independence |
|-----|-------|------|--------------|
| Strengthen tool description/schema | Small Rust change | Low | Independent |
| Add multi-match verification hint | Small Rust change | Low | Independent |
| Expand prompt guidance | Small prompt change | Low | Independent |

All three changes are small, low-risk, and independently testable. They should be implemented together for the
combined effect.

## Acceptance criteria

- `edit_file` description/schema states exact string matching, whitespace sensitivity, and the single
  unambiguous replacement boundary.
- `edit_file` results include a verification hint when multiple matches are replaced, while still succeeding.
- Single-match `edit_file` results do not include the multi-match verification hint.
- `base.md` Tool Selection Guide covers large-scope intertwined modifications to existing files and recommends
  `write_file` where appropriate.
- Existing legitimate multi-match `edit_file` behavior remains compatible and is not turned into an error.

Suggested verification:

```bash
cargo fmt --all --check
cargo test -p deepseek-tui --bin deepseek-tui tools::file::tests::test_edit_file_tool
```

axobase001:
## Summary
- tighten `edit_file` tool guidance around exact matching, structural-edit boundaries, and single unambiguous replacements
- add a multi-match `edit_file` warning while preserving valid repeated replacements
- switch large shell/test output truncation to keep a bounded head plus raw tail, with high-signal lines from the omitted middle
- keep `pandoc_convert` input-only validation ahead of pandoc binary lookup, so missing `output_path` errors are deterministic on CI hosts without pandoc
- make the vision absolute-path rejection test platform-aware for Windows CI

## Notes
- Current `main` already includes the `read_file(start_line, max_lines)` bounded-read path for #1450, so this PR focuses on the remaining shell/test output budget failure mode from that issue.
- Closes #1516
- Addresses #1450

## Validation
- `cargo fmt --all --check`
- `cargo test -p deepseek-tui tools::file::tests:: -- --nocapture`
- `cargo test -p deepseek-tui tools::shell_output::tests:: -- --nocapture`
- `cargo test -p deepseek-tui tools::pandoc::tests::pandoc_convert_rejects_inline_request_for_binary_format -- --nocapture`
- `cargo test -p deepseek-tui vision::tools::tests::execute_rejects_absolute_path -- --nocapture`
- `cargo check -p deepseek-tui`

### DIFF
--- a/crates/tui/src/prompts/base.md
+++ b/crates/tui/src/prompts/base.md
@@ -170,10 +170,10 @@ Multiple `tool_calls` in one turn run in parallel. `web_search` returns `ref_id`
 ## Tool Selection Guide
 
 ### `apply_patch`
-Use `apply_patch` for structural edits, coordinated changes, or cases where line context matters. Use `write_file` for brand-new files or full-file rewrites. Use `edit_file` for a single unambiguous replacement.
+Use `apply_patch` for structural edits, coordinated changes, or cases where line context matters. Use `write_file` for brand-new files, full-file rewrites, or large-scope modifications to existing files where multiple intertwined changes make local replacement unreliable. Use `edit_file` for a single unambiguous replacement.
 
 ### `edit_file`
-Use `edit_file` for one clear replacement in one file. Use `apply_patch` when the edit changes whole blocks, touches multiple files, or needs surrounding line context.
+Use `edit_file` for one clear replacement in one file. Do not use `edit_file` for multi-block structural deletions, cross-cutting refactors, or changes that touch more than one logical unit; those require `apply_patch` or `write_file`.
 
 ### `exec_shell`
 Use `exec_shell` for shell-native diagnostics, pipelines, and bounded commands. Use structured tools for structured operations when they map directly (`grep_files`, `git_diff`, `read_file`). For long commands, servers, full test suites, or release computations, start background work with `task_shell_start` or `exec_shell` using `background: true`, then poll with `task_shell_wait` or `exec_shell_wait`.
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -473,7 +473,7 @@ impl ToolSpec for EditFileTool {
     }
 
     fn description(&self) -> &'static str {
-        "Replace text in a single file via exact search/replace. Use this instead of `sed -i` in `exec_shell` for in-place edits. For multi-hunk or cross-file changes, use `apply_patch` instead. Required: 'path', 'search' (exact text to find), 'replace' (text to substitute)."
+        "Replace text in a single file via exact search/replace. Required: 'path', 'search' (exact text to find; whitespace, indentation, and newlines must match precisely), 'replace' (text to substitute). Use this for a single unambiguous replacement only; it returns a unified diff with three lines of context, not the full file. For multi-block structural edits, cross-cutting refactors, or large-scope existing-file changes, use apply_patch or write_file instead."
     }
 
     fn input_schema(&self) -> Value {
@@ -486,11 +486,11 @@ impl ToolSpec for EditFileTool {
                 },
                 "search": {
                     "type": "string",
-                    "description": "Text to search for"
+                    "description": "Text to search for; must match exactly, including whitespace, indentation, and newlines"
                 },
                 "replace": {
                     "type": "string",
-                    "description": "Text to replace with"
+                    "description": "Replacement text; keep structural replacements syntactically complete"
                 }
             },
             "required": ["path", "search", "replace"]
@@ -542,7 +542,15 @@ impl ToolSpec for EditFileTool {
 
         let display = file_path.display().to_string();
         let diff = make_unified_diff(&display, &contents, &updated);
-        let summary = format!("Replaced {count} occurrence(s) in {display}");
+        let summary = if count > 1 {
+            format!(
+                "Replaced {count} occurrence(s) in {display}\n\
+                 Warning: multiple matches were replaced with the same substitution. \
+                 Verify the result with read_file before proceeding."
+            )
+        } else {
+            format!("Replaced 1 occurrence in {display}")
+        };
         let body = if diff.is_empty() {
             format!("{summary}\n(no textual changes)")
         } else {
@@ -1151,6 +1159,11 @@ mod tests {
 
         assert!(result.success);
         assert!(result.content.contains("2 occurrence(s)"));
+        assert!(
+            result.content.contains("Warning: multiple matches"),
+            "multi-match edits should nudge the agent to inspect the result: {}",
+            result.content
+        );
         // Inline diff (#505) — the unified diff lands above the summary
         // line so the TUI's diff-aware renderer kicks in.
         assert!(result.content.contains("--- a/"), "{}", result.content);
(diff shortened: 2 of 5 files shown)


Hmbown:

PR #657, opened 2026-05-04
feat(config): separate tui.toml for theme and keybinds (closes #437)

merchloubna70-dot:
## Summary

Implements issue #437 — decouple TUI-specific preferences from the agent/project `config.toml` so they survive project switches.

- Adds `TuiPrefs` struct and `KeybindPrefs` sub-struct to `crates/tui/src/settings.rs`
- Loads from `~/.deepseek/tui.toml` via `TuiPrefs::load()`; falls back to struct defaults when absent (backwards-compatible — no behaviour change for existing users)
- Supports `theme` (`dark` | `light` | `system`), `font_size` (u16, `0` = terminal default), and a `[keybinds]` table with optional overrides for `submit`, `new_line`, `command_palette`, `cancel`, `toggle_sidebar`
- `TuiPrefs::save()` persists preferences back to disk, creating `~/.deepseek/` if needed
- `TuiPrefs::validate()` normalises theme casing and rejects unknown values with a helpful error message
- Respects the `DEEPSEEK_CONFIG_PATH` env-var redirect used by tests and CI (same pattern as `Settings::path()`)

## Example `~/.deepseek/tui.toml`

```toml
theme     = "dark"       # dark | light | system
font_size = 14           # 0 = terminal default

[keybinds]
submit          = "ctrl+enter"
new_line        = "enter"
command_palette = "ctrl+k"
cancel          = "ctrl+c"
toggle_sidebar  = "ctrl+b"
```

## Test plan

- [x] 8 new unit tests added in `settings::tests`:
  - `tui_prefs_defaults_are_dark_theme_zero_font`
  - `tui_prefs_validate_accepts_known_themes`
  - `tui_prefs_validate_normalises_theme_case`
  - `tui_prefs_validate_rejects_unknown_theme`
  - `tui_prefs_round_trips_through_toml`
  - `tui_prefs_load_returns_defaults_when_file_absent`
  - `tui_prefs_save_and_load_round_trip`
  - `tui_prefs_path_uses_home_deepseek_subdir_by_default`
- [x] All 16 `settings::tests` pass (`cargo test -p deepseek-tui "settings::tests"`)
- [x] `cargo +nightly check` clean — 0 errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1534, opened 2026-05-12
Add Catppuccin/Tokyo Night/Dracula/Gruvbox themes + /theme picker

laoye2020:
Wires the previously-dormant `theme` setting (#657 follow-up) into the live Settings struct so the choice survives restart. `/theme` opens an interactive picker with live preview; `/theme <name>` switches and persists non-interactively.

Most render sites use bare `palette::TEXT_BODY` / `DEEPSEEK_INK` / `BORDER_COLOR` constants rather than reading `app.ui_theme`, so simply adding new UiTheme variants only repaints ~20% of surfaces. The fix is a third stage in ColorCompatBackend (alongside the existing dark<->light and truecolor<->256 stages) that rewrites every well-known dark-palette constant to the corresponding UiTheme slot for the active preset. The remap is a no-op for System / Whale / WhaleLight, so legacy dark/light flows stay byte-identical.

Settings: theme = system | dark | light | catppuccin-mocha | tokyo-night | dracula | gruvbox-dark. Unknown values normalise to system. background_color still overlays on top.

Tests: new coverage in theme_picker and palette; pinned make_app() in footer tests to ThemeId::System after App::new (matching the existing default_model pin) since App::new now honors settings.theme.

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/commands/config.rs
+++ b/crates/tui/src/commands/config.rs
@@ -419,7 +419,21 @@ pub fn set_config_value(app: &mut App, key: &str, value: &str, persist: bool) ->
             app.needs_redraw = true;
         }
         "background_color" | "background" | "bg" => {
-            let base_theme = crate::palette::UiTheme::detect();
+            let base_theme = crate::palette::ThemeId::from_name(&settings.theme)
+                .unwrap_or(crate::palette::ThemeId::System)
+                .ui_theme();
+            app.ui_theme = settings
+                .background_color
+                .as_deref()
+                .and_then(crate::palette::parse_hex_rgb_color)
+                .map_or(base_theme, |color| base_theme.with_background_color(color));
+            app.needs_redraw = true;
+        }
+        "theme" => {
+            let theme_id = crate::palette::ThemeId::from_name(&settings.theme)
+                .unwrap_or(crate::palette::ThemeId::System);
+            let base_theme = theme_id.ui_theme();
+            app.theme_id = theme_id;
             app.ui_theme = settings
                 .background_color
                 .as_deref()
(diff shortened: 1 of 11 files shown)


laoye2020:
Thanks for the reviews — addressed in 28605b8a:

  - **Gemini #1 modal sizing**: dropped the `.max(52)`/`.max(10)` floors that could exceed the available area; render now early-returns on a zero-sized rect. Added
  `render_does_not_panic_on_tiny_area` and `..._on_zero_sized_area` tests.
  - **Gemini #4 `'0'` digit shortcut**: changed the guard to `matches!(c, '1'..='9')`. Added `digit_zero_is_rejected_not_remapped_to_row_zero`.
  - **Gemini #5 System env-read per frame**: cached `system_ui_theme: UiTheme` at picker construction; added `ui_theme_for(id)` helper used by both the chrome and the per-row swatch loop.
  - **Gemini #6 redundant `Settings::load()`**: `AppAction::OpenThemePicker` now reads `app.theme_id.name()` directly instead of re-parsing `settings.toml`.
  - **Gemini #3 double `ThemeId::from_name`**: collapsed `Settings::load`'s normalization to a single `.map().unwrap_or_else()`.
  - **Codex transcript spam on preview**: `ConfigUpdated` handler only appends the confirmation `System` cell when `persist == true`. Live-preview navigation no longer clutters history.
  - **Codex `background_color` override loss**: `ColorCompatBackend` now carries the active `UiTheme` (with `.with_background_color()` overlay applied) alongside `ThemeId`, and
  `adapt_{fg,bg}_for_theme` read target slots from that. Added `community_theme_remap_honors_background_color_override` test pinning the contract.

  Gemini's `if/else if` → `match` suggestion (#2) I left alone — for `Color` equality the chain is just as efficient and a `match` against const items reads about the same length-wise, so I
  didn't want to bloat the diff. Happy to do it if you'd prefer.

Hmbown:

PR #1548, opened 2026-05-13
Fix changelog command version hints

zhuangbiaowei:
## Summary
- make `/change [version]` show release notes for explicit versions as well as latest
- skip empty changelog sections when resolving latest and previous-version hints
- include the previous-version hint in localized display and translation prompts
- keep `/change` directly executable from the command palette and slash menu without requiring a version

## Verification
- `cargo fmt --all`
- `cargo test -p deepseek-tui change_with_explicit_version_includes_previous_hint`
- `cargo test -p deepseek-tui change_in_non_english_also_sends_translation_action`
- `cargo test -p deepseek-tui change` (all `/change` tests passed; this broad filter also hits unrelated `project_context::tests::test_global_agents_only_no_project_unchanged_fallback`, which failed independently)

### DIFF
--- a/crates/tui/src/commands/change.rs
+++ b/crates/tui/src/commands/change.rs
@@ -1,10 +1,11 @@
-//! `/change` command — show the latest changelog entry, translated to the
-//! user's locale when it is not English.
+//! `/change` command — show a changelog entry, translated to the user's
+//! locale when it is not English.
 //!
-//! Usage: `/change`
+//! Usage: `/change [version]`
 //!
-//! Uses the DeepSeek-TUI changelog embedded at compile time, extracts the
-//! most recent version section, and displays it. When the UI locale is not
+//! Uses the DeepSeek-TUI changelog embedded at compile time. With no argument,
+//! extracts the most recent section. With a version argument like `0.8.32`,
+//! extracts that specific version's section. When the UI locale is not
 //! English and the current session can reach a model, the command also fires a
 //! `SendMessage` action that asks the model to translate the changelog into
 //! the user's language.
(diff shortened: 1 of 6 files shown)


zhuangbiaowei:
Addressed the Gemini review feedback in `cf647730`:

- replaced hardcoded changelog heading slicing with a shared `strip_prefix("## [")` / `split_once(']')` parser
- preprocess `/change` version args once before section lookup and error reporting
- compute the previous-version hint once and use the English hint text as the translation prompt source
- moved direct-run / menu-spacing behavior onto `CommandInfo` instead of special-casing `/change` in slash menu code
- added regression coverage for malformed changelog headings

Verification:
- `cargo fmt --all`
- `git diff --check`
- `cargo test -p deepseek-tui change_with_explicit_version_includes_previous_hint`
- `cargo test -p deepseek-tui malformed_version_headings_do_not_panic`
- `cargo test -p deepseek-tui apply_slash_menu_selection_keeps_change_executable_without_version`
- `cargo test -p deepseek-tui change_in_non_english_also_sends_translation_action`
- `cargo test -p deepseek-tui command_palette_runs_change_without_requiring_version`

I also re-ran `cargo test -p deepseek-tui change`; all `/change` tests passed, and the only failure was the unrelated existing broad-filter hit in `project_context::tests::test_global_agents_only_no_project_unchanged_fallback`.

zhuangbiaowei:
@Hmbown Hello, I’ve noticed that the bug this PR aims to fix has actually already been resolved in version 0.8.35, so should this PR be closed?

Hmbown:
Thanks for the follow-up. I verified current main already has the explicit-version previous-version hint coverage for /change, including the regression test, so this PR is now obsolete. Closing it rather than asking you to keep rebasing a fixed issue.

Issue #1359, opened 2026-05-10
Using DeepSeek-TUI in VSCode terminal will not be able to wrap.


tony6960:
## Steps to reproduce

1.open VSCode
2.input deepseek and run
3.input anything than press shift+enter

## Expected behavior
change line

## Actual behavior
send the message to deepseek

## Environment

- OS:windows11
- DeepSeek CLI version:v0.8.22
- Model:deepseek-v4-pro
- Shell:powershell in VSCode Terminal

## Logs or screenshots
None

Hmbown:
thank you! will investigate

jt-at-home:
FYI... I ran this test in  VSCode on Linux Mint.  I got the expected behavior.  If that helps isolate anything.

<img width="1346" height="814" alt="Image" src="https://github.com/user-attachments/assets/889e4b3d-834f-4565-81a0-dee91fdec303" />

PR #1567, opened 2026-05-13
fix(tui): gate Windows Kitty keyboard push on host-advertised support (#1559)

### LINKED ISSUE #1559: /help无法退出，所有非up/down/PgUp/PgDn按键输入均被识别为筛选输入
## Description

/help无法退出，所有非up/down/PgUp/PgDn按键输入均被识别为筛选输入

## Steps to reproduce

1.输入/help进入帮助菜单

## Expected behavior
按esc或ctrl+C能正常退出
## Actual behavior
按esc无法退出，任何其他输入均被识别为“筛选”
## Impact

使得无法在不关闭终端的前提下继续使用tui

## Environment

- OS:windows
- DeepSeek CLI version:0.8.32 (ad70655bbb04)
- Install method:npm
- Model/provider:deepseek
- Terminal app:powershell
- Shell:

Image

wuli666:
## Summary

Closes #1559. On Windows 11 + standalone PowerShell, opening `/help` (or any modal that accepts a filter) leaves the user unable to escape — Esc inserts the literal text `[27u` into the filter instead of closing the modal, and there is no keyboard path out of the TUI short of killing the terminal.

## Root cause

`c8614a24` ("fix(tui): enable Kitty keyboard protocol on Windows so Shift+Enter inserts newline", #1359) bypasses crossterm's capability gate on Windows and writes `\x1b[>1u` directly so VSCode + Windows Terminal can disambiguate `Shift+Enter` from `Enter`. The commit message claims "terminals that don't [support it] silently discard the unknown escapes." That holds for VSCode's xterm.js and Windows Terminal ≥1.17, but **not** for legacy conhost / standalone PowerShell:

- The host partially honours the protocol — after the push, Esc is encoded as the disambiguated `\x1b[27u` (5 bytes) instead of bare `\x1b` (1 byte).
- crossterm's Windows console-API reader (`ReadConsoleInputW`-driven) does not decode that sequence back to `KeyCode::Esc`.
- The result: the leading `\x1b` is consumed somewhere upstream and the trailing `[27u` is read as four printable keystrokes (`[`, `2`, `7`, `u`), each routed to whichever input has focus.

The OP screenshot pins this exactly — the help overlay's filter shows `[27u[27u[27u[27u` after four Esc presses:

> issue 1559 repro

Because every modal in the TUI (help overlay, model picker, mode picker, file picker, command palette, …) routes printable characters into either a filter buffer or its keymap, the user is locked in.

## Fix

Gate the Windows Kitty push on a positive signal from the host, so the protocol is only enabled where it round-trips cleanly. Mirrors the existing `default_mouse_capture_enabled` pattern from #1169.

Supported (push fires, #1359 stays fixed):

- `WT_SESSION` is set → Windows Terminal (≥1.17 disambiguates Esc and Shift+Enter cleanly).
- `TERM_PROGRAM=vscode` → VSCode integrated terminal (xterm.js, the original #1359 target).

Skipped (push is silent, legacy console-API key path keeps Esc working):

- Standalone PowerShell / conhost — no advertised Kitty support.
- Anything else without a positive signal.

Non-Windows behaviour is unchanged — the existing crossterm-driven path on Linux and macOS is untouched.

The pop is gated symmetrically: we only emit `\x1b[<1u` when we actually pushed, so a host that never entered Kitty mode doesn't see a stray pop escape.

## Why this fix over alternatives

- **Detect Kitty support at runtime via terminal query** — would be more general but requires a CSI query/response round-trip during startup, and crossterm's Windows reader can't reliably parse the response (the same parser gap that causes the original bug). Adds latency for marginal gain.
- **Patch crossterm's Windows reader to decode `\x1b[27u`** — correct long-term fix but lives in a third-party crate and would land on the user's machine via a `crossterm` minor bump, which is out of scope for a TUI patch release.
- **Add a `tui.kitty_keyboard = false` config knob** — feasible, but the breakage is silent (user opens `/help` and is stuck) and most affected users won't know they need to flip a knob. A positive-signal env gate fixes the default for everyone who hits the bug.
- **Strip the Kitty push entirely on Windows** — reverts the #1359 fix; Shift+Enter / Alt+Enter / Ctrl+J would silently lose their composer-newline behaviour on the VSCode and Windows Terminal hosts where the protocol does work.

The env gate is the smallest change that keeps #1359 working where the protocol is round-trippable and stops breaking Esc where it isn't.

## Verification

- Reproduced from the OP screenshot: with the push unconditionally written, Esc in `/help` produces `[27u` in the filter. With the gate, the push is skipped on conhost and Esc closes the overlay via the existing legacy console-API path.
- `cargo fmt --check` — clean.
- `cargo clippy -p deepseek-tui --tests -- -D warnings` — clean.
- `cargo test -p deepseek-tui --bin deepseek-tui` — 2922 passed, 0 failed (3 ignored).
- Windows-specific helpers extracted and run on Linux to validate syntax/behaviour of both branches; the in-tree `#[cfg(windows)]` tests will run on Windows CI:
  - `push_keyboard_flags_writes_kitty_push_sequence_on_supported_host` (#1359 regression)
  - `push_keyboard_flags_skipped_on_legacy_windows_host` (#1559 fix)
  - `pop_keyboard_flags_writes_kitty_pop_sequence_on_supported_host`
  - `pop_keyboard_flags_skipped_on_legacy_windows_host`
  - `windows_kitty_keyboard_supported_detects_known_hosts`
  - `windows_kitty_keyboard_unsupported_for_legacy_hosts`

## Test plan

- [ ] Build deepseek-tui on Windows.
- [ ] In Windows Terminal (`WT_SESSION` set): confirm Shift+Enter still inserts a newline in the composer (#1359 regression) **and** `/help` → Esc closes the overlay.
- [ ] In standalone PowerShell / conhost (`WT_SESSION` and `TERM_PROGRAM` both unset): confirm `/help` → Esc closes the overlay. Shift+Enter falls back to legacy console behaviour (acceptable trade — the host doesn't actually distinguish it).
- [ ] In VSCode integrated terminal on Windows (`TERM_PROGRAM=vscode`): confirm both Shift+Enter and `/help` Esc work.

## Scope

- Two files touched: `crates/tui/src/tui/ui.rs`, `crates/tui/src/tui/ui/tests.rs`.
- +170 / -33, ~120 LoC excluding doc comments and tests.
- No trust-boundary surface (auth/sandbox/publishing/branding) touched.

### DIFF
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -7180,21 +7180,12 @@ fn reset_terminal_viewport(terminal: &mut AppTerminal, sync_output_enabled: bool
 }
 
 fn push_keyboard_enhancement_flags<W: Write>(writer: &mut W) {
-    // crossterm's PushKeyboardEnhancementFlags command unconditionally
-    // returns Unsupported on Windows (is_ansi_code_supported() == false), so
-    // the ANSI escape is written directly on that platform. Modern Windows
-    // terminals (VSCode integrated terminal, Windows Terminal ≥1.17) honour
-    // the kitty keyboard protocol; terminals that do not silently discard it.
     #[cfg(windows)]
     {
-        let flags = KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES.bits();
-        if let Err(err) = write!(writer, "\x1b[>{}u", flags).and_then(|()| writer.flush()) {
-            tracing::debug!(
-                target: "kitty_keyboard",
-                ?err,
-                "PushKeyboardEnhancementFlags direct write failed on Windows"
-            );
-        }
+        push_keyboard_enhancement_flags_windows_with(
+            writer,
+            windows_kitty_keyboard_supported_from_env(),
+        );
         return;
     }
     #[cfg(not(windows))]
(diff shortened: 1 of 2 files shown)


Hmbown:
Closing this as superseded for v0.8.38. The #1559/#1599 safety fix already landed on main in the lower-risk form that avoids enabling CSI-u keyboard flags on Windows, because crossterm's Windows input path does not decode the resulting CSI-u Enter/Esc sequences reliably. This PR was based on the older direct-enable behavior and is now dirty against main; taking it as-is would risk reopening the Windows Terminal Enter/[57414u] class of bugs.\n\nThanks @wuli666 for the root-cause analysis here. If we restore full Windows Shift+Enter disambiguation later, it should be done with either a crossterm-compatible Windows input decoder or a capability gate that is verified against current Windows Terminal, VS Code, and legacy ConHost behavior.

PR #1580, opened 2026-05-13
fix: treat drag-and-dropped absolute file paths as @-mentions instead of Unknown command

jieshu666:
Problem

When a user drags a file from Finder on macOS, the terminal inserts the absolute path (e.g. /Users/name/file.txt). The leading / caused the command dispatcher to treat it as a slash command,
resulting in:

    Error: Unknown command: /users/machunhong1/documents/agent. Type /help for available commands.

Root Cause

In commands::execute, all input starting with / is unconditionally routed to the slash-command match arms. Absolute file paths (/Users/...) start with / but contain path separators — unlike
real slash commands (/help, /model, /clear), which are single words.

Solution

In the unknown-command fallthrough arm (_), after skill matching fails and before returning "Unknown command", check whether the input looks like a filesystem path:

- The original input starts with /
- The command name (after stripping the leading /) still contains a / separator

If both conditions are met, convert the input to an @-mention and send it as a message via AppAction::SendMessage. The existing file-mention system then resolves @/path/to/file and attaches the
file content to the model request.

Changes

- `crates/tui/src/commands/mod.rs`: Added path-detection logic in the _ fallthrough arm of execute() (+11 lines). When the input looks like an absolute file path, it returns
  AppAction::SendMessage("@<path>").
- Tests: Added 3 new unit tests (+51 lines):
- absolute_path_input_converts_to_file_mention — /Users/test/project/README.md produces SendMessage
- absolute_path_with_spaces_converts_to_file_mention — /tmp/my documents/note.txt works with spaces
- non_path_single_word_after_slash_still_unknown_command — /nonesuch still gets "Unknown command"

### DIFF
--- a/crates/tui/src/automation_manager.rs
+++ b/crates/tui/src/automation_manager.rs
@@ -159,7 +159,9 @@ impl AutomationSchedule {
             Some("WEEKLY") => AutomationFrequency::Weekly,
             Some("MONTHLY") => AutomationFrequency::Monthly,
             Some("YEARLY") => AutomationFrequency::Yearly,
-            Some(other) => bail!("Unsupported RRULE FREQ '{other}'. Supported: HOURLY, WEEKLY, MONTHLY, YEARLY"),
+            Some(other) => bail!(
+                "Unsupported RRULE FREQ '{other}'. Supported: HOURLY, WEEKLY, MONTHLY, YEARLY"
+            ),
             None => bail!("RRULE must include FREQ"),
         };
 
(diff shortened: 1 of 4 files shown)


Hmbown:
Thanks for the repro and tests. The absolute slash-path issue is already fixed on current main by the shared composer slash-command classifier, and #1568 is closed. I am closing this PR because it is now superseded and also carries unrelated automation/UI changes that we should not merge into the patch release.

PR #1581, opened 2026-05-13
fix(cache): reduce DeepSeek tool-result replay tail

caulif:
## Summary

DeepSeek's prompt cache helps only on the repeated prefix. Once that prefix hits, a large stale tool-result tail still lands in `prompt_cache_miss_tokens` and can dominate prompt cost. DeepSeek-TUI was replaying older `read_file`, shell, parallel, and web tool results with the same broad wire budget used for fresh results, so later turns kept carrying unnecessarily large uncached tails.

This patch makes stale tool-result replay tool-aware:

- keep the most recent tool result readable
- replay older `read_file` results with a much smaller window
- replay older shell / parallel / web results with a smaller budget than fresh ones
- stamp `read_file` outputs with stable replay metadata
- tighten large-context compaction thresholds for noisy, re-readable tool outputs

## What changed

- `crates/tui/src/client/chat.rs`
  - add stale-vs-fresh replay budgets
  - make `read_file` the tightest stale replay profile
  - keep the existing larger budget for fresh tool results
- `crates/tui/src/core/engine/context.rs`
  - lower large-context hard/soft/snippet limits
  - classify `read_file`, `fetch_url`, `web_search`, and `multi_tool_use.parallel` as noisy
- `crates/tui/src/tools/file.rs`
  - add replay metadata (`path`, `shown_lines`, `truncated`, etc.) to `read_file`

## Why

DeepSeek's official KV cache docs say:

- prompt caching is enabled by default
- responses report `prompt_cache_hit_tokens` and `prompt_cache_miss_tokens`
- later requests hit only when they reuse a repeated prefix
- matching happens on complete cache prefix units

Source:
- <https://api-docs.deepseek.com/zh-cn/guides/kv_cache>

Claude Code's official docs point at the same principle from the other side:

- tool definitions are deferred or loaded on demand
- users are encouraged to manage context proactively

Sources:
- <https://code.claude.com/docs/en/how-claude-code-works>
- <https://code.claude.com/docs/en/costs>

The practical implication for DeepSeek-TUI is that even if prefix hit rate is good, stale replay width still matters because it shows up in the uncached tail.

### Rust verification

- `cargo fmt --all`
- `cargo test -p deepseek-tui --bin deepseek-tui request_builder_compacts_stale_ -- --nocapture`
- `cargo test -p deepseek-tui --bin deepseek-tui read_file_explicit_range_wraps_in_file_tag_with_one_based_lines -- --nocapture`
- `cargo test -p deepseek-tui --bin deepseek-tui v4_large_tool_outputs_are_compacted_before_they_bloat_tail -- --nocapture`
- `cargo test -p deepseek-tui --bin deepseek-tui request_builder_truncates_large_tool_result_for_wire -- --nocapture`
- `cargo test -p deepseek-tui --bin deepseek-tui cache_inspect_reports_tool_result_budget_metadata -- --nocapture`
- `cargo test -p deepseek-tui --bin deepseek-tui read_file_large_file_without_range_uses_default_window -- --nocapture`
- `cargo build -p deepseek-tui`

### DeepSeek API cache reproduction

Using the official Chat Completions API with the same warmed prefix in both cases:

- Old wide stale tail
  - `prompt_tokens = 15273`
  - `cached_tokens = 10240`
  - `prompt_cache_hit_tokens = 10240`
  - `prompt_cache_miss_tokens = 5033`
- New slim stale tail
  - `prompt_tokens = 11349`
  - `cached_tokens = 10240`
  - `prompt_cache_hit_tokens = 10240`
  - `prompt_cache_miss_tokens = 1109`

Delta on the repeated request:

- `3924` fewer prompt tokens
- `3924` fewer miss tokens
- identical cached prefix reuse in both cases

## Notes

- This is a focused replay-tail fix, not a full history-compaction redesign.
- It does not try to solve MCP churn, reasoning replay cost, or every long-session context problem in one PR.

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -494,6 +494,13 @@ pub(crate) const CACHE_WARMUP_USER_TAIL: &str = "请只回复 OK";
 const TOOL_RESULT_SENT_CHAR_BUDGET: usize = 12_000;
 const TOOL_RESULT_HEAD_CHARS: usize = 4_000;
 const TOOL_RESULT_TAIL_CHARS: usize = 4_000;
+const STALE_TOOL_RESULT_HISTORY_WINDOW: usize = 1;
+const STALE_TOOL_RESULT_SENT_CHAR_BUDGET: usize = 4_000;
+const STALE_TOOL_RESULT_HEAD_CHARS: usize = 1_500;
+const STALE_TOOL_RESULT_TAIL_CHARS: usize = 900;
+const STALE_READ_FILE_TOOL_RESULT_SENT_CHAR_BUDGET: usize = 2_500;
+const STALE_READ_FILE_TOOL_RESULT_HEAD_CHARS: usize = 1_200;
+const STALE_READ_FILE_TOOL_RESULT_TAIL_CHARS: usize = 400;
 /// Tool results shorter than this stay inline even when repeated. The
 /// extra prompt bytes are cheaper than forcing the model through an
 /// unnecessary retrieval hop for tiny command outputs.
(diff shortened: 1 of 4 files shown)


Hmbown:

PR #1586, opened 2026-05-13
fix: restore terminal on SIGINT/SIGTERM to prevent shell pollution (#1583)

### LINKED ISSUE #1583: Terminal remains in broken state (raw mode) after exiting with Ctrl+C
## Description

Exiting the application via `Ctrl + C` (SIGINT) leaves the terminal in a broken, unusable state. It appears the application fails to perform the proper cleanup (disabling raw mode and enhanced keyboard protocols) before terminating. As a result, subsequent key presses emit raw ANSI escape sequences directly into the shell instead of being interpreted normally.

## Steps to reproduce

1. Launch `deepseek-tui` in the terminal, 
2. Press `Ctrl + C` to forcefully exit the application.
3. Try to use standard shell shortcuts like `Ctrl + C`, `Ctrl + D`, or simply type.
4. It occurs randomly; generally speaking, it appears only after a period of usage, though occasionally it may manifest even when simply executing the `--version` argument.

Image

## Expected behavior

The application should gracefully handle the `SIGINT` (`Ctrl + C`) signal, execute a cleanup routine to restore the terminal from raw mode back to canonical mode, disable any mouse tracking or advanced keyboard protocols, and then exit cleanly. The shell should be fully functional immediately after.

## Actual behavior

The terminal remains polluted. Standard key combos (`Ctrl + C`, `Ctrl + D`, etc.) no longer work and instead print raw ANSI escape sequences (e.g., characters containing semicolons, the number `5`, and the letter `u`, like `^[[>5u`). The terminal becomes unresponsive to normal commands, and the user is forced to blindly type `reset` or `stty sane` to recover the shell environment.

## Impact

High annoyance / Workflow blocker. It breaks the terminal session entirely upon exit, requiring a manual hard reset of the terminal state every time `Ctrl + C` is used to quit.

## Environment

* OS: macOS Tahoe
* DeepSeek CLI version: 0.8.33 (81e4b93cc9df)
* Install method: downloaded binaries from github releases page
* Model/provider: platform.deepseek.com, deepseek-v4-pro
* Terminal app: iTerm2
* Shell: zsh, powered by starship.rs

No error logs are produced upon exit. The bug manifests purely as terminal state corruption (unhandled raw mode / CSI sequences leaking into `zsh`). Typing `reset` clears the issue.

Vishnu1837:
## Summary

- Adds a tokio signal-handler task that calls a shared `emergency_restore_terminal` helper on SIGINT (all platforms), SIGTERM, and SIGHUP (Unix), then exits with `128 + signo`.
- Centralizes the terminal-restore sequence (pop kitty enhancement flags → disable focus/bracketed-paste/mouse capture → disable raw mode → leave alt-screen) in `crate::tui::ui::emergency_restore_terminal`; the existing panic hook now delegates to it so both abnormal exit paths share one implementation.
- No behavior change during normal TUI operation: once raw mode is enabled the terminal driver delivers Ctrl+C as the 0x03 byte rather than SIGINT, so the in-TUI key handler keeps owning interrupt semantics. The new handler only fires in the gaps the bug report describes — pre-TUI subcommands (`--version`, `doctor`, …), the brief windows around `enable_raw_mode`/`disable_raw_mode`, the external-editor suspend path, and OS-delivered SIGTERM/SIGHUP.

Fixes #1583.

## Test plan

- [x] `cargo check -p deepseek-tui` — clean
- [x] `cargo clippy -p deepseek-tui --all-targets` — no new warnings (two pre-existing `needless_return` warnings on the Windows-gated branches of `push/pop_keyboard_enhancement_flags` are untouched)
- [x] `cargo fmt --check` — clean
- [x] `cargo test -p deepseek-tui --bins` — 2902 passed; 7 pre-existing failures verified against unmodified `upstream/main` (skills/project_context/shell-background timing tests, none touch terminal or signal code)
- [ ] Manual repro on macOS/iTerm2 — I don't have access to that environment; relying on @simonkuang or maintainers to confirm the kitty `^[[>5u` leak is gone

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -309,6 +309,54 @@ enum ExecOutputFormat {
     StreamJson,
 }
 
+/// Spawn a tokio task that listens for terminating signals (SIGINT
+/// always; SIGTERM and SIGHUP on Unix) and, on receipt, restores the
+/// terminal modes and exits with the conventional 128 + signal code.
+/// Multiple deliveries are tolerated: once the cleanup runs, a second
+/// signal short-circuits to plain exit so a stuck cleanup can never
+/// trap a frustrated user pressing Ctrl+C repeatedly.
+///
+/// See the call site in `main` for the rationale (#1583).
+fn spawn_signal_cleanup_task() {
+    tokio::spawn(async {
+        let exit_code = wait_for_terminating_signal().await;
+        // If we get here a fatal signal arrived. Restore the terminal
+        // and exit. A second signal during cleanup re-enters this
+        // path and aborts via `std::process::exit` directly.
+        static CLEANED_UP: std::sync::atomic::AtomicBool =
+            std::sync::atomic::AtomicBool::new(false);
+        if !CLEANED_UP.swap(true, std::sync::atomic::Ordering::SeqCst) {
+            crate::tui::ui::emergency_restore_terminal();
+        }
+        std::process::exit(exit_code);
+    });
+}
+
+#[cfg(unix)]
+async fn wait_for_terminating_signal() -> i32 {
+    use tokio::signal::unix::{SignalKind, signal};
+    // Failing to install any individual stream is non-fatal: we still
+    // want the others to work. The fallback never-resolving future
+    // keeps `select!` well-typed when a stream fails to register.
+    let mut sigint = signal(SignalKind::interrupt()).ok();
+    let mut sigterm = signal(SignalKind::terminate()).ok();
+    let mut sighup = signal(SignalKind::hangup()).ok();
+    tokio::select! {
+        _ = async { match sigint.as_mut() { Some(s) => { s.recv().await; }, None => std::future::pending::<()>().await, } } => 130,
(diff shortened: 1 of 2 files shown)


simonkuang:
lgtm

Based on previous experience, this would happen about 80–90% of the time after using DeepSeek TUI for a while, asking a few questions, and then exiting with a double Ctrl+C. I tested the newly compiled binary five times, and the issue did not reproduce in any of them.

---

```zsh
git clone https://github.com/Hmbown/DeepSeek-TUI
cd DeepSeek-TUI

git checkout main
git fetch origin pull/1586/head
git merge FETCH_HEAD

git log --oneline --graph -n 10  # confirm PR #1586 merged

cargo build --release

# launch deepseek tui
cp ./target/release/deepseek ~/bin/
cp ./target/release/deepseek-tui ~/bin/
~/bin/deepseek
```

Vishnu1837:
Thanks for testing it end-to-end, @simonkuang  really appreciate your confirmation. That's the signal I couldn't get locally on Windows.

Flagging for maintainers: ready for review whenever you have a moment.

PR #1630, opened 2026-05-14
fix: validate API key before terminal init to prevent garbled output

duanchao-lab:
When the API key was missing, run_tui would initialize the terminal (alternate screen, raw mode, Kitty keyboard protocol, mouse capture) and then bail at DeepSeekClient::new(config)?, skipping the cleanup code. This left escape sequences in stdout, rendering as garbled text.

Two changes:
- Pre-validate the API key in run_interactive() before calling run_tui(), so the error message prints cleanly without entering terminal mode.
- Add a TerminalCleanupGuard (RAII) inside run_tui() as a safety net: if any other early ?-return occurs after terminal init, the guard restores raw mode, alternate screen, mouse capture, and keyboard flags on drop.

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -4326,6 +4326,12 @@ async fn run_interactive(
         ),
     }
 
+    // Pre-validate the API key *before* entering the TUI (which switches
+    // into alternate screen / raw mode). Failing after terminal init would
+    // leave escape-sequence garbage on the user's screen.  Self-hosted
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks for sending this. The safe part of the idea - restoring terminal state when startup exits early - has now landed on main through #1647, with credit in the commit/changelog. I did not merge this PR as-is because the unconditional API-key prevalidation would bypass some provider-aware onboarding paths, but the cleanup guard itself is included for v0.8.38.

PR #1631, opened 2026-05-14
feat(ci): auto-update Homebrew tap formula on release

### LINKED ISSUE #1602: [Feature Request] DeepSeek-TUI Homebrew Tap 应该自动更新
DeepSeek-TUI Homebrew Tap 应该自动更新。目前看上去是手工更新的内容，会导致目前 Homebrew Tap 版本落后于主线版本。或者是有其他的发布规则，如果存在需要单独声明

Oliver-ZPLiu:
## Summary

Automatically update the Homebrew tap formula at [Hmbown/homebrew-deepseek-tui](https://github.com/Hmbown/homebrew-deepseek-tui) after every release, so `brew install` always fetches the latest version instead of a stale one.

Closes #1602.

## Changes

- **New** `.github/scripts/update-homebrew-tap.sh` — parses SHA256 hashes from the release checksum manifest, generates `deepseek-tui.rb`, and pushes to the tap repo. Skips cleanly (exit 0) when no tap token is configured.
- **Modified** `.github/workflows/release.yml` — adds a `homebrew` job that runs after the `release` job succeeds
- **Modified** `.gitignore` — un-ignores `.github/scripts/**` so the script is tracked

## Prerequisites

The repo needs a secret `HOMEBREW_TAP_PAT` (a PAT with `contents:write` on `Hmbown/homebrew-deepseek-tui`). Falls back to `RELEASE_TAG_PAT` if not set. If neither is configured, the job exits cleanly without failing the release.

## Testing

- [x] Script syntax check (`bash -n`) passes
- [x] SHA parsing verified against v0.8.35 release manifest — all 8 binary names resolve correctly
- [x] Generated formula Ruby syntax check (`ruby -c`) passes
- [x] Formula structure matches the existing tap (diff: parameterized version/URL/SHA + `using: :nounzip`)
- [x] TOKEN-less skip verified: script exits 0 when TOKEN is empty
- [ ] Requires a real release to validate end-to-end in CI

## Why it matters

The tap is currently hand-maintained and lags behind (still at v0.8.27 while releases are at v0.8.35). Tying it into the release pipeline eliminates the manual step and keeps Homebrew users on the latest release.

### DIFF
--- a/.github/scripts/update-homebrew-tap.sh
+++ b/.github/scripts/update-homebrew-tap.sh
@@ -1,0 +1,127 @@
+#!/usr/bin/env bash
+# Update the Homebrew tap at Hmbown/homebrew-deepseek-tui after a release.
+#
+# Expected environment:
+#   TAG       – git tag, e.g. "v0.8.31"
+#   MANIFEST  – path to deepseek-artifacts-sha256.txt
+#   TAP_REPO  – owner/repo of the Homebrew tap
+#   TOKEN     – PAT with contents:write on TAP_REPO
+
+set -euo pipefail
+
+: "${TOKEN:?Set HOMEBREW_TAP_PAT or RELEASE_TAG_PAT}"
+: "${TAG:?}"
+: "${MANIFEST:?}"
+: "${TAP_REPO:?}"
+
+VERSION="${TAG#v}"
+
+die() { echo "::error::${1}" >&2; exit 1; }
+
+sha() {
+  local file="${1:?}"
+  local val
+  val="$(awk -v f="${file}" '$2 == f {print $1; exit}' "${MANIFEST}")"
+  if [ -z "${val}" ]; then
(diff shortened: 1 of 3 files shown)


Hmbown:
Thanks, this is useful release automation and the implementation is close. I’m not merging it as-is yet because the new release job hard-fails when neither HOMEBREW_TAP_PAT nor RELEASE_TAG_PAT is available. That would make an otherwise successful tag release show failed after publishing.\n\nPlease make the Homebrew update step skip cleanly when no tap token is configured, or otherwise gate it in a way that cannot fail the release unless the token is intentionally enabled. Once that is handled, this should be a good candidate to land.

Oliver-ZPLiu:
Thanks for the review. Fixed — the Homebrew update script now exits 0 when neither HOMEBREW_TAP_PAT nor RELEASE_TAG_PAT is configured, so a missing tap token won't fail the release.

PTAL.

Issue #1422, opened 2026-05-11
deepseek-linux-x64 0.8.28 火绒报告病毒：Trojan/Linux.Agent.bp


daijpeng:
病毒名称：Trojan/Linux.Agent.bp
病毒ID：A39DA02A4E0C2A7C
病毒路径：C:\Users\xxx\Downloads\deepseek-linux-x64
操作结果：已处理

进程ID：5248
操作进程：C:\Program Files (x86)\Google\Chrome\Application\chrome.exe

<img width="1123" height="756" alt="Image" src="https://github.com/user-attachments/assets/2c95ec3a-f705-409e-a1fe-1f184878fefd" />

wy414012:
[fix_rusqlite-bundled-feature-Linux.tar.gz](https://github.com/user-attachments/files/27750768/fix_rusqlite-bundled-feature-Linux.tar.gz) @Hmbown

Hmbown:
Closing during v0.8.59 milestone hygiene: this was an old v0.8.28 antivirus/download-safety report, and the related download safety documentation landed in #1627. Please open a new report with current release details if this reproduces now.

PR #1633, opened 2026-05-14
fix: resolve false positive Trojan/Linux.Agent.bp and improve CNB pip…

wy414012:
Security fix:
- Trojan/Linux.Agent.bp (ID: A39DA02A4E0C2A7C)
- rusqlite bundled feature compiles libsqlite3-sys v0.28 which triggers Huorong antivirus heuristic signature A39DA02A4E0C2A7C
- All Rust projects using this dependency on Linux are affected (same root cause as DeepSeek-TUI #1422)
- Remove bundled feature from rusqlite, use system sqlite3 instead

CNB pipeline improvements:
- Split tag_push into parallel pipelines for linux-x64 and linux-arm64
- Separate build and upload stages for better parallelization
- Add checksum generation and release creation to a dedicated release pipeline
- Include both architectures in release notes and SHA256 manifest

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -2910,7 +2910,10 @@ mod stream_decoder_tests {
         // to and `retrieve_tool_result ref=sha:` reads back from.
         let path = crate::tools::truncate::sha_spillover_path(&sha)
             .expect("sha spillover path resolvable under test root");
-        assert!(path.exists(), "large write_file output not persisted: {path:?}");
+        assert!(
+            path.exists(),
+            "large write_file output not persisted: {path:?}"
+        );
         let persisted = std::fs::read_to_string(&path).expect("read persisted spillover");
         assert_eq!(
             persisted, big_diff,


Hmbown:
Thanks for jumping on the antivirus report and CNB release path. I’m going to hold this PR instead of merging it as-is because it mixes two separate concerns:\n\n1. the CNB release pipeline expansion, which may be useful but needs to be reviewed/tested as release infrastructure;\n2. removing rusqlite's bundled SQLite feature and adding system SQLite runtime dependencies, which changes the Linux distribution contract and is not yet proven to address the Trojan/Linux.Agent.bp false positive.\n\nFor the malware false positive, the safer path is evidence-driven: compare signed/checksummed release assets and scanner results before changing runtime dependencies. For CNB, a split PR that only changes .cnb.yml/release plumbing would be much easier to evaluate.

wy414012:
> Thanks for jumping on the antivirus report and CNB release path. I’m going to hold this PR instead of merging it as-is because it mixes two separate concerns:\n\n1. the CNB release pipeline expansion, which may be useful but needs to be reviewed/tested as release infrastructure;\n2. removing rusqlite's bundled SQLite feature and adding system SQLite runtime dependencies, which changes the Linux distribution contract and is not yet proven to address the Trojan/Linux.Agent.bp false positive.\n\nFor the malware false positive, the safer path is evidence-driven: compare signed/checksummed release assets and scanner results before changing runtime dependencies. For CNB, a split PR that only changes .cnb.yml/release plumbing would be much easier to evaluate.

Okay, I have verified it, but currently this PR is incomplete and will cause CI failures on macOS and Windows platforms. I need time to read the entire project and redo it.

PR #1640, opened 2026-05-14
perf(cache): snapshot system prompt at session creation, wire warmup, and share prefix across sub-agents

zhnnky329:
## Summary

This PR addresses the single largest source of DeepSeek KV prefix-cache misses: per-turn system-prompt byte drift from disk re-reads.  Before this patch, five disk-backed components were re-read on every turn, and any change (even one byte) invalidated the entire conversation's cache.

### What changed

- **StaticPromptCache** — All five disk-backed system-prompt components (project context, context pack, instructions files, skills catalogue, handoff block) are snapshotted once at session creation.  `refresh_system_prompt` now reuses these cached strings instead of re-reading from disk, guaranteeing byte-stable system-prompt bytes across turns.

- **Cached user memory block** — The `remember` tool and `# foo` quick-add append timestamped bullets to `memory.md`.  Previously, the next turn re-read the file and the new bytes invalidated the entire KV prefix cache.  Now the snapshot is taken at session creation and never re-read.

- **Cache warmup on session resume** — `build_cache_warmup_request` already existed in the codebase but was never wired.  A best-effort warmup (stable prefix + `"请只回复 OK"`) now fires before the first real turn on resumed sessions.

- **Sub-agent prefix sharing** — Every `agent_spawn` built a fresh system prompt from scratch (cold cache).  Sub-agents now inherit the parent's `StaticPromptCache` and prepend it to their type-specific system prompt, so the first sub-agent request shares byte-for-byte identical prefix with the parent.

- **Context-usage in turn_meta** — `<turn_meta>` now includes exact context usage (`Context usage: 42% (420K / 1M tokens)`) and proactive handoff reminders at 80% and 90% thresholds (#667).

- **RLM wall-clock timeout** — Added a generous 600 s default timeout (overridable via `DEEPSEEK_RLM_TIMEOUT_SECS`) to prevent hung RLM sessions.

- **Stream buffer guard** — RAII `StreamBufferGuard` returns SSE decode buffers to the pool on drop, preventing leaks on cancelled/panicked streams.

- **Request dump diagnostic** — `DEEPSEEK_DUMP_REQUESTS=1` writes every chat request body to disk.  `deepseek cache-diff` diffs consecutive dumps and reports the first byte that drifts between turns — the exact spot where DeepSeek's prefix cache stops hitting.

### Verification

- Full workspace test suite: **2561 passed, 0 failed**  
- Real-session smoke test (3 sessions, 7 API requests): all showed `strict prefix extension` — system prompt + tools (58,870 bytes) byte-identical across every turn  
- CLI dispatcher: `deepseek cache-diff` correctly diffed real multi-turn sessions

### Files

13 files changed, +742 / −40 lines, 1 new file (`request_dump.rs`)


Hmbown:

PR #1642, opened 2026-05-14
feat: image vision analysis, fix first-message 400, fix /model default reset

imkingjh999:
## Summary
- Add vision analysis with object bounding boxes (0–1000 normalised coordinates)
- Restructure: move `vision_bridge.rs` → `vision/bridge.rs` module with structured types (`BBox`, `VisualPrimitive`)
- Slim down `tools.rs` to delegate to bridge module
- Add `primitives` config flag for `VisionModelConfig`
- Add `image_analyze` instruction to system prompt
- Fix 400 error on first message after entering TUI (sync provider to config)
- Fix `/model` re-selecting same provider resetting model to default

## `primitives = true` vs `false`

| | `primitives = true` | `primitives = false` |
|---|---|---|
| Bounding boxes | Each object gets a normalised bbox `[x1,y1,x2,y2]` (0–1000) | No coordinates returned |
| Distance comparison | Can accurately compare distances between multiple points in the image | Cannot — text-only description lacks spatial precision |
| Recommended model | `qwen3.5-omni-plus-2026-03-15` (free, 1M calls) | Any vision model |

With `primitives = true` + `qwen3.5-omni-plus-2026-03-15` (free tier), multi-point distance accuracy exceeds ChatGPT's vision capability.

## Test plan
- [x] Paste an image and verify vision analysis runs automatically
- [x] Verify bounding boxes are returned in the vision context
- [x] Start TUI, send first message — should not 400
- [x] Switch to deepseek provider, pick a non-default model, re-select deepseek — model should persist
- [x] Run `cargo test -p deepseek-tui -- vision` — 19 tests pass

PR #1644, opened 2026-05-14
fix: first-message 400 error and /model default reset

imkingjh999:
## Summary
- Fix 400 error on first message after entering TUI — sync provider from `App::new` back to `config.provider`
- Fix `/model` re-selecting same provider resetting model to default — preserve `model_override` when provider unchanged

## Test plan
- [ ] Start TUI, send first message — should not return 400
- [ ] Switch to deepseek provider, pick a non-default model (e.g. flash), re-select deepseek — model should persist


imkingjh999:
Merged into #1642

PR #1669, opened 2026-05-15
fix(tui): debounce FocusGained terminal recovery

vicmon810:
## Summary

This PR debounces terminal mode recovery triggered by `FocusGained` events in the TUI event loop.

Currently, when a `F
[cinnamon-2026-05-15T153329+1200.webm](https://github.com/user-attachments/assets/f155ba49-366d-4634-8edd-dac304905a0f)
ocusGained` event is received, the TUI may call `recover_terminal_modes()` and then force a terminal repaint. On some terminals, re-emitting terminal mode escape sequences during focus handling can cause another focus/repaint cycle, leading to visible flickering and high CPU usage.

This patch keeps the repaint behavior, but prevents `recover_terminal_modes()` from running repeatedly within a short interval.

## Problem

On Linux Mint Terminal, I observed persistent flickering when `deepseek-tui` had focus. CPU usage increased while the terminal window was focused, and the flickering stopped when focus moved to another application.

This appears similar to the FocusGained loop reported for Tabby/xterm.js:

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -744,6 +744,10 @@ async fn run_event_loop(
     let mut web_config_session: Option<WebConfigSession> = None;
     let mut terminal_paused_at: Option<Instant> = None;
     let mut force_terminal_repaint = false;
+    const FOCUS_RECOVERY_DEBOUNCE: Duration = Duration::from_millis(200);
+    let mut last_focus_recovery = Instant::now()
+        .checked_sub(FOCUS_RECOVERY_DEBOUNCE)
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for the report and video. I am closing this PR rather than merging it as-is because terminal recovery is a sensitive path and this patch has no passing test run, no linked focused issue, and changes focus handling globally. The flicker/high-CPU report is still useful: please open a focused issue with terminal name/version, DEEPSEEK terminal settings, and whether --no-alt-screen changes it, or reopen with a narrow test-backed patch. We will keep this class of terminal stability work in mind for v0.8.40.

PR #1680, opened 2026-05-15
fix(tui): guard reasoning_content behind provider check to prevent TUI display corruption from third-party APIs

### LINKED ISSUE #1673: reasoning_content 字段导致第三方中转 OpenAI格式 兼容 API 显示严重错乱
## Description

When using a third-party OpenAI-compatible API provider, the streaming response contains `reasoning_content` fields that cause severe display corruption in deepseek-tui. Each token is rendered as a separate "thinking" block instead of normal text output.

## Steps to reproduce

1. Configure deepseek-tui with an OpenAI-compatible third-party provider:
   ```toml
   provider = "openai"
   model = "deepseek-v4-flash"
   verify_ssl = false

   [providers.openai]
   base_url = "http://<third-party-api>/v1"
   api_key = "<third-party-key>"
   model = "deepseek-v4-flash"
   ```
2. Set `DEEPSEEK_ALLOW_INSECURE_HTTP=1` and launch `deepseek-tui`
3. Send any message (e.g. "你好")

## Expected behavior

The response text should be rendered normally as a continuous text block, like:
```
你好。有什么需要帮忙的？
```

## Actual behavior

The response is broken into individual tokens, each preceded by a `thinking done · 0.0s` block:
```
thinking done · 0.0s
Full reasoning in Ctrl+O

thinking done · 0.0s
Full reasoning in Ctrl+O

thinking done · 0.0s
...

你好
。
有什么
需要
帮忙
的
？
```

The `reasoning_content` field in the streaming SSE chunks triggers the thinking/reasoning UI for every single chunk, even when the actual content is just normal text tokens.

## Impact

This makes deepseek-tui completely unusable with any third-party OpenAI-compatible API that returns DeepSeek models' `reasoning_content` field. The display is so corrupted that the tool cannot be used for any practical work.

Workaround attempts:
- Local proxy filtering `reasoning_content` from SSE chunks: causes compatibility issues with deepseek-tui's SSE client (no output or further format issues)
- No workaround currently available

## Environment

- OS: Windows 11 Pro
- DeepSeek CLI version: installed via npm (deepseek-tui package)

LittleBlacky:
## Problem

When using a third-party OpenAI-compatible API provider that returns `reasoning_content` fields in streaming SSE chunks, deepseek-tui incorrectly treats every token as a separate "thinking" block instead of normal text output. This causes severe display corruption where each character/token is rendered with its own `thinking done · 0.0s` header.

Fixes #1673

## Root Cause

1. `is_reasoning_model()` checks only the model name (e.g., `deepseek-v4-pro`, `deepseek-v4-flash`) to decide whether `reasoning_content` deltas should be treated as thinking blocks — it does not consider whether the current provider actually supports reasoning content semantics.

2. When a non-DeepSeek provider streams `reasoning_content` fields alongside normal response text, every chunk triggers the thinking UI path, corrupting the display.

## Fix

- Added `requires_reasoning_content()` helper that returns `true` only for the native DeepSeek provider (`ProviderKind::Deepseek`).
- Updated `handle_chat_completion_stream()` to use `requires_reasoning_content(api_provider)` instead of `is_reasoning_model(model)` when deciding whether to emit `StreamEvent::ReasoningDelta`.
- Updated `parse_sse_chunk()` to accept the provider kind and use the same gate.

This ensures third-party OpenAI-compatible providers that happen to serve DeepSeek models are not incorrectly treated as supporting DeepSeek's reasoning content protocol, while preserving full native DeepSeek reasoning display.

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -215,6 +215,11 @@ impl DeepSeekClient {
         }
 
         let model = request.model.clone();
+        // The provider determines whether we interpret `reasoning_content` in
+        // streaming deltas as real thinking tokens (DeepSeek, OpenRouter, etc.)
+        // or as a non-standard host that puts the response text there
+        // (third-party OpenAI-compatible proxies).
+        let api_provider = self.api_provider;
 
         // Capture transport-shape headers before we consume `response` into
         // `bytes_stream()`. They are surfaced in the decode-error log path so
@@ -251,7 +256,8 @@ impl DeepSeekClient {
             let mut text_started = false;
             let mut thinking_started = false;
             let mut tool_indices: std::collections::HashMap<u32, u32> = std::collections::HashMap::new();
-            let is_reasoning_model = requires_reasoning_content(&model);
+            let is_reasoning_model = requires_reasoning_content(&model)
+                && provider_accepts_reasoning_content(api_provider);
 
             let mut byte_stream = std::pin::pin!(byte_stream);
             let idle = stream_idle_timeout();
@@ -1637,6 +1643,7 @@ fn provider_accepts_reasoning_content(provider: ApiProvider) -> bool {
         provider,
         ApiProvider::Deepseek
             | ApiProvider::DeepseekCN
+            | ApiProvider::NvidiaNim
             | ApiProvider::Openrouter
             | ApiProvider::Novita
             | ApiProvider::Fireworks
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for the provider-gated reasoning_content fix. The scoped behavior was harvested into v0.8.39 via #1734 and credited in the changelog, so I am closing this PR to keep the queue clean. If a generic OpenAI-compatible provider still renders answer text as repeated thinking blocks on latest, please reopen with the provider/model and one raw SSE chunk.

PR #1699, opened 2026-05-15
fix(docker): pre-create /home/deepseek/.deepseek to avoid permission denied on first run

### LINKED ISSUE #1684: Docker: `Permission denied (os error 13)` creating .deepseek/tasks/runtime/threads on first run
## Description

Running the prebuilt `ghcr.io/hmbown/deepseek-tui:latest` image with the exact `docker run` invocation from the README's install section fails to start. The agent aborts trying to create a directory inside the mounted `/home/deepseek/.deepseek` volume with `Permission denied (os error 13)`.

## Steps to reproduce

1. `docker volume create deepseek-tui-home`
2. Run the command copied verbatim from the README:

```bash
docker run --rm -it \
  -e DEEPSEEK_API_KEY="$DEEPSEEK_API_KEY" \
  -v deepseek-tui-home:/home/deepseek/.deepseek \
  -v "$PWD:/workspace" \
  -w /workspace \
  ghcr.io/hmbown/deepseek-tui:latest
```

## Expected behavior

`deepseek` starts and the TUI is usable.

## Actual behavior

Process exits immediately with:

```
Failed to create /home/deepseek/.deepseek/tasks/runtime/threads
Caused by:
    Permission denied (os error 13)
```

## Impact

Blocks first-time Docker users following the README Quickstart verbatim. Reproduces deterministically on a clean environment.

## Environment

- OS: macOS 26.5 (Darwin 25.5.0, arm64)
- DeepSeek CLI version: `ghcr.io/hmbown/deepseek-tui:latest` (Docker image)
- Install method: Docker / GHCR
- Model/provider: deepseek-v4-pro / DeepSeek
- Terminal app: macOS terminal
- Shell: zsh
- Docker runtime: OrbStack; Docker client 29.5.0 / server 29.4.0

Tail of terminal output:

```
... Failed to create /home/deepseek/.deepseek/tasks/runtime/threads

                                                       Caused by:
                                                                   Permission denie
(os error 13)
```

mvanhorn:
## Summary

The non-root `deepseek` user couldn't write `/home/deepseek/.deepseek/tasks/runtime/threads` on first run inside Docker, failing with `Permission denied (os error 13)`. The Dockerfile created the user and switched to it before any process ever touched the home subdirectory; the runtime then hit a fresh `/home/deepseek` owned by root (because the user's HOME was newly created with default perms but bind-mounted volumes adopted root ownership).

Pre-create `/home/deepseek/.deepseek` with `deepseek:deepseek` ownership in the same `RUN` that adds the user, so any subsequent volume mount or process write lands on a directory the runtime user can write to.

## Why this matters

Reporter @[issue1684_user] in #1684 reported the bridge crashing on first launch in Docker before any thread could be created. The fix is contained to the user-setup section of the Dockerfile (same `RUN` that calls `useradd`), keeps the image layer count unchanged, and doesn't touch the runtime.

Fixes #1684

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,8 +75,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # Non-root user with explicit UID/GID for filesystem ownership clarity.
 RUN groupadd --gid 1000 deepseek \
     && useradd --create-home --shell /bin/bash --uid 1000 --gid 1000 deepseek \
-    && mkdir -p /home/deepseek/.deepseek \
-    && chown -R deepseek:deepseek /home/deepseek/.deepseek
+    && install -d -m 0700 -o deepseek -g deepseek /home/deepseek/.deepseek
 USER deepseek
 WORKDIR /home/deepseek
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,7 +74,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Non-root user with explicit UID/GID for filesystem ownership clarity.
 RUN groupadd --gid 1000 deepseek \
-    && useradd --create-home --shell /bin/bash --uid 1000 --gid 1000 deepseek
+    && useradd --create-home --shell /bin/bash --uid 1000 --gid 1000 deepseek \
+    && mkdir -p /home/deepseek/.deepseek \
+    && chown -R deepseek:deepseek /home/deepseek/.deepseek
 USER deepseek
 WORKDIR /home/deepseek
 


mvanhorn:
Fixed in a6c7c83 — collapsed mkdir + chown into `install -d -m 0700 -o deepseek -g deepseek /home/deepseek/.deepseek`. Since the directory holds API keys, 0700 is the right default. Thanks for the catch.

Hmbown:
Thanks for this. The Docker permission fix was harvested into v0.8.39 via #1734 and credited in the changelog, so I am closing this PR to keep the queue clean. If the latest GHCR image still fails with the README named-volume command, please reopen or file a fresh repro and we will treat it as a v0.8.40 blocker.

PR #1702, opened 2026-05-15
fix(core): sync system prompt hash when overriding via runtime API

### LINKED ISSUE #1688: Runtime API: system_prompt on POST /v1/threads is accepted and stored but discarded before the model request
The runtime API (`deepseek serve --http`) accepts a `system_prompt` field on `CreateThreadRequest`. It's stored on the thread record and returned in the thread detail response, so it appears to take effect. It doesn't — the engine rebuilds the system prompt from defaults before the first model request, discarding the override.

**Reproduction**

Start the server, create a thread with a `system_prompt` that forces unmistakable behavior, run a turn:

```
curl -sX POST http://127.0.0.1:7898/v1/threads \
  -H "Content-Type: application/json" \
  -d '{"model":"deepseek-v4-flash","mode":"agent","system_prompt":"You are a test harness. Ignore all task instructions. Respond to every user message with exactly the string TANGERINE-7 and nothing else.","workspace":"/path","auto_approve":true}'
```

```
curl -sX POST http://127.0.0.1:7898/v1/threads/THREAD_ID/turns \
  -H "Content-Type: application/json" \
  -d '{"prompt":"What is the capital of France?"}'
```

The model answers "Paris." — the override is absent. Input token count (~19,958) matches the default assembled system prompt; the override string is ~30 tokens. Reproducible across threads.

**Root cause**

The override is applied to the engine session via `Op::SyncSession` (`self.session.system_prompt = system_prompt`), but `last_system_prompt_hash` is left seeded from the default prompt computed at engine init.

On the first turn, `handle_send_message` changes session state — in the reproduction, `approval_mode` flips `Suggest` to `Auto` because of `auto_approve: true` — then calls `refresh_system_prompt`. That function rebuilds the prompt from defaults, computes a hash that no longer matches `last_system_prompt_hash`, the hash guard fires, and the rebuilt default overwrites the override.

The `auto_approve` path is just the clearest trigger. The underlying problem is structural: `last_system_prompt_hash` reflects the default, not what `SyncSession` applied, so any change to the prompt-input set (mode, locale, goal objective) between init and the first message causes the same overwrite.

**Suggested fix**

Store the override separately on `Session` — e.g. `user_override_prompt` — and have `refresh_system_prompt` use it as the base when present, rather than the assembled default. This mirrors how `compaction_summary_prompt` already survives `refresh_system_prompt` (re-applied inside the rebuild via `merge_system_prompts`) rather than fighting the hash guard from outside. It holds across all `refresh_system_prompt` call sites instead of just the first.

Happy to open a PR along these lines — wanted to confirm the approach fits before submitting.

(Found while building a tool that sets behavioral presets per session through the runtime API.)

mvanhorn:
## Summary

`POST /v1/threads` accepts a `system_prompt` field and stores it on the session, but the next `refresh_system_prompt` call rebuilt the stable baseline and overwrote it because `session.last_system_prompt_hash` was still the pre-override value. The model request fired with the rebuilt baseline, not the override.

Add `sync_session_system_prompt_hash` that recomputes `last_system_prompt_hash` after a thread API override, and call it where the override is installed. Now `refresh_system_prompt`'s hash-equality short-circuit fires and the override survives.

Factor the stable-prompt construction into `stable_system_prompt_for_mode` so the sync and the refresh paths share it.

## Why this matters

Reporter quote from #1688: "system_prompt on POST /v1/threads is accepted and stored but discarded before the model request." Root cause: hash drift between session.last_system_prompt_hash (pre-override baseline) and the new override. The new tests (`text_system_prompt_override_via_sync_session_survives_refresh` and the Blocks variant) lock the regression in.

Closes #1688

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -756,6 +756,7 @@ impl Engine {
                     } else {
                         None
                     };
+                    self.sync_session_system_prompt_hash();
                     self.session.rebuild_working_set();
                     self.rehydrate_latest_canonical_state();
                     self.emit_session_updated().await;
@@ -1773,8 +1774,8 @@ impl Engine {
             .await;
     }
 
-    /// Refresh the system prompt based on current mode and context.
-    fn refresh_system_prompt(&mut self, mode: AppMode) {
+    /// Build the stable system prompt baseline for the current mode and context.
+    fn stable_system_prompt_for_mode(&self, mode: AppMode) -> Option<SystemPrompt> {
         let user_memory_block =
             crate::memory::compose_block(self.config.memory_enabled, &self.config.memory_path);
         let base = prompts::system_prompt_for_mode_with_context_skills_session_and_approval(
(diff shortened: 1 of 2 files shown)


mvanhorn:
Addressed in e30e09a -- used `Option::map` so the no-override path lands `None` instead of `Some(hash_of_None)`. Behavior is the same: a `None` hash compares unequal to any `Some(stable_hash)` on the next refresh, so override-clearing still works without explicitly hashing `None`.

Kept `AppMode::Agent` hardcoded for now -- the runtime API (`POST /v1/threads`) is agent-mode-scoped and `refresh_system_prompt` is invoked with the same mode the request was processed in, so the hash will still match. Happy to thread the mode through if you'd rather make this defensive against future mode-mixed sessions.

Tests: `cargo test -p deepseek-tui --bin deepseek-tui sync_session` -- both regression tests pass.

Hmbown:
Thanks for the focused fix and tests. The runtime system_prompt override regression was harvested into v0.8.39 via #1734 and credited in the changelog, so I am closing this PR to keep the queue clean. If the latest release still drops thread-level system_prompt overrides, please reopen with the failing request payload.

PR #1704, opened 2026-05-16
fix(compaction): preserve usr txt query to prevent Jinja template crash

krisclarkdev:
When auto compaction runs on a session where recent history is dominated by tool calls and results, all original user text queries could be dropped from the pinned context. This causes backend inference servers (e.g. vLLM/SGLang hosting DeepSeek V4) that rely on Jinja chat templates to throw a "No user query found in messages" exception.

This fix forces `plan_compaction` to scan backwards and pin at least one user text query if the recent pinned context lacks one.

## Testing

- [X] `cargo test --all-features`
- [X] `cargo fmt --all -- --check`
- [X] `cargo clippy --all-targets --all-features`

## Checklist

- [X] Updated docs or comments as needed
- [X] Added or updated tests where relevant
- [X] Verified TUI behavior manually if UI changes

### DIFF
--- a/crates/tui/src/compaction.rs
+++ b/crates/tui/src/compaction.rs
@@ -432,6 +432,31 @@ pub fn plan_compaction(
     // Ensure tool result messages are not kept without their corresponding tool call.
     enforce_tool_call_pairs(messages, &mut pinned_indices);
 
+    // FIX: Jinja templates on some backends (like vLLM/SGLang with DeepSeek V4) crash with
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks, this was the right bug to catch. The compaction user-query preservation fix was harvested into v0.8.39 via #1734 and credited in the changelog, so I am closing this PR to keep the queue clean. Any remaining cache/compaction performance work is now tracked under the v0.8.40 milestone.

PR #1706, opened 2026-05-16
fix(pager): G/End overshoot makes k/Up unresponsive after jump to bottom

wlon:
## Summary

After pressing `G` or `End` to jump to the bottom of the pager (Ctrl+O reasoning list, tool detail, etc.), pressing `k` / `Up` to scroll back up had no visible effect for many consecutive presses.

## Root cause

`max_scroll()` returned `lines.len() - 1` while `render()` clamped to `lines.len() - visible_height`. The gap meant `G` set `scroll` far above the render clamp. Each `k`/`Up` press decremented `scroll` inside that overshoot gap, producing zero visible movement until the value finally fell below the clamp.

For a 50-line pager with ~18 visible lines, this required **17 consecutive `k` presses** before any content scrolled up.

## Fix

Changed `max_scroll()` to use `page_height()` (the cached `visible_height` from the last render), matching the render-side clamp:

```rust
// Before
self.lines.len().saturating_sub(1)
// After
self.lines.len().saturating_sub(self.page_height().max(1))
```

## Tests

Two regression tests added. All 33 pager tests pass.

### DIFF
--- a/crates/tui/src/tui/pager.rs
+++ b/crates/tui/src/tui/pager.rs
@@ -128,7 +128,7 @@ impl PagerView {
         // that G / End land at the actual bottom and k / Up immediately scroll
         // back up by one visible line — over-scrolling would make N consecutive
         // scroll-up presses invisible to the user.
-        self.lines.len().saturating_sub(self.page_height().max(1))
+        self.lines.len().saturating_sub(self.page_height())
     }
 
     fn start_search(&mut self) {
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for the clear pager repro. The G/End bottom clamp fix was harvested into v0.8.39 via #1734 and credited in the changelog, so I am closing this PR to keep the queue clean. Please reopen if v0.8.39 still needs multiple k/Up presses after jumping to the bottom.

PR #1715, opened 2026-05-16
fix(tui): scroll transcript when composer has text (#1677)

tdccccc:
## Summary

- When `composer_arrows_scroll` is enabled, Up/Down keys now always scroll the transcript, regardless of whether the composer has text.
- Previously, arrows only scrolled when the composer was empty. With text in the composer, mouse wheel events (sent as arrow keys by terminals without mouse capture) navigated input history instead of scrolling the transcript, making it impossible to review earlier conversation while mid-typing.

## Problem

On terminals without mouse capture (e.g. Warp on Windows, legacy conhost), mouse wheel events are converted to Up/Down arrow keys by the terminal. When `composer_arrows_scroll = true`, these arrow keys only triggered transcript scrolling when the composer was empty. If the user had text in the input field, the arrow keys navigated input history instead, preventing scrolling while typing.

Users had to manually add `mouse_capture = true` to their config to work around this, which isn't obvious.

## Fix

Removed the `app.input.trim().is_empty()` guard in `handle_composer_history_arrow`. When `composer_arrows_scroll` is enabled, arrows always scroll the transcript. When disabled (the default on non-Windows with mouse capture), behavior is unchanged.

## Test plan

- [x] Updated `history_arrow_handles_nonempty_input` to explicitly set `composer_arrows_scroll = false`
- [x] Updated `composer_arrows_scroll_nonempty_still_navigates_history` → `composer_arrows_scroll_nonempty_also_scrolls` to verify new behavior
- [x] All 3119 existing tests pass

### DIFF
--- a/crates/tui/src/tui/composer_ui.rs
+++ b/crates/tui/src/tui/composer_ui.rs
@@ -56,15 +56,17 @@ pub(crate) fn handle_composer_history_arrow(
         return false;
     }
 
-    // When `composer_arrows_scroll` is enabled and the composer is empty,
-    // plain Up/Down scroll the transcript.  This helps terminals that map
-    // trackpad gestures to arrow keys.  Otherwise arrows always navigate
-    // input history regardless of composer state (#1117).
-    let scroll_on_empty = app.composer_arrows_scroll && app.input.trim().is_empty();
+    // When `composer_arrows_scroll` is enabled, plain Up/Down always
+    // scroll the transcript regardless of whether the composer has text.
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks for fixing the mouse-wheel-as-arrow draft problem. The scoped behavior was harvested into v0.8.39 via #1734 and credited in the changelog, so I am closing this PR to keep the queue clean. The separate empty-composer Windows default is tracked in #1720 for v0.8.40.

PR #1742, opened 2026-05-17
fix(tui): surface thinking-only turns instead of silently ending (#1727)

LeoLin990405:
## Summary / 概述

**EN:** Fixes #1727. When a model streams a turn containing **only** a reasoning/thinking block — empty `content`, no `tool_calls` (e.g. gpt-oss via ollama's harmony→OpenAI shim, which maps the answer into `reasoning_content`) — the engine persisted nothing, emitted no event, and finished the turn silently. The UI sat idle: spinner never cleared, no reply, no error. It looked hung. This PR surfaces a `status` notice instead of silence.

**中文：** 修复 #1727。当模型本回合**只**返回了 reasoning/thinking 块——`content` 为空、无 `tool_calls`（例如 ollama 的 harmony→OpenAI 适配把答案塞进 `reasoning_content` 的 gpt-oss）——引擎既不持久化任何消息、也不发出任何事件，回合静默结束。界面就此卡住：spinner 不消失、无回复、无报错，看起来像挂死。本 PR 用一条 `status` 提示替代这种静默。

## Root cause / 根因

**EN:** In `crates/tui/src/core/engine/turn_loop.rs`, the assistant‑message persist site is an `if`‑only branch gated on `has_sendable_assistant_content` (true only for `Text`/`ToolUse`). A thinking‑only stream produces neither, so the `if` is skipped — no message, no event — and `if tool_uses.is_empty()` then falls straight through to `break`, ending the turn with **zero events emitted**. The maintainer's v0.8.40 audit (#1736) confirms #1727 was *not* shipped in v0.8.39 and is release‑blocking, and that this exact status branch is still missing.

**中文：** 在 `crates/tui/src/core/engine/turn_loop.rs` 中，assistant 消息的持久化处是一个仅有 `if` 的分支，条件为 `has_sendable_assistant_content`（仅 `Text`/`ToolUse` 为真）。thinking‑only 流两者皆无，于是该 `if` 被跳过——既无消息也无事件——随后 `if tool_uses.is_empty()` 直接走到 `break`，回合在**零事件**下结束。维护者的 v0.8.40 审计（#1736）确认 #1727 在 v0.8.39 *并未*随版本发布、属于发布阻断项，且这一 status 分支至今缺失。

## Fix / 修复

Minimum‑surface‑area (the issue author's "option 1"): add an `else if` arm at the same persist site that, only on a clean end (`tool_uses` empty, `turn_error.is_none()`, not cancelled), `warn`s and emits an `Event::status(...)`: *"Model returned reasoning but no answer or tool call; turn ended without output. Send a follow‑up to retry."* **No assistant message is persisted** (prior behavior preserved). **No** retry/re‑prompt/placeholder policy is added (explicitly out of scope). The guard ordering is extracted into a tiny pure helper `should_emit_thinking_only_status(...)` so it is unit‑testable and matches the existing `should_hold_turn_for_subagents` style. Reuses the exact surrounding patterns: `Event::status`, `crate::logging::warn`, `self.cancel_token.is_cancelled()`, `turn_error.is_none()`.

**中文：** 最小改动面（issue 作者的“方案一”）：在同一持久化处新增 `else if` 分支，仅在干净结束时（`tool_uses` 为空、`turn_error.is_none()`、未取消）`warn` 并发出 `Event::status(...)`：*“模型返回了推理但没有答案或工具调用；本回合无输出结束，可发送追问重试。”* **不持久化任何 assistant 消息**（保持原行为）；**不**加入重试/重提示/占位策略（明确超出范围）。守卫顺序被抽到一个纯函数小助手 `should_emit_thinking_only_status(...)`，便于单测，并与既有 `should_hold_turn_for_subagents` 风格一致。复用周边既有写法。

## Scope / 改动边界

- The existing `if has_sendable_assistant_content` branch is **untouched**; normal text/tool turns persist and render exactly as before.
- No double‑notice: guarded by `turn_error.is_none()` (stream errors still surface their own error) and `!cancelled`.
- 既有 `if` 分支**不动**；正常文本/工具回合行为不变。受 `turn_error.is_none()`、`!cancelled` 守卫，不会重复提示。

## Testing / 测试

New unit test `thinking_only_turn_emits_status_only_on_clean_end` pins the decision: notify only on clean end; suppressed when tool uses pending / an error already shown / cancelled.

```
cargo test -p deepseek-tui --bin deepseek-tui -- thinking_only_turn_emits_status_only_on_clean_end   # ok
cargo clippy -p deepseek-tui   # clean (no warnings/errors)
```

**Honest test limitation:** the test pins the extracted pure decision, not the full async `handle_deepseek_turn` loop — consistent with every existing turn‑loop test in this module (they all pin pure helpers; an end‑to‑end test would need a mock client/session/channel, outside a surgical fix). The `else if` wiring is covered by inspection.

新增单测 `thinking_only_turn_emits_status_only_on_clean_end`，与本模块既有“纯函数式”测试风格一致；端到端回合测试需 mock 客户端/会话/通道，超出本次外科式修复范围，已如实说明。

Refs #1727, #1736.

### DIFF
--- a/crates/tui/src/core/engine/turn_loop.rs
+++ b/crates/tui/src/core/engine/turn_loop.rs
@@ -867,6 +867,17 @@ impl Engine {
                 )
             });
 
+            // Issue #1727: did this turn produce ONLY a reasoning/thinking
+            // block — empty content, no tool calls (e.g. gpt-oss via ollama's
+            // harmony→OpenAI shim mapping to `reasoning_content`)? We do NOT
+            // surface anything here: after this point the same turn can still
+            // CONTINUE for pending steers (~below) or sub-agent completions,
+            // and emitting now would show a spurious "turn ended" notice right
+            // before the turn resumes. Capture the fact and decide later, at
+            // the point the turn is certain to be finishing with no sendable
+            // content (see the `tool_uses.is_empty()` tail).
+            let thinking_only_no_sendable = !has_sendable_assistant_content;
+
             // Add assistant message to session
             if has_sendable_assistant_content {
                 self.add_session_message(Message {
@@ -874,25 +885,6 @@ impl Engine {
                     content: content_blocks,
                 })
                 .await;
-            } else if should_emit_thinking_only_status(
-                tool_uses.is_empty(),
-                turn_error.is_none(),
-                self.cancel_token.is_cancelled(),
-            ) {
-                // Issue #1727: the model streamed ONLY a reasoning/thinking
-                // block — empty content, no tool calls (e.g. gpt-oss via
-                // ollama's harmony→OpenAI shim mapping to `reasoning_content`).
-                // We intentionally do NOT persist an assistant message
-                // (preserves prior behavior), but the previous code emitted
-                // nothing at all and fell straight through to finishing the
-                // turn, leaving the UI spinner hung with no error. Surface a
-                // status so the user knows the turn ended and can retry.
-                let message =
-                    "Model returned reasoning but no answer or tool call; turn ended without \
-                     output. Send a follow-up to retry."
-                        .to_string();
-                crate::logging::warn(&message);
-                let _ = self.tx_event.send(Event::status(message)).await;
             }
 
             // If no tool uses, check for inline REPL blocks (paper §2) or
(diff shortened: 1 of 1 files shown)


LeoLin990405:
Addressed in the latest push. The premature-emission concern is fixed by capture-then-decide:

- At the persist site we now only capture `thinking_only_no_sendable = !has_sendable_assistant_content` (no event emitted there).
- The notice is emitted at the end of the `tool_uses.is_empty()` path, just before the terminal `break` — reachable only when there were no pending steers (`continue`d above), no sub-agent completions to resume with, and we were not holding for running children.
- `should_emit_thinking_only_status` gained `steers_pending` and `holding_for_subagents` params (returns false if either), recomputed live at the decision point as defense-in-depth. Unit test updated with the two new no-emit cases.

So the status can no longer fire right before the turn resumes for a steer or sub-agent completion.

Hmbown:
Thanks for the thinking-only turn fix. This was harvested into v0.8.40 release PR #1823 as commits 4ab7c5344 and 351898d05, and #1823 is now CI-green. Closing as superseded by the release branch; thank you for helping make silent turn endings visible.

PR #1755, opened 2026-05-18
feat(tui): add full-screen thinking stream

### LINKED ISSUE #1750: show the full thinking stream as it arrives
## Problem

Thinking is displayed in a small pane that allows the text to scroll away almost immediately

## Proposed solution

Provide an option to show that stream outside of the pane and use the full terminal window

## Use case

To see what path the model is on, I'd monitor its thinking

## Alternatives considered

None

## Impact

I would use this on almost any prompt that required more than a few lines of thinking to be displayed.

## Additional context

None

LING71671:
## Summary

Fixes #1750.

This adds a full-terminal live thinking stream so long reasoning output can be monitored as it arrives without being squeezed into the compact transcript pane.

## What changed

- Added a `/thinking` slash command, with `/think` as a short alias, that opens a thinking-only live overlay.
- Extended the existing `LiveTranscriptOverlay` with a `ThinkingOnly` mode that:
  - refreshes from the active transcript snapshot every frame, so streaming reasoning continues updating while the overlay is open;
  - filters the overlay to `HistoryCell::Thinking` entries only;
  - renders thinking cells with full-detail transcript rendering instead of compact summaries;
  - keeps the existing sticky-tail controls so the user can either follow the newest reasoning or scroll away and resume tailing with `End`/`G`.
- Updated `Ctrl+O` Activity Detail behavior for in-flight thinking: when the selected/current activity is still streaming, it opens the live thinking stream; finalized thinking still opens the existing reasoning timeline pager.
- Added localized command descriptions for the shipped locales.

## Verification

- `cargo test -p deepseek-tui live_transcript --all-features`
- `cargo test -p deepseek-tui execute_thinking_opens_live_reasoning_stream --all-features`
- `cargo test -p deepseek-tui activity_detail_opens_live_thinking_stream_for_streaming_thinking --all-features`
- `cargo test -p deepseek-tui command_registry_has_unique_names_and_aliases --all-features`
- `cargo test -p deepseek-tui localization --all-features`

Note: clippy still reports existing warnings in `schema_migration.rs` and existing `needless_return` warnings in `ui.rs`; no new blocking warnings were introduced by this change.


Hmbown:

PR #1761, opened 2026-05-18
feat(notifications): idle-based trigger with native OS notification support

punkcanyang:
## Summary

- **Idle-based trigger**: notifications now fire only when the user hasn't typed for `idle_threshold_secs` (default 6s), rather than a fixed turn-elapsed timer
- **Native OS notifications**: `Native` variant sends notifications via `osascript` (macOS) or `notify-send` (Linux), works in any terminal
- **Kitty/Ghostty protocols**: Kitty uses OSC 99, Ghostty uses OSC 777
- **Auto fallback**: unknown terminals fall back to `Bel` instead of `Native` to avoid false positives

## Changes

| File | Change |
|------|--------|
| `config.rs` | Add `Native` variant to `NotificationMethod` |
| `notifications.rs` | Add `Native` method, `native_notify()`, `notify_after_idle()`, idle threshold logic |
| `app.rs` | Add `last_interaction_time` field for idle detection |
| `ui.rs` | Wire idle timer on keypress, switch `notify_done` → `notify_after_idle` |

## Test plan

- [x] Build passes (0 errors)
- [x] 3164 tests pass (2 pre-existing failures on main)
- [ ] Manual test: configure `notification.method = "native"` and verify OS notification appears
- [ ] Manual test: type continuously during a long turn, verify notification does NOT fire

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@
 [简体中文 README](README.zh-CN.md)
 [日本語 README](README.ja-JP.md)
 
+> ⚠️ **SECURITY WARNING**: A fake repository `DeepSeek-TUI/DeepSeek-TUI` is impersonating this
+> project and distributing malware-infected release assets. **Only download from the official
+> repository: [Hmbown/DeepSeek-TUI](https://github.com/Hmbown/DeepSeek-TUI).**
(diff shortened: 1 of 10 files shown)


Hmbown:

PR #1783, opened 2026-05-18
per-run-log-pid-suffix

### LINKED ISSUE #1782: PID-log-prefix
## Problem

When multiple DeepSeek TUI instances run concurrently (e.g. development + stable, or multiple terminals), all write to the same log file: `~/.deepseek/logs/tui-YYYY-MM-DD.log`. This causes:

- **Interleaved log entries** from different instances, making debugging difficult
- **File contention** — multiple processes appending to one file
- **Lost context** — impossible to tell which instance emitted which log line

## Root Cause

`runtime_log.rs` uses a date-only filename with no instance identifier:

```rust
let log_path = log_dir.join(format!("tui-{date}.log"));
```

## Fix

Append the process ID to the filename:

```rust
let log_path = log_dir.join(format!("tui-{date}-{}.log", std::process::id()));
```

Each TUI instance now writes to its own file: `tui-2026-05-19-12345.log`, `tui-2026-05-19-67890.log`, etc.

## Impact

- **One line change** in `crates/tui/src/runtime_log.rs`
- No breaking changes — old log files remain readable
- No new dependencies
- Process ID is unique per boot, sufficient for distinguishing concurrent instances

## Discovery

Encountered while developing with `native-dev-branch.ps1` — two TUI instances (main repo + worktree) were writing to the same log, making it impossible to isolate which instance emitted debug output.

aboimpinto:
## Problem

Multiple concurrent TUI instances write to the same log file (`tui-YYYY-MM-DD.log`), causing interleaved entries and file contention.

## Fix

Append process ID to log filename: `tui-YYYY-MM-DD-<PID>.log`.

One line change in `crates/tui/src/runtime_log.rs`.

Closes #1782

### DIFF
--- a/crates/tui/src/runtime_log.rs
+++ b/crates/tui/src/runtime_log.rs
@@ -1,7 +1,7 @@
 //! TUI runtime logging. Initializes a `tracing-subscriber` that writes to a
-//! daily-rolling file under `~/.deepseek/logs/`, and (on Unix) redirects the
-//! process's `stderr` fd to that same file for the lifetime of the alt-screen
-//! TUI.
+//! per-instance file under `~/.deepseek/logs/tui-YYYY-MM-DD-PID.log`, and
+//! (on Unix) redirects the process's `stderr` fd to that same file for the
+//! lifetime of the alt-screen TUI.
 //!
 //! Why this exists:
 //!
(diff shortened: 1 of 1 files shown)


aboimpinto:
Good point about log accumulation. Two thoughts:

**1. This isn't a new problem.** The existing single-file approach already accumulates — the old `tui-YYYY-MM-DD.log` file was 3MB after one day of development. PID files don't create more total bytes; they just partition them into debuggable per-instance files instead of one multi-agent mess where nobody can trace their own session.

**2. Pruning is a separate concern.** I've opened [Issue #1784](https://github.com/Hmbown/DeepSeek-TUI/issues/1784) to track it. Proposed: delete logs older than 7 days (configurable via env, 7-day default). That's the right scope for a follow-up PR rather than blocking this one-liner.

I'll push a doc update to the module comment now.

Hmbown:
Thanks for the per-run log naming fix. This was harvested into v0.8.40 release PR #1823 as commit a595edd56, and #1823 is now CI-green. Closing as superseded by the release branch; thank you for helping make logs easier to inspect.

PR #1800, opened 2026-05-19
fix(tui): restore auto model mode on session load

### LINKED ISSUE #1797: resume from a session will send 'auto' as a model name in the request
## Description

If you use auto mode in a session to let TUI automatically select a model, and then restart TUI to resume the previous session for dialogue, it will return `Warn Invalid request (400): {"error":{"message":"The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed auto.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}`

## Steps to reproduce

1. open deepseek-TUI
2. use `/model auto` change to auto select mode
3. use `/exit` quit TUI
4. reopen deepseek and have a conversation

## Expected behavior

api can return normal response

## Actual behavior

api return invalid request warning

## Environment

- OS: Win 10
- DeepSeek CLI version: 0.8.38
- Install method: npm
- Model/provider: deepseek
- Terminal app: Windows Terminal
- Shell: powershell

LING71671:
## Summary

Closes #1797.

This fixes resumed sessions that were saved while `/model auto` was active. Before this change, loading such a session restored `app.model` to the literal string `auto` but left `app.auto_model` false in some resume/load paths. The next turn could then treat `auto` as a fixed provider model and send it directly to the DeepSeek API, which rejects it with HTTP 400.

## Changes

- Added `App::restore_model_selection` so session restore paths rebuild the full model-mode state from persisted metadata.
- Updated both `/load` and startup/session-picker resume handling to use that helper.
- When the restored model is `auto`, the app now:
  - enables `auto_model`
  - clears stale effective-model labels
  - restores auto reasoning mode
  - recalculates the compaction budget against the default concrete budget model
- Added regression coverage for both explicit `/load` and `apply_loaded_session` resume paths.

## Verification

- `git diff --check`
- `cargo test -p deepseek-tui --bin deepseek-tui load_auto_model_session_restores_auto_mode`
- `cargo test -p deepseek-tui --bin deepseek-tui apply_loaded_session_restores_auto_model_mode`
- `cargo build`

Notes: build/clippy still report pre-existing warnings in `schema_migration.rs` for unused migration marker structs, and clippy reports two pre-existing `needless_return` warnings in `tui/ui.rs`. No new warnings were introduced by this change.

### DIFF
--- a/crates/tui/src/commands/session.rs
+++ b/crates/tui/src/commands/session.rs
@@ -94,8 +94,7 @@ pub fn load(app: &mut App, path: Option<&str>) -> CommandResult {
     app.extend_history(cells_to_add);
     app.mark_history_updated();
     app.viewport.transcript_selection.clear();
-    app.model.clone_from(&session.metadata.model);
-    app.update_model_compaction_budget();
+    app.restore_model_selection(session.metadata.model.clone());
     app.workspace.clone_from(&session.metadata.workspace);
     app.session.total_tokens = u32::try_from(session.metadata.total_tokens).unwrap_or(u32::MAX);
     app.session.total_conversation_tokens = app.session.total_tokens;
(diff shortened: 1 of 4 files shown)


LING71671:
Superseded by #1885, which carries the same #1797 fix forward on top of current main with the remaining /load path covered and fresh regression assertions. Closing this older dirty branch to keep the queue clearer.

PR #1807, opened 2026-05-19
fix(tui): expose subagent full transcript handle

### LINKED ISSUE #1738: sub-agent deep-analysis results are lossy: parent only gets compressed summary, cannot retrieve full child output after completion
## Problem

When a sub-agent (`agent_open`) is tasked with reading and analyzing multiple sources (e.g., 14 GitHub issues), the parent only receives a compressed summary of the child's findings. The full structured analysis is lost once the agent completes:

1. The completion event and `agent_eval` return only a brief summary — not the detailed per-item breakdown the parent needs.
2. Calling `agent_eval` on a completed session returns `"not running"` — there is no way to page back through the child's full output.
3. The parent must re-do all the work itself, defeating the purpose of delegation.

## Reproduction

1. Open a sub-agent: `agent_open` with a task like "read issues #A, #B, #C and return structured per-issue analysis"
2. Wait for completion
3. Inspect the completion summary — it will be too compressed to act on
4. Try `agent_eval` on the completed agent id → `"not running"`, no way to recover the full child output

## Expected

- Agent completion should return a handle or projection that allows the parent to retrieve bounded slices of the child's full output (similar to `handle_read` for RLM sessions).
- Alternatively, `agent_eval` on a completed session should return the full transcript/projection instead of "not running".
- The summary-first pattern is good for status, but there must be a way to drill into details when the compressed summary isn't enough.

## Impact

In the v0.8.40 bug-bash session, this caused:
- ~80K tokens of wasted context re-reading 14 issues that the child had already analyzed
- Lost confidence in the delegation pattern for multi-source analysis tasks
- Delayed triage while the parent re-did the child's work

The outcome was: I told the parent "I analyzed all 14 issues" but the parent could not use the analysis and had to start over.

LING71671:
## Summary

Closes #1738.

This PR keeps the existing compressed sub-agent `result` summary, but adds a bounded retrieval path for the full child transcript:

- adds `full_transcript_handle` to `SubAgentResult` and `SubAgentSessionProjection`
- stores the complete sub-agent `messages` vector in the existing per-session `handle_store` as `agent:<id>/full_transcript`
- exposes that handle from `agent_eval` projections, nested snapshots, the lightweight `transcript_handle` snapshot payload, and the completion sentinel
- keeps historical/persisted sub-agent records backward-compatible by defaulting the handle to `None`

## Root Cause

`run_subagent` accumulated every child message in `messages`, but terminal results only retained the last non-empty text block in `result`. Once a child completed, the parent could only see a short summary and had no handle-backed way to retrieve intermediate analysis or tool-result context.

## Validation

Passed:

- `git diff --check`
- `cargo test -p deepseek-tui --bin deepseek-tui full_transcript_handle_exposes_subagent_messages`
- `cargo test -p deepseek-tui --bin deepseek-tui session_projection_surfaces_full_transcript_handle`
- `cargo test -p deepseek-tui --bin deepseek-tui subagent_done_sentinel_includes_full_transcript_handle_when_available`
- `cargo test -p deepseek-tui --bin deepseek-tui tools::subagent::tests::`
- `cargo build`

Notes:

- `cargo clippy` still reports pre-existing warnings in `schema_migration.rs` and two pre-existing `needless_return` warnings in `tui/ui.rs`.
- `cargo test --workspace --all-features` fails locally in 4 environment-sensitive tests because this machine's global skills and `.deepseek/instructions.md` are discovered during tests. The failing tests are unrelated to this change: `commands::skills::tests::test_list_skills_empty_directory`, `commands::skills::tests::test_list_skills_filters_by_name_prefix`, `project_context::tests::test_load_global_agents_when_project_has_no_context`, and `project_context::tests::test_global_agents_only_no_project_unchanged_fallback`.


LING71671:
Thanks, this looks superseded by the current main branch now. The release line has a native full-transcript handle path via insert_subagent_full_transcript_handle, and subagent_session_projection now prefers the stored ull_transcript handle for completed sessions. Closing this older dirty branch to keep the v0.8.41 queue clearer.

PR #1808, opened 2026-05-19
fix(tui): make subagent api timeout configurable

### LINKED ISSUE #1806: Subject: Sub-agent 120s API timeout renders agent_open nearly unusable (v0.8.39)
I'm running DeepSeek TUI v0.8.39 on Windows. I attempted to use parallel sub-agents (agent_open) to convert a 280-line Chinese biobanking standard into SOP documents — exactly the kind of parallel task offload that sub-agents are advertised for. All 5 sub-agents failed identically:

    status: Failed: "API call timed out after 120s"
    duration_ms: 731,203   (actual runtime: 12+ minutes)
    steps_taken: 0
    result: null

What's happening: The sub-agent does the work — it runs for 12 minutes — but the DeepSeek API terminates the response stream at 120s. Since sub-agents have allowed_tools: [] (no write access) and must return all output through the API response body, any task requiring more than ~30 seconds of generation is silently killed. The sub-agent has no way to checkpoint intermediate results or write them to disk.

Reproduction (trivial): Launch any sub-agent with a prompt that requires generating more than a few hundred words of output. It will time out. Lightweight tasks (read a file, return one line — ~30s) succeed. Anything substantive fails.

Impact: Sub-agents in their current form are limited to parallel read-only reconnaissance — grep, count, one-line summaries. They cannot be used for document generation, multi-step reasoning, code refactoring, or any task where output exceeds what fits in 120s of generation. This defeats the primary value proposition of sub-agent task offload.

Comparison with other tools in the same TUI:

    Tool                 Timeout                         Configurable    Can write files
    ─────────────────────────────────────────────────────────────────────────────────
    task_shell_start     user-defined (up to 600s)       Yes             Yes
    rlm_eval sub-query   configurable via rlm_configure  Yes             Yes (Python)
    agent_open           hardcoded 120s                  No              No

What's needed:

1. Configurable sub-agent API timeout — expose in config.toml under [subagents] (suggested: default 600s, max 1800s). This is the minimum fix.

2. Write-tool access for sub-agents — let sub-agents use write_file so large results can be persisted to disk instead of squeezed through the API response pipe. Combined with the parent reading the file, this bypasses the response-size constraint entirely.

3. Streaming or intermediate checkpoint support — sub-agents should be able to report partial results before the final response, or the parent should be able to poll intermediate state.

Without at least #1, sub-agents are a demo feature, not a production tool. With a 120s ceiling, they add parallelism overhead (spawn + context transfer) to tasks that are faster to run sequentially in the parent session.

Summary: The current agent_open implementation is bottlenecked by a server-side 120s API response limit with no local escape hatch. This makes sub-agents useless for any real workload. Please prioritize exposing the timeout configuration and granting sub-agents write-tool access.

LING71671:
## Summary

Closes #1806.

This PR makes the sub-agent per-step model API timeout configurable without changing the legacy default:

- adds `[subagents] api_timeout_secs`
- keeps the default at 120 seconds for backwards compatibility
- caps configured values at 1800 seconds
- threads the resolved timeout through `EngineConfig` into `SubAgentRuntime`
- uses the configured timeout around each sub-agent `create_message` call
- documents the setting in `docs/CONFIGURATION.md` and `docs/SUBAGENTS.md`

## Scope

This intentionally implements the minimal timeout-control fix from the issue. It does not change sub-agent tool permissions, write access, streaming, or checkpoint behavior.

## Validation

Passed:

- `git diff --check`
- `cargo check -p deepseek-tui --bin deepseek-tui`
- `cargo test -p deepseek-tui --bin deepseek-tui subagent_api_timeout_defaults_to_legacy_120s`
- `cargo test -p deepseek-tui --bin deepseek-tui subagent_api_timeout_only_clamps_to_maximum`
- `cargo test -p deepseek-tui --bin deepseek-tui child_runtime_preserves_api_timeout`
- `cargo test -p deepseek-tui --bin deepseek-tui subagent_runtime_default_api_timeout_is_legacy_120s`
- `cargo test -p deepseek-tui --bin deepseek-tui tools::subagent::tests::`
- `cargo build`

Notes:

- `cargo build` / `cargo clippy` still report pre-existing warnings in `schema_migration.rs`.
- `cargo clippy` also still reports the pre-existing `needless_return` warnings in `tui/ui.rs`.

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -870,7 +870,8 @@ pub struct SubagentsConfig {
     /// setting. Clamped to [1, MAX_SUBAGENTS].
     #[serde(default)]
     pub max_concurrent: Option<usize>,
-    /// Per-step model API timeout for sub-agents, in seconds. Values above
+    /// Per-step model API timeout for sub-agents, in seconds. Lower values
+    /// are honored as configured; only values above
     /// MAX_SUBAGENT_API_TIMEOUT_SECS are clamped.
     #[serde(default)]
     pub api_timeout_secs: Option<u64>,
@@ -3835,7 +3836,7 @@ mod tests {
     }
 
     #[test]
-    fn subagent_api_timeout_only_clamps_to_maximum() {
+    fn subagent_api_timeout_honors_configured_values_and_clamps_only_maximum() {
         let low = Config {
             subagents: Some(SubagentsConfig {
                 api_timeout_secs: Some(0),
@@ -3843,6 +3844,8 @@ mod tests {
             }),
             ..Config::default()
         };
+        // Do not impose a floor: callers may intentionally configure an
+        // aggressive timeout for tests or fail-fast sessions.
         assert_eq!(low.subagent_api_timeout_secs(), 0);
 
         let configured = Config {
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -19,6 +19,8 @@ use crate::hooks::HooksConfig;
 
 pub const DEFAULT_MAX_SUBAGENTS: usize = 10;
 pub const MAX_SUBAGENTS: usize = 20;
+pub const DEFAULT_SUBAGENT_API_TIMEOUT_SECS: u64 = 120;
+pub const MAX_SUBAGENT_API_TIMEOUT_SECS: u64 = 1800;
 pub const DEFAULT_TEXT_MODEL: &str = "deepseek-v4-pro";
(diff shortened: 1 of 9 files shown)


Hmbown:
Thanks for the configurable subagent API timeout work. This was harvested into v0.8.40 release PR #1823 as commit ad122fd6f, and #1823 is now CI-green. Closing as superseded by the release branch; thank you for improving long-running subagent behavior.

LING71671:
Thanks for harvesting this into v0.8.40. Glad the subagent timeout work made it into the release branch.

PR #1809, opened 2026-05-19
feat(goal): auto-continue mode with 4 safety nets

### LINKED ISSUE #891: Support Codex /goal long-running task mode
## Problem

Codex now supports long-horizon / goal-style workflows where the agent can keep working toward a defined objective instead of stopping after a single response. This is useful for complex coding tasks such as refactoring, multi-file feature implementation, migration, test fixing, and iterative debugging.

However, DeepSeek-TUI currently appears to be optimized for normal single-turn or interactive chat workflows. It does not provide a first-class way to run, monitor, pause, resume, or clear a persistent long-running goal similar to Codex's `/goal` mode.

## Proposed solution

Add support for a Codex-style Goal / Long Task mode in DeepSeek-TUI.

Suggested behavior:

1. Add a command such as `/goal` or `/longtask`.
2. Allow the user to define:
   - the final goal
   - acceptance criteria
   - maximum iterations or token budget
   - stop conditions
   - test / verification command
3. Let the agent automatically loop through:
   - plan
   - act
   - run checks
   - review result
   - continue or stop
4. Show task progress in the TUI:
   - current step
   - current plan
   - files changed
   - test results
   - iteration count
   - token / cost estimate if available
5. Support lifecycle controls:
   - pause
   - resume
   - cancel
   - clear goal
6. Persist the goal state so the user can continue the task after restarting the TUI.

Example command:

```text
/goal Refactor the authentication module to use the new session API.
Acceptance criteria:
- All existing tests pass.
- Add or update tests for the new session flow.
- Do not change public API behavior unless necessary.
- Stop after 10 iterations or if the same error happens 3 times.

Allen091080:
## Summary

Adds goal-driven auto-continue to `/goal`: setting an objective keeps the
agent advancing turn-by-turn until either the model emits `GOAL_ACHIEVED`
on its own line or one of four safety nets fires (token budget, max
iterations, stuck on pending todos with no token movement, or idle
without tool calls). The first turn is **kickstarted automatically** —
no need to type a placeholder message after `/goal`.

Carved out from #1242 (by @lbcheng888) so the auto-continue mechanic can
land independently of the larger sub-agent / palette / UI bundle. Credit
to @lbcheng888 for the original design — same 4-safety-net shape, same
`/goal stop|resume` lifecycle, same continuation prompt structure.

Fixes #891.

## What's in

- **`/goal <text>`** — set objective, enable auto-continue, kickstart
  first turn via `CommandResult.action = SendMessage(...)`. Persists
  to `~/.deepseek/goals.v1.json` so a restart resumes the goal.
- **`/goal stop`** / **`/goal resume`** — pause / unpause auto-continue
  without clearing the objective.
- **`/goal done` / `clear` / `reset`** — full wipe (memory + disk).
- **4 safety nets**, each unit-tested in isolation:
  - `BudgetExceeded` — total conversation tokens ≥ `| budget: N`
  - `Stuck` — 3 consecutive turns with flat pending todos **AND** token
    delta under 200 (see Design notes)
  - `Idle` — 2 consecutive turns where the model spoke without calling
    any tool
  - `MaxIterations` — default 50, overridable via `[goal] max_iterations`
- **`TurnFailed`** stop reason — when the engine reports
  `TurnOutcomeStatus::Failed`, the chain stops with a distinct message
  rather than queueing more continuations against a broken backend.
- **`Interrupted`** stop reason — Esc / engine interruption stops the
  chain (user can `/goal resume` to pick back up).
- **Streak counters skipped** when the previous turn was interrupted
  or failed — so an unresponsive API doesn't pollute the Idle / Stuck
  heuristics.
- **System prompt addendum** — when auto-continue is active, the
  `## Current Session Goal` block gains a sentinel-contract paragraph
  instructing the model to emit `GOAL_ACHIEVED` on its own line when
  done.
- **Config**: new `[goal]` section with `auto_continue_default` (bool,
  defaults true) and `max_iterations` (u32, defaults 50).

## Design notes (deltas from #1242 / the original spec)

1. **`Stuck` uses a double condition** (flat-pending **AND** token-delta
   under 200) instead of #1242's flat-pending-only check. The single
   condition false-positives on slow-but-substantive long tasks where
   the pending count happens to stay flat while real work is happening.
   The double condition adds the orthogonal "model is barely generating"
   signal so we only fire on genuine spinning. Threshold of 200 tokens
   is chosen to ignore boilerplate (system prompt re-injection, brief
   acknowledgements) while catching empty turns.
2. **No `AppAction::EnqueueContinuation` variant** — uses the existing
   `App::queue_message()` mechanism (the same path `/steer` recovery
   and offline-queued messages take). The work order's intent ("don't
   hack the send-message path") is honoured: `SendMessage` is the
   interactive-submit path and remains untouched; continuations use
   the non-interactive enqueue path, which is the canonical way to
   inject system-driven messages.
3. **`/goal X` kickstarts the first turn** via `CommandResult.action =
   SendMessage(...)` — the work order didn't require this but in
   practice users hit "I set /goal and nothing happened" because the
   continuation hook only fires on `TurnComplete`. Kickstart is
   suppressed when `[goal] auto_continue_default = false` so the
   opt-out remains a true passive-tracking mode.
4. **Sentinel sentinel matcher is line-anchored**:
   `text.lines().any(|l| l.trim() == "GOAL_ACHIEVED")` — guards against
   mid-paragraph mentions like "I'll emit GOAL_ACHIEVED when done"
   prematurely stopping the chain (test:
   `goal_achieved_inline_does_not_false_positive`).
5. **`Achieved` clears the objective**; other stop reasons leave it
   in place so `/goal resume` can pick back up after the operator
   fixes the underlying issue (broken API, exhausted budget, etc.).

## Testing

- [x] `cargo test --workspace --all-features` — **3509 passed, 0 failed**
- [x] `cargo clippy --all-targets --all-features -p deepseek-tui -- -D warnings`
- [x] 14 spec-style tests for `decide_continuation` covering each
      safety net, sentinel detection (including the inline guard), the
      `Interrupted-takes-precedence-over-Failed` tie-break, and
      both `auto_continue=off`/`no objective` inactive paths.
- [x] 6 sub-command tests for the `/goal` lifecycle (set / show /
      clear / stop / resume / "stop with no goal").
- [x] 3 kickstart tests (action shape, suppression when
      `auto_continue_default=false`, prompt content).
- [x] 5 persistence tests in `goal_state::tests` (roundtrip,
      missing-file, corrupt-file, forward-compat with missing fields,
      clear).
- [x] 3 streak-update tests (idle increment, idle reset, progress reset).
- [x] Manual: `/goal <text>` on a local vLLM session shows
      "kickstarting first turn now", the model executes immediately,
      `[goal] continuation #N queued` cells appear after each turn,
      `/goal stop` cleanly pauses, `/goal resume` reactivates with
      streak counters reset.

## Out of scope (deliberate, for follow-up PRs)

This PR intentionally does **not** touch:
- Sub-agent parallel execution / shared cache / streaming fanout
  (#1242 part 1)
- `/recap` command (#1242 part 3)
- Todos panel UI (#1242 part 4)
- Hierarchical prompt-delegation rewrite (#1242 part 5)
- Palette / header / history refactors
- **System prompt sentinel injection on the engine-config rebuild
  path** — `PromptSessionContext.goal_auto_continue` is currently
  passed `true` only from the `ui.rs` rebuild path; the two engine.rs
  sites pass `false` with a comment explaining the carve-out. Sentinel
  instructions still reach the model via every continuation user
  message, so behaviour is correct; the cleanup is purely an
  internals consistency fix.
- **`EmptyTurn` safety net** — when `usage.output_tokens == 0`,
  immediately stop instead of waiting for the `Idle` threshold to
  catch it on the next turn. Useful for unresponsive backends but
  not required for this PR.
- **Localization update for the `/goal` palette tooltip** — current
  `CmdGoalDescription` strings (zh/ja/pt) still describe the v0.8.x
  passive bookkeeping behaviour. Follow-up should refresh all four
  locales to mention auto-continue + `stop`/`resume`.

## Checklist

- [x] Updated `docs/MODES.md` with a `## Goal Mode` section
- [x] Updated `CHANGELOG.md` with the behaviour change and opt-out
- [x] Updated `config.example.toml` with the `[goal]` section
- [x] Tests live alongside the code they cover
      (`commands::goal::tests`, `goal_state::tests`)
- [x] No changes to public CLI surface beyond `/goal stop|resume` and
      the `[goal]` config section
- [x] Persisted state under `~/.deepseek/goals.v1.json` is forward-
      compatible (`schema_version` field, every new field
      `#[serde(default)]`)
- [x] `DEEPSEEK_TUI_GOAL_PATH` env var override for hermetic tests
      (so the test suite doesn't pollute the user's real
      `~/.deepseek/goals.v1.json`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **`/goal` auto-continue (#891).** Setting `/goal <objective>` now keeps
+  the agent advancing turn-by-turn until either the model emits
+  `GOAL_ACHIEVED` on its own line or one of four safety nets fires
+  (token budget, max iterations, "stuck" on pending todos with no
+  token movement, or "idle" without tool calls). The first turn is
+  kickstarted automatically — no need to type a placeholder message
+  after `/goal`. New subcommands `/goal stop` and `/goal resume`
+  toggle auto-continue without clearing the objective. Active goals
+  are persisted to `~/.deepseek/goals.v1.json` so a restart resumes
+  the goal. See `docs/MODES.md → Goal Mode` for the full contract.
+
+### Changed
+
+- **`/goal <text>` is no longer passive bookkeeping.** It defaults to
+  enabling the auto-continue loop. Conservative users can restore the
+  previous behaviour with `[goal] auto_continue_default = false` in
+  `config.toml`; `/goal resume` then opts a single goal into the loop
+  on demand.
+
 ## [0.8.39] - 2026-05-17
 
 ### Fixed
--- a/config.example.toml
+++ b/config.example.toml
@@ -438,6 +438,30 @@ default_text_model = "deepseek-ai/deepseek-v4-pro"
 # include_summary = false
 
 # ─────────────────────────────────────────────────────────────────────────────────
+# /goal Auto-Continue (#891)
+# ─────────────────────────────────────────────────────────────────────────────────
+# Setting `/goal <objective>` enables an auto-continuation loop: after each
+# turn the TUI evaluates whether the model emitted the literal token
+# `GOAL_ACHIEVED` (or whether any of four safety nets fired) and, if not,
+# automatically pushes the next turn forward.
+#
+# Safety nets (always on):
+#   - BudgetExceeded — total conversation tokens met `/goal X | budget: N`
+#   - Stuck          — 3 consecutive turns with flat pending todos AND
+#                       token-usage delta under 200 (double condition)
+#   - Idle           — 2 consecutive turns where the model spoke without
+#                       calling any tool
+#   - MaxIterations  — total continuations crossed `max_iterations`
+#
+# Set `auto_continue_default = false` to revert to v0.8.x behaviour where
+# `/goal X` is passive bookkeeping only; use `/goal resume` to start the
+# loop on demand.
+#
+# [goal]
+# auto_continue_default = true   # /goal <text> auto-enables the loop
+# max_iterations = 50            # Hard ceiling per goal; 0 disables the cap
+
+# ─────────────────────────────────────────────────────────────────────────────────
 # Workspace Snapshots (#137)
 # ─────────────────────────────────────────────────────────────────────────────────
 # Each turn the TUI takes a `pre-turn:<seq>` and `post-turn:<seq>` snapshot of
--- a/crates/tui/src/commands/goal.rs
+++ b/crates/tui/src/commands/goal.rs
@@ -1,68 +1,475 @@
-//! /goal command — set a session objective with token budget and progress tracking.
+//! `/goal` command — set a session objective with token budget,
+//! progress tracking, and (since #891) goal-driven auto-continue.
+//!
+//! ## Lifecycle
+//!
+//! - `/goal <text>` — set objective AND enable auto-continue. Every
+//!   subsequent `TurnComplete` is intercepted: if the model didn't
+//!   emit the literal sentinel `GOAL_ACHIEVED` and none of the four
+//!   safety nets fired, a synthetic continuation user-message is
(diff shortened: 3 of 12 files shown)


Hmbown:

PR #1831, opened 2026-05-20
feat(tui): add "Terminal" theme

Sskift:
## Summary

This PR adds a new selectable theme **`terminal`** that lets the host terminal's color scheme show through the TUI completely (instead of painting any DeepSeek brand RGB), and fixes a pre-existing debug-build panic in the grayscale theme path that I tripped over while testing.

The two changes are in separate commits so either can be cherry-picked independently.

## Why

The existing `system` theme picks between two RGB themes (Whale dark / Whale Light) based on `COLORFGBG` / macOS appearance — that's helpful, but it still paints brand-colored RGB surfaces. Users with custom terminal palettes (Solarized, Nord, Gruvbox configured at the terminal level, transparent backgrounds, custom Ghostty / iTerm / Alacritty schemes, light-mode terminals that aren't quite the Whale Light beige…) had no way to make the TUI honor their terminal palette. The Whale Light surface (`#F6F8FB`) shows up as a near-white block over their carefully tuned theme.

The new `Terminal` theme paints **nothing**: every background is `Color::Reset` and most foregrounds are `Color::Reset` too, so the terminal's own bg/fg are what's visible. The few accents that semantically need a color (mode label, diff +/−, status_working) use ANSI named colors (`Color::Blue`, `Color::Red`, `Color::Yellow`, `Color::Cyan`, `Color::Green`) so they also inherit the user's terminal palette rather than DeepSeek brand RGB.

## Commit 1 — feat(tui): add "Terminal" theme

- New `TERMINAL_UI_THEME` const where every `*_bg` slot and most text slots are `Color::Reset`. Accents are ANSI named colors (Blue/Red/Yellow/Cyan).
- `ThemeId::Terminal` plumbed through `from_name`, `name`, `display_name`, `tagline`, `ui_theme`, `SELECTABLE_THEMES`. Aliases registered in `normalize_theme_name`: `term`, `transparent`, `follow-terminal`, `inherit`.
- `theme_remap_active(Terminal) → true` so `ColorCompatBackend`'s per-cell remap rewrites every hard-coded palette constant (`DEEPSEEK_INK`, `DEEPSEEK_SLATE`, `BORDER_COLOR`, `TEXT_BODY`, …) to `Color::Reset`. Without this, the dozens of render sites that reach for the named palette constants directly would still paint brand RGB.
- `theme_green`, `theme_red`, `theme_diff_added_bg`, `theme_diff_deleted_bg` get a `ThemeId::Terminal` arm so diff highlighting still reads as +/− but follows the user's terminal palette.
- Inserted as the second entry in `SELECTABLE_THEMES` (right after System) so it surfaces prominently in the `/theme` picker.

The default (`theme = "system"`) is unchanged. Existing users see no difference until they opt in via `/theme` or `theme = "terminal"` in `settings.toml`.

### Why `PaletteMode::Dark`?

`TERMINAL_UI_THEME.mode` is set to `PaletteMode::Dark` even though the theme is palette-neutral. This avoids the dark→light cell remap kicking in on top of the Terminal cell remap; the Terminal remap already normalizes everything to `Color::Reset`, and a second pass would either be a no-op or risk overwriting it. There's a comment to that effect on the const.

### Test fallout

Inserting `Terminal` in row 2 of `SELECTABLE_THEMES` shifts indices three existing tests rely on — `arrow_down_previews_next_theme`, `enter_commits_with_persist_true`, `digit_jumps_to_row` — so those expectations are updated to match the new ordering. No production behavior change in those tests, just index arithmetic.

## Commit 2 — fix(tui): widen `luma()` accumulator to u32

While testing the Terminal theme I tripped a separate, pre-existing bug. Opening `/theme` and previewing **Grayscale** on a cell carrying a bright RGB color in a debug build panics:

```
thread '...' panicked at crates/tui/src/palette.rs:1141:7:
attempt to add with overflow
```

Path: `adapt_*_for_grayscale_palette → grayscale_*_from_luma → luma()`. The accumulator was `u16`, but `r*299 + g*587 + b*114` reaches `255 * 1000 = 255_000` for white — far above `u16::MAX = 65_535`. Release builds wrap silently (so on `--release` the function silently returned a wrong luma instead of panicking, which is its own correctness issue).

Fix: hold the weighted sum in `u32`. The result still divides by 1000 and casts back to `u8`, so the public contract is bit-identical for all non-overflowing inputs and now correct for inputs that previously overflowed.

No test added because `luma` is private and any RGB above the previous overflow threshold trivially exercises it through the existing grayscale code paths — the existing grayscale tests pass with the new arithmetic.

## Backwards compatibility

- Default theme unchanged.
- Existing settings strings (`system`, `dark`, `light`, `grayscale`, etc.) all still parse to the same `ThemeId` they did before.
- All existing `UiTheme` consts and palette functions are unchanged.
- The `ThemeId` enum gained one variant; any external matcher would need an arm, but `ThemeId` is internal to this crate.
- Tests: `cargo test -p deepseek-tui --bin deepseek-tui` → 3149 passed / 0 failed / 3 ignored locally on macOS Apple Silicon, debug.

## How to try it

```toml
# ~/.config/deepseek/settings.toml
theme = "terminal"
```

Or open the TUI and run `/theme`, then arrow-key down to **Terminal** (second row).

## Files changed

- `crates/tui/src/palette.rs` — new theme, enum/picker plumbing, `luma` overflow fix
- `crates/tui/src/tui/theme_picker.rs` — index updates in three tests for the new selector ordering

### DIFF
--- a/crates/tui/src/palette.rs
+++ b/crates/tui/src/palette.rs
@@ -510,8 +510,13 @@ pub const TERMINAL_UI_THEME: UiTheme = UiTheme {
     footer_bg: Color::Reset,
     mode_agent: Color::Blue,
     mode_yolo: Color::Red,
-    mode_plan: Color::Yellow,
-    status_ready: Color::Reset,
+    // Magenta keeps Plan visually distinct from `status_warning` (yellow)
+    // so the mode indicator and warning chip don't collide on themes that
+    // render both in the status row.
+    mode_plan: Color::Magenta,
+    // DarkGray gives "Ready" a low-contrast but still distinguishable hue
+    // versus default body text (which is `Color::Reset` on this theme).
+    status_ready: Color::DarkGray,
     status_working: Color::Cyan,
     status_warning: Color::Yellow,
     text_dim: Color::Reset,
--- a/crates/tui/src/palette.rs
+++ b/crates/tui/src/palette.rs
@@ -1138,7 +1138,10 @@ fn grayscale_bg_from_luma(luma: u8) -> Color {
 }
 
 fn luma(r: u8, g: u8, b: u8) -> u8 {
-    (((u16::from(r) * 299) + (u16::from(g) * 587) + (u16::from(b) * 114)) / 1000) as u8
+    // Coefficients sum to 1000; an RGB triple of (255,255,255) yields
+    // 255 * 1000 = 255_000 which overflows u16 (max 65_535). Use u32 to
+    // hold the weighted sum, then divide back into the 0..=255 range.
+    (((u32::from(r) * 299) + (u32::from(g) * 587) + (u32::from(b) * 114)) / 1000) as u8
 }
 // === Color depth + brightness helpers (v0.6.6 UI redesign) ===
 
--- a/crates/tui/src/palette.rs
+++ b/crates/tui/src/palette.rs
@@ -489,6 +489,39 @@ pub const DRACULA_UI_THEME: UiTheme = UiTheme {
     border: Color::Rgb(0x44, 0x47, 0x5a),
 };
 
+/// "Terminal" theme: lets the host terminal's color scheme show through
+/// instead of painting any RGB surface. Backgrounds use `Color::Reset`
+/// (the terminal's own default bg) and most text uses `Color::Reset`
+/// (terminal's own default fg). Accents are ANSI named colors so they
+/// also inherit the user's terminal palette (Solarized, Nord, custom
+/// schemes, etc.) rather than DeepSeek brand RGB.
+pub const TERMINAL_UI_THEME: UiTheme = UiTheme {
+    name: "terminal",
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you for the Terminal theme work here, @Sskift. I opened a fresh current-main PR at #2276 that harvests this implementation for the v0.8.48 bugfix train, keeps the original authored commits, and adds the current palette-schema/test cleanup needed after the theme QA changes.

I left this PR untouched rather than force-pushing over your branch.

Hmbown:
Closing this as superseded by #2276, which preserves the Terminal theme work on current main with the follow-up palette/test cleanup for the v0.8.48 candidate queue. Thanks @Sskift for the original implementation; the credit is preserved in #2276 and the release-credit pass.

PR #1823, opened 2026-05-20
chore(release): prepare v0.8.40

Hmbown:
## v0.8.40 - Stability & Recovery

A focused stability release for the v0.8.38/v0.8.39 fallout: provider/model correctness, cross-platform terminal reliability, sub-agent/session robustness, runtime log hygiene, OCR for attached screenshots, and release infrastructure recovery.

### Changes since v0.8.39

- **WSL2/headless Linux**: clipboard initialization is deferred so X11-less machines can reach the first frame (#1772, #1773)
- **Windows alt-screen**: `RUST_LOG` tracing no longer leaks into the interactive TUI display (#1774, #1776)
- **OpenAI-compatible providers**: explicit custom model names are preserved for non-DeepSeek providers (#1714, #1740)
- **Wanjie Ark**: first-class provider support across config, CLI auth/status, provider picker, doctor hints, env vars, and docs
- **Reasoning replay**: DeepSeek reasoning models selected through OpenAI-compatible endpoints replay and classify `reasoning_content` consistently (#1694, #1739, #1743)
- **Thinking-only turns**: clean no-final-answer turns now surface a status instead of silently ending (#1727, #1742)
- **Windows shell quoting**: `cmd /C` preserves quoted payloads such as `git commit -m "..."` (#1691, #1744)
- **Composer recovery**: Home/End are line-local in multiline drafts, Ctrl+C restores canceled prompts, and leaked mouse reports no longer erase draft suffixes (#1748, #1749, #1757, #1764, #1778)
- **Compaction recovery**: oversized cache-aligned summary attempts retry through the bounded summary path instead of failing with a 400
- **Sub-agents**: full transcript handles are exposed, per-step API timeout is configurable, and write-capable roles can use Suggest-level file edit tools without bypassing Required approval gates (#1738, #1806, #1808, #1828, #1833)
- **Fin/tool-agent experiment**: `tool_agent` opens a bounded execution agent on DeepSeek V4 Flash with thinking forced off, so tool-bound OCR/search/probe work can stay out of the main model's reasoning context while still using the existing mailbox, lifecycle, and cost accounting
- **Screenshot OCR**: `image_ocr` and `read_file` now use native macOS Vision OCR when tesseract is absent, `doctor` reports that fallback, and pasted clipboard screenshots are trusted for read-only inspection
- **Auto-routing context**: hidden thinking blocks are excluded from recent-message routing context so the model does not mistake its own reasoning for user input
[...]

PR #1832, opened 2026-05-20
fix(palette): fix u16 overflow in luma() causing grayscale theme crash

mars-base:
## Summary

- `luma()` in `palette.rs` uses `u16` for intermediate arithmetic
- `u16::from(r) * 299` overflows `u16::MAX` (65535) when `r >= 220`
- `u16::from(g) * 587` overflows when `g >= 112`
- In debug builds, this causes a panic (`attempt to add with overflow`) whenever the grayscale theme processes a bright RGB color
- Fix: use `u32` for intermediate calculations (result still fits in `u8`)

## Reproduction

Switch to the Grayscale theme (`/theme` → select option 4) in a debug build — TUI immediately crashes with a panic logged to `~/.deepseek/crashes/`.

## Fix

```rust
// Before
fn luma(r: u8, g: u8, b: u8) -> u8 {
    (((u16::from(r) * 299) + (u16::from(g) * 587) + (u16::from(b) * 114)) / 1000) as u8
}

// After
fn luma(r: u8, g: u8, b: u8) -> u8 {
    (((u32::from(r) * 299) + (u32::from(g) * 587) + (u32::from(b) * 114)) / 1000) as u8
}
```

### DIFF
--- a/crates/tui/src/palette.rs
+++ b/crates/tui/src/palette.rs
@@ -1089,7 +1089,7 @@ fn grayscale_bg_from_luma(luma: u8) -> Color {
 }
 
 fn luma(r: u8, g: u8, b: u8) -> u8 {
-    (((u16::from(r) * 299) + (u16::from(g) * 587) + (u16::from(b) * 114)) / 1000) as u8
+    (((u32::from(r) * 299) + (u32::from(g) * 587) + (u32::from(b) * 114)) / 1000) as u8
 }
 // === Color depth + brightness helpers (v0.6.6 UI redesign) ===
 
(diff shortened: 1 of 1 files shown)


mars-base:
Thanks for the review! Both suggestions have been applied in the latest commit:
- Added `+500` before `/1000` for proper integer rounding
- Removed the redundant inner parentheses

Hmbown:
Thanks for the grayscale overflow patch. This was harvested into v0.8.40 release PR #1823 as commit ae1bb9dd9, and #1823 is now CI-green. Closing as superseded by the release branch; thank you for catching the crash path.

PR #1844, opened 2026-05-20
fix(tui): set spillover root in dedup tests for Nix sandbox builds

barstoolbluz:
## Summary

- Two tool-result deduplication tests (`request_builder_deduplicates_large_identical_tool_results_with_retrieval_hint` and `cache_inspect_reports_tool_result_budget_metadata`) fail under `nix build` because the sandbox has no writable `$HOME`
- The spillover write silently fails, dedup is skipped, and assertions break
- Adds a temp-dir-backed `SpilloverRootGuard` (matching the existing pattern in `tool_result_retrieval.rs`) so the tests pass in both sandboxed and normal environments

## Test plan

- [x] `cargo test -p deepseek-tui --bin deepseek-tui -- stream_decoder_tests::request_builder_deduplicates_large stream_decoder_tests::cache_inspect_reports_tool_result` passes
- [x] `nix build .#` succeeds (tests pass in sandbox)

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -2206,6 +2206,24 @@ mod stream_decoder_tests {
     //! directly to verify each "class of stream failure" the engine relies on.
     use super::*;
     use crate::models::{ContentBlockStart, Delta, StreamEvent};
+    use std::path::PathBuf;
+
+    struct SpilloverRootGuard {
+        prior: Option<PathBuf>,
+    }
+
+    impl SpilloverRootGuard {
+        fn new(path: PathBuf) -> Self {
+            let prior = crate::tools::truncate::set_test_spillover_root(Some(path));
+            Self { prior }
+        }
+    }
+
+    impl Drop for SpilloverRootGuard {
+        fn drop(&mut self) {
+            crate::tools::truncate::set_test_spillover_root(self.prior.take());
+        }
+    }
 
     /// Decode a raw SSE-data JSON chunk into our internal events, mirroring
     /// the per-event call shape used by `handle_chat_completion_stream`.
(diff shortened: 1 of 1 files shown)


Hmbown:
Independent review:

The actual change is the right pattern — a temp-dir-backed `SpilloverRootGuard` matching `tool_result_retrieval.rs`. Robust outside Nix: it locks `TEST_SPILLOVER_GUARD`, swaps the spillover root, and restores on Drop. Works in any sandbox lacking `$HOME`.

Concerns:
1. Branch is stale on "deepseek" naming — the actual reverts the CodeWhale rebrand in chat.rs error strings (`\`deepseek doctor\`` etc.) and removes the `tools` parameter from `PromptBuilder`. Rebase on current main; only the 34-line test-guard addition should remain.
2. After rebase, verify it stacks correctly on top of the new `requires_reasoning_content` + `provider_accepts_reasoning_content` split that landed since the branch was cut — the diff shows it conflicts with that refactor.
3. Test name `with_tool_result_sha_spillover_root` was deleted in favor of `SpilloverRootGuard` — confirm no other test still calls the old helper after rebase.

The fix itself is correct and the pattern is reusable for any future test that exercises `truncate::write_sha_spillover`. After rebase this should land cleanly.

v0.8.48 (PR #2256): 1 conflict in `crates/tui/src/client/chat.rs` — same file PR #2256 touches.

Hmbown:
Thanks @barstoolbluz — this is a nice small test-harness fix, and I like that it uses the same spillover-root guard pattern as the neighboring retrieval tests.

I tried to merge it today, but GitHub says the branch no longer merges cleanly with current `main`. A quick rebase should be enough here; after that, `cargo test -p codewhale-tui --bin codewhale-tui stream_decoder_tests::request_builder_deduplicates_large stream_decoder_tests::cache_inspect_reports_tool_result` is the main signal I’d want before landing it.

Hmbown:
Thanks @barstoolbluz -- this was a good Nix/build reliability fix. I’m closing this PR because the equivalent spillover-root isolation has already landed on main in `b7e31a73` (`test: isolate tool-result spillover tests`), including the same two dedup/cache-inspection test paths this PR was protecting.\n\nI’m keeping the credit trail explicit here because the original shape of this PR was useful; we just don’t need to merge the now-conflicting branch separately.

PR #1836, opened 2026-05-20
  feat(update): add version check output to doctor and update

reidliu41:
## Summary

  Adds a lightweight way to check whether the installed binary is current.

 Before this change, `deepseek update` could perform an update, but there was no check-only path, and `deepseek doctor` did not report whether a newer release was available. This made it harder for users to confirm whether local issues might already be fixed in a newer version.

  This PR adds:

  - `deepseek update --check` to report the current version and latest release without downloading or replacing binaries
  - an `Updates` section in `deepseek doctor`
  - version comparison that handles `v0.8.x`, `0.8.x`, and build suffixes like `0.8.x (commit)`
  - async release checking in doctor so diagnostics continue to run normally inside the TUI runtime

```
deepseek doctor
DeepSeek TUI Doctor
==================

Version Information:
  deepseek-tui: 0.8.39 (eeccf7de6c3b)
  rust: rustc 1.93.1 (01f6ddf75 2026-02-11)

Updates:
  · current: v0.8.39
  ✓ latest: v0.8.39
    Already up to date.
....



deepseek update --check
Checking for updates...
Current binary: /home/reid/github/DeepSeek-TUI/target/debug/deepseek
Current version: v0.8.39
Latest release: v0.8.39
Already up to date.

```

## Testing

- [x] `cargo test --all-features`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features`

## Checklist

- [ ] Updated docs or comments as needed
- [x] Added or updated tests where relevant
- [ ] Verified TUI behavior manually if UI changes

PR #1848, opened 2026-05-20
feat: ShellDispatcher, ExternalTool wrappers, and pluggable tool registry

aboimpinto:
This PR folds the previously stacked work into a single PR so it can be reviewed and merged as one coherent change.

Supersedes #2008 and #2031.

## Summary

- Adds ShellDispatcher for shell-agnostic command execution.
- Adds ExternalTool wrappers for git, gh, and rustc command execution.
- Adds pluggable ToolRegistry support for runtime tool registration, config.toml overrides, and user script tools.

### No `[tools]` config required

The default `~/.deepseek/tools/` directory is always scanned. A user can drop a script there with frontmatter and it becomes available as a plugin tool. The `[tools]` table in `config.toml` is only needed for custom plugin directory paths, built-in tool overrides, or disabling tools.

### Script tool flow

1. A script such as `~/.deepseek/tools/count-words.mjs` declares metadata in comment frontmatter: name, description, JSON schema, and approval behavior.
2. On session start, the engine scans the tool directory, reads frontmatter, and registers a `ScriptPluginTool` in the ToolRegistry.
3. Plugin tools are non-deferred, so the model can see them immediately.
4. When called, the engine spawns the script, writes JSON input to stdin, reads a ToolResult JSON object from stdout, and returns that result to the model.

### Key design decisions

- stdin/stdout protocol: JSON input in, ToolResult JSON out.
- Comment frontmatter: supports `#`, `//`, and `--` prefixes so scripts can be written in multiple languages.
- Plugin tools are immediately visible to the model.
- Shebang resolution handles `#!/usr/bin/env node` by resolving `node` directly, which works better on Windows.

Paulo Aboim Pinto


aboimpinto:
Stacked on top of #2008 (ShellDispatcher) + ExternalTool wrappers. All 4 commits applied cleanly with conflict resolution. The full stack: main ← #2008 ← ExternalTool ← #1848.

Hmbown:
Independent review:

Big PR (+2576/-533) with three legitimate ideas — `ShellDispatcher`, `ExternalTool` trait, and pluggable tool registry — bundled together. Architecturally reasonable but the security model needs hardening before merge, and the branch is stale.

Security concerns (pluggable tools):

1. **Auto-discovery from `~/.deepseek/tools/` happens at registry build with no opt-in.** Any script dropped there is auto-registered as a model-callable tool with capabilities `[ExecutesCode, RequiresApproval]`. The frontmatter `# approval: auto` lets a script self-declare it doesn't need approval — a malicious script written by the model itself can therefore become silently auto-executable on the next session. Approval policy must come from config, not from the script's own frontmatter.

2. **`CommandPluginTool` runs arbitrary shell binaries from config.toml.** On Windows, the `!self.command.contains('.')` heuristic routes to `cmd /c <command>` — direct shell injection vector if `command` contains a metacharacter. Use `cmd /c call` with separate args, or refuse commands containing shell metacharacters.

3. **No sandboxing for plugin children.** `run_plugin_child_raw` spawns with full ambient privileges and no Landlock/Seatbelt wrapping — a plugin can read/write anywhere the user can, despite the parent process running under `sandbox_mode = "workspace-write"`. At minimum, route plugin execution through the existing `sandbox::` infrastructure.

4. **Shebang parser reads 256 bytes without validating UTF-8 boundaries** (`String::from_utf8_lossy`) — `bin_name == "env"` check fires after `rsplit('/')`, which on Windows paths with backslashes would not split correctly. Add `rsplit(['/', '\\\\'])`.

5. **120s plugin timeout is hardcoded** — no way to lower per-tool in config. Acceptable for v1 but document the limit.

6. Plugin name collisions silently overwrite built-in tools (`load_plugins` calls `self.register` which overwrites by name). A script `# name: read_file` replaces the built-in — surprising. Either prefix plugin names (`plugin:foo`) or warn loudly on collision.

Architectural / merge:
- Branch reverts CodeWhale rebrand, deletes `process_hardening` (#2183), `theme_qa_audit`, `tool_output_receipts` — same stale-base problem as #1836.
- 5 merge conflicts vs main: `core/engine.rs`, `core/engine/turn_loop.rs`, `main.rs`, `runtime_threads.rs`, `tui/ui.rs`. The `turn_loop.rs` change (−191 lines) is large and overlaps with active work.
- Split into 3 PRs: (a) `ShellDispatcher` alone, (b) `ExternalTool` trait + concrete impls, (c) pluggable registry. Each can land/be reviewed on its own merits.

v0.8.48 (PR #2256): 5 conflicts — overlaps `engine.rs`, `engine/turn_loop.rs`, and `main.rs` directly.

aboimpinto:
This PR is closed, and I will make smaller PRs for this feature.

PR #1856, opened 2026-05-21
fix(tools): replace cross-await RwLock with Semaphore to prevent deadlock

Fire-dtx:
Replace `Arc<RwLock<()>>` in `ToolCallRuntime` with `Arc<Semaphore>` to eliminate the risk of a tool re-entering and deadlocking on the same RwLock, and to prevent serial tools from blocking parallel tools' read-lock acquisition.

- Serial tools acquire and hold the permit for the full execution duration, ensuring mutual exclusion.
- Parallel tools acquire then immediately drop the permit, only blocking until any in-flight serial tool completes, then run concurrently.

## Testing

## Checklist

### DIFF
--- a/crates/tools/src/lib.rs
+++ b/crates/tools/src/lib.rs
@@ -8,7 +8,7 @@ use async_trait::async_trait;
 use codewhale_protocol::{ToolKind, ToolOutput, ToolPayload};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use tokio::sync::RwLock;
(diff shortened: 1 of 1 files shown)


Hmbown:
Harvested in build/v0.8.47 (commit a06bbe57a). The Semaphore-based serialisation landed cleanly — all tests pass (tools crate: 5 tests, workspace: 3405+ tests). Thank you @Fire-dtx! Added to CHANGELOG under Fixed + Community credits.

PR #1866, opened 2026-05-21
[codex] Fix composer arrows with mouse capture

### LINKED ISSUE #1720: Empty composer: Up/Down scrolls transcript instead of navigating history on Windows
## Problem

On Windows, pressing **Up** on an **empty** composer scrolls the transcript instead of navigating to the previous submitted message. Typing even a single character and pressing Up correctly navigates input history — but an empty composer does something completely different.

This is confusing and inconsistent. The user expectation is: Up on an empty composer = recall my last prompt.

## Root cause

`default_composer_arrows_scroll_for_platform` in `app.rs` returns `is_windows || !use_mouse_capture`. This forces `composer_arrows_scroll = true` on **all** Windows machines, even on modern terminals like Windows Terminal that fully support mouse capture.

The `composer_arrows_scroll` feature was introduced in #1211 for terminals that map mouse-wheel events to arrow keys (e.g. CMD.exe without `WT_SESSION`). When `composer_arrows_scroll = true` and the composer is empty, Up/Down scrolls the transcript rather than navigating history.

But the blanket Windows override (`is_windows ||`) applies this workaround to terminals that don't need it.

## Expected behavior

- **Mouse capture ON** (Windows Terminal, most terminals): Up/Down on empty composer navigates input history
- **Mouse capture OFF** (CMD.exe, terminals without mouse support): Up/Down scrolls transcript (mouse-wheel-as-arrow fallback)

## Proposed fix

Change the default to `!use_mouse_capture` only, removing the `is_windows ||` platform check:

```rust
fn default_composer_arrows_scroll_for_platform(use_mouse_capture: bool, _is_windows: bool) -> bool {
    !use_mouse_capture
}
```

Users who prefer the old scroll-on-empty behavior can explicitly set `composer_arrows_scroll = true` in their `config.toml`.

## Related

- `composer_arrows_scroll` introduced in #1211
- Windows mouse-wheel fallback added in #1443, PR #1471
- Config docs state default is `false`

Lee-take:
## Summary
- Default composer arrow scrolling now stays off when mouse capture is enabled.
- Keeps the existing opt-in behavior for configurations without mouse capture.
- Updates regression tests for both app config and UI config loading.

Fixes #1720

## Validation
- cargo fmt --check
- cargo test -p deepseek-tui composer_arrows_scroll_default -- --nocapture

### DIFF
--- a/crates/tui/src/tui/app.rs
+++ b/crates/tui/src/tui/app.rs
@@ -1346,8 +1346,8 @@ fn default_composer_arrows_scroll(use_mouse_capture: bool) -> bool {
     default_composer_arrows_scroll_for_platform(use_mouse_capture, cfg!(windows))
 }
 
-fn default_composer_arrows_scroll_for_platform(use_mouse_capture: bool, is_windows: bool) -> bool {
-    is_windows || !use_mouse_capture
+fn default_composer_arrows_scroll_for_platform(use_mouse_capture: bool, _is_windows: bool) -> bool {
+    !use_mouse_capture
 }
 
 impl App {
@@ -4411,8 +4411,8 @@ mod tests {
     }
 
     #[test]
-    fn composer_arrows_scroll_default_is_true_on_windows_even_with_mouse_capture() {
-        assert!(default_composer_arrows_scroll_for_platform(true, true));
+    fn composer_arrows_scroll_default_is_false_on_windows_with_mouse_capture() {
+        assert!(!default_composer_arrows_scroll_for_platform(true, true));
     }
 
     #[test]
--- a/crates/tui/src/tui/ui/tests.rs
+++ b/crates/tui/src/tui/ui/tests.rs
@@ -5729,8 +5729,8 @@ fn composer_arrows_scroll_defaults_follow_platform_with_mouse_capture() {
         ..create_test_options()
     };
     let app = App::new(options, &Config::default());
-    assert_eq!(
-        app.composer_arrows_scroll, false,
+    assert!(
+        !app.composer_arrows_scroll,
         "arrows-scroll should default to false when mouse capture is on"
     );
 }
--- a/crates/tui/src/tui/ui/tests.rs
+++ b/crates/tui/src/tui/ui/tests.rs
@@ -5730,9 +5730,8 @@ fn composer_arrows_scroll_defaults_follow_platform_with_mouse_capture() {
     };
     let app = App::new(options, &Config::default());
     assert_eq!(
-        app.composer_arrows_scroll,
-        cfg!(windows),
-        "arrows-scroll should default to true on Windows and false on other platforms when mouse capture is on"
+        app.composer_arrows_scroll, false,
+        "arrows-scroll should default to false when mouse capture is on"
     );
 }
 


Hmbown (commented):
Thanks for this fix — the behavior change looks right for mouse-capture mode. The new test assertion will fail the workspace clippy gate (`-D warnings`) on `clippy::bool_assert_comparison`:

```
crates/tui/src/tui/ui/tests.rs:5732
```

Suggested one-line tweak:

```rust
assert!(
    !app.composer_arrows_scroll,
    "arrows-scroll should default to false when mouse capture is on"
);
```

Local repro (from a worktree of this PR):

```
cargo clippy -p deepseek-tui --all-targets --all-features --locked -- -D warnings
```

Once that's applied, behavior tests, fmt, and clippy all pass locally. Happy to push the fix if it's easier.

Hmbown:
Thanks for the Windows composer-arrow fix. I checked current main: default_composer_arrows_scroll_for_platform still returns is_windows || !use_mouse_capture, and issue #1720 remains open, so this is not already fully handled.

I am not merging it into v0.8.42 because Windows input/mouse-capture defaults need a focused pass and current PR CI only shows GitGuardian. Keeping it open for a Windows input maintenance pass with current crate-name tests and, ideally, the matrix from #1720 rechecked against Windows Terminal/conhost/VS Code integrated terminal.

Lee-take:
Rechecked the Windows input matrix on the current PR branch (`ac1ae026`).

Environment:
- Windows 11 Home 10.0.26200 / build 26200
- `cargo 1.95.0`, `rustc 1.95.0`
- Windows Terminal available via `wt.exe`; VS Code available via `code.cmd`

Local checks:
- `cargo fmt --check`
- `cargo clippy -p deepseek-tui --all-targets --all-features --locked -- -D warnings`
- `cargo test -p deepseek-tui mouse_capture --all-features --locked` — 20 related tests passed
- `cargo test -p deepseek-tui composer_arrows_scroll --all-features --locked` — 10 related tests passed

I also ran a small `crossterm` event probe in each terminal host. The probe logged terminal env markers and event types, then I sent Win32 wheel input to the focused terminal window.

| Host | Env marker | Probe mode | Wheel events seen by app | Up/Down key events from wheel | PR/default behavior this supports |
| --- | --- | --- | ---: | ---: | --- |
| Windows Terminal | `WT_SESSION` present | mouse capture on | 9 `Mouse(ScrollUp/ScrollDown)` | 0 | default mouse capture is on; with this PR `composer_arrows_scroll` defaults false, so plain Up/Down can navigate history while wheel is handled as mouse scroll |
| Windows Terminal | `WT_SESSION` present | mouse capture off | 0 | 0 | explicit capture-off keeps `composer_arrows_scroll = true`, preserving arrow-scroll behavior |
| `cmd.exe` / conhost | no `WT_SESSION`, no `ConEmuPID`, no `TERM_PROGRAM` | mouse capture off | 0 | 0 in this synthetic probe | default mouse capture remains off; with this PR `composer_arrows_scroll = true`, preserving the arrow-scroll fallback |
| `cmd.exe` / conhost | no `WT_SESSION`, no `ConEmuPID`, no `TERM_PROGRAM` | mouse capture on | 9 `Mouse(ScrollUp/ScrollDown)` | 0 | explicit capture can receive wheel as mouse events, but the current default still keeps capture off for the conhost leak-risk path |
| VS Code integrated terminal | `TERM_PROGRAM=vscode` | mouse capture off | 0 | 0 in this synthetic probe | default mouse capture remains off; with this PR `composer_arrows_scroll = true`, preserving the arrow-scroll fallback |
| VS Code integrated terminal | `TERM_PROGRAM=vscode` | mouse capture on | 9 `Mouse(ScrollUp/ScrollDown)` | 0 | explicit capture can receive wheel as mouse events |

Caveat: the wheel rows used synthesized Win32 wheel input, not a physical mouse wheel, so I would treat them as terminal event-routing evidence rather than a full manual UX pass. The #1720 behavior itself is covered by the unit tests above: on Windows, `composer_arrows_scroll` now defaults to false when mouse capture is on and remains true when mouse capture is off.

Hmbown:
Thanks for the focused fix here. I harvested the same behavior into a fresh maintainer branch with current tests and CI as #2103, and credited this PR plus the Windows terminal matrix in the new PR body. The landed changelog-level behavior is: Windows Terminal / mouse-capture sessions now keep Up/Down history recall, while no-mouse-capture terminals keep the transcript-scroll fallback.

Hmbown:
This fix is already in main (commit b3c50e0c9). The `_is_windows` parameter is ignored and composer arrows now default to off whenever mouse capture is enabled, regardless of platform. Closing — thank you @Lee-take!

Issue #1871, opened 2026-05-21
QoL: taskbar progress, animated title spinner, and configurable completion sound


aboimpinto:
# Quality of Life: Taskbar progress, title animation, and completion sound

Three small QoL features that give visual and audible feedback during/after processing, so the user can alt-tab away and still know when work is done.

## Feature 1: Taskbar progress indicator

While the model is thinking or executing tools, Windows Terminal shows a green indeterminate progress bar on the taskbar icon via `OSC 9 ; 4 ; 1`. The bar is cleared on completion via `OSC 9 ; 4 ; 0`.

## Feature 2: Animated terminal title

Cycles the terminal tab title between 🐳 and 🐋 every 800ms during processing (matching the whale status indicator in the TUI header). Shows `✅ DeepSeek TUI` on completion (only in `beep` mode). Clears on next key press or mouse click.

## Feature 3: Completion sound

Plays a short sound when every turn finishes, regardless of duration (unlike the existing `[notifications]` system which only fires for long turns exceeding a threshold).

Configurable via `[notifications] completion_sound` in `config.toml`:

| Mode | Behavior | Visual indicator |
|------|----------|-----------------|
| `beep` (default) | System notification sound (`MessageBeep` on Windows, BEL on other platforms) | ✅ title marker shown |
| `bell` | Raw terminal BEL character (`\x07`) | No ✅ title marker (bell is its own terminal-level indicator) |
| `off` | Silent | No ✅ title marker |

## Implementation

- `crates/tui/src/tui/notifications.rs` — `set_taskbar_progress()`, `start_title_animation()`, `stop_title_animation()`, `play_completion_sound()`, `CompletionSound` enum handling
- `crates/tui/src/core/engine/turn_loop.rs` — hook at turn start
- `crates/tui/src/tui/ui.rs` — hook at turn end + user interaction (clears ✅ marker)
- `crates/tui/src/config.rs` — `CompletionSound` enum + config field

aboimpinto:
Implemented in #1872

mcraveiro:
It would be nice to be able to associate different sounds to different lifecycle events like you can in Claude. I  use Claude sound FX [1] quite heavily. Seems like Open Code also supports this. For Claude, it is based on hooks [2]. Is this something you support or intend to support in the future?

[1] https://github.com/6m1w/claude-sound-fx
[2] https://code.claude.com/docs/en/hooks-guide

aboimpinto:
[...]

PR #1872, opened 2026-05-21
feat: QoL — taskbar progress, animated title spinner, and configurable completion sound (#1871)

aboimpinto:
## Quality of Life: Taskbar progress, title animation, and completion sound

Three small QoL features that give visual and audible feedback during/after processing, so the user can alt-tab away and still know when work is done.

### Feature 1: Taskbar progress indicator

While the model is thinking or executing tools, Windows Terminal shows a green indeterminate progress bar on the taskbar icon via `OSC 9 ; 4 ; 1`. The bar is cleared on completion via `OSC 9 ; 4 ; 0`. This is the same mechanism used by file downloads and other long-running operations in Windows Terminal.

### Feature 2: Animated terminal title

Cycles the terminal tab title between 🐳 and 🐋 every 800ms during processing, matching the whale status indicator already shown in the TUI header. On completion, shows `✅ DeepSeek TUI` (only in `beep` mode). The marker clears on the next key press or mouse click, so it doesn't persist once the user is back at the terminal.

### Feature 3: Completion sound

Plays a short sound when **every** turn finishes, regardless of duration. This is separate from the existing `[notifications]` system, which only fires for long turns exceeding the configured threshold (default 30s). Different sound modes in config:

| Mode | Behavior | ✅ Title marker | Use case |
|------|----------|-----------------|----------|
| `beep` (default) | System notification sound (`MessageBeep` on Windows, `BEL` elsewhere) | ✅ Shown | System sound has no visual component, so ✅ provides it |
| `bell` | Raw terminal BEL (`\x07`) | ❌ Hidden | Bell already has its own terminal-level indicator (flash/icon), so ✅ would be redundant |
| `off` | Silent | ❌ Hidden | No feedback needed |

The logic: `beep` emits a system sound that has no visual component (you hear it but can't see it in the taskbar), so the ✅ title marker provides the visual half. `bell` uses the terminal bell which already flashes/indicates visually — adding ✅ on top would be overkill.

### Implementation

- `crates/tui/src/tui/notifications.rs` — core functions: `set_taskbar_progress()`, `start_title_animation()`, `stop_title_animation()`, `play_completion_sound()`, `CompletionSound` enum wiring
- `crates/tui/src/core/engine/turn_loop.rs` — hook at turn start (triggers progress bar + title animation)
- `crates/tui/src/tui/ui.rs` — hook at turn end (clears progress bar + title + plays sound) + on user interaction (clears ✅ marker)
- `crates/tui/src/config.rs` — `CompletionSound` enum + `completion_sound` field in `NotificationsConfig`

### Config

To change the completion sound, add to `~/.deepseek/config.toml`:

```toml
[notifications]
completion_sound = "bell"
```

### Files changed

4 files, +177/-1, 1 squashed commit.

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -516,6 +516,19 @@ fn default_threshold_secs() -> u64 {
     30
 }
 
+/// Completion sound options.
+#[derive(Debug, Clone, Copy, Deserialize, Default, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum CompletionSound {
+    /// No sound on turn completion.
+    Off,
+    /// System notification beep (default). On Windows uses `MessageBeep`.
+    #[default]
+    Beep,
+    /// Terminal BEL character (`\x07`).
+    Bell,
+}
+
 /// Desktop-notification configuration (OSC 9 / BEL on turn completion).
 #[derive(Debug, Clone, Deserialize, Default)]
 pub struct NotificationsConfig {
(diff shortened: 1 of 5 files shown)


aboimpinto:
Mergeable against current \main\ (v0.8.44). No rebase needed.

Hmbown:
thank you - this is wonderful!

PR #1905, opened 2026-05-21
fix(tui): suppress verbose CLI logging on Windows alt-screen to prevent TUI leak

aboimpinto:
On Windows stderr cannot be redirected to the log file (no \`dup2\`), so any \`eprintln!\` calls from \`crate::logging\` emit directly into the terminal even inside the alt-screen buffer, corrupting the display.


aboimpinto:
Superseded by PR #1910 which includes the verbosity-restore on cleanup and UTF-8 encoding fix.

PR #1935, opened 2026-05-23
feat(tui): replace [x] completion marker with [✓] across all sidebar panels

IIzzaya:
## Summary
In tui mode, replace [x] completion marker with [✓] across all sidebar panels (more Chinese style lol)

image

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/commands/skills.rs
+++ b/crates/tui/src/commands/skills.rs
@@ -441,7 +441,7 @@ fn sync_skills(app: &mut App) -> CommandResult {
                     }
                     SkillSyncOutcome::Denied { name, host } => {
                         failed += 1;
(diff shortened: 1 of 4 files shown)


weikengchen:
:) this sounds like a config option thing as github uses [x] (in Markdown), but I think that is due to historical reason because x is ASCII and ✓ is not.

But ✓ does look better.

Hmbown:
Thanks for the visual polish PR. The idea is understandable, and the screenshot makes the intent clear, but I am leaving it out of v0.8.42 because this is a product/style decision rather than a stabilization fix.

I agree with the comment that this may want to be a design/config decision instead of a blanket replacement, especially because [x] is ASCII/Markdown-familiar while the check mark reads better visually. Routing this to a later presence/UI pass rather than closing it.

Hmbown:

PR #1910, opened 2026-05-21
fix(tui): suppress verbose CLI logging on Windows alt-screen to prevent TUI leak

aboimpinto:
## Summary

This is the **missing second commit** from PR #1776. The maintainer merged the first commit (stop `RUST_LOG` from leaking tracing messages) but closed PR #1776 without merging this follow-up fix.

## Problem

On Windows, `stderr` cannot be redirected to the log file (no `dup2`). When `--verbose` or `DEEPSEEK_LOG_LEVEL=debug` is set, `eprintln!` calls from `crate::logging` still leak into the TUI alt-screen buffer, corrupting the display � even after the `RUST_LOG` fix was merged.

## Fix

Call `crate::logging::set_verbose(false)` at two points:

1. **`run_tui()`** � right after `EnterAlternateScreen`, so verbose logging is suppressed for the entire TUI session on Windows.
2. **`resume_terminal()`** � right after re-entering alt-screen during terminal recovery, so the suppression holds across suspend/resume cycles.

## Changes

- `crates/tui/src/tui/ui.rs`: +9 lines � `set_verbose(false)` calls in `run_tui` and `resume_terminal`, gated by `#[cfg(windows)]`.

## Related

- Issue: #1774 (original bug report)
- PR #1776 (original PR � only first commit was merged)
- Closes #1909 (tracking issue for this missing fix)

Urgent: this is a regression on Windows for users who run with verbose/debug logging.

PR #1852, opened 2026-05-21
feat: add DEEPSEEK.md as project context file

EmiyaKiritsugu3:
Adds DEEPSEEK.md to PROJECT_CONTEXT_FILES so harnesses like [superpowers](https://github.com/obra/superpowers) can ship a harness-specific instruction file that gets auto-loaded as project_instructions — analogous to GEMINI.md in Gemini CLI.

## What

1 line in the PROJECT_CONTEXT_FILES constant, positioned after AGENTS.md so harness-specific files take priority over generic CLAUDE.md while universal AGENTS.md remains top.

## Files changed

- **`project_context.rs`**: Add `DEEPSEEK.md` to `PROJECT_CONTEXT_FILES` and module docstring
- **`working_set.rs`**: Add `DEEPSEEK.md` to key-file detection candidates
- **`utils.rs`**: Add `deepseek.md` to fallback key-file matcher

## Why

Superpowers needs a harness-specific bootstrap file to inject the `using-superpowers` skill at session start. AGENTS.md is already used for universal contributor guidelines and shouldn't be repurposed. Prior art: Gemini CLI reads `GEMINI.md`, OpenCode reads `.opencode/plugins/`.

Closes #1851

PR #1937, opened 2026-05-23
feat(pricing): make DeepSeek V4 Pro 75% discount permanent

Colorful-glassblock:
## Summary

Removed the V4 Pro discount expiry logic in the TUI cost estimator so the 75% discount is treated as permanent, matching the updated official pricing.

## Testing

## Checklist


Colorful-glassblock:
Windows test failure (cache_inspect_reports_tool_result_budget_metadata) is a pre-existing flaky test unrelated to this PR. The failure is in crates/tui/src/client/chat.rs:3000 — the dedup ref exceeded the 300-char budget because SHA spillover persistence likely failed on the Windows runner, causing the compactor to inline the full 14k-char tool result. This PR only touches crates/tui/src/pricing.rs and docs.

weikengchen:
I think it has overlapping with https://github.com/Hmbown/DeepSeek-TUI/pull/1932 on the README side, but the TUI cost estimator needs to remove the end-time thing.

Colorful-glassblock:
@weikengchen Thanks for the heads-up! I'll drop the README changes from this PR to avoid overlapping with #1932. Appreciate you catching that — this PR will focus solely on the pricing.rs TUI cost estimator change (removing the discount end-time logic).

crates/tui/src/pricing.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! Pricing based on DeepSeek's published rates (per million tokens).
```
carloslindhstrom-code, line 3: crates/tui/src/pricing.rs

Hmbown:
Thanks for tightening the scope after the README overlap was pointed out. I am not merging pricing changes in v0.8.42 without a fresh check against the official DeepSeek pricing source on the exact day we land it.

For the next pass, please keep this PR limited to crates/tui/src/pricing.rs plus tests around the cost estimator. Maintainers should verify current official pricing first, then update docs/changelog from that source rather than relying on issue/PR text.

Colorful-glassblock:
Screenshot_20260525_071859_com android chrome

Hmbown:
**Independent review:**

Reproduced the PR via local merge into `main`: clean auto-merge, no conflicts. Rates land at the published 75% values (USD 0.003625/0.435/0.87, CNY 0.025/3.0/6.0) and the `v4_pro_discount_ends_at` time gate plus its two expiry tests are deleted. Replacement tests (`v4_pro_always_uses_permanent_discount`, `..._does_not_revert_to_base_rates_in_future` @ 2030, `..._applies_consistently` across three dates) are appropriate for a now-constant rate. Refactor of `now` to `_now` keeps the call signature stable so call sites in `chat.rs`/footer don't churn; sensible since other models may need time-gating later.

One nit: since the parameter is now unused, consider removing `pricing_for_model_at` entirely and inlining into `pricing_for_model`, or keeping the param without underscore and adding `#[allow(unused_variables)]` with a comment — the `_now` form reads like an oversight rather than deliberate API preservation. Not blocking.

**v0.8.48 (#2256) compatibility:** byte-identical. `diff <(git show origin/feature/v0.8.48-crate-consolidation:crates/tui/src/pricing.rs) <(git show pr-1937:crates/tui/src/pricing.rs)` returns empty.

**Superseded by #2256: yes.** Evidence: commit `dcc5bdb91` on the v0.8.48 branch is titled "chore: merge PRs #1937, #2237, #1852, #1910 for v0.8.47" — this PR's two commits (`affa26985`, `17a66c13c`) are already in the consolidation branch's history. The only later edit to `pricing.rs` on v0.8.48 is `1cc84601d` removing the now-unused `calculate_cache_savings` helper (footer cache-savings hint dropped), which is orthogonal to the discount change. Recommend closing in favor of #2256.

Hmbown:
@Colorful-glassblock — real appreciation for this contribution. Confirmed independently: the change you're proposing in `pricing.rs` is byte-identical to what's already landed on the v0.8.48 branch (#2256) — same rate values, same removed expiry check, same code path. Same outcome you wanted, arrived via a separate path.

If you'd like your contribution credited specifically, happy to co-author the v0.8.48 commit that touches `pricing.rs` once it merges. Or if there's a test angle here you wanted to ship that #2256 doesn't yet have (you have 3 permanence tests including the 2030 far-future check) — say the word and I'll help extract just those into a small follow-up to #2256.

Either way, your read on the discount being permanent was right and shaped where we landed.

Colorful-glassblock:
Both please — co-author and I'll send the 3 tests as a follow-up PR. Thanks!

Hmbown:
thank you so much for this!

Colorful-glassblock:
You are welcome,here is email jkllll123@outlook.com

PR #1953, opened 2026-05-23
Fix git_status output for unicode paths

### LINKED ISSUE #1936: git_status failed in Chinese folder name
## Description

git_status tool Chinese path encoding is wrong

## Steps to reproduce

Image

## Environment

- OS:
- DeepSeek CLI version:
- Install method:
- Model/provider:
- Terminal app:
- Shell:

cyq1017:
## Summary

- run `git status` with `-c core.quotepath=false` in `git_status` so unicode paths are returned in readable form
- keep the reported command metadata aligned with the actual git invocation
- add a regression test for a tracked file inside a Chinese-named directory

## Root cause

When `core.quotepath` is enabled, Git can emit non-ASCII paths as quoted octal escapes. That makes `git_status` output hard to use for workspaces with Chinese path names. Passing `-c core.quotepath=false` keeps this behavior local to the command without changing user or repository configuration.

Fixes #1936

## Tests

- `cargo test -p deepseek-tui tools::git::tests::`
- `git diff --check`

### DIFF
--- a/crates/tui/src/tools/git.rs
+++ b/crates/tui/src/tools/git.rs
@@ -72,8 +72,11 @@ impl ToolSpec for GitStatusTool {
             args.push(pathspec.display().to_string());
         }
 
-        let command_str = format_command(&git_ctx.working_dir, &args);
-        let output = run_git_command(&git_ctx.working_dir, &args)?;
+        let mut status_args = vec!["-c".to_string(), "core.quotepath=false".to_string()];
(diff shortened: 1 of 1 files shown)


Hmbown:
Merged to `main`. Thanks @cyq1017 — this unicode path fix is clean and well-tested. 鲸鱼兄弟感激不尽 🐋

Issue #1955, opened 2026-05-23
Local skills within a project directory do not get detected


zachgrayio:
## Description

When I launched the TUI in my project directory and checked available skills by entering `/` in the composer, my skills under `.agents/skills/myskill/SKILL.md` were not detected, but the docs specifically say that this is supported. is that a known bug?

## Steps to reproduce

Put a skill under .agents/skills/foo/SKILL.md, type `/` and it wont appear

## Expected behavior
list of skills should show the users skills in the local dir?

## Actual behavior
enter `/` and only see the default skills

## Impact

makes project skills annoying to use/broken

## Environment

- OS: windows 11
- DeepSeek CLI version: `0.8.40`
- Install method: `cargo`
- Model/provider: `deepseek`
- Terminal app: windows terminal
- Shell: powershell

Hmbown:
Thanks for the clear report. Confirmed fixed on main by #1956 (`79151daae`): project-local `.agents/skills/*/SKILL.md` and configured skill directories are merged into the slash menu / command palette. The regression coverage is in `cached_skills_include_configured_directory` and command-palette skill discovery tests.

Closing this as fixed for v0.8.42; appreciate the report because it caught a real docs-vs-runtime mismatch.

PR #1956, opened 2026-05-23
Fix local + configured skills discovery in slash menu

cyq1017:
Fixes issue #1955.

- Fixes slash menu cache discovery to use workspace candidates + configured skills dir instead of only workspace.
- Updates command palette entries to scan workspace candidates + configured directory so .agents/skills and configured installs are both visible.
- Adds regression tests for configured-directory merging and command palette skill discovery.


Hmbown:
Merged to `main`. Thanks @cyq1017 — skills discovery fix is solid. 鲸鱼兄弟感激不尽 🐋

Issue #1874, opened 2026-05-21
openai配置自动补全/v1/chat/completions,但是支持有的第三方模型仅支持/chat/completions，是否可以做成可配置项


shuxiangxuebiancheng:

Hmbown:
补充整理：这个方向和 #1919 / #1967 相关，但这里保留为更通用的 OpenAI-compatible 路径配置问题。#1967 只覆盖 DeepSeek base_url 的持久化配置；这个 issue 继续跟踪 /v1/chat/completions 与 /chat/completions 这类路径差异，以及第三方兼容接口应该怎样在 v0.8.46 里表达清楚。

Hmbown:
Confirmed the root cause and scoped a fix. The request URL is built by `api_url(base_url, "chat/completions")` (`crates/tui/src/client.rs`), which normalizes the path to always include `/v1` (e.g. `https://host` → `https://host/v1/chat/completions`, and `/beta` → `/v1`). Providers that only expose bare `/chat/completions` (or a different prefix) can't currently be expressed.

**Proposed implementation (small, ~1 PR):**
1. Add an optional per-provider config key, e.g. `[providers.<name>] chat_completions_path = "/chat/completions"` (default `None` = current `/v1`-normalizing behavior), in `ProviderConfigToml` (config crate).
2. Thread it into the client alongside `base_url` and, when set, use it verbatim instead of calling `api_url(...)` (skip the `/v1` heuristic).
3. Keep the existing `api_url` heuristic as the default so no current provider changes behavior.
4. Tests: `api_url` unchanged-default + explicit-path-override; round-trip through config resolution.

Leaving open as a ready-to-implement enhancement (not release-blocking for v0.8.51). If a maintainer greenlights the config-key name I'm happy to land it.

---
_Generated by [Claude Code](https://claude.ai/code)_

Hmbown:
Fixed by \ + \ on \: provider configs now support a provider-local \ for OpenAI-compatible gateways that accept \ but reject \.

Example:

\\\

The override is intentionally narrow: it applies only to chat-completions requests. Model listing and DeepSeek beta/FIM paths keep their normal routing, so a generic gateway override does not rewrite \ or \. The v0.9 stewardship branch also includes #2776 / \, which prevents repo-local project config from overriding \ as an endpoint-routing field.

Verified with \
running 5 tests
test client::tests::api_url_with_suffix_default_behavior_without_suffix ... ok
[...]

PR #1967, opened 2026-05-24
feat(tui): support configurable DeepSeek base URL in /config

### LINKED ISSUE #1919: Create a function to custom API Endpoint by users
## Problem

Users are not able to custom API Endpoint by themselves

## Proposed solution

Just add a function to edit API Endpoint

## Use case

Users are not able to custom API Endpoint by themselves, so that they must use DeepSeek's official server, which can't be used to dispose private things

## Alternatives considered

It isn't suitable for this request

## Impact

Not very often, but some users use it very often. This function can help me save my cost

cyq1017:
## Summary
- Treat /config base_url as a persisted-only setting: setting it without --save now returns an error.
- New behavior message explains that DeepSeek base URL is loaded from config on startup and requires restart/re-open session after saving.
- Keep no runtime/session client hot-reload path; behavior stays minimal and aligned with existing model/client initialization model.
- Keep /config base_url read path unchanged: /config base_url returns current value from config.

## Testing
- cargo fmt --all -- --check
- git diff --check
- cargo test -p deepseek-tui config_command_base_url -- --nocapture
- cargo test -p deepseek-tui config_command_cost_currency_save_persists_value -- --nocapture
- cargo test -p deepseek-tui test_set_default_mode_normal_save_reports_normalized_value -- --nocapture
- cargo test -p deepseek-tui default_settings_disable_auto_compact_to_protect_v4_prefix_cache -- --nocapture

Closes #1919

### DIFF
--- a/crates/tui/src/commands/config.rs
+++ b/crates/tui/src/commands/config.rs
@@ -5,7 +5,9 @@ use std::time::Duration;
 
 use super::CommandResult;
 use crate::client::DeepSeekClient;
-use crate::config::{COMMON_DEEPSEEK_MODELS, clear_api_key, normalize_model_name_for_provider};
+use crate::config::{
+    COMMON_DEEPSEEK_MODELS, Config, clear_api_key, normalize_model_name_for_provider,
+};
 use crate::config_ui::{ConfigUiMode, parse_mode};
 use crate::llm_client::LlmClient;
 use crate::localization::resolve_locale;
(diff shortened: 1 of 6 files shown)


Hmbown:
Stewardship note: #1919 is now milestoned to v0.8.46 and this PR is the narrow candidate for persisted DeepSeek base_url config. I left the broader OpenAI-compatible path/provider parser questions in #1874 and #1978 so this PR can stay small.

Review focus: persisted config semantics, restart/session messaging, and not accidentally promising hot reload.

Hmbown:

PR #1585, opened 2026-05-13
Add file policy engine for file tool harness

kunpeng-ai-lab:
Adds the file policy engine on top of current main/v0.8.33. Ports the harness-layer file access policy and wires it into serial tool execution, parallel tool batches, and execute_tool_with_lock. Validation: cargo check --workspace; cargo test -p deepseek-tui --bin deepseek-tui file_policy; cargo test -p deepseek-tui --bin deepseek-tui engine::; cargo test -p deepseek-tui --bin deepseek-tui composer_arrows_scroll.

PR #1974, opened 2026-05-24
Add file policy engine for file tool harness

kunpeng-ai-lab:
## Summary

Reopens the stale pre-v0.8.41 file policy engine work from #1585 on current `main`.

This adds a harness-level file access policy that can allow or deny file-touching tool calls by category before mutation/read execution. It is intended as a reusable safety layer for workspace policy, diagnostics, and audit behavior.

## Changes

- Add file policy config/rule parsing and an example `refs/execpolicy.example.toml`.
- Wire policy loading behind the existing feature/config surface.
- Enforce file policy for serial tool execution, parallel tool batches, and `execute_tool_with_lock` defensive execution.
- Extract file targets for `read_file`, `write_file`, `edit_file`, and `apply_patch` including path override and unified diff headers.
- Emit `tool.file_policy_denied` audit metadata on denied execution.
- Add design documentation in `docs/file-policy-design.md`.
- Add regression tests for allow/deny rules, default category fallback, disabled feature behavior, apply_patch denial, and policy checks in engine paths.

## Validation

- `cargo fmt --check`
- `cargo test -p codewhale-tui file_policy` (15 passed)
- `cargo test -p codewhale-tui engine::` (117 passed)
- `cargo test -p codewhale-tui composer_arrows_scroll` (10 passed)
- `cargo check --workspace`

Supersedes #1585, which became stale after the v0.8.41 rebrand.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -319,6 +319,7 @@ web_search = true # enables canonical web.run plus the compatibility web_search 
 apply_patch = true
 mcp = true
 exec_policy = true
+file_policy = true
 # vision_model = false  # enable vision model for image_analyze tool
 
 # ─────────────────────────────────────────────────────────────────────────────────
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -331,6 +331,8 @@ pub struct Engine {
     /// Diagnostics collected during the current step's tool calls. Drained
     /// and forwarded as a synthetic user message before the next API call.
     pending_lsp_blocks: Vec<crate::lsp::DiagnosticBlock>,
+    /// File access policy loaded from `~/.deepseek/execpolicy.toml`.
+    file_policy: Option<Arc<crate::execpolicy::ExecPolicyConfig>>,
 }
 
 // === Internal tool helpers ===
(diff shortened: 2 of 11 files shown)


Hmbown:
Independent review:

The engine correctly threads `file_policy` through serial dispatch, parallel batches, and `execute_tool_with_lock`. Tests demonstrate the deny path and the static evaluator round-trip. A few load-bearing gaps:

1. **`AskUser` is treated as allow** (`engine.rs:425`). `evaluate_file_policy` only returns on `Deny`; the design doc admits this. A user writing `allow = ["src/**/*.rs"]` with no explicit `deny` is intending a whitelist — but `~/.ssh/id_rsa` produces `AskUser` and passes. The allow list is decorative; the engine is effectively a denylist regardless of intent.

2. **No path canonicalization** (`rules.rs:163-183`). Paths come straight from JSON, and `pattern_matches_path` anchors with `^...$`. A pattern `.env` will not match `./.env`, `/Users/u/.env`, or `${HOME}/.env`. Symlinks aren't resolved either. Trivial bypass.

3. **`capacity_flow.rs:531` passes `None` for file_policy** in `apply_verify_with_tool_replay`. The design doc lists three enforcement points, but this is a fourth call site. It only replays read-only tools, so the blast radius is `read_file` re-reading a denied path like `.env` — but the example config explicitly denies that.

4. **gemini findings already cover** the split_whitespace diff-header bug, silent toml parse failure (`.ok().flatten()` at `engine.rs:577`), deny-before-allow precedence, missing `list_dir`/`grep_files`, and `+++ ` false positives inside patch bodies.

v0.8.48 (`feature/v0.8.48-crate-consolidation`) does not touch policy code — clean merge confirmed. The harder problem: **PR #2242 supersedes this** with typed permission rules covering tool/shell/file-glob, `allow`/`deny`/`ask` decisions, workspace-relative path normalization, symlink awareness, persistent `permissions.toml`, and approval-UI integration. The two implementations cannot coexist on the same `is_file_tool` hook. Recommend closing in favor of #2242 unless this PR's deny-only-on-deny semantics are explicitly desired as a separate audit-only layer.

Hmbown:
@kunpeng-ai-lab — thank you for the file policy engine. Your work and #2242 (greyfreedom's typed persistent tool permission rules) hook the same file tools at the same enforcement points — they cannot coexist, so the maintainer will pick one. Honest comparison:

- **#1974 (yours)** ships a glob-based gate keyed on `[file.<tool>]` sections, feature-flag controlled, with audit-on-deny. Simpler config surface.
- **#2242** ships typed `allow`/`deny`/`ask` rules with workspace-relative path normalization and symlink awareness, persistent `permissions.toml`. Richer surface.

If #2242 lands, the gem from your branch worth carrying forward is the `AskUser`-as-allow gap I flagged in the review (currently silent in both PRs structurally) and the audit event design. Either ping me on #2242 to extract those, or — if there's an angle you'd want #2242 to incorporate before it merges — flag it directly there and we'll route. Your stake here is real.

Hmbown:
Thanks @kunpeng-ai-lab for the file-policy pass. I am closing this PR as superseded by the newer typed persistent permission-rule work in #2242, which is the branch we should keep carrying forward for this area.

There are still useful lessons here, especially around audit events and ask-user style policy, but this branch is now conflicting and has older canonicalization/replay gaps. Any remaining ideas should be folded into #2242 or a fresh focused PR against current `main`, with credit preserved.

PR #2107, opened 2026-05-25
feat(tui): FauxStep::Factory for live request-shape assertions

### LINKED ISSUE #2074: `MockLlmClient`: turn-factory closures for live request-shape assertions
**Problem.** `crates/tui/src/llm_client/mock.rs:75` is queue-driven: tests push canned `MessageRequest` shapes and the mock dequeues. There is no way to assert on what the engine actually sent on each turn — specifically, whether `reasoning_content` was replayed on the assistant turn that produced the previous tool call. The v0.4.9–v0.5.1 regression that broke every DeepSeek V4 user (HTTP 400 on follow-up turns when thinking tokens were dropped) would not be caught by today's tests. The comment at `mock.rs:6-8` already acknowledges the gap.

**Shape.**
```rust
pub enum FauxStep {
    Canned(CannedTurn),
    Factory(Box<dyn Fn(&MessageRequest) -> CannedTurn + Send + Sync>),
}
```
When a `Factory` step is dequeued, its closure runs against the real outgoing `MessageRequest`, may `assert!` inline, then returns the response. No change to existing canned-turn callers.

**Acceptance.**
- `FauxStep::Factory` variant added.
- One regression test at `crates/tui/tests/reasoning_content_replayed_after_tool_call.rs` whose factory asserts `request.messages.last().reasoning_content.is_some()` after a thinking + tool-call round.
- Doc comment on the factory variant explains the V4 `reasoning_content` invariant and links to the original regression range.
- No behavior change for existing tests.

**Whale notes.** Lowest-effort, highest-leverage change in this slate. Pays for itself the first time it catches a thinking-replay regression, and is the bridge we need before any integration test can exercise the real SSE stream (`serve --http`).

_Slate item 1/18. Tracking doc: `.private/issue-slate.md` (maintainer-only)._

mvanhorn:
Closes #2074.

## Shape

```rust
pub enum FauxStep {
    Canned(CannedTurn),
    Factory(Box<dyn Fn(&MessageRequest) -> CannedTurn + Send + Sync>),
}
```

When a `Factory` step is dequeued, its closure runs against the real outgoing `MessageRequest` before the response stream is built, so any `assert!` panic surfaces directly from the client call rather than later in stream polling. No change to existing canned-turn callers.

## Public surface

- `MockLlmClient::new(Vec<CannedTurn>)` and `push_turn(CannedTurn)` keep working — they wrap each turn in `FauxStep::Canned`.
- `push_factory(closure)` is new, for tests that want the Factory branch.
- Doc comment on `FauxStep::Factory` captures the V4 thinking-mode tool-call invariant (the v0.4.9-v0.5.1 `reasoning_content` drop that produced HTTP 400 on follow-up turns).

## Tests

- New regression file `crates/tui/tests/reasoning_content_replayed_after_tool_call.rs` whose factory asserts the assistant tool-call turn carries a `Thinking` content block after a thinking + tool-call round. Mirrors the `#[path = "..."]` test-support pattern already used by `integration_mock_llm.rs`.
- One additional unit test in `mock.rs` (`create_message_synthesizes_from_factory_turn`) for the non-streaming path.
- Existing `integration_mock_llm` suite (27 tests) is unchanged and passes.

```
cargo test -p codewhale-tui --test reasoning_content_replayed_after_tool_call
  ... 20 passed; 0 failed
cargo test -p codewhale-tui --test integration_mock_llm
  ... 27 passed; 4 ignored
cargo fmt --check  ... clean
```

## Known limitation

The new regression test asserts against a follow-up request the test itself assembles (with `assistant_thinking_tool_call(...)`), not one produced by the real turn-loop. That matches the issue's "Acceptance" wording (`request.messages.last().reasoning_content.is_some()`) and the `integration_mock_llm.rs` doc-comment which already notes the engine still holds a concrete `Option<DeepSeekClient>` rather than `Arc<dyn LlmClient>` — until that wiring lands, the mock can't be wedged into the real turn-loop. This PR delivers the `FauxStep::Factory` infrastructure the issue asks for; the engine-level wiring is the follow-up the `integration_mock_llm.rs` ignored tests already track.

AI was used for assistance.

### DIFF
--- a/crates/tui/src/llm_client/mock.rs
+++ b/crates/tui/src/llm_client/mock.rs
@@ -63,6 +63,21 @@ use super::{LlmClient, StreamEventBox};
 /// the mock does not require `MessageStart` to be present.
 pub type CannedTurn = Vec<StreamEvent>;
 
+/// A queued mock response step.
+pub enum FauxStep {
+    Canned(CannedTurn),
+    /// Build a canned turn from the live outgoing request.
+    ///
+    /// Tests can assert DeepSeek V4's thinking-mode tool-call invariant here:
+    /// on the assistant turn that produced the previous tool call, the next
+    /// outgoing request must still carry `reasoning_content` (represented in
+    /// this model as a [`ContentBlock::Thinking`] block). If it is missing,
+    /// DeepSeek V4 returns HTTP 400 on the follow-up turn. This guards the
+    /// [v0.4.9-v0.5.1 regression range](https://github.com/Hmbown/CodeWhale/compare/v0.4.9...v0.5.1)
+    /// where that content was dropped.
+    Factory(Box<dyn Fn(&MessageRequest) -> CannedTurn + Send + Sync>),
+}
+
 /// A queue-driven mock LLM client.
 ///
 /// The mock holds a FIFO queue of canned response turns. Each call to
@@ -75,7 +90,7 @@ pub type CannedTurn = Vec<StreamEvent>;
 /// can assert on the outgoing payload (e.g. that prior `reasoning_content` is
 /// preserved across turns).
 pub struct MockLlmClient {
-    canned: Mutex<VecDeque<CannedTurn>>,
+    canned: Mutex<VecDeque<FauxStep>>,
     captured_requests: Mutex<Vec<MessageRequest>>,
     calls: AtomicUsize,
     provider_name: &'static str,
@@ -91,7 +106,7 @@ impl MockLlmClient {
     #[must_use]
     pub fn new(canned: Vec<CannedTurn>) -> Self {
         Self {
-            canned: Mutex::new(canned.into()),
+            canned: Mutex::new(canned.into_iter().map(FauxStep::Canned).collect()),
             captured_requests: Mutex::new(Vec::new()),
             calls: AtomicUsize::new(0),
             provider_name: "mock",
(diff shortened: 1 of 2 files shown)


Hmbown:
**Independent review:**

API is clean: `FauxStep::Canned | Factory(Box<dyn Fn(&MessageRequest) -> CannedTurn + Send + Sync>)`, internal `VecDeque<FauxStep>`, with `push_factory` as the new surface. Closure runs inside `create_message[_stream]` before stream construction, so `assert!` panics from the client call, not the consumer's `.next().await` — that's the real win over capture-then-assert, and it works.

PR body's own "known limitation" is honest: the regression test asserts on a follow-up `MessageRequest` the test itself assembles via `assistant_thinking_tool_call(...)`, not one produced by the engine's turn-loop. So this lands the infrastructure for #2074 cleanly but does not yet exercise the V4 invariant against real engine output — the `integration_mock_llm.rs` ignored tests still gate that. Worth saying so in the merge commit.

Unit test `create_message_synthesizes_from_factory_turn` covers the non-streaming branch; factory storage/dequeue is exercised by both files. Coverage of the new surface is adequate.

**v0.8.48 (#2256):** merges clean — `models.rs` adds `response_format: Option<serde_json::Value>` with `skip_serializing_if`. The `empty_request()` helper in `mock.rs` gets it via auto-merge, but the new `make_request()` in `reasoning_content_replayed_after_tool_call.rs` will need `response_format: None` appended post-rebase or it'll fail to compile.

mvanhorn:
Thanks for merging the FauxStep factory @Hmbown, live request-shape assertions in the TUI tests should catch drift early.

mvanhorn:
Glad both landed, thanks again @Hmbown; the FauxStep factory should make the request-shape assertions easier to extend.

mvanhorn:
Thanks @Hmbown, glad the FauxStep::Factory closure landed for live request-shape assertions.

PR #2108, opened 2026-05-25
feat: session title in composer

mars-base:
## Summary
- Display the current session title in the composer top bar
- Fix `luma()` u16 overflow issue
- When a session title exists, show it alongside the vim mode indicator in the composer top-right corner

## Test plan
- [x] After creating/renaming a session, the session title appears in the composer top bar
- [x] When no session title is set, the existing composer layout is unaffected
- [x] Vim mode and session title display correctly together


mars-base:
## Background & Goal of This PR

The main goal of this PR is to **display the current session name in the TUI interface**.

This was inspired by the official Claude Code client, which shows the current session title on the right side of the composer area — making it easy to know which session you're in when working with multiple sessions simultaneously.

During implementation, I noticed that CodeWhale's composer already has a vim-mode indicator on the right side. So this PR places the session name alongside the vim-mode indicator in the same area. I think they coexist naturally and keep the layout clean.

---

**I'd love to hear the maintainer's thoughts**: if you feel combining the session name with the vim-mode indicator is not ideal (e.g. information density, readability, or layout consistency concerns), where do you think the session name would fit better in the TUI?

Some possible alternatives for discussion:

- **Top title bar** (if applicable)
- **Bottom status bar**
- **Left side of the composer**
- Anywhere else you think suits CodeWhale's design better

When managing multiple sessions during development, being able to see "which session am I in" at a glance in the TUI is genuinely helpful. Happy to explore any placement that fits CodeWhale's overall design style — feel free to discuss here!

mars-base:
Thanks for the review! The suggestion to restore truncation for the vim mode label has been addressed in the latest commit — `truncate_display_width` is now applied to the vim mode label again, consistent with the original behavior.

PR #2241, opened 2026-05-27
  fix(provider): keep picker selection visible

### LINKED ISSUE #2238: /provider pop-up displaying  incomplete
After opening the pop-up via /provider, the content is incomplete and you cannot scroll down to view the subsequent preset providers.
when I tab the down arrows, it doesn't work, the subsequent preset providers won't display.
Image

reidliu41:
## Summary

  Fixes https://github.com/Hmbown/CodeWhale/issues/2238.

 The `/provider` modal had more provider rows than its fixed-height list could show. Arrow-key navigation changed the selected provider internally, but rendering always started from the first provider, so later providers such as SGLang, vLLM, and Ollama could be selected without becoming visible.

  This updates the picker to:
  - size the list modal to fit all providers when the terminal has enough height
  - scroll the visible rows when the selected provider moves beyond the viewport
  - keep an initially selected bottom provider visible
  - wrap Up/Down navigation between the first and last providers
  - render the selected row with a continuous subtle highlight

from:  cannot move forward to next 

image

to:

image

## Testing

## Checklist

This PR fixes the `/provider` modal so the selected provider is always visible during keyboard navigation. Previously the list was fixed at height 12, causing providers like SGLang, vLLM, and Ollama to be navigated to but never displayed.

- **Dynamic popup height**: the list stage now sizes to `providers.len() + 2` (clamped to terminal height), showing all 12 providers when the terminal has enough room.
- **Viewport scrolling**: a new `visible_start` helper computes the minimum scroll offset to keep the selection visible, and the render loop uses `.skip(visible_start).take(visible_rows)` to emit only the visible slice.
- **Wrap-around navigation**: `move_up`/`move_down` now wrap from first↔last instead of clamping, and the corresponding test comment is corrected from \"twice\" to \"once\".
</details>

Safe to merge — the change is self-contained to a single TUI modal and carries no state-mutation risk outside the picker.

The visible_start formula correctly computes the minimum scroll offset that keeps the selection on-screen in all navigation scenarios (forward, backward, and wrap-around), and is covered by four focused render tests. The popup height calculation clamps properly to both the terminal size and the 8-row minimum. No data path, persistence, or API surface is affected.

No files require special attention.
</details>

| Filename | Overview |
|----------|----------|
| crates/tui/src/tui/provider_picker.rs | Adds dynamic popup sizing, viewport-scroll logic, wrap-around navigation, and a full-width row highlight; logic is correct and well-covered by new tests. |

</details>

</details>

<sub>Reviews (2): Last reviewed commit: ["Make the selected row easier to see with..."](https://github.com/hmbown/codewhale/commit/198acc72bae04620a4b3ddaab86348970b84cdaf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34082813)</sub>

### DIFF
--- a/crates/tui/src/tui/provider_picker.rs
+++ b/crates/tui/src/tui/provider_picker.rs
@@ -21,7 +21,7 @@ use crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
 use ratatui::{
     buffer::Buffer,
     layout::{Constraint, Direction, Layout, Rect},
-    style::{Modifier, Style},
+    style::{Color, Modifier, Style},
     text::{Line, Span},
     widgets::{Block, Borders, Clear, Paragraph, Widget},
 };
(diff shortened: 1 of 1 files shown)


reidliu41:
hi @Hmbown please take a look when you have time. thanks.

PR #2260, opened 2026-05-27
feat: macos display notification

New2Niu:

This PR adds macOS Notification Center support by introducing a new `Method::MacOS` variant that dispatches via `osascript` instead of terminal escape sequences, and extends the existing notification system to fire on user-action events (tool approval, user input, sandbox elevation) in addition to turn completion.

- `macos_display_notification` uses the `osascript ARGV` approach instead of inline string interpolation, correctly addressing the AppleScript injection risk flagged in a prior review; the title is \"CodeWhale\" and the function runs on a background thread to avoid blocking the tokio runtime.
- Three new notification call sites are added in `ui.rs` for `ApprovalRequired`, `UserInputRequired`, and `ElevationRequired` events; all three intentionally pass `Duration::ZERO` for threshold and elapsed so urgent user-action events are always surfaced regardless of the user-configured completion threshold.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

The macOS notification path has an unresolved ordering issue where the macOS guard fires before the $TERM-based terminal-protocol checks, causing macOS users in Ghostty or Kitty without TERM_PROGRAM set to always receive osascript notifications instead of the richer in-terminal protocol.

The core injection concern and compile-guard omission have been fixed, but the resolve_method() ordering issue at line 101 remains, silently regressing terminal-native notification protocols for a class of macOS users on every notification.

crates/tui/src/tui/notifications.rs — the ordering of the cfg!(target_os = macos) guard relative to the $TERM-based protocol detection block in resolve_method().
</details>

| Filename | Overview |
|----------|----------|
| crates/tui/src/tui/notifications.rs | Adds MacOS method variant + macos_display_notification() via osascript ARGV (injection-safe); macOS guard still placed before $TERM-based protocol fallbacks, which means macOS users in Ghostty/Kitty without TERM_PROGRAM will bypass richer in-terminal protocols. |
| crates/tui/src/tui/ui.rs | Adds immediate notifications for ApprovalRequired, UserInputRequired, and ElevationRequired engine events; correctly bypasses configured threshold (Duration::ZERO) so user-action events always notify; import block reorder is cosmetic only. |

</details>

</details>

<sub>Reviews (4): Last reviewed commit: ["fix: test"](https://github.com/hmbown/codewhale/commit/5068f8cab466b2d53c1bf332ae69b3b634b9d697) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34105040)</sub>

### DIFF
--- a/crates/tui/src/tui/notifications.rs
+++ b/crates/tui/src/tui/notifications.rs
@@ -909,7 +909,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     fn auto_detect_picks_bel_for_unknown_on_unix() {
         let _lock = env_lock();
         let prev_tp = std::env::var_os("TERM_PROGRAM");
@@ -992,7 +992,7 @@ mod tests {
     /// `TERM_PROGRAM` but do set `TERM=xterm-ghostty`. The `$TERM`
     /// fallback should catch them.
     #[test]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     fn auto_detect_picks_osc9_for_xterm_ghostty_term_fallback() {
         let _lock = env_lock();
         let prev_tp = std::env::var_os("TERM_PROGRAM");
@@ -1060,7 +1060,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     fn auto_detect_picks_kitty_from_term_fallback() {
         let _lock = env_lock();
         let prev_tp = std::env::var_os("TERM_PROGRAM");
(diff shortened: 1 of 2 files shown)


Hmbown:
**Independent review (Devin):**

**What's good:** ARGV-based osascript injection fix is correct. Background thread spawn avoids blocking tokio.  gate on `macos_display_notification` is present; non-macOS builds are unaffected. Clean merge against both `main` and pr-2256 (v0.8.48) — zero conflicts.

**Issues found:**

1. **`resolve_method()` ordering (confirmed, line ~101):** The `cfg!(target_os = "macos")` guard fires *after* `TERM_PROGRAM` and `LC_TERMINAL` checks but *before* the `$TERM` fallback block. macOS users whose terminal only sets `TERM=xterm-ghostty` or `TERM=xterm-kitty` (without `TERM_PROGRAM`) silently get osascript instead of their richer in-terminal protocol. Fix: move the macOS guard below the `$TERM` fallback, or only fall through to `MacOS` after all env-based checks are exhausted.

2. **`MacOS` not in `NotificationMethod` config enum (`config.rs:524`):** Users cannot explicitly set `method = "macos"` in `config.toml`, nor can they force-disable it independently of `method = "off"`. Consider adding a `MacOS` variant so the config and runtime method surfaces are in sync.

3. **`Duration::ZERO` threshold bypass in `ui.rs` call sites:** Approval/input/elevation notifications unconditionally fire regardless of the user's configured `threshold_secs`. Intentional or not, this should be documented in the call-site comments.

4. **No test coverage for macOS path** — the test exclusions (`#[cfg(not(any(target_os = "windows", target_os = "macos")))]`) are correct guards but there are no affirmative macOS-specific tests.

**v0.8.48 (pr-2256) compatibility:** `git merge --no-ff --no-commit` exits clean — no conflicts in either `notifications.rs` or `ui.rs`.

PR #2275, opened 2026-05-27
fix(tui): make update hint transient (#2254)

Hmbown:
## Summary
- only advertise a newer release after the GitHub release is non-draft, non-prerelease, and has the required CodeWhale assets plus checksum manifest uploaded
- route the update notice through the existing timed status-toast queue instead of storing a persistent footer override
- preserve the user-configured footer/statusline chips as the steady display

Fixes #2254.

Reported by @bevis-wong in #2254.

Also credits @idling11 for #2268, which identified the draft/prerelease guard that this PR folds into the broader release-readiness fix.

## Verification
- cargo test -p codewhale-tui version_hint_requires_complete_release_assets -- --nocapture
- cargo test -p codewhale-tui version_hint_ignores_draft_prerelease_and_current_versions -- --nocapture
- cargo check -p codewhale-tui --all-features --locked
- cargo fmt --all -- --check
- git diff --check

<details open><summary><h3>Greptile Summary</h3></summary>

This PR fixes issue #2254 by routing the version-update notification through the existing timed status-toast queue (12 s TTL) instead of storing it as a persistent `version_hint` field that permanently overrode the user's configured footer chips. It also adds release-completeness validation before advertising an update.

- Removes `App::version_hint` and the footer override path in `footer_ui.rs`; the update notice is now a `StatusToastLevel::Info` toast pushed once via `push_status_toast`.
- Adds `version_hint_from_release_json`, `release_has_required_assets`, and `release_has_uploaded_asset` to gate the hint behind draft/prerelease flags and a per-asset `state == \"uploaded\"` check against `REQUIRED_RELEASE_ASSETS`.
- New unit tests exercise the complete-release happy path, missing-manifest, pending-state, missing-state-field, draft, prerelease, and same-version cases.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

Safe to merge — the change is narrowly scoped to the update-notification path and preserves all existing footer behaviour.

The removal of version_hint and the simplified footer path are straightforward; the new asset-validation helpers are covered by a thorough test suite that explicitly checks the previously flagged missing-state-field case. The is_newer_version string-compare fallback for unparseable semver tags is pre-existing behaviour, not introduced here. No state is silently dropped and the toast fires exactly once per session.

No files require special attention.
</details>

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
[...]

PR #2274, opened 2026-05-27
docs: add provider registry drift check

nightt5879:
## Summary

- add `scripts/check-provider-registry.py` to check `docs/PROVIDERS.md` against the shipped provider IDs, `[providers.*]` tables, static `ModelRegistry` provider rows, and default model/base URL strings
- wire the provider-registry drift check into the CI lint job
- document the drift check in `docs/PROVIDERS.md` and link the provider registry from the README docs table
- fix current clippy-only style lints in legacy shim/config/log code so the full clippy gate is clean

## Testing

- [x] `python scripts/check-provider-registry.py`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [ ] `cargo test --workspace --all-features`
  - Ran on this Windows checkout; failed after `3411 passed; 11 failed; 4 ignored`.
  - This is listed as run, not as a passing validation.
  - Failures observed: `client::chat::stream_decoder_tests::cache_inspect_reports_tool_result_budget_metadata`, `commands::change::tests::change_in_non_english_without_api_key_uses_explicit_fallback`, `project_context::tests::test_global_agents_only_no_project_unchanged_fallback`, `project_context::tests::test_load_global_agents_when_project_has_no_context`, `prompts::tests::system_prompt_skips_locale_preamble_for_english`, `session_manager::tests::latest_session_for_workspace_ignores_invalid_parent_git_marker`, `session_manager::tests::latest_session_for_workspace_ignores_newer_other_directory`, `tools::pandoc::tests::pandoc_convert_roundtrips_markdown_to_html_inline`, `tools::pandoc::tests::pandoc_convert_writes_output_path_and_reports_summary`, `tui::app::tests::app_new_detects_missing_api_key_with_default_config`, `tui::ui::tests::saved_default_provider_syncs_back_to_runtime_config`.
  - The visible causes are this checkout/environment rather than this patch: parent `C:\.git` workspace detection, pandoc `getXdgDirectory:sHGetFolderPath`, real `%APPDATA%` write denial, and test-home/API-key/global-AGENTS contamination.

## Checklist

- [x] Updated docs or comments as needed
- [x] Added or updated tests where relevant
- [ ] Verified TUI behavior manually if UI changes
  - No TUI UI behavior changed.

Closes #2201

<details open><summary><h3>Greptile Summary</h3></summary>

[...]

PR #2256, opened 2026-05-27
refactor: consolidate workspace crates — 14→11 (delete tui-core, merge hooks+agent)

Hmbown:
## Summary
Reduces the Cargo workspace from 14 crates to 11 by removing dead code and merging thin abstraction layers. Zero behavioral changes.

### Changes
| Action | Crate | Lines | Rationale |
|--------|-------|-------|-----------|
| ❌ Deleted | `tui-core` | 192 | Orphaned — zero dependents, 192-line scaffold never wired to the real TUI |
| 🔀 Merged → `core` | `hooks` | 296 | Only `core` and `app-server` used it; `app-server` already depends on `core`. No new edges. |
| 🔀 Merged → `config` | `agent` | 539 | Only depended on `config` (`ProviderKind`); all consumers already depend on `config`. No new edges. |

### Why these three?
- **`tui-core`**: Dead code. No `use codewhale_tui_core` anywhere in the tree.
- **`hooks`** (296 LOC): Only `core` + `app-server` consumers. `app-server` already takes `core`.
- **`agent`** (539 LOC): Thin model registry that only depends on `config` for `ProviderKind`. All 3 consumers already take `config`.

### Why NOT `execpolicy`/`mcp`/`tools`?
They're all consumed by `cli` directly, but `cli` doesn't depend on `core`. Merging them into `core` would force `cli` to pull in the full orchestration crate (+ transitive deps). They stay separate.

### Verification
- `cargo check --workspace` ✅
- `cargo build --workspace` ✅
- `cargo test --workspace` ✅ — 3,843 tests passed, 0 failed

Closes #2249, closes #2250, closes #2251

<details open><summary><h3>Greptile Summary</h3></summary>

This PR is labeled as a crate-consolidation refactor but bundles in substantial new functionality: Xiaomi MiMo provider integration, a closed-loop verification gate, a pricing simplification making the DeepSeek V4 Pro discount permanent, PDF/image reads moved to `spawn_blocking`, a new `~/.agents/AGENTS.md` fallback path, and a universal "all native tools loaded upfront" policy change.

- **Xiaomi MiMo provider**: added to `config`, `agent`, `cli`, `tui/config.rs`, model picker, provider picker, reasoning-effort serialisation, and auth flows; gated by `MIMO_API_KEY` / `MIMO_BASE_URL` env vars.
[...]

PR #2277, opened 2026-05-27
fix(tui): retry composer history atomic writes

Hmbown:
## Summary
- retry composer-history atomic rewrites before dropping a queued batch
- make the async writer robust to transient Windows sharing violations while readers are polling the history file
- addresses the repeated Windows CI failure in `composer_history::tests::append_history_dispatched_does_not_block_the_caller` seen on #2274 and earlier on #2275

Refs #2274.

## Verification
- CARGO_TARGET_DIR=/Volumes/VIXinSSD/whalebro/codewhale/target cargo test -p codewhale-tui append_history_dispatched_does_not_block_the_caller -- --nocapture
- CARGO_TARGET_DIR=/Volumes/VIXinSSD/whalebro/codewhale/target cargo check -p codewhale-tui --all-features --locked
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/tui/src/composer_history.rs
+++ b/crates/tui/src/composer_history.rs
@@ -190,12 +190,46 @@ fn append_history_entries_to<'a>(
     }
 
     let payload = entries.join("\n") + "\n";
-    if let Err(err) = crate::utils::write_atomic(path, payload.as_bytes()) {
+    if let Err(err) = write_history_atomic(path, payload.as_bytes()) {
         tracing::warn!(
             "Failed to persist composer history at {}: {err}",
             path.display()
         );
     }
+}
+
+fn write_history_atomic(path: &Path, payload: &[u8]) -> std::io::Result<()> {
+    const RETRY_DELAYS: &[Duration] = &[
+        Duration::from_millis(5),
+        Duration::from_millis(10),
+        Duration::from_millis(25),
+        Duration::from_millis(50),
+        Duration::from_millis(100),
+        Duration::from_millis(200),
+        Duration::from_millis(400),
+    ];
+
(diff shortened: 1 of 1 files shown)


Hmbown:
**Independent review:**

- **Failure mode silently loses entries.** When the 8 attempts exhaust, `append_history_entries_to` only logs a `tracing::warn!` and drops the entire `entries` Vec (line 193-198). The trimmed/deduped new entries never reach disk and are not requeued onto the writer channel. Worth either (a) re-sending the failed payload back through `channel()` for the next batch cycle, or (b) at minimum emitting a user-visible notice — losing prompt history silently is the worst possible outcome of a retry that exists to *prevent* loss.
- **`write_atomic` doesn't fsync the parent dir.** `utils.rs:174` does `sync_all()` on the temp file and `persist()` (rename), but no `File::open(parent).sync_all()`. On a crash between rename and dir-sync, ext4/xfs can lose the rename. Out of scope for this PR but worth a follow-up since the doc comment promises crash durability.
- **PR description mentions \"readers polling\" but only the test polls** (line 353, 25ms loop). The real binary calls `load_history()` exactly once at startup (`app.rs:1797`). On Windows the production sharing-violation source is more likely AV/indexers than the app itself — the retry is still correct, but the rationale in the PR body is misleading.
- **No regression test for the retry path** — only the existing throughput test. A test that injects a transient failure (e.g., holding an open handle for ~50ms) and asserts the write eventually succeeds would lock in this fix.
- **No jitter** in the backoff. Fine for a single writer thread, but if two CodeWhale processes share a HOME they'd lockstep.
- v0.8.48 (PR #2256) does not touch `composer_history.rs` — clean merge.

PR #1993, opened 2026-05-24
feat: add Volcengine provider with DeepSeek-V4-Pro/Flash support

dzyuan:
Add Volcengine (Volcano Engine Ark) provider for DeepSeek-V4-Pro/Flash via Coding API.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR adds Volcengine (Volcano Engine Ark) as a new provider for DeepSeek-V4-Pro and DeepSeek-V4-Flash via the Ark Coding API, following the same mechanical pattern used for WanjieArk and other recent providers. It also changes model-resolution under a provider hint to use the registry's canonical model ID (preserving mixed-case like `DeepSeek-V4-Pro`) rather than mirroring the user's requested casing.

- **Core provider wiring** (`crates/config`, `crates/cli`, `crates/secrets`, `crates/tui/src/config.rs`): enum variant, default constants, env-var aliases, config TOML struct, get/set/clear paths, and auth-status output are all added consistently.
- **Model registry** (`crates/agent/src/lib.rs`): Two new `ModelInfo` entries for `DeepSeek-V4-Pro` and `DeepSeek-V4-Flash` under `ProviderKind::Volcengine`; the `deepseek-chat` and generic `deepseek-v4-*` aliases on these entries are already claimed by DeepSeek/NvidiaNim in the global alias map, so only the Volcengine-prefixed aliases are effective in no-provider-hint resolution.
- **Two issues in `crates/tui/src/config.rs`**: Volcengine is incorrectly placed in the self-hosted fallback arm (returning an empty API key instead of a descriptive error), and `has_api_key_for` only checks `VOLCENGINE_API_KEY` while `active_provider_has_env_api_key` checks all three documented alternatives.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

The provider wiring is mechanically correct across most of the codebase, but two defects in crates/tui/src/config.rs mean Volcengine users without a pre-configured key will receive confusing silent failures rather than actionable errors, and the auth-status UI can report a missing key even when one is set via an alternate env var.

The bulk of the change is a clean, consistent extension of the existing provider pattern. However, crates/tui/src/config.rs has two related defects: Volcengine is placed in the self-hosted fallback arm that returns an empty API key string, and has_api_key_for only checks the primary env var while the other two advertised alternatives are ignored. Both produce incorrect behavior for Volcengine users right now.

[...]

PR #2394, opened 2026-05-31
test(tui): include Volcengine in provider picker expectation

Hmbown:
## Summary
- include Volcengine Ark in the provider picker regression test expectation after #1993 landed
- unblocks macOS CI on provider-unrelated PRs that already contain current main

## Verification
- cargo fmt --all -- --check
- git diff --check
- cargo test -p codewhale-tui picker_lists_all_providers -- --nocapture

### DIFF
--- a/crates/tui/src/tui/provider_picker.rs
+++ b/crates/tui/src/tui/provider_picker.rs
@@ -474,6 +474,7 @@ mod tests {
                 "OpenAI-compatible",
                 "AtlasCloud",
                 "Wanjie Ark",
+                "Volcengine Ark",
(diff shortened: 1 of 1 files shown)


Hmbown:
Merged this tiny CI repair to unblock the queue after #1993 added Volcengine Ark. Local targeted proof passed (`cargo fmt --all -- --check`, `git diff --check`, and `cargo test -p codewhale-tui picker_lists_all_providers -- --nocapture`), and the branch checks that had completed were green; this fixes the same provider-picker drift that was causing unrelated macOS PR runs to fail.

PR #2318, opened 2026-05-28
feat(hooks): allow message_submit to transform submitted text

AresNing:
## Summary

Implements phase 1 of #1364: mutable `message_submit` hooks.

- Adds a structured stdin/stdout path for non-background `message_submit` hooks
- Allows hooks to replace submitted text via stdout JSON: `{"text":"..."}`
- Allows hooks to block submission with exit code `2`
- Keeps background `message_submit` hooks observer-only
- Preserves existing env vars and keeps `shell_env` stdout `KEY=VALUE` semantics unchanged
- Applies transformed text consistently to history, file mention expansion, API messages, and engine dispatch
- Adds an RFC documenting the planned PR split for the remaining `turn_end` and subagent lifecycle work

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets --all-features`
  - Passed with two pre-existing clippy warnings in unrelated files:
    - `crates/tui/src/commands/config.rs:476`
    - `crates/tui/src/runtime_log.rs:177`
- [x] `cargo test --workspace --all-features`
  - Ran with isolated `HOME` plus existing Cargo/Rustup caches to avoid user-global AGENTS/skills prompt injection in prompt tests.
- [x] `cargo test -p codewhale-tui --bin codewhale-tui hooks::tests::`
- [x] `cargo test -p codewhale-tui --bin codewhale-tui dispatch_user_message_`
- [x] `cargo check`

## Checklist

- [x] Updated docs or comments as needed
- [x] Added or updated tests where relevant
- [ ] Verified TUI behavior manually if UI changes

## Non-goals

- Does not implement `turn_end` / `agent_idle`
- Does not implement subagent lifecycle hooks
- Does not mutate tool args or tool results
- Does not apply JSON stdout semantics to all hook events

<details open><summary><h3>Greptile Summary</h3></summary>

This PR implements phase 1 of #1364, upgrading `message_submit` hooks from observer-only to a mutable pipeline: non-background hooks now receive the submitted text as a JSON stdin payload and can replace it via `{"text":"..."}` stdout or block submission with exit code `2`. The deadlock risk from synchronous stdin writes is resolved by moving all pipe I/O (stdin write, stdout read, stderr read) onto dedicated threads, with pipe-reader joins deliberately skipped on timeout to honour the configured hook deadline.

[...]

PR #2434, opened 2026-05-31
feat(hooks): harvest message_submit transform hooks

### LINKED ISSUE #1364: Hooks need mutation rights on user submit and a turn-end event

Hmbown:
Summary:
- Harvests the phase-1 `message_submit` transform hook work from #2318.
- Allows non-background `message_submit` hooks to replace submitted text with stdout JSON, block with exit code 2, or continue as observer-only background hooks.
- Preserves the existing `shell_env` stdout contract and routes transformed text through history, file mention expansion, API messages, and engine dispatch.
- Hardens the continued-failure path so `continue_on_error = true` surfaces stderr/stdout as a transient TUI warning instead of failing silently.

Credit:
- Original implementation by @AresNing in #2318, preserved as the first two commits on this branch.
- Thanks @esinecan for #1364 and the concrete portable-extension use case, and @LeXwDeX for the compatibility notes in that issue.

Scope:
- Phase 1 only: mutable `message_submit` hooks.
- Does not close #1364; `turn_end` and subagent lifecycle hooks remain follow-up work.

Issues:
- Refs #1364
- Supersedes the mergeable phase-1 slice of #2318 if this lands.

Validation:
- `cargo fmt --all`
- `git diff --check`
- `CARGO_TARGET_DIR=/Volumes/VIXinSSD/codewhale-target/harvest2318 cargo test -p codewhale-tui hooks::tests::message_submit --all-features --locked`
- `CARGO_TARGET_DIR=/Volumes/VIXinSSD/codewhale-target/harvest2318 cargo test -p codewhale-tui dispatch_user_message_surfaces_continued_message_submit_stderr --all-features --locked`
- `CARGO_TARGET_DIR=/Volumes/VIXinSSD/codewhale-target/harvest2318 cargo test -p codewhale-tui dispatch_user_message_uses_transformed_message_submit_text --all-features --locked`
- `CARGO_TARGET_DIR=/Volumes/VIXinSSD/codewhale-target/harvest2318 cargo test -p codewhale-tui dispatch_user_message_blocked_by_message_submit_hook_does_not_start_turn --all-features --locked`

### DIFF
--- a/crates/tui/src/commands/hooks.rs
+++ b/crates/tui/src/commands/hooks.rs
@@ -45,7 +45,7 @@ fn events() -> CommandResult {
         (HookEvent::SessionEnd, "fires once on graceful shutdown"),
         (
             HookEvent::MessageSubmit,
-            "fires when the user submits a turn (before model dispatch)",
+            "fires before model dispatch; can transform or block submitted text",
         ),
         (
             HookEvent::ToolCallBefore,
--- a/crates/tui/src/hooks.rs
+++ b/crates/tui/src/hooks.rs
@@ -14,8 +14,9 @@
 #[allow(unused_imports)]
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
+use serde_json::json;
 use std::collections::HashMap;
-use std::io::Read;
+use std::io::{Read, Write};
 use std::path::PathBuf;
 use std::process::{Command, Stdio};
 use std::time::{Duration, Instant};
@@ -423,6 +424,24 @@ pub struct HookResult {
     pub error: Option<String>,
 }
 
+/// Result of running mutable `message_submit` hooks.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MessageSubmitOutcome {
+    /// No hook changed the submitted text.
+    Unchanged,
+    /// One or more hooks replaced the submitted text.
+    Replaced(String),
+    /// A hook intentionally blocked the submission.
+    Blocked { reason: String },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum MessageSubmitStdout {
+    Unchanged,
+    Replaced(String),
+    Invalid(String),
+}
+
 /// Executor for running hooks
 #[derive(Debug, Clone)]
 pub struct HookExecutor {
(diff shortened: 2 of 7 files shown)


esinecan:
Amazing everyone 🫶 I will be putting it to use right away

PR #2458, opened 2026-05-31
docs(protocol): add doc comments to all public types

HUQIANTAO:
Add doc comments to all public types in the protocol crate, covering all request/response types, event frames, tool payloads, and approval model.


Hmbown:
Thanks for the protocol docs. I pushed a maintainer follow-up that runs rustfmt and merges current main into the branch so the already-landed secrets docs are no longer replayed as stale stacked context. Local checks passed in the PR worktree: cargo fmt --all --check, cargo test -p codewhale-protocol --all-features --locked, and git diff --check.

PR #2493, opened 2026-06-01
 fix(config): allow missing tui status items

### LINKED ISSUE #2483: bug(config): `tui.status_items` is wrongly required — config.toml fails to parse when `[tui]` exists without it
## Summary

On v0.8.48, launching `codewhale-tui` fails to start if `config.toml` has a `[tui]` section that does not include a `status_items` key:

```
Error: Failed to parse config file: ~/.deepseek/config.toml

Caused by:
    TOML parse error at line 1, column 1
      |
    1 | api_key = "sk-..."
      | ^
    missing field `status_items`
```

This breaks any pre-existing config carried over from older releases (e.g. after the DeepSeek-TUI → CodeWhale upgrade), since older `[tui]` blocks never wrote `status_items`.

## Repro

Minimal `config.toml`:

```toml
api_key = "sk-xxxx"

[tui]
locale = "zh-Hans"
mouse_capture = false
```

Run `codewhale-tui` (or `codewhale doctor`) → parse fails with `missing field status_items`.

## Root cause

`crates/tui/src/config.rs` (around line 669):

```rust
#[serde(deserialize_with = "deser_status_items")]
pub status_items: Option<Vec<StatusItem>>,
```

The field is *intended* to be optional — the doc comment says "`None` (the field missing from `config.toml`) means use the built-in default order", and `deser_status_items` itself starts with `Option::deserialize(deserializer)?`, so it already handles absence.

However, serde treats a field with `#[serde(deserialize_with = ...)]` as **required** unless `#[serde(default)]` is also present. The custom deserializer never gets a chance to return `None` because serde errors on the missing key first.

## Fix

Add `default` to the serde attribute (one line):

```rust
#[serde(default, deserialize_with = "deser_status_items")]
pub status_items: Option<Vec<StatusItem>>,
```

Verified locally against v0.8.48 (`e0160cc6`): with this change the same config parses cleanly, `status_items` correctly falls back to `None` (default footer order), and `codewhale doctor` loads the config without error.

## Notes

- The user-side workaround (`status_items = []` under `[tui]`) is not equivalent — per the docs, `Some([])` means "show nothing in the footer", whereas a missing field should mean "use the default order". Only the `#[serde(default)]` fix restores the documented behavior.
- Other `Option<...>` fields in `TuiConfig` that lack a custom `deserialize_with` are unaffected (serde defaults bare `Option` fields to `None`); the bug is specific to the `deserialize_with` + no-`default` combination.

Environment: v0.8.48 (`e0160cc6f0ba`), rustc 1.95.0, Linux.

reidliu41:
## Summary

  Fixes #2483.

  Configs with a `[tui]` table failed to parse unless `status_items` was also present:

  ```text
  missing field `status_items`
```

  That broke older or hand-written configs such as:

  [tui]
  locale = "zh-Hans"
  mouse_capture = false

  status_items is intended to be optional: missing means use the default footer order, while an empty list means
  hide all footer items. The custom deserializer needed serde(default) so serde can pass missing fields through as
  None.

## Testing

## Checklist

Adds `serde(default)` to the `status_items` field on `TuiConfig` so that configs containing a `[tui]` section without `status_items` no longer fail to parse with "missing field `status_items`".

- The one-line change is the correct, idiomatic fix: combining `serde(default)` with `deserialize_with` tells serde to produce `None` (the `Option`'s default) when the field is absent, while still routing present values through the custom `deser_status_items` deserializer.
- A new unit test validates the previously broken case directly.
</details>

Safe to merge — the change is minimal and correct, fixing a parse regression without altering any existing deserialization path.

Adding serde(default) to an Option field is the idiomatic fix for this class of serde issue. When the field is absent serde now yields None directly; when it is present the custom deser_status_items function is called exactly as before. The new test exercises the previously failing case and the existing test suite continues to cover the non-missing path. No logic changed beyond the missing-field handling.

No files require special attention.
</details>

| Filename | Overview |
|----------|----------|
| crates/tui/src/config.rs | Adds `serde(default)` to `TuiConfig::status_items` and a regression test; the fix correctly handles the missing-field case without touching the custom deserializer logic. |

</details>

</details>

<details open><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A{status_items field\npresent in TOML?} -->|No - field absent| B["serde default kicks in\nfield value = None"]
    A -->|Yes - field present| C["Call deser_status_items"]
    C --> D["Deserialize as Option of Vec of String"]
    D --> E["Filter unknown variants\nvia StatusItem::from_key"]
    E --> G["Return Some Vec of StatusItem"]
    B --> H["status_items = None\nuse built-in default order"]
    G --> I["status_items = Some items\nuse configured order"]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(config): allow missing tui status it..."](https://github.com/hmbown/codewhale/commit/b0e3f29bf5e506d8eee2e5d75b5a6da3604e5a50) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34875244)</sub>

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -666,7 +666,7 @@ pub struct TuiConfig {
     ///
     /// Edited interactively via `/statusline`; persisted to `tui.status_items`
     /// in `~/.deepseek/config.toml`.
-    #[serde(deserialize_with = "deser_status_items")]
+    #[serde(default, deserialize_with = "deser_status_items")]
     pub status_items: Option<Vec<StatusItem>>,
     /// Emit OSC 8 hyperlink escape sequences around URLs in the transcript so
     /// supporting terminals (iTerm2, Terminal.app 13+, Ghostty, Kitty,
@@ -8813,5 +8813,15 @@ model = "deepseek-ai/deepseek-v4-pro"
         assert_eq!(items[1], StatusItem::Model);
         assert_eq!(items[2], StatusItem::Cost);
         assert_eq!(items[3], StatusItem::Status);
+    }
+
+    #[test]
+    fn status_items_deser_allows_missing_field() {
+        let toml_str = r#"
+            locale = "zh-Hans"
+            mouse_capture = false
+        "#;
+        let tui: TuiConfig = toml::from_str(toml_str).expect("missing status_items should parse");
+        assert_eq!(tui.status_items, None);
     }
 }


Hmbown:
Thanks @reidliu41 — harvested this config compatibility fix into main for v0.8.49 in c1f74b3b2, covering issue #2483. Closing this PR as integrated and credited in the changelog.

PR #2499, opened 2026-06-01
feat(tools): byte-level schema canonicalize for prefix-cache stability

HUQIANTAO:
## Summary

When MCP servers return tool schemas, the field order within each schema object and the order of entries in `required` / `dependentRequired` arrays can vary across reconnections. This causes the serialized tool catalog bytes to change even when the logical schema is unchanged, busting DeepSeek's KV prefix cache.

This PR adds `schema_canonicalize::canonicalize_schema` which recursively:

- Sorts every `required` array alphabetically
- Sorts every `dependentRequired` sub-array alphabetically
- Rebuilds object keys in alphabetical order
- Recurses into all nested objects and arrays

The canonicalize step runs after `schema_sanitize` in `build_api_tools`, so each tool's `input_schema` is first cleaned then byte-stabilized. The existing `OnceLock` api_cache pins the result, ensuring the tool catalog bytes are identical across reads and across process restarts.

## Motivation

This is the highest-value, lowest-risk optimization identified in the cache audit. The `serde_json::Value::Object` backed by `IndexMap` (via `preserve_order` feature) preserves insertion order from the MCP server, which can differ between reconnections. By canonicalizing after sanitize, we guarantee that the same logical schema always produces the same bytes, preventing unnecessary prefix cache busts.

## Changes

- **New file**: `crates/tui/src/tools/schema_canonicalize.rs` (199 lines including 8 unit tests)
- **Modified**: `crates/tui/src/tools/registry.rs` - add `canonicalize_schema` call after `sanitize` in `build_api_tools`
- **Modified**: `crates/tui/src/tools/mod.rs` - add module declaration

## Testing

8 unit tests covering:
1. `sorts_required_array` - basic required array sorting
2. `equivalent_ordering_matches` - different field orders produce identical bytes
3. `sorts_dependent_required` - dependentRequired sub-array sorting
4. `recursive_into_properties` - nested schema canonicalization
5. `preserves_non_required_array_order` - non-required arrays keep semantic order
6. `handles_empty_schema` - empty object edge case
7. `handles_deeply_nested` - deeply nested schema recursion
8. `key_order_is_alphabetical_after_canonicalize` - object key ordering

All existing tests continue to pass. No new dependencies added (uses existing `serde_json`).

## Risk Assessment

Extremely low risk:
- Zero new dependencies
- Only modifies the tool schema serialization path (already cached by `OnceLock`)
- No API changes
- No behavior changes for providers that don't use `preserve_order`
- Fail-safe: if canonicalize somehow breaks a schema, `schema_sanitize` already runs first

### DIFF
--- a/crates/tui/src/tools/mod.rs
+++ b/crates/tui/src/tools/mod.rs
@@ -41,6 +41,7 @@ pub mod remember;
 pub mod revert_turn;
 pub mod review;
 pub mod rlm;
+pub mod schema_canonicalize;
 pub mod schema_sanitize;
 pub mod search;
 pub mod shell;
--- a/crates/tui/src/tools/registry.rs
+++ b/crates/tui/src/tools/registry.rs
@@ -16,6 +16,7 @@ use serde_json::Value;
 use crate::client::DeepSeekClient;
 use crate::models::Tool;
 
+use super::schema_canonicalize;
 use super::schema_sanitize;
 use super::spec::{
     ApprovalRequirement, ToolCapability, ToolContext, ToolError, ToolResult, ToolSpec,
@@ -224,6 +225,7 @@ impl ToolRegistry {
             .map(|tool| {
                 let mut schema = tool.input_schema();
                 schema_sanitize::sanitize(&mut schema);
+                schema_canonicalize::canonicalize_schema(&mut schema);
                 Tool {
                     tool_type: None,
                     name: tool.name().to_string(),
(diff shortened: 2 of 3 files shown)


Hmbown:
Harvested into `codex/v0.8.50-triage`. This was a good low-risk prefix-cache stability slice, and I pulled it in for the release branch.

I fixed the formatting failure locally with `cargo fmt --all` and amended the cherry-pick while preserving authorship. Focused schema canonicalization tests and `codewhale-tui` clippy pass on the triage branch.

HUQIANTAO:
Closing: this slice was harvested upstream (per the maintainer comments) — the work is in main, no need to keep the open PR alive. Thanks for the review!

PR #2501, opened 2026-06-01
feat(client): add in-process LLM response LRU cache

HUQIANTAO:
## Summary

Add an in-process LRU cache for LLM responses to deduplicate identical non-streaming, tool-free requests. When the same request is sent multiple times (model retries, app-server batch processing, user resubmissions), the cache returns the cached response immediately without an API call.

## Motivation

Codewhale currently has zero client-side response caching. Every identical request triggers a full API round-trip, adding 1-3 seconds of latency and consuming tokens. For common scenarios like model retries after network errors or repeated queries in app-server batch processing, this represents a significant optimization opportunity.

## Changes

- **New dependency**: `lru = "0.12"` (single file, zero transitive dependencies, pure std)
- **New file**: `crates/tui/src/llm_response_cache.rs` (262 lines including 7 unit tests)
- **Modified**: `crates/tui/src/client/chat.rs` - Integrate cache lookup/put in `create_message_chat`
- **Modified**: `crates/tui/src/main.rs` - Add module declaration

### Design

**Scope**: Only non-streaming, tool-free requests are cached. Streaming responses and tool-carrying requests are excluded to avoid complexity and side-effect issues.

**Cache key**: SHA-256(provider | model | base_url | system_hash | messages_hash)

**Capacity**: 256 entries (~1 MB at 4 KB/response average)

**Singleton**: Global `OnceLock<LlmResponseCache>` initialized on first use

### Integration

The cache is integrated into `create_message_chat()`:

1. Before the API call: compute cache key and check for a hit
2. If hit: return cached response immediately (zero latency)
3. If miss: make API call, cache the response, return it

Cache hits still count usage tokens for accurate cost tracking.

## Testing

7 unit tests covering:
1. `make_key_different_inputs_produce_different_keys` - Key uniqueness
2. `make_key_same_inputs_produce_same_key` - Key determinism
3. `put_and_get_returns_cached_response` - Basic cache hit
4. `get_miss_returns_none` - Cache miss
5. `capacity_evicts_oldest` - LRU eviction
6. `hit_rate_tracks_correctly` - Hit/miss counters
7. `stats_reports_size_and_capacity` - Stats snapshot

All existing tests continue to pass.

## Expected Impact

- **Latency**: Zero latency on cache hits (vs 1-3s for API calls)
- **API calls**: 30-50% reduction for repeated queries
- **Memory**: ~1 MB overhead (256 entries × 4 KB average)
- **Cost**: Proportional reduction in token costs for cached requests

## Risk Assessment

Low risk:
- Cache miss falls through to normal API path (no behavior change)
- Only caches non-streaming, tool-free requests (conservative scope)
- LRU eviction prevents unbounded memory growth
- Single new dependency (lru) with zero transitive deps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1011,6 +1011,7 @@ dependencies = [
  "ignore",
  "image",
  "libc",
+ "lru 0.12.5",
  "multimap",
  "objc2",
  "objc2-foundation",
@@ -1920,6 +1921,12 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "foldhash"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
+
+[[package]]
+name = "foldhash"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
@@ -2162,13 +2169,24 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
+version = "0.15.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
+dependencies = [
+ "allocator-api2",
+ "equivalent",
+ "foldhash 0.1.5",
+]
+
+[[package]]
+name = "hashbrown"
 version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
 dependencies = [
  "allocator-api2",
  "equivalent",
- "foldhash",
+ "foldhash 0.2.0",
 ]
 
 [[package]]
@@ -2922,6 +2940,15 @@ dependencies = [
  "rangemap",
  "time",
  "weezl",
+]
+
+[[package]]
+name = "lru"
+version = "0.12.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "234cf4f4a04dc1f57e24b96cc0cd600cf2af460d4161ac5ecdd0af8e1f3b2a38"
+dependencies = [
+ "hashbrown 0.15.5",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Hmbown:
Thanks for pushing this. I reviewed the code path for the v0.8.50 triage branch, and I’m going to leave this out of the release slice for now.

The direction is good, but the current cache key is too narrow for production: it keys on provider/model/base URL/system/messages, while `create_message_chat()` also shapes the wire request with fields like `max_tokens`, `temperature`, `top_p`, `tool_choice`, `thinking`, and `reasoning_effort`. Two requests with the same messages but different generation controls could incorrectly reuse one response. The PR also adds `lru = "0.12"` while the workspace already has `lru 0.16`, which splits the lockfile, and CI is currently red.

The safest next slice would be to key from the canonical wire body after provider/model normalization and reasoning/tool transformations, use the existing `lru` version, and add regression tests proving temperature/top_p/max_tokens/reasoning changes do not collide. This is very much in CodeWhale’s cache-maximal spirit; it just needs the correctness guardrails before it belongs in a release branch.

Hmbown:
Thanks @HUQIANTAO. I reviewed this again for the v0.9.0 stewardship train. The canonical-body key placement is useful, so if we harvest a response-cache slice we should credit #2501 and this update.

I am not going to merge it as-is for the release branch yet. The remaining blockers are cache identity and behavior: the key needs provider/base URL/path suffix/account context in addition to the final wire body, the new `lru = "0.12"` would add a second LRU version while the current lock already has `lru 0.16.4`, repeated stochastic/tool-free calls would change semantics unless this is deterministic or explicit opt-in, and cached `MessageResponse` usage data could make spend accounting look like a fresh API call.

A future narrow harvest should keep the post-canonical-body placement idea in `crates/tui/src/client/chat.rs`, rewrite the cache key and eligibility rules, and add regressions for provider/base URL separation, sampling params, stochastic bypass or opt-in behavior, and usage accounting on hits.

Hmbown:
Update: I opened the narrow response-cache harvest as #2805 against the v0.9.0 stewardship branch. This keeps the useful #2501 canonical-wire-body idea, but tightens the release slice to explicit deterministic non-streaming/tool-free requests, includes provider/base URL/path/account context in the key, reuses the existing `lru 0.16` lock entry, and zeroes usage on cache hits so spend counters are not double-counted.

Leaving this source PR open until #2805 finishes checks and lands, then I will close it as superseded with credit preserved.

Hmbown:
Thanks again @HUQIANTAO. The safe response-cache slice from this PR landed in #2805, merged as `e03f2ab22509f9c6d2d7c1a8a6a42dd35ae7d457`, with credit preserved in the harvest PR/changelog and a mappable `Co-authored-by: HUQIANTAO <58421104+HUQIANTAO@users.noreply.github.com>` trailer.

The landed slice keeps the canonical-wire-body cache-key idea from #2501, while tightening eligibility to explicit deterministic non-streaming/tool-free requests, adding provider/base URL/path/account context to the key, using the existing `lru 0.16` lock entry, and zeroing usage on cache hits so local spend counters are not double-counted.

Closing this source PR as superseded by #2805. Thanks for pushing the cache direction and for updating the key placement after review.

PR #2518, opened 2026-06-01
refactor(client): enforce append-only invariant on Session.messages via compile-time visibility

HUQIANTAO:
## Summary

Make the `Session.messages` field private and expose controlled accessor methods that enforce the append-only invariant required by DeepSeek's KV prefix cache. This converts the existing soft invariant (documented in comments) into a hard compile-time invariant that prevents entire classes of prefix-cache-breaking mutations.

## Background

DeepSeek's KV prefix cache relies on the first N bytes of the message array remaining identical across turns. Any mutation that changes the byte sequence of historical messages (insert, remove, reorder) invalidates the cache from that point onwards, forcing a full re-prefill at miss prices.

The current codebase maintains this invariant through convention: `add_message()` appends, `messages_with_turn_metadata()` returns stored messages unchanged, and comments warn against insert/remove. But the `messages` field is `pub`, which means any future contributor can write `self.session.messages.insert(0, ...)` or `self.session.messages.remove(0)` to fix a bug, and the compiler will happily accept it. The prefix cache breakage would only surface as a performance regression in production, not as a test failure.

## What changed

The `messages` field is now private. Six controlled accessors replace direct field access:

| Method | Visibility | Purpose |
|--------|-----------|---------|
| `add_message(m)` | `pub` | Append-only path for new messages |
| `messages()` | `pub` | Read-only slice access |
| `messages_mut()` | `pub(crate)` | Mutable access for engine internals |
| `remove_oldest_message()` | `pub(crate)` | Controlled prefix trim during emergency overflow |
| `replace_messages(msgs)` | `pub(crate)` | Bulk replacement for compaction/purge |
| `truncate_messages(idx)` | `pub(crate)` | Tail truncation for emergency recovery |

All call sites in `engine.rs`, `turn_loop.rs`, and `capacity_flow.rs` were updated to use the new API. The `messages_with_turn_metadata()` function now calls `self.session.messages().to_vec()` instead of `self.session.messages.clone()`.

## Why compile-time enforcement matters

Comments and tests are soft guards: they catch violations after the fact, during code review or CI. A private field with restricted access is a hard guard: the compiler rejects invalid mutations before the code can even be tested. This matters especially for prefix cache stability because:

1. Cache breakage is invisible in tests (the model still responds, just slower).
2. The breakage compounds over turns (every subsequent turn pays the miss price).
3. The root cause is hard to diagnose (requires comparing byte sequences of the full message array).

## Testing

All 104 engine tests and 4 session tests pass unchanged. This is a pure visibility refactor with zero logic changes.

## Files changed

- `crates/tui/src/core/session.rs`: Make `messages` private, add 6 accessor methods
- `crates/tui/src/core/engine.rs`: Update ~25 call sites to use new API
- `crates/tui/src/core/engine/turn_loop.rs`: Update ~8 call sites to use new API
- `crates/tui/src/core/engine/capacity_flow.rs`: Update ~10 call sites to use new API
- `crates/tui/src/core/engine/tests.rs`: Update ~20 test assertions to use new API


HUQIANTAO:
Closing: the append-only invariant refactor is invasive (touches the engine, capacity_flow, turn_loop, and tests) and the rebase onto v0.8.52 main runs into major conflicts because upstream callers still access session.messages directly. The prefix-cache-stability argument is sound and worth keeping the branch alive for, but a follow-up PR will need a fresh design pass on the accessors to land cleanly. Thanks for the review!

PR #2504, opened 2026-06-01
[codex] v0.8.50 triage harvest

Hmbown:
## Summary

v0.8.50 triage harvest after v0.8.48/v0.8.49 release stewardship. This branch does **not** tag, publish, create a GitHub Release, or republish `deepseek-tui`.

## Newly Harvested In This Push

- #2573 from @idling11 for #2571: bounded foreground `exec_shell` reader drains so timeout/pipe-handle cleanup cannot wedge later tools behind the global tool lock. Local harvest also covers normal-exit inherited-pipe cleanup with a foreground orphan-pipe regression.
- #2045 from @ZhulongNT: Windows NSIS installer plus classroom/lab deployment checklist, with maintainer hardening so the release workflow builds `CodeWhaleSetup.exe` from the real Windows binaries, the installer compiles under NSIS 3.12, docs use live asset names, and PATH add/remove is exact-entry based. Refs #1987.
- #2521 from @cyq1017: `/context` uses the effective routed model window, so DeepSeek V4 routes show the 1M-token budget while legacy DeepSeek routes keep the 128K fallback.
- #2562 from @cyq1017 for #2561: npm wrapper `--version` prefers the installed binary version before falling back to package metadata.
- #2537 from @cyq1017 for #2533: truncated subagent tool calls and repeated truncated responses produce model-visible errors and are capped.
- #2559 from @cyq1017 for #2369: config migration from legacy DeepSeek paths now reports a user-visible notice.
- #2527 from @HUQIANTAO: sidebar panels use theme colors and force a full repaint after theme switches.
- #2563 from @cyq1017 for #2494: session listings include absolute timestamps.
- #2564 from @cyq1017 for #2556: documents the read-only receipt-export boundary.
- #2536 from @cyq1017 for #2380: turn metadata exposes the current effective model.
- #2569 from @lucaszhu-hue: harvested the safe AtlasCloud user value as exact provider-hinted `vendor/model-id` pass-through, without freezing a volatile provider catalog or blanket capability flags.
- Local release bump to 0.8.50 for test builds only.

## Additional Local Fixes In This Push

- Fixed multiline composer Up/Down behavior from the WeChat/Chinese UX report: holding Up/Down at the first/last line of a multiline draft no longer replaces the draft with prompt history.
- Clarified the English DeepSeek footer balance chip from `bal` to `balance`, after #2567 showed it could be confused with session spend.

## Earlier Harvested Work Still In Branch

[...]

PR #2519, opened 2026-06-01
test(client): add plan mode toggle byte-stability invariant test

HUQIANTAO:
## Summary

Add a dedicated test that verifies the tool catalog remains byte-stable across Plan mode toggles and deferred tool activations. This converts the existing soft invariant (documented in comments and verified by ordering tests) into a hard byte-level assertion that catches any future regression in catalog construction determinism.

## Background

DeepSeek's KV prefix cache includes the tools array in the immutable prefix. Any byte-level change in the tools array forces a full re-prefill on the next turn. The existing tests verify tool ordering (alphabetical within partitions, deferred tools at the tail), but they don't verify that the serialized bytes are identical across repeated catalog builds.

This matters for two scenarios:

1. **Mode toggles**: Users switch between Plan, Agent, and YOLO modes during a session. Each mode builds a different tool catalog (Plan excludes execution tools). If the catalog construction has any non-determinism (e.g., HashMap iteration order, timestamp-dependent logic), toggling between modes could produce different byte sequences for the same mode, busting the prefix cache.

2. **Deferred tool activation**: When ToolSearch activates a deferred tool mid-session, the tool must be appended to the tail of the catalog without reordering the head. The existing `active_tool_list_pushes_deferred_activations_to_the_tail` test verifies the ordering, but not the byte-level stability of the head.

### New test: `plan_mode_toggle_preserves_catalog_byte_stability`

Verifies three invariants:

1. **Same-mode determinism**: Building the catalog twice for Plan mode produces identical JSON bytes. Same for Agent mode. This catches any non-determinism in catalog construction.

2. **Cross-mode head stability**: Non-deferred tools common to Plan and Agent modes appear in the same order. Plan mode excludes execution tools, but the tools that are present in both modes must have stable byte positions.

3. **Deferred activation tail-append**: Activating a deferred tool mid-session appends to the tail without reordering the catalog head. The test builds a catalog with a deferred tool, activates it, and verifies the head prefix is preserved.

### New doc comment on `build_model_tool_catalog`

Documents the catalog-head stability invariant: the head of the catalog (non-deferred tools) must remain byte-identical across mode toggles for tools common to both modes.

## Testing

The new test passes. All existing tool catalog tests continue to pass.

## Files changed

- `crates/tui/src/core/engine/tool_catalog.rs`: Add doc comment documenting the catalog-head stability invariant
- `crates/tui/src/core/engine/tests.rs`: Add `plan_mode_toggle_preserves_catalog_byte_stability` test

### DIFF
--- a/crates/tui/src/core/engine/tests.rs
+++ b/crates/tui/src/core/engine/tests.rs
@@ -1051,6 +1051,137 @@ fn turn_tool_registry_builder_keeps_plan_mode_read_only_for_files() {
     );
 }
 
+/// Plan mode toggle must not change the byte representation of the tool
+/// catalog head. DeepSeek's KV prefix cache includes the tools array in
+/// the immutable prefix; if toggling between Plan and Agent mode changes
+/// the tool bytes, every mode switch forces a full re-prefill.
+///
+/// This test verifies two invariants:
+/// 1. Building the catalog twice for the same mode produces identical bytes.
+/// 2. The head of the catalog (non-deferred tools) preserves its order
+///    when deferred tools are activated mid-session.
+#[test]
+fn plan_mode_toggle_preserves_catalog_byte_stability() {
+    let always_load = HashSet::new();
+
+    // Build catalog for Plan mode twice — must be byte-identical.
+    let plan_native = vec![
+        api_tool("read_file"),
+        api_tool("list_dir"),
+        api_tool("write_file"),
+        api_tool("edit_file"),
+        api_tool("exec_shell"),
+    ];
+    let plan_mcp = vec![api_tool("mcp_search"), api_tool("mcp_write")];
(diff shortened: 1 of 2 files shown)


Hmbown:
Hey @HUQIANTAO — the plan mode byte-stability test has been harvested into v0.8.50 (#2504)! This is a great addition — catching catalog-head reordering before it busts the prefix cache is exactly the kind of defense-in-depth we want. The doc comment on \`build_model_tool_catalog\` documenting the stability invariant is a nice touch too. Thank you! 🐋

HUQIANTAO:
Closing: this slice was harvested upstream (per the maintainer comments) — the work is in main, no need to keep the open PR alive. Thanks for the review!

PR #2520, opened 2026-06-01
feat(client): add cross-session prompt base section disk cache

HUQIANTAO:
## Summary

Add a new `prompt_persist.rs` module that caches the immutable base section of the system prompt on disk for cross-session reuse. When the SHA-256 of the base section matches a cached entry, the entire base-section assembly can be skipped on session start, providing byte-identical bytes to the API immediately.

## Background

DeepSeek's KV prefix cache matches byte sequences from the start of the system prompt. The base section (mode prompt, project context, skills, context management, compaction template) is stable across sessions for the same workspace. Currently, every new session rebuilds this section from scratch, which means:

1. The base section bytes are re-serialized every session (CPU waste).
2. If any intermediate state changes (e.g., a skill file is touched between sessions), the bytes might differ slightly, busting the server-side cache.

By caching the base section on disk and reusing it when the hash matches, we ensure byte-identical output across sessions for the same workspace. This is the infrastructure layer; wiring it into the prompt assembly pipeline (splitting `base_section()` + `volatile_section()`) will be done in a follow-up.

### Storage layout

```text
~/.codewhale/prompt_cache/
  <system_hash>.bin   — the serialized base section text
  <system_hash>.meta  — JSON metadata (workspace path, mtime, timestamp)
```

### Cache key

The cache key is the SHA-256 of the base section text, computed by the existing `PrefixFingerprint::compute` mechanism. This ensures that any change in the base section content produces a different cache key.

### Invalidation

Cache entries are invalidated by:

1. **Hash mismatch**: If the base section text changes (e.g., project context updated), the SHA-256 changes and a new entry is created.
2. **Workspace mtime change**: If the workspace directory's mtime changes (files added/removed), the cache is considered stale even if the hash matches (guards against hash collisions).
3. **Age-based eviction**: Entries older than a configurable threshold are evicted lazily.

### API

Three public functions:

- `load_cached_base_section(hash, workspace) -> Option<String>`: Try to load a cached base section.
- `save_cached_base_section(hash, text, workspace)`: Save a base section to disk.
- `evict_stale_entries(max_age_secs)`: Clean up old cache entries.

## Why this matters

The DeepSeek service-side prefix cache has an entry lifetime. When the base section bytes are identical across sessions, the server can reuse its cached KV states for the entire base section, giving ~90% discount on cached tokens. Currently, each new session starts with a fresh base section that must be prefilled from scratch on the server side.

This is the single highest-leverage optimization for cross-session cache hit rate: the base section is the largest stable block in the system prompt, and making it byte-identical across sessions maximizes the server-side cache reuse.

## Testing

4 unit tests:

- `save_and_load_round_trip`: Verify save/load cycle works correctly.
- `load_returns_none_for_missing_cache`: Verify graceful handling of cache miss.
- `load_returns_none_for_wrong_workspace`: Verify workspace-scoped invalidation.
- `evict_preserves_fresh_entries`: Verify fresh entries survive eviction.

## Files changed

- `crates/tui/src/prompt_persist.rs`: New module (254 lines)
- `crates/tui/src/main.rs`: Register the new module

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -52,6 +52,7 @@ mod prefix_cache;
 mod pricing;
 mod project_context;
 mod project_doc;
+mod prompt_persist;
 mod prompt_zones;
 mod prompts;
 mod purge;
--- a/crates/tui/src/prompt_persist.rs
+++ b/crates/tui/src/prompt_persist.rs
@@ -1,0 +1,258 @@
+//! Cross-session persistence for the immutable base section of the system
+//! prompt.
+//!
+//! ## Why
+//!
+//! DeepSeek's KV prefix cache matches byte sequences from the start of the
+//! system prompt. The base section (mode prompt, project context, skills,
+//! context management, compaction template) is stable across sessions for
+//! the same workspace. By caching this section on disk and reusing it when
+//! the SHA-256 matches, we can skip the entire base-section assembly on
+//! session start and immediately provide byte-identical bytes to the API.
+//!
+//! This is especially valuable for the DeepSeek service-side prefix cache:
+//! when the base section bytes are identical across sessions, the server
+//! can reuse its cached KV states for the entire base section, giving
+//! ~90% discount on cached tokens.
+//!
+//! ## Cache layout
+//!
+//! ```text
+//! ~/.codewhale/prompt_cache/
+//!   <system_hash>.bin   — the serialized base section text
+//!   <system_hash>.meta  — JSON metadata (workspace path, mtime, timestamp)
+//! ```
+//!
+//! The cache key is the SHA-256 of the base section text, computed by
+//! `PrefixFingerprint::compute`. The metadata file includes the workspace
+//! path and its mtime so that workspace changes invalidate the cache even
+//! if the base section hash happens to collide (extremely unlikely with
+//! SHA-256, but cheap to guard against).
+
+use std::fs;
+use std::path::{Path, PathBuf};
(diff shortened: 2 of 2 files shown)


Hmbown:
Thanks @HUQIANTAO. I reviewed this for the v0.9.0 stewardship branch and am closing it as superseded rather than merging it as-is.

The cross-session prompt-base cache idea is useful, but this branch adds `prompt_persist` infrastructure that is not wired into prompt assembly, writes tests through the real `~/.codewhale/prompt_cache` path, and now bypasses the newer prompt-cache surfaces that landed in #2801 (`CacheWarmupKey`, `PromptInspection`, request-time runtime metadata, and the existing project-context cache path).

If we rebuild persistent prompt-base caching later, the safer shape is to build around those current surfaces, use `codewhale_config::codewhale_home()` with isolated test roots and atomic writes, and key by canonical workspace/runtime prompt identity. We should credit #2520 and your persistence proposal if that future disk-cache slice lands.

Hmbown:
Closing this now as superseded, following the stewardship note above. The persistence/cache idea is still useful and remains credited to @HUQIANTAO in the v0.9 trail, but this branch is not the shape we should merge into the current release train because it is unwired and predates the newer #2801 prompt-cache/inspection surfaces.

If we rebuild disk-backed prompt-base caching later, we should credit #2520 and use the current `CacheWarmupKey` / `PromptInspection` / runtime-metadata path with isolated test roots and atomic writes.

PR #2527, opened 2026-06-01
fix(tui): sidebar panels now use theme colors instead of hardcoded Whale dark palette

HUQIANTAO:
## Summary

This patch fixes an inconsistency in the TUI sidebar rendering where panel content remained in the original theme's colors after a theme switch, while the section chrome (borders, background, title) correctly updated. The root cause was a split color-authority architecture where sidebar section frames followed the active theme but panel body text continued to reference hardcoded `palette::*` constants from the Whale dark palette.

### Symptom

When a user switches themes via `/theme` or the theme picker (e.g., from Matrix to Claude, from Claude to Dracula), the right sidebar panels including Tasks, Work, Agents, and Context would persist their original colors. The section frames (borders, background) would update immediately, creating a jarring visual mismatch where the outer chrome had changed but the inner content had not. Theme-switching to Matrix from a light theme would appear to work correctly only because Matrix happens to share similar color values with the Whale dark palette constants.

The sidebar content would eventually correct itself — either after the first conversation turn triggered a full transcript repaint, or after a terminal window resize event forced a complete buffer invalidation. This timing-dependent behavior made the bug appear intermittent and difficult to reproduce reliably.

### Root Cause Analysis

The sidebar rendering pipeline has two independent color-resolution paths that diverged during earlier theme-architecture work:

**Path A (correct):** `render_sidebar_section()` in `sidebar.rs:98` resolves section chrome colors (border, background, title, padding) through `deepseek_theme::Theme::for_palette_mode(app.ui_theme.mode)`. This path correctly follows theme changes because it reads from the active `UiTheme`'s `PaletteMode` and maps it to a corresponding `deepseek_theme::Theme`.

**Path B (broken):** Each panel-body rendering function (`task_panel_lines()`, `work_panel_lines()`, `subagent_panel_lines()`, `render_context_panel()`, `tool_status_marker()`, `agent_status_marker()`) directly referenced `palette::DEEPSEEK_SKY`, `palette::TEXT_MUTED`, `palette::TEXT_DIM`, `palette::STATUS_WARNING`, `palette::STATUS_SUCCESS`, `palette::STATUS_ERROR`, and `palette::STATUS_RUNNING`. These are compile-time constants from the Whale dark palette that never change regardless of the active theme.

The reason this discrepancy persisted without detection is that the theme picker provides live preview by swapping `app.ui_theme` and redrawing — but the hardcoded constants are evaluated at the cell-painting level, not at the theme-resolution level. The `ColorCompatBackend` cell remapper in `color_compat.rs` handles transcript-area cell remapping but does not intercept sidebar panel rendering because the sidebar uses direct `Span::styled()` calls that bypass the compat layer.

### Detailed Color Mapping

Each hardcoded constant and its theme-aware replacement:

| Constant | Field | Semantic Role |
|---|---|---|
| `DEEPSEEK_SKY` / `DEEPSEEK_BLUE` | `theme.accent_primary` | Primary accent for section labels, link indicators |
| `TEXT_MUTED` | `theme.text_muted` | Secondary descriptive text in panel rows |
| `TEXT_DIM` | `theme.text_dim` | Lowest-contrast text, status metadata |
| `TEXT_SOFT` | `theme.text_soft` | Soft-emphasis body text |
| `STATUS_WARNING` | `theme.warning` | Warning/amber state for task cards |
| `STATUS_SUCCESS` | `theme.success` | Success/green state for completed items |
| `STATUS_ERROR` | `theme.error_fg` | Error/red state for failed items |
| `STATUS_RUNNING` / `STATUS_LIVE` | `theme.tool_running` | Active/running indicator for in-progress tools |

### Changes

All sidebar panel rendering functions now accept a `&UiTheme` parameter and resolve colors from the active theme's fields. The theme reference flows through the call chain:

`render_sidebar_section()` (receives `&UiTheme` from `draw_sidebar_inner()` which reads `app.ui_theme`) → passes theme to:
- `task_panel_lines(&self, theme)`
- `work_panel_lines(&self, theme)`
- `subagent_panel_lines(&self, theme)`
- `render_context_panel(&self, theme)`
- `tool_status_marker(theme)` — helper for inline status indicators
- `agent_status_marker(theme)` — helper for agent card status badges

### Removed Dead Code

The original `cursor_position_from_thread_name()` function was already replaced in a prior commit with `thread_name_for_cursor()` but was left orphaned with a dead-code annotation. This commit removes it entirely to satisfy clippy's `dead_code` check.

## Future Direction: User-Defined Custom Themes

Tracking issue: will be opened as a follow-up PR referencing this one.

### Motivation

The current theme system supports 12 built-in presets selected from a compile-time enum. Users with specific visual preferences (accessibility needs, brand alignment, personal aesthetic) cannot adjust colors beyond picking one of the fixed options. Terminal users, especially those working in custom color schemes or with specific contrast requirements, benefit from the ability to define personal themes that persist across application updates.

### Planned Architecture

The proposal introduces a custom-theme subsystem with the following properties:

1. **Disk storage:** Custom themes live as JSON files under `~/.codewhale/themes/<name>.json`. Each file is a complete `UiTheme` definition serialized with hex-color-string fields. This directory is not part of the application bundle and survives version upgrades.

2. **Runtime loading:** At application startup, `palette::load_custom_themes()` scans the themes directory and populates a `HashMap<String, UiTheme>` registry. These themes are appended to the dynamic `selectable_themes()` list after the 12 built-in entries.

3. **Guided creation flow:**
   - User triggers `/theme new` or selects "Create Custom Theme" from the theme picker
   - CodeWhale prompts the user to describe their desired theme in natural language (e.g., "a warm amber and cream library theme with brown accents, moderate contrast, dark mode")
   - CodeWhale generates a complete `UiTheme` JSON definition from the description, reasoning across all 40 color fields based on the user's stated preferences
   - The generated theme is applied as a live preview, allowing the user to inspect every UI surface
   - If satisfied, the user confirms with Enter and the theme is persisted to `~/.codewhale/themes/<name>.json` and appears in the picker
   - If unsatisfied, the user presses Esc to discard and may iterate on the description

4. **Deletion:** In the theme picker, custom themes display a `[Del]` affordance in the bottom hint bar. Pressing the Delete key on a highlighted custom theme prompts for confirmation and, if confirmed, removes the JSON file from disk and the entry from the picker list. Built-in themes cannot be deleted. A `/theme delete <name>` slash command provides an alternative CLI path.

5. **No compilation required:** The serialized `UiTheme` JSON format means custom themes are pure data, not code. Users do not need a Rust toolchain or a rebuild cycle. The theme definition is loaded and deserialized at runtime into the same `UiTheme` struct used by built-in themes.

6. **Version resilience:** Custom themes are stored outside the application directory. When CodeWhale is updated, the `~/.codewhale/themes/` directory is never touched by the installer or updater. If a future version adds new color fields to `UiTheme`, the deserializer applies sensible defaults for missing keys and emits a warning rather than failing.

### Technical Design Decisions

- **`UiThemeCustom` intermediate type:** A separate serde-enabled struct (`UiThemeCustom`) bridges the JSON representation and the internal `UiTheme`. Each color field is a hex string (`"#RRGGBB"`) rather than a `ratatui::style::Color` enum variant, ensuring human-readability and editability. The conversion is infallible because hex string parsing is well-defined.

- **`ThemeId::Custom(String)` variant:** Extends the existing enum rather than replacing it. Built-in themes continue to use their compile-time variant, preserving all existing match arms and avoiding a cascade of downstream changes. The custom variant carries the theme's unique name (the filename stem).

- **Delete-key binding in the picker:** The theme picker already handles keyboard navigation with Vim-style `j`/`k` alongside arrow keys. Adding a Delete handler follows the same pattern and only activates when the cursor is on a custom theme entry, preventing accidental deletion of built-in presets.

- **No new crate dependencies:** The implementation reuses the existing `serde` and `serde_json` dependencies already present in `Cargo.toml`. File I/O uses `std::fs` with atomic-write semantics (write to temp file, rename) to prevent corruption on crash.

### Files to be Modified (Future PR)

- `crates/tui/src/palette.rs` — `ThemeId::Custom(String)`, `UiThemeCustom`, `load_custom_themes()`, dynamic theme list
- `crates/tui/src/commands/config.rs` — `/theme new`, `/theme delete`
- `crates/tui/src/tui/theme_picker.rs` — dynamic theme enumeration, Delete key handler, custom-theme affordance
- New: `crates/tui/src/custom_theme.rs` — theme JSON schema, file I/O, `UiThemeCustom` ↔ `UiTheme` conversion

## Files Changed

- `crates/tui/src/tui/sidebar.rs`: All panel rendering functions now accept `&UiTheme` and use theme fields instead of hardcoded `palette::*` constants. Removed dead `cursor_position_from_thread_name()` function.

## Testing

All 44 sidebar unit tests pass. Manual verification across all 12 themes confirms that sidebar panels now immediately reflect the active theme on switch, without requiring a conversation turn or window resize.

### DIFF
--- a/crates/tui/src/tui/app.rs
+++ b/crates/tui/src/tui/app.rs
@@ -1486,6 +1486,11 @@ pub struct App {
     pub session_started_at: chrono::DateTime<chrono::Utc>,
     /// Whether the UI needs to be redrawn.
     pub needs_redraw: bool,
+    /// When true, the next draw will be a full repaint (terminal clear +
+    /// all cells redrawn) instead of a ratatui incremental diff. Used by
+    /// theme switches where the diff engine may miss color-only changes
+    /// in sidebar cells that were previously rendered with palette constants.
+    pub force_next_full_repaint: bool,
     /// When the current thinking block started (for duration tracking).
     pub thinking_started_at: Option<Instant>,
     /// Whether context compaction is currently in progress.
@@ -2074,6 +2079,7 @@ impl App {
             decision_card: None,
             session_started_at: chrono::Utc::now(),
             needs_redraw: true,
+            force_next_full_repaint: false,
             thinking_started_at: None,
             is_compacting: false,
             is_purging: false,
--- a/crates/tui/src/tui/sidebar.rs
+++ b/crates/tui/src/tui/sidebar.rs
@@ -286,20 +286,21 @@ fn work_panel_lines(
     content_width: usize,
     max_rows: usize,
     palette_mode: palette::PaletteMode,
+    ui_theme: &palette::UiTheme,
 ) -> Vec<Line<'static>> {
     let theme = Theme::for_palette_mode(palette_mode);
     let mut lines: Vec<Line<'static>> = Vec::with_capacity(max_rows.max(4));
 
-    push_work_goal_lines(summary, content_width, max_rows, &mut lines);
+    push_work_goal_lines(summary, content_width, max_rows, &mut lines, ui_theme);
 
     if summary.state_updating && lines.len() < max_rows {
         lines.push(Line::from(Span::styled(
             "Work state updating...",
-            Style::default().fg(palette::TEXT_MUTED),
+            Style::default().fg(ui_theme.text_muted),
         )));
     }
 
-    push_work_checklist_lines(summary, content_width, max_rows, &mut lines);
+    push_work_checklist_lines(summary, content_width, max_rows, &mut lines, ui_theme);
     push_work_strategy_lines(summary, content_width, max_rows, &mut lines, &theme);
 
     if summary.cycle_count > 0 && lines.len() < max_rows {
@@ -309,14 +310,14 @@ fn work_panel_lines(
                 summary.cycle_count,
                 summary.cycle_count.saturating_add(1)
             ),
-            Style::default().fg(palette::TEXT_MUTED),
+            Style::default().fg(ui_theme.text_muted),
         )));
     }
 
     if lines.is_empty() {
         lines.push(Line::from(Span::styled(
             work_panel_empty_hint(content_width),
-            Style::default().fg(palette::TEXT_MUTED).italic(),
+            Style::default().fg(ui_theme.text_muted).italic(),
         )));
     }
 
@@ -328,6 +329,7 @@ fn push_work_goal_lines(
     content_width: usize,
     max_rows: usize,
     lines: &mut Vec<Line<'static>>,
+    theme: &palette::UiTheme,
 ) {
     let Some(objective) = summary.goal_objective.as_deref() else {
         return;
@@ -339,11 +341,11 @@ fn push_work_goal_lines(
     let icon = if summary.goal_completed { "✓" } else { "◆" };
     let status_style = if summary.goal_completed {
         Style::default()
-            .fg(palette::STATUS_SUCCESS)
+            .fg(theme.success)
             .add_modifier(ratatui::style::Modifier::BOLD)
     } else {
         Style::default()
-            .fg(palette::STATUS_WARNING)
+            .fg(theme.warning)
             .add_modifier(ratatui::style::Modifier::BOLD)
     };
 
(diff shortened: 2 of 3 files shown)


HUQIANTAO:
Closing: the sidebar theme-switch fix was already merged into main as commits 556e0b46 and b1cc344d (v0.8.50). The Windows-only test failure on this branch was environmental, not code-related. Thanks for the review!

PR #2555, opened 2026-06-02
test(mcp): close stale-session mock responses cleanly

cyq1017:
Problem
- The streamable HTTP stale-session retry test has failed on macOS CI in otherwise unrelated draft PRs with `connection closed before message completed`.
- The test mock server wrote complete HTTP responses and then let the socket drop immediately, which is sensitive to runner timing.

Change
- Flush and cleanly shut down the mock TCP stream after each response in that test.
- Keep the production MCP retry path unchanged.

Verification
- cargo fmt --check
- cargo test -p codewhale-tui --all-features --locked mcp::tests::streamable_http_stale_session_reconnects_and_retries_tool_call -- --nocapture
- Repeated the same targeted test 3 more times
- git diff --check

### DIFF
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -4460,6 +4460,12 @@ mod tests {
         use tokio::io::{AsyncReadExt, AsyncWriteExt};
         use tokio::net::TcpListener;
 
+        async fn write_response(socket: &mut tokio::net::TcpStream, response: &[u8]) {
+            socket.write_all(response).await.unwrap();
+            socket.flush().await.unwrap();
+            socket.shutdown().await.unwrap();
+        }
+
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @cyq1017, this is a good release-hardening cleanup. I harvested it into #2504 as `908a25d0f` after confirming the diff stays test-only and leaves the production MCP retry path unchanged.

Local verification on the release branch:
- `for i in 1 2 3; do cargo test -p codewhale-tui --all-features --locked mcp::tests::streamable_http_stale_session_reconnects_and_retries_tool_call -- --nocapture || exit 1; done`
- `git diff --check`

Hmbown:
This PR was harvested into v0.8.50. Closing as already landed.

PR #2559, opened 2026-06-02
fix(config): report legacy config migration

cyq1017:
## Problem

The first-run config migration copies `~/.deepseek/config.toml` into `~/.codewhale/config.toml`, but the successful migration is only logged. Users who keep editing the legacy file can miss that the active copy moved.

## Change

- return the copied legacy and primary config paths from `migrate_config_if_needed`
- print a one-time stderr notice when the migration actually copies the file
- keep the migration non-fatal and preserve the existing no-overwrite behavior

## Verification

- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1964,18 +1964,34 @@ pub fn default_config_path() -> Result<PathBuf> {
     Ok(primary)
 }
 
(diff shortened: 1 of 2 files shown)


Hmbown:

Issue #2494, opened 2026-06-01
mac+ item2 用户使用问题汇总


wywsoor:
macOs + item2 用户使用问题汇总：
以下问题是我自己的使用的过程遇见的问题，有的问题我查了好多地方没找到对应的解决方案，暂时决定停用 tui ，但是我希望我遇见的问题可以帮你们，话不多说开始罗列问题。

1 快捷键问题现在帮助里的快捷键都是win用户的快捷键，有好多东西macos对不上号。
2 会话交谈换行问题，我在交谈过程难免会使用复制一些带有换行符的消息，现在如果发送会直接变成多次提问直接发送。
3 终止提问的问题，比如我问了一句错的问题如何停止，目前没找都如何停止 ctrl + c 只会关闭会话 再次打开还是一样会继续刚才的错误问题
4 选择历史会话问题 我想知道我上周五最后一次会话说了啥我应该如何找回。
5 回滚机制问题 如果我发现交谈了几次越改越偏差 如果想回滚到指定节点使用 /restore(貌似是这个命令) 会提示10条带有id的选项，我如何对应那个是需要的回滚节点，并且我也找不到会话的id。
6 回滚只有10条可选项太少
7 我看可以把会话导出成为md文件 但是md文件里有没有对应id 并且内容也没有太大参考意义。

------------------------------

Summary of macOS + Item2 User Issues: The following are issues I encountered while using the app. For some problems, I searched extensively but couldn't find solutions, so I've temporarily decided to discontinue using Item2. However, I hope my experience can help you. Without further ado, here's a list of the issues:

1. Keyboard Shortcuts: The keyboard shortcuts in the help section are for Windows users, and many things don't match up for macOS.

2. Conversation Line Breaks: During conversations, I inevitably copy messages containing line breaks. Now, sending them results in multiple questions being sent simultaneously.

3. Terminating Questions: For example, how do I stop asking an incorrect question? I haven't found a way to stop it yet. Ctrl+C only closes the conversation; reopening it resumes the same incorrect question.

4. Selecting History Conversations: How can I retrieve what I said in my last conversation last Friday?

5. Rollback Mechanism Issue: If I find that the changes after several sessions are increasingly inconsistent, and I want to roll back to a specific node using `/restore` (or something similar), it will prompt me with 10 options, each with an ID. How do I identify which node I need to roll back to, and I can't find the session ID?

6. Only 10 Rollback Options are Too Few

[...]

PR #2563, opened 2026-06-02
fix(tui): show session timestamps in listings

cyq1017:
Problem
- Session listings only showed relative ages, which makes it hard to find a specific older conversation by date.
- The workspace-scoped resume paths also needed a clearer pointer that `--last`/`--continue` choose the current workspace, while explicit ids can resume other directories.

Change
- Show an absolute UTC last-updated timestamp alongside the relative age in `codewhale sessions` and the TUI session picker.
- Document the difference between all-session listing and current-workspace resume paths.

Verification
- `cargo test -p codewhale-tui absolute_updated_timestamp --all-features --locked -- --nocapture`
- `cargo test -p codewhale-tui session_picker --all-features --locked -- --nocapture`
- `git diff --check`

Refs #2494

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ codewhale doctor --json                           # machine-readable diagnostics
 codewhale setup --status                          # read-only setup status
 codewhale setup --tools --plugins                 # scaffold tool/plugin dirs
 codewhale models                                  # list live API models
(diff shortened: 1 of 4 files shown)


Hmbown:

Issue #2556, opened 2026-06-02
Proposal: read-only conformance receipt export for runtime thread/turn replay


ctxyao:
Hi CodeWhale maintainers,

I have been looking at the public CodeWhale runtime surface as a local supervisor integration point. The existing runtime model already has most of the ingredients that downstream tools would need for auditability: durable `ThreadRecord` / `TurnRecord` / `TurnItemRecord`, append-only events, usage aggregation, approval events, sandbox boundaries, and workspace rollback.

Would a small read-only receipt export fit the project roadmap?

The narrow goal would be to let local supervisors verify one completed turn without screen-scraping terminal output:

- reasoning-block custody without exporting raw reasoning by default
- tool-call to tool-result lineage
- cache/usage evidence when available, with explicit `unavailable` when provider-specific hit/miss is not stored
- event sequence lineage across replay/resume
- approval, sandbox, YOLO/auto-approve, command/file-change side-effect boundaries
- rollback/snapshot availability

One possible local-only surface:

```text
codewhale receipt export --thread <thread_id> --turn <turn_id> --format json
GET /v1/threads/{thread_id}/turns/{turn_id}/receipt
```

This should be read-only and reuse the existing local runtime auth boundary. It should not call providers, execute tools, write memory, write target files, or expose API keys. For privacy, raw reasoning should stay omitted by default; a receipt can use counts, hashes/refs, and item/event ids instead.

Example sketch:

```json
{
  "schema_id": "codewhale.conformance-receipt/v0",
  "thread": {
    "id": "thr_...",
    "model": "deepseek-v4-pro",
    "mode": "agent",
    "auto_approve": false
  },
  "turn": {
    "id": "turn_...",
    "status": "completed"
  },
  "reasoning_custody": {
    "raw_reasoning_exported": false,
    "reasoning_blocks_observed": true,
    "reasoning_block_hashes": ["sha256:..."]
  },
  "tool_lineage": {
    "tool_call_count": 1,
    "tool_result_count": 1,
    "unmatched_tool_call_ids": [],
    "unmatched_tool_result_ids": []
  },
  "cache_evidence": {
    "available": true,
    "cached_tokens": 0,
    "reasoning_tokens": 0,
    "prompt_cache_hit_tokens": null,
    "prompt_cache_miss_tokens": null
  },
  "source_event_lineage": {
    "first_seq": 1,
    "last_seq": 42,
    "event_count": 42,
    "missing_event_ranges": []
  },
[...]

PR #2564, opened 2026-06-02
docs(runtime): outline receipt export boundary

cyq1017:
Problem
- #2556 proposes a local read-only receipt export for completed runtime turns, but the contract needs a small docs-first boundary before code lands.
- The current runtime API docs list threads, turns, approvals, and events, but do not describe what a future receipt may claim or must avoid claiming.

Change
- Add `docs/RECEIPTS.md` with non-goals, candidate local surfaces, current durable data sources, a draft JSON shape, and deterministic builder rules.
- Link the protocol note from `docs/RUNTIME_API.md` and mark the receipt endpoint as proposed only.

Verification
- `git diff --cached --check`
- diff secret scan; only the existing safety phrase "secret leakage" appeared as a false-positive match
- Manual cross-check against `ThreadRecord`, `TurnRecord`, `TurnItemRecord`, and `RuntimeEventRecord`

Refs #2556

### DIFF
--- a/docs/RECEIPTS.md
+++ b/docs/RECEIPTS.md
@@ -1,0 +1,139 @@
+# Runtime Receipts
+
+This document sketches a future read-only receipt export for completed runtime
+turns. It is a protocol note, not an implemented endpoint.
+
+The goal is to let a local supervisor audit one completed turn without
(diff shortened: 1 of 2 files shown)


Hmbown:

Issue #2583, opened 2026-06-02
the engine have stopped error in v0.8.50 still


Artenx:
<img width="1626" height="1188" alt="Image" src="https://github.com/user-attachments/assets/c719e574-1c3b-474e-9a8b-7ffa958e37ec" />

Artenx:
<img width="2744" height="1240" alt="Image" src="https://github.com/user-attachments/assets/eea6970c-db50-4205-80a7-f6610e131cd0" />

gordonlu:
导致 "Turn stalled" 的原因应该是 engine 在 turn 中途意外退出了。有两种方式能确认具体原因：

---

**方法一：看崩溃报告（推荐，不需要复现）**

每次 engine 崩溃时都会自动保存一个报告文件。打开终端，跑：

```bash
ls ~/.codewhale/crashes/
```

如果有文件，跑这个查看内容（找类似文件名的文件，把文件名换成实际的）：

```bash
cat ~/.codewhale/crashes/{时间戳}-engine-event-loop.log
```

把内容发到这里，就能知道具体是哪行代码崩了。

如果没有这个目录或文件，说明 engine 可能不是代码崩溃，而是被系统kill掉的（比如内存不足 OOM）。

------

**方法二：看日志（需要再复现一次）**

先设置环境变量再启动：

```bash
RUST_LOG=panic=trace codewhale
```

正常使用直到 "Turn stalled" 再次出现。然后退出程序，在终端输出里找包含下面关键字的行：

```
ERROR panic: Task 'engine-event-loop' panicked at
```

如果能找到，把那一行和它附近的几行发出来。

---

Hmbown:
Hardened on the v0.8.51 branch (`claude/busy-mayer-b66rA`, commit `5249723`).

**What was happening:** a panic inside the turn loop unwound through `engine.run()` and was caught by `spawn_supervised("engine-event-loop")`, which wrote the crash dump and let the **entire** engine task exit. The UI never received `TurnComplete`, so it sat on "working"/"Turn stalled" forever, and every later turn was dead too — matching "the engine have stopped".

**Fix:** the turn call is now wrapped in `catch_unwind`. A panic surfaces as a *failed* `TurnComplete` with an actionable message instead of killing the loop, and the engine stays alive for the next turn. The crash dump is still written to `~/.codewhale/crashes/`.

[...]

PR #2585, opened 2026-06-02
fix: detect engine task death mid-turn and recover UI immediately

gordonlu:
## Problem

When the engine task panics (or exits unexpectedly) between `TurnStarted` and `TurnComplete`, the event channel `tx_event` is dropped and the mpsc channel closes silently. The UI's event loop uses `while let Ok(event) = rx.try_recv()` — on `Err(TryRecvError::Disconnected)` the loop simply exits without any recovery. The turn remains stuck with `is_loading=true` and `runtime_turn_status=Some("in_progress")` until the 300-second `TURN_STALL_WATCHDOG_TIMEOUT` fires in `reconcile_turn_liveness()`.

The root cause is in `spawn_supervised` (`utils.rs`): it wraps the future in `catch_unwind`, logs the panic, writes a dump file, and exits — but never sends an `EngineEvent::Error` over the channel. The UI has no way to know the engine died.

## Fix

Add a post-loop check for `TryRecvError::Disconnected` after the existing `while let Ok(event)` event processing loop (`ui.rs`). If the channel is disconnected while `app.is_loading` is true, the UI immediately:
- Finalizes in-flight streaming thinking / assistant / tool cells
- Resets `is_loading`, `runtime_turn_status`, `turn_started_at`, `dispatch_started_at`, etc.
- Pushes an error toast: "Engine process has terminated unexpectedly."

This reduces the recovery window from **300 seconds to ~1 frame** (~16ms).

## Testing

`engine_event_channel_disconnect_recovers_mid_turn_ui_state` — creates a real mpsc channel, drops the sender to simulate engine death, sets up the app in mid-turn state, applies the recovery logic, and verifies the UI state is fully cleaned up.

## Related

Fixes the symptom described in #2583. The root cause of *why* the engine panics can be diagnosed via crash dumps in `~/.codewhale/crashes/`.

This PR adds an immediate recovery path for when the engine task dies (panics or exits) mid-turn, eliminating the previous 300-second stall window. It hooks into the existing event-loop tick by checking `rx.is_closed()` after the `try_recv` drain loop and resets all in-flight UI state when a disconnect is detected while the app is in a loading/compacting/purging state.

- **`ui.rs`**: Adds a post-drain-loop disconnect guard inside the engine-event read scope; clears all per-turn app flags (`is_loading`, `is_compacting`, `is_purging`, streaming indices, agent progress, etc.), drains pending steers into the visible queue, pushes an error toast, and sets `needs_redraw`.
- **`tests.rs`**: Adds `engine_event_channel_disconnect_recovers_mid_turn_ui_state` which creates a real mpsc channel, drops the sender to simulate engine death, and verifies the resulting app state matches expectations.
</details>

Safe to merge — the disconnect check is one-shot and self-disabling, uses no await points inside the recovery block, and the guard condition prevents re-firing after cleanup.

The recovery block is logically correct: `rx.is_closed()` reliably returns `true` when all senders are dropped (confirmed by tokio 1.49 docs), the while-drain loop runs first so no buffered events are skipped, and the guard `(is_loading || is_compacting || is_purging)` ensures the block only fires when there is actually live turn state to clean up. Previously-flagged concerns around `needs_redraw` and pending-steer loss are addressed in this version.

No files require special attention beyond what was already discussed in prior review threads.
</details>

| Filename | Overview |
|----------|----------|
| crates/tui/src/tui/ui.rs | Adds a one-shot post-drain disconnect check using `rx.is_closed()`; correctly clears all per-turn UI state, drains pending steers, and triggers an immediate redraw. Guard condition prevents re-firing after the first recovery. |
| crates/tui/src/tui/ui/tests.rs | New test verifies the disconnect recovery behavior end-to-end. The test body inlines the production recovery logic verbatim rather than calling an extracted function, so future divergence won't be caught by the test. |

</details>

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant E as Engine Task
    participant CH as mpsc Channel
    participant UI as UI Event Loop
    participant APP as App State

    E->>CH: TurnStarted
    UI->>CH: try_recv() loop processes TurnStarted
    UI->>APP: "is_loading = true"

    Note over E: Panic / unexpected exit
    E--xCH: tx_event dropped (channel closed)

    loop Each UI tick
        UI->>CH: try_recv() drain loop exits
        UI->>CH: rx.is_closed()?
        CH-->>UI: true
        alt is_loading or is_compacting or is_purging
            UI->>APP: finalize streaming cells
            UI->>APP: reset all per-turn flags
            UI->>APP: drain pending_steers to queue
            UI->>APP: push error toast
            UI->>APP: "needs_redraw = true"
        end
    end
```
</details>

<sub>Reviews (6): Last reviewed commit: ["fix: preserve pending steers on engine d..."](https://github.com/hmbown/codewhale/commit/649d7bc9ab12f10a53edecd866f670bb180ff7c0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35105260)</sub>

### DIFF
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -2242,9 +2242,7 @@ async fn run_event_loop(
             // UI state immediately instead of waiting for the 300-second
             // TURN_STALL_WATCHDOG_TIMEOUT.  Use is_closed() rather than
             // try_recv() so the probe never consumes a valid event.
-            if (app.is_loading || app.is_compacting || app.is_purging)
-                && rx.is_closed()
-            {
+            if (app.is_loading || app.is_compacting || app.is_purging) && rx.is_closed() {
                 streaming_thinking::finalize_current(app);
                 app.finalize_streaming_assistant_as_interrupted();
                 app.finalize_active_cell_as_interrupted();
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -2262,6 +2262,14 @@ async fn run_event_loop(
                 app.runtime_turn_id = None;
                 app.dispatch_started_at = None;
                 app.user_scrolled_during_stream = false;
+
+                // Preserve pending steers: same hard-fail recovery as
+                // TurnComplete::Failed — surface them in the visible queue
+                // so they are not silently lost.
+                for msg in app.drain_pending_steers() {
+                    app.queue_message(msg);
+                }
+
                 app.push_status_toast(
                     "Engine process has terminated unexpectedly.",
                     StatusToastLevel::Error,
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -33,8 +33,6 @@ use tracing;
 #[cfg(target_os = "windows")]
 use windows::Win32::System::Console::{GetConsoleMode, GetStdHandle, SetConsoleMode};
 
-use tokio::sync::mpsc::error::TryRecvError;
-
 use crate::audit::log_sensitive_event;
 use crate::automation_manager::{AutomationManager, AutomationSchedulerConfig, spawn_scheduler};
 use crate::client::{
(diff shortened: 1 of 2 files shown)


Hmbown:
This was harvested into the v0.8.51 release — thank you for the contribution!

Issue #2488, opened 2026-06-01
当文件目录太深，>= 6 层的时候使用 @ 或者 control + p 无法检索到文件


simuusang:
## Steps to reproduce

1.
2.
3.

## Environment

- OS:
- codewhale version:
- Install method:
- `codewhale doctor` summary:
- Model/provider:
- Terminal app:
- Shell:

Hmbown:
Thanks @simuusang. This is useful timing-wise because v0.8.49 added configurable mention walk depth after v0.8.48: in `~/.codewhale/config.toml`, `[tui] mention_walk_depth = 0` disables the depth cap, or you can set a larger number than the default `6`.

I am leaving this open rather than closing it as “configuration exists,” because the product question is whether deep-file projects need a better default or clearer UI hint. If you can retest on v0.8.49 with `mention_walk_depth = 0`, that would tell us whether the remaining issue is only the default limit or also the search/ranking path.

Hmbown:
A UX slice for this is now in the v0.8.50 harvest branch #2504.

The branch harvested @cyq1017's #2530 as `a97675824` / `29f57665e`. Deep path-like `@` completion misses now show a targeted `mention_walk_depth` hint instead of the generic no-match status, so users can discover `/config set mention_walk_depth 0 --save` when they need unlimited depth. The actual autocomplete search behavior and default depth are unchanged.

Verified locally with:
- `cargo test -p codewhale-tui --all-features --locked try_autocomplete_file_mention_no_match -- --nocapture`
- `cargo clippy -p codewhale-tui --all-targets --all-features --locked -- -D warnings`
- `./scripts/release/check-versions.sh`

Live #2504 CI is green across Ubuntu, macOS, Windows, lint, version drift, CodeQL, mobile smoke, npm smoke, and GitGuardian.

Leaving this issue open because the branch adds discoverability for the depth cap, not a broader change to default search depth or indexing behavior. Thanks @simuusang for the clear report.

simuusang:
<img width="1403" height="1015" alt="Image" src="https://github.com/user-attachments/assets/a8f20dd5-51f4-428f-af04-b2cc533ed575" />

我更新到了0.8.50尝试搜索了一下还是不行，在我的项目里面有个叫 room_chat_shell.dart 的文件，通过搜索没办法直接搜到。

[...]

PR #2593, opened 2026-06-02
fix(tui): honor mention walk depth in file picker

cyq1017:
Problem
- `mention_walk_depth = 0` already disables the depth cap for `@` completions, but the Ctrl+P file picker still used its fixed depth-6 walk.
- That leaves deep files discoverable through one composer path but not the other.

Change
- Thread the configured mention walk depth into the file picker.
- Keep the default depth unchanged, and keep the existing candidate cap, ranking, relevance boosts, and ignore behavior.
- Update the configuration docs to state that the setting applies to both `@` completion and Ctrl+P.

Verification
- `cargo test -p codewhale-tui --all-features --locked picker_honors_configured_unlimited_walk_depth -- --nocapture`
- `cargo test -p codewhale-tui --all-features --locked file_picker -- --nocapture`
- `git diff --check`

Refs #2488

### DIFF
--- a/crates/tui/src/tui/file_picker.rs
+++ b/crates/tui/src/tui/file_picker.rs
@@ -1,10 +1,10 @@
 //! Fuzzy file-picker modal (Ctrl+P).
 //!
 //! Opens an overlay populated with workspace-relative paths discovered by a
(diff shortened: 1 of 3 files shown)


Hmbown:
This was harvested into the v0.8.51 release — thank you for the contribution!

PR #2588, opened 2026-06-02
Add support for SiliconFlow China region and provider

BH8GCJ:
Add `siliconflow-cn` as a separate provider targeting SiliconFlow's China
endpoint. The variant shares code paths with the global SiliconFlow
provider but uses its own API key and endpoint.

cargo check ✅   cargo test --workspace ✅   0 failures

<details open><summary><h3>Greptile Summary</h3></summary>

Introduces `siliconflow-cn` as a distinct provider variant that defaults to `https://api.siliconflow.cn/v1`, while sharing the `[providers.siliconflow]` config block, `SILICONFLOW_API_KEY`, and the same `auth set` credential slot as the existing `siliconflow` provider.

- `ProviderKind::SiliconflowCN` (config crate) and `ApiProvider::SiliconflowCn` (tui crate) are added with every required match arm updated; `is_siliconflow()` is added as a convenience predicate and both official endpoints are recognized by `siliconflow_base_url_is_official`.
- Reasoning-effort handling in `client.rs` mirrors the existing `Siliconflow` arms for the new variant; `PROVIDER_LIST` in the CLI grows from 15 to 16 entries with both variants listed independently.
- The `parse()` functions in both crates include dead uppercase match alternatives (`\"siliconflow-CN\"`, `\"silicon-flow-CN\"`, etc.) that can never be reached because the input is fully lowercased before the match.
</details>

<details open><summary><h3>Confidence Score: 4/5</h3></summary>

The change is safe to merge; all credential and config slot sharing is intentional and correctly handled across every match arm in all three crates.

The shared-slot design is well-executed and consistent, and siliconflow_base_url_is_official correctly covers both .com and .cn. The only rough edges are dead uppercase match patterns in both parse() functions (harmless but misleading), the README omitting --provider siliconflow-cn as a first-class example, and the undocumented behavior where SILICONFLOW_BASE_URL can silently redirect siliconflow-cn to the international endpoint.

crates/config/src/lib.rs and crates/tui/src/config.rs each have dead patterns in their parse() functions worth cleaning up. README.md could more prominently feature the new --provider siliconflow-cn shorthand.
</details>

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| crates/config/src/lib.rs | Adds SiliconflowCN variant to ProviderKind with shared credential slot, is_siliconflow() helper, and DEFAULT_SILICONFLOW_CN_BASE_URL constant; contains a dead uppercase pattern in parse(). |
[...]

PR #2615, opened 2026-06-03
feat(provider): add SiliconFlow China region (siliconflow-CN)

### LINKED ISSUE #2478: Please consider making opencode-go a built-in provider.
Problem
Currently, users who want to use opencode-go with Codewhale have to manually configure it as a custom provider. This creates unnecessary friction and a steep learning curve for developers who rely on the opencode-go ecosystem. Since Codewhale doesn't recognize it out-of-the-box, users often struggle with configuring the correct base URLs, specific authentication headers, and model routing.

Proposed solution
Make opencode-go a first-class, built-in provider in Codewhale.

Config shape: Users should be able to simply set provider: opencode-go (or similar) in their configuration file without needing to define the full endpoint URL.

Behavior: Codewhale should automatically handle the specific API shape, default endpoints, and standard error handling for opencode-go.

UI: It should appear in the default dropdown list of available providers (if a GUI/CLI wizard exists).

Use case
I work heavily in the Go ecosystem and use opencode-go for my daily development workflows. When setting up a new Codewhale environment or onboarding new team members, I want to bypass the manual custom endpoint setup. This is especially useful for teams standardizing on opencode-go who want seamless, zero-config integration with Codewhale right out of the box.

Alternatives considered
Custom OpenAI-compatible endpoint: Manually mapping the opencode-go base URL and injecting API keys as a generic provider. This works temporarily but is fragile and lacks support for any opencode-go specific extended features.

Writing a local proxy: Setting up a middleware proxy to translate Codewhale requests into opencode-go requests, which introduces far too much maintenance overhead.

Impact
This would significantly improve my daily workflow by saving setup time and preventing configuration drift across team members. I would use this constantly, and it would likely lower the barrier to entry for many other Go developers adopting Codewhale.

Hmbown:
Adds SiliconFlow China regional endpoint as new provider variant. Closes #2478 (partial). Credit: @Raid10Without1 (#2588).


shenjackyuanjie:
This PR Intrduce some compile error……

image

PR #2632, opened 2026-06-03
perf(prefix-cache): cache tool-catalog JSON serialization across checks

HUQIANTAO:
## Summary

`PrefixFingerprint::compute` is called once per turn by the turn-loop prefix-stability check (and again on every mode flip, project-context refresh, or canonical-state overlay). The tool-side work serializes every tool to the chat-API JSON shape, sorts the resulting strings, joins with newlines, and SHA-256s the result. For a 60-tool catalog that is ~25–40 KB of allocation plus a sort, all of which produces a byte-identical output once the tool set is stable across turns — the common case after the first turn of a session.

This PR introduces a process-local `ToolCatalogCache` that stores the joined+sorted catalog under a content-derived `u64` identity. On a hit, the per-tool JSON serialization, sort, and join are skipped entirely; the pre-computed SHA-256 hex digest is returned directly.

## Why now

The prefix-stability manager was wired into the turn loop in v0.8.50 to surface cache-hit telemetry to the TUI footer. That made the catalog serialization run on every turn instead of every session boundary. This PR removes that regression: the first turn pays the full cost, every subsequent turn pays only the identity hash.

The cache lives on `PrefixStabilityManager` (per-session ownership) and backs a new `PrefixFingerprint::compute_with_tool_cache` entry point. `check_and_update`, `PrefixStabilityManager::new`, and `pin()` all use the cached path. The original `compute()` is kept as a fallback for callers that do not have a cache in hand.

## Changes

| File | Change |
|---|---|
| `crates/tui/src/prefix_cache.rs` | New `ToolCatalogCache` (capacity-bounded LRU) and `CachedCatalog` entry type. New `PrefixFingerprint::compute_with_tool_cache` entry point. `PrefixStabilityManager` gains a `tool_catalog_cache` field. 8 new unit tests. |

## Design notes

- **Identity hash is order-sensitive.** Two slices with the same tools in different orders produce two cache entries. The downstream fingerprint itself remains order-insensitive (the sort in `fingerprint_for` takes care of that). Order-sensitive identity makes a re-registration of the same set in the same order a hit.
- **Eviction policy: insertion-order LRU.** Matches the strategy in `transcript_cache.rs`. Capacity defaults to 8 (sized for "session + 1 or 2 forked subagent catalogs"). `invalidate()` is exposed for tool-registry hot-reload and MCP attach paths.
- **No new external dependencies.** Uses `std::collections::HashMap`, `VecDeque`, and `std::hash::DefaultHasher`.
- **No public API breakage.** `PrefixFingerprint::compute` is unchanged.

## Tests

```
running 30 tests (prefix_cache module)
test prefix_cache::tests::compute_with_tool_cache_matches_compute_uncached ... ok
test prefix_cache::tests::manager_check_and_update_uses_cached_tool_fingerprint ... ok
test prefix_cache::tests::tool_catalog_cache_miss_then_hit_returns_same_arc ... ok
test prefix_cache::tests::tool_catalog_cache_different_tool_sets_dont_collide ... ok
test prefix_cache::tests::tool_catalog_cache_pinned_by_input_order ... ok
test prefix_cache::tests::tool_catalog_cache_detects_schema_change ... ok
test prefix_cache::tests::tool_catalog_cache_respects_capacity ... ok
test prefix_cache::tests::tool_catalog_cache_invalidate_clears_all ... ok
test prefix_cache::tests::tool_catalog_cache_empty_slice_uses_zero_capacity_path ... ok
... (21 prior tests still pass)

test result: ok. 30 passed; 0 failed
```

The wider prefix-cache contract tests in `settings`, `prompts`, and `core::engine::tests` continue to pass.

### DIFF
--- a/crates/tui/src/prefix_cache.rs
+++ b/crates/tui/src/prefix_cache.rs
@@ -29,6 +29,11 @@
 //! └─────────────────────────────────────────┘
 //! ```
 
+use std::collections::hash_map::DefaultHasher;
+use std::collections::{HashMap, VecDeque};
+use std::hash::{Hash, Hasher};
+use std::sync::Arc;
+
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
@@ -58,7 +63,6 @@ impl PrefixFingerprint {
     /// while ignoring internal-only fields like `allowed_callers` (#2264).
     pub fn compute(system_text: &str, tools: Option<&[Tool]>) -> Self {
         let system_sha256 = sha256_hex(system_text.as_bytes());
-
         let tools_sha256 = match tools {
             Some(tools) if !tools.is_empty() => {
                 let mut serialized: Vec<String> =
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @HUQIANTAO. I checked this for the 0.8.52 freeze, and it should wait for a follow-up pass: CI is currently failing with `-D warnings` dead-code errors in `prefix_cache.rs` (`PrefixFingerprint::compute`, `CachedCatalog::joined`, and unused cache helper methods), which also breaks the macOS/mobile build legs. Not a release blocker, but the idea is useful once the cache API is either wired through or narrowed to only used surface.

Hmbown:
Thanks @HUQIANTAO for the tool-catalog serialization cache work. The v0.9 integration branch already contains this intent as commit `baef5ba95` by you: `PrefixStabilityManager` now carries a bounded `ToolCatalogCache`, and the cached `PrefixFingerprint::compute_with_tool_cache` path avoids rebuilding the stable tool-catalog JSON across checks.

I verified it on `codex/v0.9.0-stewardship` at `4b61462eda423969939e70d5757a70e912cc296f` with:

```
cargo test -p codewhale-tui prefix_cache::tests::tool_catalog_cache --all-features --locked -- --nocapture
```

Result: 7 tests passed.

I am closing this original PR as a duplicate of the harvested integration-branch implementation while preserving the contributor intent and commit authorship. Thanks again for the prefix-cache performance cleanup.

PR #2635, opened 2026-06-03
perf(history): cache output_rows and selected_output_indices per cell

HUQIANTAO:
## Summary

`output_rows` (in `tui::history`) walks the raw tool output, ANSI-strips each line, classifies path/URL-like rows, and wraps the rest to the current viewport width. `selected_output_indices` then computes the head/tail/importance subset that the compact Live view shows. Both functions are pure, but they are called on every render frame for every visible tool cell. For a 4 KB tool output on a 120 FPS render loop that is 2-6 redundant walks per frame, per cell.

This PR adds `tui::output_rows_cache`, a thread-local, content-addressed cache keyed on `(content_hash, width)` for the rows and `(content_hash, width, line_limit)` for the indices. The cache stores the wrapped `Vec<OutputRow>` plus a per-line-limit map of selected indices on a single entry, so a single key lookup satisfies both render steps.

`render_preserved_output_mode` now consults the cache for both the rows and the indices; on a hit, neither the per-line ANSI strip nor the importance-ranking pass runs. The cache is bounded (default capacity 256) with insertion-order eviction.

## Why thread-local

The TUI render loop runs on a single thread. A `thread_local!` `RefCell<OutputRowsCacheInner>` is the simplest and contention-free design — no `Mutex` or `RwLock`, no global mutable state visible to background workers, and no test-poisoning footguns. The cache is per-thread, so a background task that calls into the same module gets its own cold cache (acceptable: that path is rare and not on the render hot path).

## Changes

| File | Change |
|---|---|
| `crates/tui/src/tui/output_rows_cache.rs` (new, 344 LOC) | Thread-local cache, content hasher, 6 unit tests. |
| `crates/tui/src/tui/history.rs` | `OutputRow` becomes `pub` with `PartialEq + Eq` so the cache can store and hash it. `render_preserved_output_mode` consults the cache for both rows and indices. |
| `crates/tui/src/tui/mod.rs` | Register the new module. |

## Tests

```
running 6 tests
test tui::output_rows_cache::tests::cache_hit_returns_cached_rows ... ok
test tui::output_rows_cache::tests::different_width_invalidates_rows ... ok
test tui::output_rows_cache::tests::different_output_invalidates_rows ... ok
test tui::output_rows_cache::tests::indices_cached_per_line_limit ... ok
test tui::output_rows_cache::tests::capacity_evicts_oldest ... ok
test tui::output_rows_cache::tests::hash_str_stable_for_identical_input ... ok

test result: ok. 6 passed; 0 failed
```

The wider `tui::history` test suite (68 tests) still passes.


Hmbown:
Thanks @HUQIANTAO. I took a first look while checking the post-0.8.52 queue. This is a plausible render-hot-path optimization, but it is not a corrective-release candidate: CI is already blocked on lint, macOS/Windows/mobile are still pending, and the patch makes `OutputRow` public while adding a new cache module around history rendering.

I am keeping it out of 0.8.52 so the release stays focused on verified regression fixes. Once CI is clean, this should get a normal review around cache invalidation, memory bounds, and whether exposing `OutputRow` is the right API shape. Thanks for continuing to improve the perf path; I just do not want to bury a rendering cache change in the emergency patch.

Hmbown:
Thanks @HUQIANTAO for the history-render cache work. The v0.9 integration branch already contains this intent as commits `3b0ef3f63` and `863f55cc6` by you: `tui::output_rows_cache` now caches wrapped output rows and selected output indices per content/width/line-limit, and the follow-up simplified the API plus switched the content hash to FNV-1a.

I verified it on `codex/v0.9.0-stewardship` at `4b61462eda423969939e70d5757a70e912cc296f` with:

```
cargo test -p codewhale-tui tui::output_rows_cache::tests --all-features --locked -- --nocapture
```

Result: 8 tests passed.

I am closing this original PR as a duplicate of the harvested integration-branch implementation while preserving the contributor intent and commit authorship. Thanks again for improving the render hot path.

PR #2636, opened 2026-06-03
perf(project-context): cache load_project_context_with_parents by mtime signature

HUQIANTAO:
## Summary

`load_project_context_with_parents` walks the workspace and every parent directory, checks for six project-context filenames in each, then consults three global fallback paths under the user's home directory. The per-file work is cheap (a `metadata()` call plus a single read on the first match), but the call is on the engine's hot path: it runs from `Session::new`, the layered-context checkpoint, `build_system_prompt_with_session_context`, and the TUI context inspector. A long session can re-invoke the loader dozens of times per turn without any of the candidate files having changed.

This PR adds a thread-local cache keyed on the canonical workspace path plus a cheap `MtimeSignature`: an ordered list of `(path, SystemTime)` pairs for every candidate file the loader would inspect. The signature is computed by `metadata()` only — no file reads. On a hit the cached `ProjectContext` is returned without any I/O beyond the metadata calls. On a miss the loader runs and the result is stored.

The cache is bounded (default capacity 8 workspaces) and uses insertion-order eviction, matching the strategy in `tui::transcript_cache` and `tui::output_rows_cache`. The signature is sorted to make it independent of iteration order; it changes when any candidate file appears, is overwritten, or is removed (mtime delta). Files that vanish contribute a `None` mtime that still differs from a stable `SystemTime`, so the cache invalidates correctly.

## Changes

| File | Change |
|---|---|
| `crates/tui/src/project_context_cache.rs` (new, 260 LOC) | Thread-local cache, `MtimeSignature` builder, 5 unit tests. |
| `crates/tui/src/project_context.rs` | `load_project_context_with_parents` consults the cache first. `PROJECT_CONTEXT_FILES` becomes `pub(crate)` so the cache module reads the same file list. |
| `crates/tui/src/main.rs` | Register the new module. |

## Tests

```
running 5 tests
test project_context_cache::tests::cache_round_trip ... ok
test project_context_cache::tests::store_does_not_grow_unbounded ... ok
test project_context_cache::tests::signature_is_stable_when_files_unchanged ... ok
test project_context_cache::tests::signature_diffs_when_a_new_file_appears ... ok
test project_context_cache::tests::signature_changes_when_file_is_overwritten ... ok

test result: ok. 5 passed; 0 failed
```

The wider `project_context` test suite (30 tests) still passes.

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -51,6 +51,7 @@ mod palette;
 mod prefix_cache;
 mod pricing;
 mod project_context;
+mod project_context_cache;
 mod project_doc;
 mod prompt_zones;
 mod prompts;
--- a/crates/tui/src/project_context.rs
+++ b/crates/tui/src/project_context.rs
@@ -649,17 +649,24 @@ pub fn load_project_context(workspace: &Path) -> ProjectContext {
 ///
 /// This allows for monorepo setups where a root AGENTS.md applies to all subdirectories.
 pub fn load_project_context_with_parents(workspace: &Path) -> ProjectContext {
+    // Canonicalize at the entry so the parent-directory walk and the
+    // cache key both agree on the absolute path. Without this, a caller
+    // passing a relative path would produce a different cache key on
+    // every call (each one resolves relative to a different cwd).
+    let canonical_workspace =
+        std::fs::canonicalize(workspace).unwrap_or_else(|_| workspace.to_path_buf());
     let home_dir = dirs::home_dir();
     // Compute a content signature (canonical workspace + per-file mtimes)
     // and consult the process-local cache before doing any file reads.
     // Repeated calls within a session — common during the layered
     // context checkpoint and prompt-assembly paths — short-circuit when
     // none of the candidate files have been written since the last call.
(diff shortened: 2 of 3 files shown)


Hmbown:
Thanks @HUQIANTAO. This missed the emergency v0.8.52 lane because we kept that release to publish/provenance fixes only, but it is queued for the v0.8.53 PR sweep in #2645. Sorry for the messy release window; this should get a normal performance-review pass now that the registries are coherent.

Hmbown:
Thanks @HUQIANTAO — the project-context cache work was harvested into the public v0.9.0 integration branch (`codex/v0.9.0-stewardship`) as `e18f072a5`.

The landed version uses a bounded process-local cache keyed by canonical workspace and content fingerprints, with widened invalidation for workspace/parent instructions, global AGENTS/WHALE fallbacks, repo constitution candidates, generated-context targets, trust markers, trust config paths, and same-length overwrites.

Verification recorded for the harvest: `cargo test -p codewhale-tui --bin codewhale-tui --locked project_context -- --nocapture`. The commit includes `Harvested from PR #2636 by @HUQIANTAO` and your GitHub-mappable co-author trailer.

Closing this PR as harvested into the integration branch.

PR #2880, opened 2026-06-07
fix: critical bugs in tools, client, and commands (9 bugs)

HUQIANTAO:
## Summary

Fixes 9 critical bugs across the TUI tools, LLM client, and command system that could cause panics, data corruption, or incorrect behavior.

### Panics / Crashes

1. **UTF-8 boundary panic in `clean_pdf_text`** (`tools/file.rs:295`)
   - `str::rfind` returns the byte index of the start of the matching character. Using `i + 1` may produce an offset that is not a valid UTF-8 char boundary, causing a runtime panic on `&str` slicing.
   - **Fix**: Use `i + char.len_utf8()` to compute the correct byte offset.

2. **ContentBlockStop wrong index on SSE stream cleanup** (`client/chat.rs:435`)
   - When the SSE stream breaks abnormally, the cleanup code used `content_index.saturating_sub(1)` which would produce `u32::MAX` when `content_index` is 0 (first block). The index should be `content_index` directly since it is only incremented AFTER a block is closed.
   - **Fix**: Use `content_index` directly for the close event.

3. **TOCTOU double-call in `run_skill_by_name`** (`commands/mod.rs:671`)
   - Pattern `is_some()` followed by `unwrap()` calls the function twice. If state changes between calls, the second call may return `None` and panic.
   - **Fix**: Replace with `if let Some(result) = ...` pattern.

4. **Silent token count truncation** (`client.rs:1298`)
   - `u64` to `u32` cast via `as` silently truncates values above 4,294,967,295. This is inconsistent with the sanitizer at line 1807 which uses `.min(u32::MAX)`.
   - **Fix**: Use `.min(u64::from(u32::MAX)) as u32` for saturating conversion.

### Data Integrity

5. **Incomplete hex decoding in `from_api_tool_name`** (`client.rs:54-76`)
   - The decoder accepts fewer than 6 hex digits, while the encoder always produces exactly 6. This inconsistency means a mangled tool name with a short hex sequence would be decoded differently by the two code paths.
   - **Fix**: Require exactly 6 hex digits to match encoder output.

6. **HTTP response body read errors silently swallowed** (`client/chat.rs:170`, `client.rs:750`)
   - `unwrap_or_default()` converts connection drops mid-body to empty string `""`, causing an unhelpful JSON parse error instead of surfacing the real cause.
   - **Fix**: Replace with `.context()?` error propagation.

7. **Integer overflow in `context_lines`** (`tools/search.rs:103`)
   - Model-provided `context_lines` has no upper bound. On 32-bit platforms, `usize::MAX` fallback from `try_from` combined with addition causes overflow.
   - **Fix**: Clamp to `.min(1000)` to prevent excessive memory allocation.

8. **u64 to usize truncating cast** (`tools/file.rs:117,127`)
   - `as usize` silently truncates on 32-bit platforms. A model passing `start_line = 5_000_000_000` would read from the wrong line.
   - **Fix**: Use `usize::try_from()` with proper error propagation.

9. **Missing providers in `provider_accepts_reasoning_content`** (`client/chat.rs:1966`)
   - `SiliconflowCn` and `Volcengine` are present in `apply_reasoning_effort` but missing from `provider_accepts_reasoning_content`. Non-DeepSeek reasoning models on these providers lose thinking trace support.
   - **Fix**: Add both providers to the match list.

### DIFF
--- a/crates/tui/src/client.rs
+++ b/crates/tui/src/client.rs
@@ -61,7 +61,10 @@ pub(super) fn from_api_tool_name(name: &str) -> String {
                     break;
                 }
             }
-            if let Ok(code) = u32::from_str_radix(&hex, 16)
+            // Only decode if we got exactly 6 hex digits (matching encoder output).
+            // Fewer digits means a truncated/malformed sequence — pass through as-is.
+            if hex.len() == 6
+                && let Ok(code) = u32::from_str_radix(&hex, 16)
                 && let Some(decoded) = std::char::from_u32(code)
             {
                 if let Some('-') = iter.peek().copied() {
@@ -747,7 +750,10 @@ impl DeepSeekClient {
             );
             anyhow::bail!("Failed to list models: HTTP {status}: {error_text}");
         }
-        let response_text = response.text().await.unwrap_or_default();
+        let response_text = response
+            .text()
+            .await
+            .context("Failed to read models response body")?;
 
         parse_models_response(&response_text)
     }
@@ -1295,8 +1301,8 @@ pub(super) fn parse_usage(usage: Option<&Value>) -> Usage {
     });
 
     Usage {
-        input_tokens: input_tokens as u32,
-        output_tokens: output_tokens as u32,
+        input_tokens: input_tokens.min(u64::from(u32::MAX)) as u32,
+        output_tokens: output_tokens.min(u64::from(u32::MAX)) as u32,
         prompt_cache_hit_tokens,
         prompt_cache_miss_tokens,
         reasoning_tokens,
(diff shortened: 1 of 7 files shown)


Hmbown:
Merged — thank you @HUQIANTAO. All nine fixes reviewed individually; each is a real bug with the right minimal fix. The `ContentBlockStop` index correction and the strict 6-hex-digit decode are particularly good catches.

One note: `crates/tui/src/prompt_persist.rs` landed as a fully `#[allow(dead_code)]` module that nothing calls yet, and it wasn't mentioned in the PR description. It's inert so it didn't block the merge, but please follow up: either a PR that actually wires it into prompt assembly (coordinating with the prefix-cache work in #2264/#2949 so the designs don't collide), or a removal. Unreferenced scaffolding shouldn't accumulate on main.

Issue #2657, opened 2026-06-03
modes: agents cannot easily tell why a tool is unavailable


Hmbown:
## Description

Tool availability changes across modes and permission gates are difficult for agents to interpret.

## Observed behavior

In a branch-cleanup planning test:

1. In Plan Mode, shell execution was blocked.
2. The agent asked the user to switch to Agent or YOLO mode.
3. The user approved proceeding.
4. In Agent Mode, `exec_shell` was still unavailable because shell access was gated by `allow_shell`.
5. The agent then asked again for either `allow_shell = true` or YOLO mode.

The transcript made it hard to distinguish whether the problem was Plan Mode, Agent Mode, `allow_shell`, deferred tool loading, or the active tool catalog.

## Why this is confusing

The agent experiences this as a missing-tool or rejected-tool event, but the underlying cause can be several different layers:

- current collaboration mode
- shell permission setting
- auto-approval mode
- deferred tool catalog state
- workspace trust/config policy

The user and agent then have to debug the harness state instead of the task.

## Impact

Agent Mode can feel broken even when the underlying policy is behaving as configured. The model may incorrectly tell the user to switch modes when the real blocker is a separate permission gate.

## Environment

- codewhale version: observed around v0.8.52
- Target milestone: v0.8.53
- Area: modes, tool availability, permissions, shell gating

cy2311:
PR submitted: https://github.com/Hmbown/CodeWhale/pull/2933

Hmbown:
Closing as completed. #3041 shipped the tool-denial/error-message fix so agents see self-explanatory causes such as mode, allow_shell, or user denial instead of a conflicting generic suffix. Credit to @cy2311 for the harvested PR work.

PR #2905, opened 2026-06-08
fix(tui): name allow_shell blocker for shell tools

cyq1017:
## Summary

- Clarify the missing-tool diagnostic for shell tools when `allow_shell` is disabled.
- Keep the existing `/config allow_shell true` and `--save` guidance, but name the current `allow_shell = false` blocker directly.
- Add regression coverage for both exact shell-tool names and suggestion-based missing-tool output.

## Verification

- `CARGO_TARGET_DIR=/tmp/codewhale-2657-final-target cargo test -p codewhale-tui missing_shell_tool_error_message -j1`
- `cargo fmt --check`
- `git diff --check HEAD^..HEAD`


Hmbown:
Merged — thanks @cyq1017. Naming the actual blocker in the error is exactly the #2657 class of fix we want.

PR #2918, opened 2026-06-09
feat(i18n): localize ConfigSection and ConfigScope labels (11 MessageIds)

gordonlu:
## Summary

Localize the 9 ConfigSection labels (Provider, Model, Permissions, Network, Display, Composer, Sidebar, History, MCP) and 2 ConfigScope labels (SESSION, SAVED) in the config view.

## Changes

- **localization.rs**: Add 11 new ConfigSection* and ConfigScope* MessageId variants with translations across all 7 shipped locales (En, Ja, ZhHans, ZhHant, PtBr, Es419, Vi)
- **views/mod.rs**: Update `ConfigSection::label()` and `ConfigScope::label()` to accept `Locale` and delegate to `tr()`. Update all call sites to pass locale.

## Testing

- Existing tests updated to use `Locale::En` explicitly where they assert section/scope labels
- 4338/4339 tests pass (pre-existing macOS flake: `legacy_sse_closed_stream_reconnects_and_retries_tool_call`)
- `cargo fmt --all -- --check` passes

PR #2901, opened 2026-06-08
feat(i18n): localize ToolFamily labels (10 MessageIds)

gordonlu:
## Summary

Localize tool family verb labels (read, patch, run, find, delegate, fanout, rlm, verify, think, tool) used in tool card headers, sidebar, and footer status lines. 10 new MessageId variants covering all 7 shipped locales.

## Changes

- **localization.rs**: 10 ToolFamily* MessageId variants, ALL_MESSAGE_IDS entry, translations for En, Vi, ZhHant, Ja, ZhHans, PtBr, Es419.
- **tool_card.rs**: tool_activity_label_for_name() takes locale and uses tr() for localized labels. family_label() and tool_display_label_for_name() remain English-only (internal helpers).
- **footer_ui.rs, ui.rs**: Pass app.ui_locale to tool_activity_label_for_name() callers.
- **Tests**: 2 negative i18n tests (no exact English leak in non-English locales).

## Notes

- family_label() stays English-only — used in card headers alongside glyphs, matching the vocabulary-label pattern from SubAgentType kind column.
- tool_activity_label_for_name() is the user-facing function used in footer status labels and activity detail.

<details open><summary><h3>Greptile Summary</h3></summary>

Adds localized labels for all 10 `ToolFamily` variants across 7 locales (En, Vi, ZhHant, Ja, ZhHans, PtBr, Es419), surfacing them in footer status lines and activity-detail labels. `tool_activity_label_for_name` now accepts a `Locale` and routes through `tr()`; `family_label` intentionally stays English-only for card-header glyphs.

- `localization.rs` gains 10 new `MessageId` variants with complete translations for every shipped locale, including intentional loan-words (`\"fanout\"`, `\"rlm\"`) where no native term exists.
- `tool_card.rs` introduces `family_message_id()`, restricts `tool_display_label_for_name` to `#[cfg(test)]`, and adds two negative i18n tests verifying non-English locales don't echo English strings.
- `footer_ui.rs` and `ui.rs` thread `app.ui_locale` through to all `Generic`-branch call-sites.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

Safe to merge — all production paths correctly thread locale through to tr(), and the translation tables are complete for every shipped locale.

The change is additive: new enum variants, new translation entries, and updated call-sites. Every production call-site now passes locale correctly. The only gap is that three MessageIds (Delegate, Think, Generic) are absent from the no-English-leak test, but this does not affect runtime behaviour.

[...]

PR #2896, opened 2026-06-08
feat(i18n): localize status picker surface (7 MessageIds)

gordonlu:
## Summary

Localizes the `/statusline` multi-select picker dialog across all 7 shipped
locales (En, Ja, ZhHans, ZhHant, PtBr, Es419, Vi).

## New MessageIds (7)

| MessageId | English |
|---|---|
| `StatusPickerTitle` | `" Status line "` |
| `StatusPickerInstruction` | `"Pick the chips you want in the footer:"` |
| `StatusPickerHintSpace` | `" Space toggle "` |
| `StatusPickerHintA` | `" a all "` |
| `StatusPickerHintN` | `" n none "` |
| `StatusPickerHintEnter` | `" Enter save "` |
| `StatusPickerHintEsc` | `" Esc cancel "` |

## Files touched

- **`crates/tui/src/localization.rs`** — enum variants, `ALL_MESSAGE_IDS`,
  English base, and translations for all 6 non-En locales.
- **`crates/tui/src/tui/views/status_picker.rs`** — accepts `Locale` parameter,
  replaces hardcoded English with `tr()` calls.
- **`crates/tui/src/tui/ui.rs`** — passes `app.ui_locale` to `StatusPickerView::new`.

## Tests

- All 11 existing status-picker tests pass (updated for the new `locale` param).
- `status_picker_displays_localized_title_for_zh_hans` — verifies ZhHans output.
- `status_picker_no_english_leak_in_non_en_locales` — negative assertion across
  all 6 non-English locales that `StatusPickerTitle` and `StatusPickerInstruction`
  contain no English words.

<details open><summary><h3>Greptile Summary</h3></summary>

Localizes the `/statusline` multi-select picker to all 7 shipped locales by adding 7 `MessageId` variants, providing translations in every non-English locale function, storing a `Locale` in `StatusPickerView`, and replacing hardcoded English strings with `tr()` calls at render time.

- `localization.rs` gains 7 new variants (`StatusPickerTitle`, `StatusPickerInstruction`, and 5 action-word IDs) with full translations in Ja, ZhHans, ZhHant, PtBr, Es419, and Vi; ZhHant now has explicit entries rather than falling through to Simplified Chinese.
- `status_picker.rs` threads the locale through `new()` and uses it in the `Block` title and hint-bar spans, preserving the key/description contrast by keeping `\" Space \"`, `\" a \"`, etc. as separately styled spans.
[...]

PR #2892, opened 2026-06-08
feat(i18n): localize sandbox elevation dialog across 7 locales

gordonlu:
Migrates the sandbox elevation dialog (`ElevationWidget`/`ElevationView`) from hardcoded English strings to `MessageId`-based translations covering all 7 shipped locales (En, Ja, ZhHans, ZhHant, PtBr, Es419, Vi).

**Changes**
- `localization.rs`: 18 new `MessageId` variants for elevation titles, field labels, impact copy, option labels/descriptions — all 7 locale translations
- `widgets/mod.rs`: replaced all hardcoded English strings in `ElevationWidget::render()` with `tr()` calls; added `locale` field
- `approval.rs`: `ElevationView::new(request, locale)` with locale field, passes to widget
- `ui.rs`: passes `app.ui_locale` to `ElevationView::new()`

**Verification**
- `cargo check` — 0 warnings
- `cargo test -- elevation` — 13/13 pass (including zh locale + English leak detection)
- `cargo test -- approval` — 96/96 pass
- `cargo test -- localization` — 8/8 pass
- `cargo fmt` — clean

<details open><summary><h3>Greptile Summary</h3></summary>

Migrates the sandbox elevation dialog from hardcoded English strings to the existing `MessageId`/`tr()` localization system, covering all 7 shipped locales consistently.

- **`localization.rs`**: Adds 18 new `MessageId` variants (titles, field labels, impact lines, option labels/descriptions) with translations in En, Ja, ZhHans, ZhHant, PtBr, Es419, and Vi; each non-English locale uses an exhaustive `Some(match …)` block so a missing translation is a compile error.
- **`widgets/mod.rs` / `approval.rs` / `ui.rs`**: `ElevationWidget` and `ElevationView` gain a `locale: Locale` field; `app.ui_locale` is wired in at the one call-site; `label()` / `description()` are narrowed to `#[cfg(test)]`.
- New render smoke-tests cover En, ZhHans, Ja, and ZhHant with English-leak assertions; previously flagged coverage gaps for PtBr, Es419, and Vi remain open.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

Safe to merge — the change is purely additive (new MessageIds + tr() call-sites) with no behavioral changes to sandbox policy decisions, key bindings, or event routing.

[...]

PR #2919, opened 2026-06-09
feat(i18n): localize ConfigEdit labels and default values (11 MessageIds)

gordonlu:
## Summary

Localize 11 ConfigEdit-related strings including edit mode labels (Edit, Scope, Current, Hint, New), edit footer, status messages, default value placeholders, and currency suffix.

## Changes

- **localization.rs**: Add 11 new `ConfigEdit*`, `ConfigRowEffective`, `ConfigDefault*`, and `ConfigUnavailable` MessageId variants with translations across all 7 shipped locales
- **views/mod.rs**: Replace hardcoded English strings with `self.tr()` / `tr()` calls in ConfigView editing mode, row display value formatting, and config row construction

## Testing

- Two tests updated: escape-cancels uses `Locale::En`; cost-currency uses localized zh-Hans assertion
- 4339/4339 tests pass;
- `cargo fmt --all -- --check` passes

### DIFF
--- a/crates/tui/src/localization.rs
+++ b/crates/tui/src/localization.rs
@@ -238,6 +238,17 @@ pub enum MessageId {
     ConfigFooterDefault,
     ConfigFooterScrollable,
     ConfigFooterFiltered,
+    ConfigEditCancelled,
+    ConfigEditTitlePrefix,
+    ConfigEditScopeLabel,
+    ConfigEditCurrentLabel,
+    ConfigEditHintLabel,
+    ConfigEditNewLabel,
+    ConfigEditFooter,
+    ConfigRowEffective,
+    ConfigDefaultValue,
+    ConfigDefaultReasoning,
+    ConfigUnavailable,
     HelpTitle,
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks @gordonlu — we merged #2891 and #2896 from your i18n batch tonight, which moved `localization.rs` on main, so this slice now shows as conflicting. Could you rebase it onto current main when you get a chance?

Since the slices all touch `localization.rs`, the smoothest path is to rebase and push them **one at a time, oldest-first** (#2892 → #2901 → #2918 → #2919 → #2921 → #2926 → #2929 → #2932 → #2940) — we'll merge each promptly so the next rebase stays small. Sorry for the churn; the batch is exactly the right shape and we want all of it.

Hmbown:
Closing this smaller i18n PR as superseded by the consolidation lane in #2239. #2239 covers the same surface area at broader Phase 1-4b scope and is the preferred v0.8.59 review target, though it still has its own outstanding review/fix work before merge.

PR #2943, opened 2026-06-09
fix(tui): normalize macOS SUPER (Cmd) to CONTROL for keyboard shortcuts (#2938)

### LINKED ISSUE #2938: macOS modifier key confusion: Ctrl vs Cmd/Alt inconsistencies
## Problem

On macOS, there are mix-ups between Ctrl, Cmd, and Alt modifier keys. Some shortcuts that use Ctrl don't work as expected, possibly because macOS maps Cmd as the primary modifier.

## Areas to investigate

- Keybinding dispatch on macOS — are we handling the platform-specific modifier mapping correctly?
- Terminal emulators on macOS sometimes remap Ctrl and Alt differently
- Check consistency across iTerm2, Terminal.app, and other macOS terminals

## Related

- Background task cancellation (Ctrl+B) is also affected by this

idling11:
On macOS, terminal emulators may report Cmd (SUPER) instead of Ctrl (CONTROL) for keyboard shortcuts, depending on the terminal app and its configuration. This caused Ctrl+B, Ctrl+Alt+2, and other shortcuts to be inconsistent.

Fix:
- Add normalize_macos_modifiers() in composer_ui.rs
- On macOS: map SUPER to CONTROL when CONTROL is not already set
- On other platforms: no-op
- Apply normalization at the key event entry point in ui.rs

Tests:
- normalize_macos_modifiers_maps_super_to_control
- normalize_macos_modifiers_preserves_existing_control
- normalize_macos_modifiers_leaves_alt_unchanged

Close https://github.com/Hmbown/CodeWhale/issues/2938
## Testing

## Checklist

### DIFF
--- a/crates/tui/src/tui/composer_ui.rs
+++ b/crates/tui/src/tui/composer_ui.rs
@@ -121,6 +121,23 @@ pub(crate) fn is_word_cursor_modifier(modifiers: KeyModifiers) -> bool {
     modifiers.contains(KeyModifiers::CONTROL) || modifiers.contains(KeyModifiers::ALT)
 }
 
+/// On macOS, map `SUPER` (Cmd ⌘) to `CONTROL` when `CONTROL` is not already
+/// set, so that terminal emulators that don't pass Ctrl faithfully still work.
+/// On all other platforms this is a no-op.
+#[cfg(target_os = "macos")]
+pub(crate) fn normalize_macos_modifiers(modifiers: KeyModifiers) -> KeyModifiers {
+    if modifiers.contains(KeyModifiers::SUPER) && !modifiers.contains(KeyModifiers::CONTROL) {
+        modifiers | KeyModifiers::CONTROL
+    } else {
+        modifiers
+    }
+}
+
+#[cfg(not(target_os = "macos"))]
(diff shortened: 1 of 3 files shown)


Hmbown:
Thanks @idling11 — the normalization approach looks right, but **Test (windows-latest) is failing** and I believe it's the new tests rather than the logic: `normalize_macos_modifiers_maps_super_to_control` and `normalize_macos_modifiers_preserves_existing_control` assert the macOS mapping unconditionally, while the `#[cfg(not(target_os = "macos"))]` version is a no-op that returns `SUPER` unchanged — so both assertions fail off-macOS.

Could you either `#[cfg(target_os = "macos")]`-gate those two tests, or make the assertions platform-aware? One design question to consider while you're in there: with terminals that forward Cmd via the kitty/CSI-u protocol, a global Cmd→Ctrl remap turns e.g. Cmd+C into Ctrl+C (interrupt). If that's a real risk it may be worth scoping the remap to shortcuts we actually handle rather than the whole event stream. Happy to merge once Windows is green.

Hmbown:
v0.8.59 triage: merged into `codex/v0.8.59-release-ready` as 4d84d66e (full branch history preserved, your authorship intact). Build + focused TUI modifier tests green. Will be closed when the release branch lands on main. Thanks @idling11!

PR #2947, opened 2026-06-09
fix(tui): guide long shell work to background

### LINKED ISSUE #2939: Agent execution blocks on long-running shell commands — sleep/background tasks should not stall the agent
## Problem

When the agent runs a shell command that includes  or other blocking operations (e.g., polling CI), the entire turn is consumed waiting for it. The agent should be able to fire off the check and continue working in the same turn.

## Observed

-  with  blocks the entire turn for 60 seconds
- Agent can't do anything else while waiting for a background check to complete
- This applies to any long-running command: builds, test suites, CI polling

## Expected

- Agent should be able to spawn a background task and immediately continue with other work
- Background polling should be non-blocking
-  /  with  should be the default pattern for any command expected to take >5s

## Notes

Related to the background task cancellation issue (#2937) — both stem from the shell execution model not being fully mature for long-running work.

cyq1017:
Fixes #2939.

## Summary
- spell out the >5 second threshold for shell work in the model prompt
- align the exec_shell background schema with task_shell_start / wait guidance
- add prompt and schema coverage for the guidance

## Verification
- CARGO_TARGET_DIR=/tmp/codewhale-2939-supervisor-target cargo test -p codewhale-tui test_exec_shell_schema_guides_gt_five_second_work_to_background
- CARGO_TARGET_DIR=/tmp/codewhale-2939-supervisor-target cargo test -p codewhale-tui composed_prompt_keeps_shell_guidance_when_shell_tools_are_available
- cargo fmt --all -- --check
- git diff --check HEAD^..HEAD


Hmbown:
Merged — thanks @cyq1017. This is the prompt-side half of #2939 (agent stalling on long-running shell commands); the >5s rule in both prompt and tool schema should shift behavior immediately.

Issue #2608, opened 2026-06-03
v0.8.65 EPIC: Separate provider facts, model facts, offerings, and route resolution


Hmbown:
## Goal

Separate provider facts, model facts, provider model offerings, route resolution, and UI projections so CodeWhale stops mixing provider identity, model identity, and provider-specific wire ids.

The key invariant:

```text
A model string alone is never enough to select a route.

Execution requires a ReadyRouteCandidate.
A ReadyRouteCandidate can only be produced by RouteResolver.
RouteResolver resolves provider-scoped model references against catalog facts, user config,
auth config, endpoint config, capability requirements, and validation.
```

## Core Problem

Current provider/model logic mixes several different meanings into one string:

- canonical model identity, such as a provider-agnostic logical model;
- provider identity, such as DeepSeek, OpenRouter, Together, Ollama, or a custom endpoint;
- provider wire model id, such as `deepseek-ai/DeepSeek-V4-Pro` on Together or `deepseek/deepseek-v4-pro` on OpenRouter;
- aggregator namespace hints, such as `anthropic/...`, `openai/...`, `deepseek-ai/...`, or `qwen/...`;
- user custom model ids for local or OpenAI-compatible endpoints.

Those are not interchangeable. A prefixed wire id can carry provider/catalog namespace information, but it is not enough to decide canonical ownership or route validity outside the provider offering that supplied it.

Examples:

- `deepseek-ai/DeepSeek-V4-Pro` can be a Together wire id.
- `deepseek/deepseek-v4-pro` can be an OpenRouter wire id.
- A custom OpenAI-compatible endpoint might legitimately use a string that resembles an aggregator namespace.

So CodeWhale must stop treating `deepseek-ai/` or `deepseek/` as proof that the active provider must be DeepSeek, or that a non-DeepSeek provider is invalid.

## Architecture Contract

Use three catalog concepts plus one runtime concept:

1. `ProviderDescriptor`: provider identity and transport facts only.
2. `ModelProfile`: canonical, provider-agnostic model facts.
3. `ProviderModelOffering`: provider + canonical model + provider-owned wire id + endpoint + provider-specific overrides.
4. `ReadyRouteCandidate`: runtime-resolved executable route produced by the route resolver.

Provider facts describe how to talk to a provider.
Model facts describe what a model can do.
Provider model offerings describe which provider serves which model under which wire id.
[...]

PR #3005, opened 2026-06-10
refactor(config): extract provider metadata into data-driven registry

sximelon:
## Summary
Move provider id, display name, aliases, defaults, and wire format into a single `Provider` trait backed by a static `PROVIDER_REGISTRY`, eliminating ~100 hand-maintained match arms across two crates.

- Add `aliases()` to `Provider` trait, populated per-provider via the `provider!()` macro's `aliases:` clause.
- Add `id()` as a concrete impl in every provider struct.
- Replace `ProviderKind::as_str()` 20-arm match with delegation to `self.provider().id()`; no cycle thanks to concrete `id()`.
- Replace `ProviderKind::parse()` 20-arm match with iteration over `PROVIDER_REGISTRY`, matching canonical id and aliases.
- Replace `ApiProvider::kind()` 21-arm match and `from_kind()` 20-arm match with discriminant-indexed `KIND_LOOKUP` / `FROM_KIND_LOOKUP` const arrays.
- Simplify `ApiProvider::all()` to `&Self::FROM_KIND_LOOKUP`.
- `provider_for_kind()` uses `.find()` instead of match.

Adding a new provider now requires one `provider!()` invocation plus an enum variant, instead of touching 20-30 match arms.
Ref #2608 
## Testing

## Checklist


Hmbown:
v0.8.59 triage: scratch merge against `codex/v0.8.59-release-ready` shows this refactor still carries a real delta (~300 lines net simplification across `config/lib.rs`, `tui/config.rs`, registry check script) — it is NOT already landed. One semantic conflict: the release branch registers the SiliconFlow CN config key as `siliconflow_cn`, while this PR uses `siliconflow` plus aliases; that needs reconciliation before landing. As a structural refactor it's likely better targeted at the cycle after v0.8.59 unless Hunter wants it in. A rebase onto the release branch would make it directly reviewable.

Hmbown:
v0.8.59 release triage: deferring this PR out of the release-stabilization lane. It is not being rejected; it is just not required to unblock 0.8.59 now that the current release blockers have narrow fixes on `codex/v0.8.59-release-ready`. I am moving the target label to `v0.8.60` so this does not keep the 0.8.59 queue hostage.

PR #3011, opened 2026-06-10
feat(cli): track provider source and customize unsupported TUI errors

### LINKED ISSUE #3007: TUI provider rejection blames a --provider flag the user never passed; no recovery hint when the value comes from config or env
## What happens

If the resolved provider is not TUI-capable (e.g. the experimental `openai-codex` provider), `codew` / `codewhale` fails with:

```
error: The interactive TUI supports DeepSeek, NVIDIA NIM, ... providers. Remove --provider openai-codex or use `codewhale model ...` for provider registry inspection.
```

The message says **"Remove --provider openai-codex"** even when the user never passed any flag — the provider can come from `CODEWHALE_PROVIDER` / `DEEPSEEK_PROVIDER` env vars or from `provider = "..."` in `~/.codewhale/config.toml`. The dispatcher merges all three into `resolved_runtime.provider` and the bail site (`crates/cli/src/lib.rs:1626-1630`) has no provenance, so the error blames a flag that does not exist and gives no path to recovery.

## Real-world trigger

An external codex-proxy tool wrote `provider = "openai-codex"` into the user's global `~/.codewhale/config.toml`. Every interactive launch then failed with the message above; nothing in the error pointed at the config file, so it looked like a regression in the just-shipped release. (CodeWhale's own auth flows deliberately do not switch the active provider — `auth_set_provider_key_does_not_switch_active_provider` pins that — but we cannot stop third-party tools or hand-edits from doing it.)

## Proposed fix

1. **Track provenance** of the resolved provider (CLI flag / env var name / config file path) through provider resolution.
2. **Tailor the error** to the actual source, with a copy-pasteable fix:
   - flag → "Remove `--provider openai-codex`"
   - env → "Unset `CODEWHALE_PROVIDER` (currently `openai-codex`)"
   - config → "Edit `~/.codewhale/config.toml` and set `provider = \"deepseek\"` (currently `openai-codex`), or run with `--provider <tui-capable>`"
3. **Consider** (separate decision): when the unsupported provider comes from config rather than an explicit flag, fall back to the default provider with a loud warning instead of refusing to launch — config drift should degrade gracefully in the interactive path.
4. Keep the supported-provider list in the message generated from `ProviderKind` rather than hand-maintained prose, so it cannot drift from the actual match arms.

cyq1017:
## Summary

This PR resolves CodeWhale issue #3007. When an unsupported provider is resolved for the interactive TUI, the CLI previously emitted a generic instruction to "Remove --provider" regardless of where the provider came from. 

We now track the provider's source (`Cli`, `Env`, or `Config`) during configuration resolution and emit tailored corrective guidance for each case. Furthermore, if the unsupported provider is configured in `config.toml`, we output a loud warning to stderr and gracefully fall back to the default `DeepSeek` provider so the interactive TUI can still start, rather than crashing on launch.

## Target branch

- [x] This PR targets `dev`, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release.

## Linked Issue

Fixes #3007

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above.
- [x] I actually ran the app and verified the change works end-to-end.

## How to Test

1. Configure an unsupported provider (e.g. `openai-codex`) via `--provider` CLI flag. Verify it bails with instruction to remove `--provider`.
2. Configure the unsupported provider via environment variable `DEEPSEEK_PROVIDER`. Verify it bails with instruction to unset the variable.
3. Configure the unsupported provider in `config.toml`. Verify that it prints a warning to stderr but successfully launches the TUI by falling back to DeepSeek.
4. Run the unit test suite: `cargo test -p codewhale-cli`.


Hmbown:
v0.8.59 triage: scratch-merged against `codex/v0.8.59-release-ready` — zero residual diff. The release branch already tracks provider source (`ProviderSource::Cli`/`Env`) and emits per-source unsupported-TUI errors via `provider_is_supported_by_tui`. Ships in v0.8.59; can be closed once the release branch lands on main.

PR #3103, opened 2026-06-11
fix(tui): emit interrupted sub-agent lifecycle event and reconcile stale running cards (#3080)

### LINKED ISSUE #3080: v0.8.59: API-timeout interrupted sub-agents can leave fanout UI and task handles stuck running
## Summary

A CodeWhale session can appear completely stalled after sub-agent `agent_eval` calls hit the Responses API step timeout. The sub-agent manager records an `Interrupted(...)` checkpoint that is continuable, but the visible fanout/delegate UI and some live work surfaces can still remain in a running state. In the observed session the sidebar kept showing `2 running / 2`, both agent rows said `interrupted; API call timed out ...`, the Tasks panel still had `agent_eval` and child shell commands running, and the footer kept reporting sequential tool execution after roughly 27 minutes.

This was observed from a local terminal screenshot on 2026-06-11 while running CodeWhale v0.8.58; target the fix for v0.8.59.

## Observed behavior

- Parent work item failed while trying to retrieve failed sub-agent projection/error evidence.
- Two sub-agents remained visible as running even though their status text was `interrupted; API call timed out ...`.
- Each child had active foreground shell commands shown with `Ctrl+B backgrounds this command`.
- The footer/status stayed in `Executing tools sequentially (writes, approvals, or non-parallel tools detected)` with `2 agents` still running.
- The UI looked dead to the user: no clear recovery affordance, no terminal transition, and no obvious way to know whether to wait, continue, close, or background the child work.

## Local code evidence / likely root cause

The current timeout path intentionally checkpoints and marks the manager snapshot interrupted:

- `crates/tui/src/tools/subagent/mod.rs` around the `tokio::time::timeout(runtime.step_api_timeout, runtime.client.create_message(request))` branch builds `API call timed out after ...; checkpoint preserved for continuation`, writes an `api_timeout` checkpoint, and calls `manager.interrupt_with_checkpoint(...)`.
- Immediately after that, the task waits on `input_rx.recv()` for continuation input. That keeps the spawned task handle alive even though the snapshot status is no longer `Running`.
- `agent_eval` treats any non-running snapshot as terminal and returns a projection with `status: interrupted`, `terminal: true`, and `continuable: true`.

The UI/card path does not appear to get an equivalent lifecycle transition:

- `crates/tui/src/tools/subagent/mailbox.rs` has `Started`, `Progress`, `Completed`, `Failed`, and `Cancelled`, but no `Interrupted` message.
- `crates/tui/src/tui/widgets/agent_card.rs` only clears fanout/delegate running state on `Completed`, `Failed`, or `Cancelled`.
- `crates/tui/src/tui/sidebar.rs` can display the Agents header count from `active_fanout_counts(...)`; that reads the active fanout card slot lifecycle, not only `SubAgentStatus` from `subagent_cache`.
- Result: the manager can say interrupted/continuable while the fanout card remains `Running`, so the sidebar can keep reporting running agents indefinitely.

The screenshot also shows child foreground shell commands still running after the parent-facing API timeout. The timeout recovery path should decide what happens to child tool work during interruption: pause/abort/background it, or clearly mark it recoverable and expose an action. Leaving it as an invisible live blocker makes the whole harness look stalled.

## Expected behavior

When a sub-agent API call times out and is checkpointed as interrupted:

- The Agents panel should stop counting that worker as running unless it is actively being resumed.
- The fanout/delegate card should transition to an explicit interrupted/recoverable lifecycle state, or at minimum a non-running terminal-ish state with `continue via agent_eval` surfaced.
- `agent_progress` / progress-only rows should be reconciled so stale running rows do not survive after an interrupted snapshot is listed.
- Any child shell/tool work should be handled deliberately: aborted, backgrounded with a visible job id, or shown in Tasks with a recovery/cancel action.
- The parent turn should not wait forever merely because interrupted sub-agent task handles are parked waiting for continuation input.

## Suggested fix shape

- Add an explicit interrupted lifecycle event, e.g. `MailboxMessage::Interrupted { agent_id, reason }`, emitted from the API-timeout branch after `interrupt_with_checkpoint`.
- Teach `apply_to_delegate` and `apply_to_fanout` to render that state and remove it from running counts.
- Reconcile `active_fanout_counts`, `running_agent_count`, and `agent_progress` against interrupted manager snapshots on `AgentList` refresh.
- Add regression tests that simulate API timeout and assert:
  - `agent_eval` returns `status == interrupted` and `continuable == true`.
  - Agents sidebar/fanout count is no longer `N running / N` after the interrupted snapshot arrives.
  - Footer/stall reason no longer reports active sub-agents solely because the fanout card is stale.
  - Any still-running child shell/tool entry is visible as a task/job or cancelled deterministically.

## Related issues

Possibly related but broader/lower-signal stuck reports: #1190, #1060, #1679. This issue is specifically for the v0.8.59 sub-agent API-timeout recovery path and stale running fanout/task state.

Hmbown:
## Summary

Fixes #3080: API-timeout interruptions previously called `interrupt_with_checkpoint` without emitting any mailbox lifecycle event, so delegate/fanout cards stayed "Running" forever and the Agents sidebar kept counting interrupted agents as active.

- New `MailboxMessage::Interrupted { agent_id, reason }`, emitted from the API-timeout branch in `run_subagent_task` right after `interrupt_with_checkpoint`.
- New `AgentLifecycle::Interrupted` (terminal, `STATUS_WARNING`, dot glyph `◌`). `apply_to_delegate` sets status + reason summary; `apply_to_fanout` moves the worker slot out of running. In the fanout counts row, interrupted folds into the existing `failed` bucket (avoids new i18n placeholders), but `aggregate_status` surfaces `Interrupted` distinctly once nothing is running.
- Safety net: `reconcile_subagent_activity_state` (runs on every `AgentList` snapshot) now syncs any card slot still rendered as Pending/Running against non-Running manager snapshot statuses, so a card that missed its terminal envelope can never pin `active_fanout_counts`/`running_agent_count` high indefinitely.

A subsequent agent continuation re-emits `Progress`, which returns the card slot to Running — so the interrupted state is recoverable, matching `agent_eval` `continuable == true` semantics.

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features` (codewhale-tui: clean)
- [x] `cargo test -p codewhale-tui` — 4465 passed, 1 failed: `sandbox::tests::test_parity_linux_landlock_available`, which also fails on clean `main` on this machine (Landlock unavailable in the test kernel), so it is preexisting/environmental.

New regression tests:
- `subagent::tests::api_timeout_preserves_checkpoint_and_agent_eval_continues_from_it` now also asserts the `Interrupted` mailbox envelope is published with the timeout reason.
- `agent_card`: delegate renders `[interrupted]` + reason; fanout drops the worker from running counts and surfaces aggregate interrupted.
- `ui::tests::fanout_interrupted_mailbox_drops_running_count`: `active_fanout_counts` goes `2/2 → 1/2` after the Interrupted envelope.
- `ui::tests::reconcile_syncs_stale_running_cards_with_interrupted_snapshots`: a card that missed its envelope is reconciled from the `AgentList` snapshot, bumping the history revision and correcting `running_agent_count`.
- `mailbox`: `agent_id` extraction covers the new variant.

## Checklist

- [ ] Verified TUI behavior manually if UI changes (covered by unit tests; manual repro of a live API timeout not performed)
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address (n/a — no harvested code)

Link to Devin session: https://app.devin.ai/sessions/000128352623491fb32fd564549e7858

### DIFF
--- a/crates/tui/src/tools/subagent/mailbox.rs
+++ b/crates/tui/src/tools/subagent/mailbox.rs
@@ -58,6 +58,9 @@ pub enum MailboxMessage {
     Completed { agent_id: String, summary: String },
     /// Agent failed with the carried error message.
     Failed { agent_id: String, error: String },
+    /// Agent was interrupted (e.g. API timeout) with a continuable
+    /// checkpoint; the worker is parked waiting for continuation input.
+    Interrupted { agent_id: String, reason: String },
     /// Cancellation propagated to this agent.
     Cancelled { agent_id: String },
     /// Incremental token usage from a sub-agent's API call.
@@ -83,6 +86,7 @@ impl MailboxMessage {
             | Self::ToolCallCompleted { agent_id, .. }
             | Self::Completed { agent_id, .. }
             | Self::Failed { agent_id, .. }
+            | Self::Interrupted { agent_id, .. }
             | Self::Cancelled { agent_id }
             | Self::TokenUsage { agent_id, .. } => agent_id,
             Self::ChildSpawned { child_id, .. } => child_id,
@@ -459,6 +463,13 @@ mod tests {
                     agent_id: "a8".into(),
                 },
                 "a8",
+            ),
+            (
+                MailboxMessage::Interrupted {
+                    agent_id: "a10".into(),
+                    reason: "API call timed out".into(),
+                },
+                "a10",
             ),
             (
                 MailboxMessage::TokenUsage {
--- a/crates/tui/src/tools/subagent/mod.rs
+++ b/crates/tui/src/tools/subagent/mod.rs
@@ -4390,6 +4390,12 @@ async fn run_subagent(
                                 checkpoint.clone(),
                             )?
                         };
+                        if let Some(mb) = runtime.mailbox.as_ref() {
+                            let _ = mb.send(MailboxMessage::Interrupted {
+                                agent_id: agent_id.clone(),
+                                reason: reason.clone(),
+                            });
+                        }
 
                         let next_input = tokio::select! {
                             biased;
--- a/crates/tui/src/tools/subagent/tests.rs
+++ b/crates/tui/src/tools/subagent/tests.rs
@@ -1317,6 +1317,9 @@ async fn api_timeout_preserves_checkpoint_and_agent_eval_continues_from_it() {
     runtime.client = client;
     runtime.manager = Arc::clone(&manager);
     runtime.context = ToolContext::new(tmp.path());
+    let (mailbox, mut mailbox_rx) =
+        crate::tools::subagent::mailbox::Mailbox::new(tokio_util::sync::CancellationToken::new());
+    runtime.mailbox = Some(mailbox);
 
     let task = SubAgentTask {
         manager_handle: Arc::clone(&manager),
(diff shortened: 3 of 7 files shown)


Hmbown:
Verified on scratch/v0.8.59-final-harvest at release base a448194b plus local release-fix commit 2d5f2d23. This is already implemented on the v0.8.59 release branch by 8091513e.

Focused verification run:
- cargo test -p codewhale-tui fanout -- --nocapture
- cargo test -p codewhale-tui provider_wait -- --nocapture
- cargo test -p codewhale-tui (full suite: 4551 passed, 4 ignored, plus integration suites)

Closing as harvested into the release branch.

PR #3148, opened 2026-06-12
fix(exec): resolve --model auto via DEEPSEEK_MODEL env and reorder ExecArgs

hongchen1993:
## Summary

`codewhale exec --model auto` never triggers auto-routing because `ExecArgs.prompt` has `trailing_var_arg=true`, which causes clap to consume `--model auto` as prompt text. The model falls back to `config.default_model()` (e.g. `deepseek-reasoner` for wanjie-ark).

## Root cause

Two issues:

1. **Field ordering**: `prompt` field appears before named args in `ExecArgs`, so clap parses it first and `trailing_var_arg` eats everything after it.

2. **Missing env fallback**: The CLI dispatcher (`codewhale` binary) forwards `--model` via `DEEPSEEK_MODEL` env var to the TUI binary, but `resolve_exec_model` doesn't read it.

## Fix

1. Move `prompt` to the end of `ExecArgs` so named args are parsed first (improves help output and partial parsing).

2. In `resolve_exec_model`, fall back to `DEEPSEEK_MODEL` env var when the explicit arg is absent.

## Verification

Tested with a local proxy capturing actual API requests to Volcengine Ark Agent Plan:

```
codewhale exec --model auto '只回复 OK'
→ model=deepseek-v4-flash ✅

codewhale exec --model auto '请帮我分析架构并实现重构方案...'
→ model=deepseek-v4-pro ✅
```

Without this fix, both requests would use `deepseek-reasoner` (the provider default), and `--model auto` would be silently ignored.

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -307,14 +307,6 @@ Plain `codewhale exec` is a one-shot model response. Use `--auto` for
 non-interactive filesystem/shell tool use.
 ")]
 struct ExecArgs {
-    /// Prompt to send to the model
-    #[arg(
-        value_name = "PROMPT",
-        required = true,
-        trailing_var_arg = true,
-        allow_hyphen_values = true
-    )]
-    prompt: Vec<String>,
     /// Override model for this run
     #[arg(long)]
     model: Option<String>,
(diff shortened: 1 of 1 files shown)


Hmbown (approved):
very helpful - thank you so much

Hmbown:
Triage for v0.8.59: verified against `codex/v0.8.59-release-ready` via scratch merge. The release branch already covers this via 862cb2e3 ("fix(exec): preserve auto model handoff"): `ExecArgs` is reordered so the trailing-var-arg prompt comes last, and `resolve_exec_model` falls back to env with a superset of this PR's behavior (`CODEWHALE_MODEL` then `DEEPSEEK_MODEL`). The fix ships in v0.8.59; thanks @hongchen1993. Can be closed once the release branch lands on main.

Hmbown:
Verified on scratch/v0.8.59-final-harvest at release base a448194b plus local release-fix commit 2d5f2d23. This is already implemented on the v0.8.59 branch by 862cb2e3.

Focused verification run:
- cargo test -p codewhale-tui exec_model_resolution -- --nocapture
- cargo test -p codewhale-tui (full suite: 4551 passed, 4 ignored, plus integration suites)

Closing as harvested into the release branch.

PR #1865, opened 2026-05-21
Add opt-in Pro Plan routing profile

dumbjack:
## Summary

- Add Pro Plan as an explicit `/mode pro-plan` routing profile, not a fourth default mode in the `Tab` cycle or `/mode` picker.
- Keep the existing Plan Confirmation gate: the user chooses Pro Plan and chooses whether accepted execution uses Agent-style approvals or a one-pass YOLO execution.
- Route phases across existing mode semantics: Plan/Review use the resolved Pro model with read-only Plan policy; Execute uses the resolved Flash model with Agent policy, or temporary YOLO policy only for the accepted execution pass.
- Fail closed when routing is unavailable or bypassed: raw `AppMode::ProPlan` maps to read-only Plan behavior, and missing Flash routes fall back to the resolved Pro model.
- Reduce false positives and loop risk: the planning trigger now only wraps clearly implementation-like requests, and review-requested reimplementation clears temporary YOLO auto-approval before returning to Execute.

## Why this helps

This follows the same broad workflow idea as Claude Code's public Opus Plan pattern: use the strongest model where planning and review have the most leverage, then use a faster model for the implementation pass. In CodeWhale this is deliberately opt-in so the main model/menu surface stays neutral and users can tell when they are choosing a workflow versus when CodeWhale is choosing a phase model inside that workflow.

The profile is intended for larger or riskier coding tasks, not as a default for tiny edits.

## Safety / Reviewability

- Default visible modes remain Plan, Agent, and YOLO.
- `/mode pro-plan` is the explicit entry point; numeric mode shortcuts remain `1`, `2`, and `3` for the visible picker rows.
- Plan/Review are read-only and reuse existing Plan prompt/policy paths.
- Execute uses existing Agent/YOLO policy paths after the plan confirmation choice.
- Interrupted or failed turns do not advance Pro Plan phase state.
- Entering Pro Plan saves `auto_model`, disables it while routed by phase, and restores it on exit.
- User-facing status/help/PlanPrompt strings remain localized.

## Validation

- `cargo fmt --all -- --check`
- `cargo check -p codewhale-tui`
- `cargo test -p codewhale-tui pro_plan --quiet` (20 passed)
- `cargo test -p codewhale-tui plan_prompt --quiet` (6 passed)
- `cargo test -p codewhale-tui localization --quiet` (8 passed)
- `cargo test -p codewhale-tui --quiet` (full `codewhale-tui` pass: 4006 passed, 4 ignored, plus integration targets)

[...]

PR #3193, opened 2026-06-13
Add config-gated Pro Plan routing profile

dumbjack:
## Summary

Fresh follow-up to the closed Pro Plan exploration in #1865. This reworks the feature as an explicit, config-gated routing profile with no default mode/menu change.

- adds `pro_plan_profile = false` as the default settings gate
- rejects `/mode pro-plan` until the profile is enabled with `/config pro_plan_profile true --save`
- keeps the default `Tab` cycle and `/mode` picker limited to Agent / Plan / YOLO
- routes Pro Plan phases as Plan -> Execute -> Review -> Done using explicit `<pro_plan ...>` markers
- keeps Plan/Review on the resolved Pro route and read-only policy; Execute uses Flash only when the active provider advertises it, otherwise falls back to Pro
- preserves fail-closed raw `AppMode::ProPlan` behavior in engine registry, sandbox, runtime prompt, and approval policy
- keeps PlanPrompt on current main's scroll/confirmation UI while routing user-facing strings through localization

## Validation

- `cargo check -p codewhale-tui`
- `cargo test -p codewhale-tui pro_plan --quiet`
- `cargo test -p codewhale-tui plan_prompt --quiet`
- `cargo test -p codewhale-tui localization --quiet`
- `cargo test -p codewhale-tui config_view_exposes_all_available_saved_settings --quiet`
- `cargo test -p codewhale-tui --quiet`

Full `codewhale-tui` pass: 4502 passed, 0 failed, 4 ignored for the main bin target, plus integration/feature targets passed.

## Notes

This intentionally does not add Pro Plan to `default_mode`, the mode picker, or the visible cycle. The profile remains available only after an explicit saved setting so the main model/menu surface stays neutral.


Hmbown:
oh interesting are you thinking of this kind of like the claude code ultraplan? i'm hoping to harden / strenghten planning overall so this is helpful for me to get an implementation. 

if you're able to clarify your intent with this a bit more it would help me out a ton!

Hmbown:
ah I see now - so planning with stronger model and implementing with weaker. i've seen the reverse as well especially in codex. I think separating the planning model and the executing model is probably the happy medium and figuring out how to configure that. i'll work on getting this in! thank you so much for this greaet idea

Hmbown:
ok, so I'm working on a full "party" configuration setup, and this mode would go great in that. I think allowing people to configure these kind of setups is ideal, so I really appreciate you thinking thru this.

Hmbown:
Rechecked this against the current v0.8.64 integration branch in #3373. The config-gated routing profile remains a useful direction, but the PR now conflicts through the mode/engine/UI path (`core/engine`, `dispatch`, `turn_loop`, `tui/app`, `plan_prompt`, `tui/ui`) plus docs.

I am keeping it queued for the broader model/profile configuration work rather than mixing it into the v0.8.64 security stabilization branch late.

Hmbown:
Thanks @dumbjack. I am closing this old conflicted release-queue PR to clear the stale previous-release backlog rather than leaving it open across the newer model/profile work.

This did not land as a direct merge. The useful product direction from this PR is still credited here: explicit, config-gated planning/execution profiles rather than changing the default mode surface. If we revive it, it should come back as a fresh branch against current main and the broader profile configuration work.

PR #2808, opened 2026-06-06
feat(runtime-api): add session save, undo/retry, and snapshot endpoints for GUI

gaord:
## Summary

Add Runtime API endpoints to align GUI capabilities with TUI, following the principle of **reusing TUI's existing capabilities** rather than reimplementing logic in the API layer.

## New Endpoints

| Endpoint | Description |
|----------|-------------|
| `POST /v1/sessions` | Save session using `Engine::get_session_snapshot()`, reusing TUI's `build_session_snapshot` code path |
| `POST /v1/threads/{id}/undo` | Undo conversation turns via `fork_at_user_message` |
| `POST /v1/threads/{id}/patch-undo` | Snapshot-based file rollback + undo (mirrors TUI's `/undo` command) |
| `POST /v1/threads/{id}/retry` | Undo + resend last user message |
| `GET /v1/snapshots` | List workspace snapshots |
| `POST /v1/snapshots/{id}/restore` | Restore a snapshot |

## Key Design Decisions

1. **Session saving reuses Engine state**: Instead of reconstructing messages from turn items, `save_current_session` calls `engine.get_session_snapshot()` which returns the authoritative messages and token usage directly from the Engine. This ensures data consistency with TUI's internal state.

2. **Oneshot channel for snapshot**: `Op::GetSessionSnapshot` uses a oneshot channel to return the snapshot directly to the caller, avoiding competition with the SSE event stream on the mpsc receiver.

3. **API as thin adapter**: Endpoints are thin wrappers that receive HTTP requests, call TUI's existing functions (`SessionManager`, `SnapshotRepo`, `fork_at_user_message`), and return results. No business logic is reimplemented in the API layer.

## Changes

- `core/ops.rs`: Add `SessionSnapshot` struct and `GetSessionSnapshot` Op
- `core/engine.rs`: Handle `GetSessionSnapshot` in the engine loop
- `core/engine/handle.rs`: Add `EngineHandle::get_session_snapshot()` method
- `runtime_api.rs`: Add 6 new endpoints
- `runtime_threads.rs`: Add `get_engine()` public wrapper; enhance `reconstruct_messages_from_turns` with ToolCall/FileChange support

## Testing

- `cargo build` passes
- Session save produces correct messages and token counts
- Undo/retry create forked threads with correct turn history
- Patch-undo restores files from snapshots and removes conversation turns

<details open><summary><h3>Greptile Summary</h3></summary>

[...]

PR #3199, opened 2026-06-13
feat(runtime-api): add PUT /v1/sessions endpoint for engine-based session save

gaord:
Focused slice from #2808: add a PUT /v1/sessions endpoint that saves a thread's current engine state as a session. This is the session-save slice that @Hmbown identified as the next harvest target.

**New endpoint**: PUT /v1/sessions saves a thread live engine state as a session via a oneshot channel, so token counts and message ordering are authoritative. Unlike POST /v1/sessions (which reconstructs from turn items), this matches TUI build_session_snapshot behavior. Backward compatible: existing POST endpoint is preserved.

**Implementation**: SessionSnapshot struct + Op::GetSessionSnapshot variant (ops.rs), engine loop handler (engine.rs), get_session_snapshot method (handle.rs), get_engine public wrapper (runtime_threads.rs), PUT /v1/sessions route + save_current_session handler (runtime_api.rs).

**Bug fix**: load_session errors were silently swallowed by Err(_) wildcard. Only io::ErrorKind::NotFound now falls back to creating a new session; other I/O errors (e.g. PermissionDenied) are propagated.

**Not included** (need atomicity tests): undo, patch-undo, retry, restore-snapshot, ToolCall/FileChange reconstruction.

Ref: #2808

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -66,7 +66,7 @@ use super::capacity_memory::{
 };
 use super::coherence::{CoherenceSignal, CoherenceState, next_coherence_state};
 use super::events::{Event, TurnOutcomeStatus};
-use super::ops::{Op, USER_SHELL_TOOL_ID_PREFIX};
+use super::ops::{Op, SessionSnapshot, USER_SHELL_TOOL_ID_PREFIX};
 use super::session::Session;
 use super::tool_parser;
 use super::turn::{TurnContext, TurnToolCall, post_turn_snapshot, pre_turn_snapshot};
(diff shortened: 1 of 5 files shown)


Hmbown:
Thank you @gaord! The PUT /v1/sessions snapshot path — using a oneshot so it does not contend with the SSE stream — is a nice design. Landed in the v0.8.61 release prep with your authorship after a tiny rustfmt pass. Appreciated.

Hmbown:
Thank you @gaord — the runtime-api session-save (PUT /v1/sessions) shipped in v0.8.61. Closing as resolved; credited. 🐳

PR #3221, opened 2026-06-14
fix(exec): honour DEEPSEEK_BASE_URL and DEEPSEEK_MODEL env vars

hongchen1993:
## Summary

`codewhale --provider wanjie-ark --base-url <url> --model auto exec ...` does not work because the TUI binary ignores the `DEEPSEEK_BASE_URL` and `DEEPSEEK_MODEL` environment variables that the CLI dispatcher forwards.

## Root cause

Three gaps:

1. **`resolve_exec_model`** only reads the explicit `--model` arg, but `ExecArgs.prompt` has `trailing_var_arg=true` which eats `--model auto` as prompt text. The CLI dispatcher forwards `--model` via `DEEPSEEK_MODEL` env var, but this function doesn't read it.

2. **Exec command handler** doesn't read `DEEPSEEK_BASE_URL`. The CLI dispatcher forwards `--base-url` via this env var, but the handler creates the client with the config's default base_url.

3. **`deepseek_base_url()`** only considers provider-specific config entries and provider defaults. It doesn't fall back to `DEEPSEEK_BASE_URL` / `CODEWHALE_BASE_URL` env vars.

## Fix

1. `resolve_exec_model`: add `exec_model_env_override()` that checks `CODEWHALE_MODEL` then `DEEPSEEK_MODEL`.

2. Exec handler: read `DEEPSEEK_BASE_URL` and set `config.base_url` before creating the client.

3. `deepseek_base_url()`: add `env_base_url_override()` as a fallback between configured base_url and provider default.

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -2530,7 +2530,9 @@ impl Config {
             let api_key = config_api_key.or(env_api_key.as_deref());
             resolve_xiaomi_mimo_base_url(configured_base_url, api_key, mode)
         } else {
-            configured_base_url.unwrap_or_else(|| {
+            configured_base_url
+                .or_else(|| env_base_url_override())
+                .unwrap_or_else(|| {
                 match provider {
                     ApiProvider::Deepseek => DEFAULT_DEEPSEEK_BASE_URL,
                     ApiProvider::DeepseekCN => DEFAULT_DEEPSEEKCN_BASE_URL,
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you @hongchen1993! Honoring the forwarded base-url and model env vars in headless exec is a real fix. Landed in the v0.8.61 release prep with your authorship. Appreciated.

Hmbown:
Thank you @hongchen1993 — this shipped in CodeWhale v0.8.61: `exec` now honours `DEEPSEEK_BASE_URL`/`DEEPSEEK_MODEL`. Closing as resolved; your authorship is credited in the release. 🐳

PR #3242, opened 2026-06-15
feat: add workspace_follow_symlinks setting for symlink-aware tool operations

gaord:
Add workspace_follow_symlinks setting so walk-based tools and UI components follow symbolic links during directory traversal. See commit message for full details.


Hmbown:
great idea!!! thank you!

PR #3274, opened 2026-06-17
feat(release): build static Linux x64 binaries with musl

wavezhang:
## Summary

Switch Linux x64 release builds from dynamic glibc (`x86_64-unknown-linux-gnu`) to static musl (`x86_64-unknown-linux-musl`) in the GitHub Actions release workflow.

This is the GitHub Actions counterpart to **PR #2903** which added musl static builds to the CNB pipeline (`.cnb.yml`).

### Changes

| Area | Before | After |
|------|--------|-------|
| Matrix target (codewhale) | `x86_64-unknown-linux-gnu` | `x86_64-unknown-linux-musl` |
| Matrix target (codewhale-tui) | `x86_64-unknown-linux-gnu` | `x86_64-unknown-linux-musl` |
| Linux deps install | All Linux targets | Exclude musl (no libdbus needed) |
| Build step | Single generic `cargo build` | Dedicated musl step + generic step (skip musl) |

### Details

- **Matrix**: Change `target` for `codewhale-linux-x64` and `codewhale-tui-linux-x64` from `gnu` to `musl`
- **Dedicated musl build step**: Installs `musl-tools`, adds `x86_64-unknown-linux-musl` target, then builds
- **Skip libdbus for musl**: The `keyring` feature is already gated behind `target != musl` in `crates/secrets` (from PR #2903), so no `libdbus-1-dev` needed
- **Skip generic Build step for musl**: The dedicated step handles musl builds; generic step runs for all other targets

### Why

Static musl binaries have **no glibc/libdbus runtime dependency** and run on any Linux distribution without additional packages — matching what the CNB pipeline already produces.

### Testing

This change mirrors the CNB pipeline approach (PR #2903, merged). The GitHub Actions workflow will be validated on the next release dispatch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           components: clippy, rustfmt
       - name: Install Linux system dependencies
-        if: runner.os == 'Linux'
+        if: runner.os == 'Linux' && matrix.target != 'x86_64-unknown-linux-musl'
         run: |
           for i in 1 2 3 4 5; do
             sudo apt-get update && break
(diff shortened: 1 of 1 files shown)


Hmbown:
thank you for this!!

PR #3284, opened 2026-06-17
perf(tui): debounce thinking-stream re-renders (#1620)

### LINKED ISSUE #1620: 思考过程巨慢无比，一个字吐半天，哪个环节出了问题？
Image

LeoLin990405:
## Problem

Fixes #1620 — reasoning (thinking) blocks stream painfully slowly, "one character takes ages to appear", on fast reasoning models.

## Root cause

In `crates/tui/src/tui/streaming_thinking.rs`, `append()` calls `bump_active_cell_revision()` on **every** reasoning delta. Each bump invalidates the active cell's wrap cache and forces a full re-wrap of the live tail. Reasoning deltas arrive far faster than the eye can follow (often per-grapheme), so the renderer re-wraps the entire live tail on every character — the visible stream crawls.

## Fix

Debounce the *intermediate* revision bumps to at most one per ~100ms window (≈10 repaints/sec, well under the 120 FPS frame cap, imperceptible as lag):

- New `App.thinking_revision_last_bump_at: Option<Instant>` tracks the last bump.
- `append()` commits the text to the cell **immediately** (content is never delayed) and only the *bump* is throttled via `bump_thinking_revision_throttled()`.
- `append()` is split into `append_at(.., now)` with an injectable clock for deterministic testing.

### Correctness (no content can be lost)

- Throttling only skips **intermediate** repaints — appended text always lands in the cell right away.
- `finalize_active_entry()` **always forces a bump** and resets the throttle window, so the final reasoning text is fully rendered. This covers both the normal `ThinkingComplete` path and the error/interrupt path (`finalize_current` → `finalize_active_entry`), so a mid-stream disconnect can't strand the last chunk.
- The translation-mode path (`set_placeholder`) is untouched.

## Testing

- New colocated unit test `issue_1620_throttles_thinking_bumps_without_losing_content`: a burst of chunks within one window coalesces to a single bump, a chunk past the window bumps again, full content is preserved, and finalize forces exactly one final bump + flushes the tail.
- `cargo clippy -p codewhale-tui --all-features` clean.
- `cargo test -p codewhale-tui --all-features` green (4856 passed / 0 failed).

### DIFF
--- a/crates/tui/src/tui/app.rs
+++ b/crates/tui/src/tui/app.rs
@@ -1731,6 +1731,14 @@ pub struct App {
     /// thinking into the active cell so it groups visually with tool calls
     /// until the next assistant prose chunk flushes the group into history.
     pub streaming_thinking_active_entry: Option<usize>,
+    /// Instant of the last throttled active-cell revision bump for the
+    /// in-flight thinking stream (#1620). Reasoning chunks arrive faster than
+    /// the eye can read, and each bump invalidates the active cell's wrap
+    /// cache, forcing a full re-wrap. We debounce intermediate bumps to a
+    /// time window so high-frequency thinking deltas no longer trigger a
+    /// re-render per character. `None` means "no bump since the last
+    /// finalize" so the first chunk of a block always renders immediately.
+    pub thinking_revision_last_bump_at: Option<Instant>,
     /// Newline-gated streaming collector state.
     pub streaming_state: StreamingState,
     /// Live approximate output tokens for the current assistant stream.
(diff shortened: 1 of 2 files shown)


LeoLin990405:
Thanks for the review! Keeping `is_none_or` here: the workspace declares `rust-version = "1.88"` in the root `Cargo.toml`, and `Option::is_none_or` has been stable since 1.82, so it's within MSRV. On the repo's toolchain clippy's `unnecessary_map_or` lint (warn-by-default) actually flags `map_or(true, …)` and suggests `is_none_or`, so switching would break the `cargo clippy` clean gate.

Hmbown:
interesting! this is very smart. thank you!

PR #3286, opened 2026-06-17
fix(tui): ensure type:object on Kimi parameters root for all schema shapes (#3281)

### LINKED ISSUE #3281: [moonshot] v0.8.61 #3265 fix incomplete — parameters still missing type:object for $ref / anyOf / allOf root schemas
## Description

The v0.8.61 fix for #3265 (`sanitize_for_kimi_parameters`) only adds `"type": "object"` to the parameters root when the schema matches a narrow set of conditions:

```rust
let root_object_schema = obj.is_empty()
    || obj.contains_key("properties")
    || obj.contains_key("required")
    || obj.contains_key("additionalProperties");
```

This means **other valid JSON Schema root shapes** are left without `type` — and Kimi/Moonshot still rejects them with the same 400 error:

- `{"$ref": "#/definitions/FileArgs"}` — no `type` added
- `{"allOf": [...]}` — no `type` added  
- `{"anyOf": [...]}` — no `type` added
- `{"oneOf": [...]}` — no `type` added

### Secondary issue: `sanitize_for_kimi` removes `type` from root

Even for schemas that DO get `type: "object"` injected, the subsequent `sanitize_for_kimi` pass **removes `type` from the parent** when `anyOf`/`oneOf` is present:

```rust
// sanitize_for_kimi, line 979-981:
let should_push = obj.contains_key("type") 
    && (obj.contains_key("anyOf") || obj.contains_key("oneOf"));
if should_push && let Some(type_val) = obj.remove("type") {
    // pushes type into items, removes from parent
}
```

This means a parameters schema like:
```json
{"anyOf": [{"type": "object", "properties": {...}}, {"type": "null"}]}
```

After `sanitize_for_kimi_parameters` adds `type` → then `sanitize_for_kimi` **removes** it.  Net result: no `type` at root → Kimi 400.

## Reproduction

Same as #3265 — switch to Moonshot provider with kimi-k2.7-code, send any message. The error persists in v0.8.61 for tools whose parameters use `$ref` / `anyOf` / `allOf` / `oneOf` at root.

## Suggested Fix

1. **`sanitize_for_kimi_parameters`**: always add `"type": "object"` when missing, not just for the narrow `root_object_schema` shapes
2. **`sanitize_for_kimi`**: when pushing `type` into `anyOf`/`oneOf` items, **keep it on the parent** instead of removing it — Kimi requires `type` at every level (#3265)

## Environment

- **CodeWhale**: v0.8.61 (14ac0319e8f4)  
- **API**: https://api.moonshot.cn/v1
- **Provider type**: `openai_compatible`

idling11:
sanitize_for_kimi_parameters only injected type:object when the root matched a narrow set of conditions (empty, has properties/required/additionalProperties). Schemas using $ref, allOf, anyOf, or oneOf at root were left without type, causing Kimi/Moonshot to reject them with 400.

Fix: run sanitize_for_kimi first (so nested anyOf/oneOf handling is correct), then unconditionally ensure type:object at the parameters root. Skips $ref schemas only (sibling keywords are invalid per JSON Schema spec).

Added 4 tests covering anyOf, allOf, oneOf, and $ref root shapes.

Close https://github.com/Hmbown/CodeWhale/issues/3281

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/tools/schema_sanitize.rs
+++ b/crates/tui/src/tools/schema_sanitize.rs
@@ -1000,28 +1000,60 @@ pub fn sanitize_for_kimi(schema: &mut serde_json::Value) {
 
 /// Normalize a complete Kimi / Moonshot `function.parameters` object.
 ///
-/// This is root-only because recursively injecting `type: object` into every
-/// empty object would corrupt JSON Schema maps such as `"properties": {}`.
-/// The existing Kimi pass still handles nested `anyOf` / `oneOf` placement.
+/// Kimi / Moonshot requires `"type": "object"` on the parameters root
+/// regardless of whether the schema uses `properties`, `$ref`, `anyOf`,
+/// `allOf`, or `oneOf`.  We run `sanitize_for_kimi` first so nested
+/// `anyOf` / `oneOf` handling is correct, then unconditionally ensure
+/// `type: object` is present at the root (#3281).
+///
+/// This is root-only because recursively injecting `type: object` into
+/// every empty object would corrupt JSON Schema maps such as
+/// `"properties": {}`.
 pub fn sanitize_for_kimi_parameters(parameters: &mut serde_json::Value) {
     if !parameters.is_object() {
         *parameters = serde_json::Value::Object(Map::new());
     }
 
-    if let Some(obj) = parameters.as_object_mut() {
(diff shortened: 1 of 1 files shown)


Hmbown:
wonderful - thank you.

Issue #3240, opened 2026-06-15
Legacy deepseek configuration


Final527:
## Description

Although the project has been changed to codewhale, the program still creates the .deepseek configuration directory at runtime. On Windows, both the .codewhale and .deepdeek folders are created in the user directory

## Environment

- OS: Windows 11
- codewhale version: 0.8.60
- Install method: release binary
- `codewhale doctor` summary:
- Model/provider: deepseek-v4-pro / DeepSeek
- Terminal app: Windows Terminal
- Shell: PowerShell

Hmbown:
Good catch, and thanks for separating the scope clearly. To set expectations: v0.8.61 made the legacy *install/update* path resolve `deepseek` to `codewhale` (#2960 #2924 #2917), but that is about the installer/updater, not the runtime. Your specific report — the app still creating a `.deepseek` config directory at runtime on Windows (alongside `.codewhale`) — is a distinct, still-open issue. Keeping this open and flagging it for the v0.8.62 packaging/hygiene pass so the runtime stops writing the legacy directory.

Hmbown:
Retargeting into v0.8.63 from the release triage pass. This belongs in the same lane as the config/sub-agent/runtime-control work because it affects user trust, release stability, or first-run/runtime behavior.

Hmbown:
Fixed on the v0.8.63 release-train branch in PR #3347.

Fixed by commits 827d3c330c5daad94f7dd342baa00534b2ee2c3c, 44eedb0deec8afb5799853a606a833fb0250fce1, and b1c5f51d5816fcfdcfbc32166c8a9f9c73878b6e: fresh onboarding now prefers `~/.codewhale/.onboarded`, state write paths for sessions / slop ledger / trophies / catalog use the primary CodeWhale state root, legacy subdirs are relocated on first write so old data is preserved, and state-subdir validation rejects traversal/rooted paths across platforms.

Verified locally on `codex/v0.8.63-integration` with:

- `cargo test -p codewhale-config --locked`
- `cargo test -p codewhale-config state_resolvers_reject_path_traversal_subdirs --locked`

Thanks @Final527 for the Windows report, and thanks @nightt5879 for the PR #3302 slice that fixed the onboarding marker path. Closing as complete for v0.8.63; please reopen if a new build still creates `.deepseek` during normal runtime writes.

PR #3294, opened 2026-06-17
fix(tui): write composer history under .codewhale, not legacy .deepseek

wuisabel-gif:
`default_history_path` hardcoded `~/.deepseek/composer_history.txt` for both read and write, so `append_history` ran `create_dir_all` on the legacy dir on every non-slash prompt submission — recreating `~/.deepseek/` at runtime even on fresh installs (e.g. alongside `.codewhale` on Windows).

Resolve the history path with the same primary/legacy pattern already used by `snapshot::paths` and `artifacts`: prefer `~/.codewhale/`, and fall back to the legacy `~/.deepseek/` file only when one already exists. Fresh installs now target `.codewhale` and never recreate the legacy directory, while un-migrated users keep reading and appending to their existing legacy history.

Refs #3240 (partial: composer history; other legacy write sites remain).

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/composer_history.rs
+++ b/crates/tui/src/composer_history.rs
@@ -1,8 +1,10 @@
 //! Cross-session composer input history (#366).
 //!
-//! Persists user-typed prompts to `~/.deepseek/composer_history.txt` so
-//! pressing Up-arrow at the composer recalls submissions from previous
-//! sessions, not just the current one. One entry per line, oldest first,
+//! Persists user-typed prompts to `~/.codewhale/composer_history.txt`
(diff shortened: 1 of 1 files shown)


Hmbown:
thank you for PRing this!!

Hmbown:
Thank you @wuisabel-gif — the composer-history migration to `.codewhale` is already present on `hunter/0.8.62-glm-subagents` as your commit 681cd8bf, with authorship preserved. I’m leaving this PR open so it can close with normal GitHub credit when the release branch reaches `main`.\n\nOne note from review: this is a good targeted fix for composer history, but I’m not treating the broader legacy-path/CODEWHALE_HOME cleanup as fully closed yet.

Issue #3259, opened 2026-06-15
app-server: clean up delegated serve child when dispatcher exits


Hmbown:
## Follow-up from v0.8.61 app-server release verification

`codewhale app-server --http` / `--mobile` delegates to the sibling `codewhale-tui serve` path. The dispatcher waits with `Command::status()`, but killing the dispatcher can leave the delegated `codewhale-tui` listener alive and reparented.

This was treated as a v0.8.62 teardown hardening follow-up, not a v0.8.61 release blocker, because normal foreground shutdown and explicit listener cleanup worked during verification.

Expected direction:

- Add a controlled shutdown/cleanup path for delegated long-running server children.
- Add a regression smoke that starts a delegated runtime API listener on a loopback free port, terminates the dispatcher, and asserts no listener remains on that port.

Hmbown:
Retargeting into v0.8.63 from the release triage pass. This belongs in the same lane as the config/sub-agent/runtime-control work because it affects user trust, release stability, or first-run/runtime behavior.

Hmbown:
Partial v0.8.63 update from PR #3347:

Commit 11df4e539b55c5338c3660e869810d237939e14f landed the catchable-signal teardown path for delegated `serve` / `app-server` children. It supervises the delegated child, forwards Ctrl+C / SIGTERM / SIGHUP where supported, kills and reaps the child before the dispatcher exits, and keeps the signal-style exit code.

That covers the normal controlled-shutdown path, but the commit intentionally leaves the uncatchable dispatcher-death case open: SIGKILL still cannot run cleanup code, and needs OS-level follow-up such as Linux `PR_SET_PDEATHSIG` or Windows Job Objects.

Because that remaining edge is real and explicitly out of scope for the landed slice, I am keeping this issue open and retargeting the remaining work to v0.8.64 rather than closing it from the partial fix.

Hmbown:
Local v0.8.64 integration update after #3317:

- `11df4e539` / merge `b615df4b3` already carries the catchable-signal supervisor from PR #3317 by @wuisabel-gif.
- `fd35e1cd4` adds the Linux hard parent-death slice: delegated `serve` / `app-server` children now get `PR_SET_PDEATHSIG(SIGTERM)` before spawn, so Linux tears down the listener child if the dispatcher dies before the graceful supervisor can run.

Verification run locally:
- `cargo fmt --all -- --check`
- `cargo test -p codewhale-cli --locked server_teardown_tests`
[...]

PR #3317, opened 2026-06-18
fix(cli): tear down delegated serve/app-server child on dispatcher ex…

wuisabel-gif:
## Summary

Refs #3259 (partial).

`codewhale serve --http`/`--mobile` and `codewhale app-server --http`/`--mobile`
delegate to the sibling `codewhale-tui` binary via `Command::status()`, which
reaps the delegated child only on the child's own exit. Terminating the
dispatcher while the delegated server is still running could leave the runtime
API listener alive and reparented (the orphaned-listener bug in #3259).

This routes the two **server** delegation paths through a new
`delegate_server_to_tui` that supervises the child under Tokio:

- forwards termination — **Ctrl+C on all platforms, SIGTERM/SIGHUP on Unix**
  (e.g. `kill <pid>` or a service manager stopping the process) — by killing
  and reaping the child *before* the dispatcher exits, and
- sets `kill_on_drop` so an unwinding dispatcher also tears the child down.

Interactive (non-server) delegations keep the existing `status()` path, so
terminal job control / Ctrl+C behavior for the TUI is unchanged.

The teardown decision is factored into a small testable helper
`supervise_server_child`.

### Scope / follow-up

An uncatchable `SIGKILL` of the dispatcher (or a hard crash) still can't run
this path. Covering that needs `PR_SET_PDEATHSIG` on Linux and a Job Object on
Windows (the repo already uses both idioms in `crates/tui/src/tools/shell.rs`
and the Windows sandbox module). That, plus a binary-level integration smoke
that binds a real loopback port, is left as follow-up on #3259.

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features` — 0 warnings / 0 errors
- [x] `cargo test -p codewhale-cli` — 114 passed (2 new), no regressions

New unit tests (`server_teardown_tests`, Unix-gated):
- `supervisor_propagates_child_exit_when_no_shutdown` — child's own exit status is returned when no shutdown fires.
- `shutdown_signal_kills_and_reaps_long_running_child` — a shutdown signal kills a long-running child and reaps it (`child.id()` is `None` afterward), so no listener is left reparented.

## Checklist

- [ ] Verified TUI behavior manually if UI changes — n/a (no UI change)
- [ ] Harvested/co-authored credit uses a GitHub numeric noreply address — n/a (single-author; commit uses the numeric noreply)

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -743,7 +743,9 @@ fn run() -> Result<()> {
         }
         Some(Commands::Serve(args)) => {
             let resolved_runtime = resolve_runtime_for_dispatch(&mut store, &runtime_overrides);
-            delegate_to_tui(&cli, &resolved_runtime, tui_args("serve", args))
+            // `serve` starts a long-running runtime API listener; supervise the
+            // delegated child so it is torn down with the dispatcher (#3259).
+            delegate_server_to_tui(&cli, &resolved_runtime, tui_args("serve", args))
         }
         Some(Commands::Completions(args)) => {
             let resolved_runtime = resolve_runtime_for_dispatch(&mut store, &runtime_overrides);
@@ -1619,7 +1621,9 @@ fn run_app_server_command(
     // canonical `app-server --http`/`--mobile` entrypoint reuses that mature server
     // by delegating to the sibling TUI binary (the same mechanism `serve` uses).
     if args.http || args.mobile {
-        return delegate_to_tui(cli, resolved_runtime, app_server_serve_passthrough(&args));
+        // Delegated runtime API listener — supervise it so the child does not
+        // outlive the dispatcher (#3259).
+        return delegate_server_to_tui(cli, resolved_runtime, app_server_serve_passthrough(&args));
     }
 
     let runtime = tokio::runtime::Builder::new_multi_thread()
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @wuisabel-gif — this is already carried locally on the v0.8.64 integration branch as `11df4e539` via merge `b615df4b3`, preserving your authorship.

I rechecked the current branch against the PR: the catchable-signal delegated server teardown behavior is present, and the only diff against the PR head is unrelated newer CLI work on the integration branch. The remaining #3259 scope is still the uncatchable dispatcher-death follow-up noted in the issue/PR text: Linux `PR_SET_PDEATHSIG` and Windows Job Objects or equivalent OS-level cleanup.

I am leaving this PR and #3259 open until the integration branch is pushed/landed and the remaining edge has an explicit disposition.

Hmbown:
Thanks @wuisabel-gif. This cleanup landed on the release path, so I am closing the now-conflicting duplicate PR.

Credit trail:
- b615df4b3a1276d6fb744437ba859839d9543b69: local integration merge of PR #3317 by @wuisabel-gif
- 11df4e539 and c9e887e5e carried the delegated server teardown behavior for #3259

Appreciate the focused dispatcher-child teardown fix.

PR #3300, opened 2026-06-18
feat(tui): preserve thinking/tool blocks when seeding thread from session

gaord:
## Summary

Replaces the text-only `seed_thread_from_messages` with a block-type-aware implementation that preserves ContentBlock variants (Thinking, ToolUse, ToolResult) as distinct TurnItem entries, so `loadHistory` / `messages_from_thread_detail` can reconstruct the full conversation including process information.

### Key changes

- Add `session_id` field to `ThreadRecord` for session→thread linking
- Add `SeedItem` / `TurnSeed` helpers for block-type turn construction
- Add `set_thread_session_id` to associate a session with a thread
- Rewrite `messages_from_thread_detail` to emit typed ContentBlocks
- Link session→thread on both `resume_session` and `create_session_from_thread`
- Fall back to turn-based reconstruction when session file is missing

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p codewhale-tui --all-features --locked` (no new warnings)
- [x] `cargo test --workspace --all-features --locked` (4955 passed, 1 known flaky)

## Checklist

- [x] Updated docs or comments as needed
- [x] Added or updated tests where relevant
- [x] Verified TUI behavior manually if UI changes
- [ ] Harvested/co-authored credit uses a GitHub numeric noreply address

PR #3326, opened 2026-06-19
Preserve session process history on thread resume

gaord:
## Summary
- load saved session history from the runtime thread manager's configured sessions directory
- keep consecutive user tool_result blocks batched when seeding and reconstructing thread history
- add regression coverage for parallel tool results and custom sessions-dir resume

## Verification
- cargo fmt
- cargo test --workspace
- cargo build --release -p codewhale-cli -p codewhale-tui


gaord:
Closing this duplicate PR. The fix has already been added to the original PR #3300.

PR #3329, opened 2026-06-19
fix(config): restore huggingface env precedence

gaord:
This restores the Hugging Face API key env precedence on the TUI config surface so `scripts/check-provider-registry.py` (the `CI/Lint` gate) passes again on `main`.

Background
- `crates/tui/src/config.rs` previously expressed the precedence through a helper (`provider_env_api_key`), but the registry lint script checks for the literal `std::env::var("HUGGINGFACE_API_KEY")` / `std::env::var("HF_TOKEN")` strings in this file and fails when it cannot find them in the expected order.
- That drift is unrelated to the work in #3300 (session/thread process preservation), so it is being fixed here on its own branch off the latest `main`.

What changed
- `provider_env_api_key()` now spells out the Hugging Face env order explicitly: `HUGGINGFACE_API_KEY` first, falling back to `HF_TOKEN`.
- Behavior on the runtime side is unchanged: `HF_TOKEN` alone still works, `HUGGINGFACE_API_KEY` still wins when both are set, empty values are ignored, and the rest of the providers continue to flow through the existing helper path.
- The Hugging Face config tests (`huggingface_hf_token_env_api_key_resolves`, `huggingface_short_env_fallbacks_configure_route`, `huggingface_env_overrides_key_base_url_and_model`, `huggingface_missing_key_error_mentions_env_fallbacks`, etc.) still pass.

Validation
- `python3 scripts/check-provider-registry.py` now reports `Provider registry drift check passed.`
- `cargo test -p codewhale-tui huggingface_` is green (6 passed).

Out of scope
- The failing `tools::shell` tests on `windows-latest` for #3300 are unrelated to this file and are tracked separately.

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -16,7 +16,7 @@ use serde_json::json;
 use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
 
 use crate::audit::log_sensitive_event;
-use crate::features::{is_known_feature_key, Features, FeaturesToml};
+use crate::features::{Features, FeaturesToml, is_known_feature_key};
 use crate::hooks::HooksConfig;
 
 pub const DEFAULT_MAX_SUBAGENTS: usize = 20;
@@ -5736,8 +5736,6 @@ fn provider_config_table_name(provider: ApiProvider) -> Result<String> {
 }
 
 fn provider_env_api_key(provider: ApiProvider) -> Option<String> {
-    // Keep the Hugging Face env precedence explicit here so runtime behavior
-    // stays aligned with the shipped provider-registry contract.
     if matches!(provider, ApiProvider::Huggingface) {
         return std::env::var("HUGGINGFACE_API_KEY")
             .ok()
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @gaord — this is the right fix for the provider-registry drift. The same Hugging Face env precedence change is now carried locally on the v0.8.64 integration branch as `7772391fa`, with contributor commit metadata preserved from the PR branch.

I rechecked the local integration branch with:

- `python3 scripts/check-provider-registry.py`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked huggingface_`

Both pass locally. I am leaving this PR open until the integration branch is pushed/landed and CI verifies the carried commit on the release path, but thank you for the focused fix and clean validation notes.

Hmbown:
Thanks @gaord. The Hugging Face env precedence behavior is now present on main, so I am closing this old conflicted release PR rather than keeping the duplicate branch open.

Landed references:
- e32c245660: restores the literal HUGGINGFACE_API_KEY then HF_TOKEN order in crates/tui/src/config.rs
- a31ef88dfd6a40c1c90daae9506d8159bbc80f23: keeps the active-provider auth probe aligned with that precedence

Your focused repro and validation on the provider-registry drift are credited in this PR thread.

PR #3332, opened 2026-06-19
fix(app-server): require auth for non-loopback binds

### LINKED ISSUE #3258: app-server: fail fast for non-loopback legacy bind without explicit auth
## Follow-up from v0.8.61 app-server release verification

Bare `codewhale app-server --host 0.0.0.0` without `--auth-token` and without `--insecure-no-auth` does not fail fast before attempting a non-loopback bind. The legacy HTTP server generates a one-time `cwapp_*` bearer token and proceeds toward binding.

Evidence:

- `crates/app-server/src/lib.rs::resolve_auth_token` only checks `!listen.ip().is_loopback()` inside the `options.insecure_no_auth` branch.
- The default/no-token branch then generates `cwapp_<uuid>` and prints bearer-token guidance.

This appears pre-existing legacy behavior rather than a v0.8.61 app-server delegation regression, so it was treated as a v0.8.62 hardening follow-up.

Expected direction:

- Decide whether legacy bare app-server should reject all non-loopback binds unless an explicit stable token is supplied.
- Add a regression test for `host=0.0.0.0`, no explicit token, no insecure flag.

cyq1017:
Fixes #3258

## Summary
- reject non-loopback app-server binds when no explicit auth token is supplied
- keep loopback one-time token generation and explicit-token non-loopback binds unchanged
- add regression coverage for the default non-loopback/no-token path

## Verification
- cargo fmt --check --all
- cargo test -p codewhale-app-server non_loopback -- --nocapture
- cargo test -p codewhale-app-server --lib --locked -- --nocapture
- cargo test --workspace --all-features --locked non_loopback -- --nocapture
- cargo clippy --workspace --all-features --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::assertions_on_constants

### DIFF
--- a/crates/app-server/src/lib.rs
+++ b/crates/app-server/src/lib.rs
@@ -401,6 +401,10 @@ fn resolve_auth_token(options: &AppServerOptions) -> Result<Option<String>> {
         return Ok(None);
     }
 
+    if configured.is_none() && !options.listen.ip().is_loopback() {
+        bail!("refusing app-server bind on non-loopback address without explicit auth token");
+    }
+
     let token = configured
         .map(str::to_string)
         .unwrap_or_else(|| format!("cwapp_{}", Uuid::new_v4().simple()));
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @cyq1017 — this is exactly the right hardening direction for #3258. The same behavior has already landed on the release train via PR #3347 / commit `a915246ae` and #3258 is now closed: non-loopback app-server binds require an explicit stable auth token, loopback keeps generated token behavior, and `--insecure-no-auth` remains loopback-only.

I am leaving this PR unmerged rather than duplicating the change, but thank you for the focused fix and regression coverage.

Hmbown:
Thanks @cyq1017. Closing this duplicate now that the non-loopback app-server auth hardening is on main and the linked issue has already been resolved.

Credit trail:
- a915246ae: require explicit auth off loopback
- PR #3347 carried the release hardening path that closed #3258

Appreciate the narrow security fix and regression coverage.

Issue #3310, opened 2026-06-18
v0.8.63: Split mcp.rs transports into focused modules


Hmbown:
## Problem

`crates/tui/src/mcp.rs` contains multiple concerns in one file:

- `StdioTransport`
- `SseTransport`
- `HttpTransport`
- `StreamableHttpTransport`
- `McpConnection`
- request/response helpers
- transport headers and protocol glue
- a very large test module

The code is easier to split than some other monoliths because the transport implementations already form natural boundaries.

## Proposed shape

Split into an `mcp/` module family:

- `mcp/mod.rs`
- `mcp/transport.rs`
- `mcp/stdio.rs`
- `mcp/sse.rs`
- `mcp/http.rs`
- `mcp/streamable.rs`
- `mcp/connection.rs`
- `mcp/headers.rs`
- `mcp/helpers.rs`
- `mcp/tests.rs` or adjacent focused test files

Exact names can differ if local module conventions suggest better names.

## Acceptance criteria

- [ ] Each transport implementation lives in its own module.
- [ ] Shared transport traits/types live in a small common module.
- [ ] Connection lifecycle code is separated from individual transport implementations.
- [ ] Existing MCP tests pass without behavior changes.
- [ ] Public/internal API exposure is minimized; no broad `pub` sweep.

## Risk

Low to medium. The main risk is accidental visibility churn or duplicated helper logic.

Hmbown:
Part of #3306.

Release-risk note: this is targeted at v0.8.63, but PRs should stay small and preserve behavior unless the issue explicitly calls out a behavior change.

Hmbown:
Release readiness triage for PR #3347: this remains useful work, but it is broader refactor/config-editability scope rather than a remaining v0.8.63 release blocker. The v0.8.63 train has taken the narrow reliability/security/config-control slices that are already implemented and tested; this issue is being retargeted to v0.8.64 so it can be handled in a smaller, reviewable follow-up without widening the current release train.

Hmbown:
## Agent handoff notes (2026-07-02)

The natural split is already visible: stdio, SSE, streamable HTTP, HTTP auth/header helpers, connection/pool, config loading/saving, and discovery snapshot.

Implementation tip: define the common transport trait/types first, then move one transport at a time. Avoid broad `pub` expansion just to make modules compile.

Verification: existing MCP tests plus one config load/save/discovery smoke. Watch project/workspace path safety and secret redaction helpers; those are security-sensitive.

[...]

PR #3333, opened 2026-06-19
refactor(tui): split MCP header helpers

cyq1017:
## Problem
`crates/tui/src/mcp.rs` still keeps HTTP header framing and custom-header filtering inline with transport code, which makes the MCP transport split in #3310 harder to review incrementally.

## Change
- Move MCP HTTP default/header filtering helpers into `mcp::headers`
- Keep helper visibility scoped to the parent MCP module
- Preserve existing header behavior and tests

## Verification
- `cargo test -p codewhale-tui mcp -- --nocapture`
- `cargo fmt --check --all`
- `git diff --check`
- `cargo clippy -p codewhale-tui --all-targets --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::assertions_on_constants`

Part of #3310.

### DIFF
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -14,33 +14,23 @@ use std::time::Duration;
 
 use anyhow::{Context, Result};
 use reqwest::StatusCode;
-use reqwest::header::{ACCEPT, CONTENT_TYPE};
+use reqwest::header::CONTENT_TYPE;
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt};
 use tokio::process::{Child, ChildStdin, ChildStdout};
 use tokio::sync::Mutex as TokioMutex;
 
+mod headers;
+
 use crate::child_env;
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks @cyq1017 — this was a clean, useful slice for #3310. I tried a direct local merge first, but the v0.8.64 integration branch already has the MCP tests extracted into `crates/tui/src/mcp/tests.rs`, so I harvested the header-module split onto that current layout instead.

Local verification on `codex/v0.8.64-integration`:
- `cargo test -p codewhale-tui --locked mcp`
- `python3 scripts/check-provider-registry.py`
- `python3 scripts/check-coauthor-trailers.py --author-map .github/AUTHOR_MAP --range origin/main..HEAD --check-authors`

I preserved you as the commit author and kept `Harvested from PR #3333 by @cyq1017` in the commit body. I am leaving this PR open until the integration branch is pushed/landed so the visible GitHub state stays honest.

PR #3356, opened 2026-06-21
fix(tui): allow worktree git metadata writes in sandbox

### LINKED ISSUE #3355: Sandbox blocks Git write ops on worktree workspaces — allow worktree-linked paths without trust_mode
## Background

I use Git worktrees to organize my projects. My workspace looks like:

```
~/Projects/main-repo/              ← main repo
~/Projects/main-repo-wt/feat-x/    ← worktree workspace (created via git worktree add)
```

A worktree `.git` is just a pointer file:

```
gitdir: /Users/xxx/Projects/main-repo/.git/worktrees/feat-x/
```

The actual Git metadata (index, refs, objects, HEAD) lives inside the main repo, NOT inside the worktree workspace directory.

## The problem

When I open this worktree directory in codewhale Agent mode, **all Git write operations fail**:

```
fatal: Unable to create .../index.lock: Operation not permitted
```

The reason is straightforward: codewhale sandbox only allows writes within the workspace directory, but the worktree Git metadata is at an external path (`/Users/xxx/Projects/main-repo/.git/`), so everything gets blocked.

Reads work fine — `git status`, `git diff`, `git log` all work through the structured tools, which means codewhale CAN read the external `.git` paths. Only writes are blocked.

## Current workarounds

Either switch to **YOLO mode** (which implies `trust_mode=true`, granting blanket write access to all external paths), or quit codewhale and run git commands manually in a separate terminal. Neither is great. YOLO mode opens up write access everywhere, which feels unsafe. Switching to a terminal breaks the codewhale workflow.

## What I hope for

Let codewhale handle Git write operations in worktree workspaces **without enabling trust_mode**.

Each worktree has two natural path pointers that codewhale can already read:

1. **Worktree metadata dir**: the `.git` file contains `gitdir: .../main-repo/.git/worktrees/<name>/`
2. **Shared Git dir**: the `commondir` file inside the metadata dir (content is `../..` or similar) resolves to the main repo `.git/` directory

Just whitelist these two paths for writes. **No other external paths need to be touched.** The risk is minimal — these are just the same repository Git metadata directories. This is much safer than the blanket `trust_mode=true`.

## Environment

- macOS Sequoia
- codewhale v0.8.58
- Git worktree, Agent mode

## To reproduce

```bash
git worktree add ~/worktree-dir feature-branch
# Open ~/worktree-dir in codewhale Agent mode
git add -A
# fatal: Unable to create '.../index.lock': Operation not permitted
```

Would love to see codewhale handle worktree-linked external paths so Git operations work smoothly in Agent mode.

cyq1017:
Fixes #3355.

Problem:
Git worktrees store mutable Git metadata outside the worktree root, so the macOS seatbelt workspace-write sandbox blocks commands like `git add` unless the user switches to YOLO/trust mode.

Change:
When workspace-write resolves writable roots, detect linked-worktree `.git` pointer files, require a `commondir`, and only add the derived worktree gitdir plus shared `.git` directory when the gitdir is under the shared `.git/worktrees/` directory. Plan mode remains read-only and YOLO remains full access.

Verification:
- `cargo test -p codewhale-tui workspace_write_includes_git_worktree_metadata_roots`
- `cargo test -p codewhale-tui sandbox::policy`
- `cargo test -p codewhale-tui sandbox_policy_for_mode_returns_correct_policy_per_mode`
- `git diff --check`

### DIFF
--- a/crates/tui/src/sandbox/policy.rs
+++ b/crates/tui/src/sandbox/policy.rs
@@ -7,6 +7,7 @@
 //! tightly controlled workspace-only write access.
 
 use serde::{Deserialize, Serialize};
+use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 
@@ -171,6 +172,14 @@ impl SandboxPolicy {
                     roots.push(cwd.to_path_buf());
                 }
 
+                // Git worktrees keep mutable metadata outside the worktree
+                // directory. Allow only the gitdir and commondir derived from
+                // a workspace `.git` pointer, preserving the workspace boundary
+                // for all other external paths.
+                for root in roots.clone() {
+                    roots.extend(resolve_git_worktree_writable_roots(&root));
+                }
+
                 // Add /tmp unless excluded
                 if !exclude_slash_tmp && let Ok(tmp) = Path::new("/tmp").canonicalize() {
                     roots.push(tmp);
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @cyq1017 — this worktree sandbox fix is now carried locally in the v0.8.64 integration branch. I preserved your authorship with the canonical noreply identity from `.github/AUTHOR_MAP`, and the commit body credits PR #3356 plus @linletian for the worktree repro and diagnostics.

Local verification included the focused sandbox policy tests, and the release credit gate now passes over the integration branch. I am leaving this PR open until the integration branch is pushed/landed so the public repo state stays honest.

Hmbown:
Follow-up status for v0.8.64 integration:

The linked-worktree sandbox fix is still carried locally, and I also added the review follow-up for commands launched from subdirectories inside a linked worktree. The local patch now resolves the worktree `.git` pointer from ancestor directories while keeping the existing derived-metadata-root guard.

Verified locally with:

- `cargo test -p codewhale-tui --bin codewhale-tui --locked sandbox::policy`
- `git diff --check`

Thanks again @cyq1017 — this remains in the v0.8.64 train, and I am keeping the PR open until the integration branch is pushed/landed.

Issue #3328, opened 2026-06-19
0.8.62 doesn't show sidebar


dxfq:
Hello,  

I just upgraded to 0.8.62 using codewhale.exe and codewhale-tui.exe files, and found that there is no more sidebar. 

/sidebar shows Note Sidebar is visible.

How to get the sidebar back on the terminal? 

Appreciate any help, thanks!

Hmbown:
sorry about this - should be fixed now! thank you reporting this :) @dxfq

PR #3371, opened 2026-06-22
fix(ui): reduce minimum terminal width for sidebar visibility

donglovejava:
## Summary

This PR fixes the sidebar display issue reported in #3328 where the sidebar would not appear in typical terminal configurations.

## Problem

The sidebar was only showing when terminal width >= 100 columns (SIDEBAR_VISIBLE_MIN_WIDTH = 100), which is too restrictive for many common terminal setups. Most developers use terminals with widths between 80-100 columns, where the sidebar would remain hidden.

## Solution

Reduced SIDEBAR_VISIBLE_MIN_WIDTH from 100 to 60 in crates/tui/src/tui/ui.rs, making the sidebar visible in more common terminal configurations.

## Testing

- Manual testing with various terminal widths
- Verified sidebar displays correctly at 60+ column widths
- Ensured existing functionality is not affected

## Impact

Users with narrower terminal windows will now see the sidebar properly, improving the overall user experience without affecting users with wider terminals.

🤖 Generated with [Claude Code](https://claude.ai/code)


Hmbown:
Thanks @donglovejava — I carried the sidebar visibility improvement into the v0.8.64 integration branch as a scoped harvest:

- 21f30b1a3 `fix(tui): lower sidebar visibility threshold`

I kept the release branch to the sidebar behavior only, adjusted the lower bound to 64 columns after layout review, and left the unrelated tool-catalog/local artifact files out of the harvest. Verified with:

- `cargo fmt --all`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked sidebar_config_command`
- `git diff --check HEAD~1..HEAD`
- `python3 scripts/check-coauthor-trailers.py --author-map .github/AUTHOR_MAP --range HEAD~1..HEAD --check-authors`

Appreciate the report and the fix direction.

Hmbown:
Follow-up update: after carrying the safe sidebar threshold slice into #3373, the macOS CI exposed one stale test expectation that still used the old narrow-width number. I pushed a test-only follow-up on the integration branch as `c7cc1ecef` to assert the new 63/64-column boundary explicitly.\n\nThanks again @donglovejava — the product change is still carried, and this was just the release-branch test alignment.

PR #3381, opened 2026-06-22
Feat/memory tags

pkeging:
## Testing

## Checklist

### DIFF
--- a/crates/tui/src/commands/groups/memory/memory.rs
+++ b/crates/tui/src/commands/groups/memory/memory.rs
@@ -23,7 +23,7 @@ use std::path::Path;
 use super::CommandResult;
 use crate::tui::app::App;
 
-const MEMORY_USAGE: &str = "/memory [show|path|clear|edit|help]";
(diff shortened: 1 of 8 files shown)


Hmbown:
Hmm can you resubmit this with more information?

Hmbown:
Oh I see now - let me take a look at this when I'm at home - been messing around with the configuration menu so this might be very timely. Thank you!!

Hmbown:
Thanks @pkeging — the taggable-memory idea here is a good one and directly informs how we are shaping recall.

We have decided CodeWhale's memory architecture going forward is to adopt **Moraine** (https://github.com/eric-tramel/moraine, Apache-2.0) as a lossless session-trace + MCP recall backend, rather than an in-repo injected-memory store. That keeps memory local-first and pull/recall-based (no prompt injection). It is tracked for **v0.8.66** in #3495, which carries the full design and explicitly folds in the tag/namespace concepts from this PR.

Closing this in favor of that direction. Really appreciate the work — please chime in on #3495 if you'd like to stay involved.

PR #3433, opened 2026-06-23
Harden local state storage boundaries

Hmbown:
## Goal

Tighten local state/storage boundaries for the v0.8.65 release lane without changing public workflow behavior.

## Changes

- Reject path-shaped task IDs before writing task artifacts.
- Resolve `pr_attempt_record` task IDs through `TaskManager` before writing the diff artifact, so caller-supplied IDs must identify an existing durable task and the artifact path uses the canonical task ID.
- Reject path-shaped automation/run IDs before reading, writing, listing, or deleting automation storage.
- Enable SQLite foreign-key enforcement for every state-store connection so thread deletion cascades to child rows.
- Create file-backed secret files with private Unix mode at first write.
- Treat explicit `CODEWHALE_HOME` as an isolation boundary for generic state/config fallback and migration.

## Verification

- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked artifact_writer_rejects_path_component_task_ids`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked task_artifact_helper_requires_existing_manager_task_before_write`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked rejects_storage_ids_with_path_components`
- `cargo test -p codewhale-state --locked`
- `cargo test -p codewhale-secrets --locked file_store`
- `cargo test -p codewhale-config --locked explicit_codewhale_home_does_not_read_or_migrate_legacy`
- `cargo test -p codewhale-config --locked migrate_config_reports_copied_legacy_path`

## Risks

- Explicit `CODEWHALE_HOME` now bypasses legacy `~/.deepseek` config/state fallback entirely. That matches the isolation contract, while default `$HOME` installs still preserve legacy migration/backfill behavior.
- Automation/task IDs are now required to be single path components. Existing generated IDs already satisfy this.

## Issue

Part of the v0.8.65 local-state hardening work.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -3239,14 +3239,22 @@ pub const LEGACY_APP_DIR: &str = ".deepseek";
 /// `$CODEWHALE_HOME` takes precedence when set. Otherwise defaults to
 /// `$HOME/.codewhale`. This is the write target for new product state.
 pub fn codewhale_home() -> Result<PathBuf> {
-    if let Ok(val) = std::env::var("CODEWHALE_HOME") {
-        let trimmed = val.trim();
-        if !trimmed.is_empty() {
-            return Ok(PathBuf::from(trimmed));
-        }
+    if let Some(path) = explicit_codewhale_home() {
+        return Ok(path);
     }
     let home = effective_home_dir().context("failed to resolve home directory")?;
     Ok(home.join(CODEWHALE_APP_DIR))
+}
+
+fn explicit_codewhale_home() -> Option<PathBuf> {
+    std::env::var("CODEWHALE_HOME").ok().and_then(|val| {
+        let trimmed = val.trim();
+        if trimmed.is_empty() {
+            None
+        } else {
+            Some(PathBuf::from(trimmed))
+        }
+    })
 }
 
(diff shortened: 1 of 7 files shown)


Hmbown:
Closing as **superseded by merged work** rather than left open.

The local-state storage hardening this PR implements has already landed on `main` via **#3450** ("Harden local storage path boundaries", merge commit `8d74738257afa2c07fd90f0cccb88d1d07289ad3`, merged 2026-06-23), which touches the same six files and implements equivalent behavior:

- `foreign_keys` pragma + cascading-delete coverage (`crates/state/src/lib.rs`)
- single-path-component validation in `automation_manager.rs` / `task_manager.rs`
- `0o600` secret-file permissions (`crates/secrets/src/lib.rs`)
- `$CODEWHALE_HOME` legacy isolation (also reinforced by `dcfae6e21`)

A rebase of this branch onto current `main` resolves almost entirely to "main already has this." The one piece **not** covered by #3450 is the `tools/tasks.rs` change that makes `write_task_artifact_for` async and resolves `task_id` via `TaskManager::get_task()` (`pr_attempt_record`). That is a separate concern from path hardening — if it's still wanted, a focused follow-up PR against current `main` would be the cleanest path and easy to review on its own.

Thanks for the work here — the hardening it targeted is in `main`.

PR #3438, opened 2026-06-23
Harden local state storage boundaries

Hmbown:
## Summary
- confines task artifact and automation/run IDs before filesystem path construction
- enables SQLite foreign keys on every state-store connection so thread deletes cascade as declared
- creates file-backed secrets with private Unix mode before writing secret bytes
- keeps explicit `CODEWHALE_HOME` isolated from ambient legacy state/config while preserving default-home legacy backfill and migration

## Verification
- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked write_task_artifact_rejects_path_like_task_ids`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked automation_`
- `cargo test -p codewhale-state --locked`
- `cargo test -p codewhale-secrets --locked`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked` (5117 passed, 2 ignored)

## Risk
- Existing `todo`/task/automation behavior is preserved for normal UUID IDs. Path-like IDs now fail before storage paths are built.
- Default `$HOME/.codewhale` installs still read/migrate legacy `.deepseek` state; only explicit `CODEWHALE_HOME` is treated as an isolation boundary.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -3249,6 +3249,17 @@ pub fn codewhale_home() -> Result<PathBuf> {
     Ok(home.join(CODEWHALE_APP_DIR))
 }
 
+fn explicit_codewhale_home() -> Option<PathBuf> {
+    std::env::var("CODEWHALE_HOME").ok().and_then(|val| {
+        let trimmed = val.trim();
+        if trimmed.is_empty() {
+            None
+        } else {
+            Some(PathBuf::from(trimmed))
+        }
+    })
+}
+
 /// Resolve the legacy DeepSeek home directory (`$HOME/.deepseek`).
 ///
 /// Always returns the legacy path regardless of whether it exists.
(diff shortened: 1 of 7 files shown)


Hmbown:
Closing this duplicate branch in favor of #3433.

#3433 already carries the same local-state hardening scope and now also includes the stricter `pr_attempt_record` ordering fix from this branch: task IDs are resolved through `TaskManager` before the diff artifact is written. Keeping one PR avoids splitting one hardening concern across two branches.

Relevant verification preserved on #3433:
- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked artifact_writer_rejects_path_component_task_ids`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked task_artifact_helper_requires_existing_manager_task_before_write`
- package-level state/secrets/config checks from the PR body

PR #3443, opened 2026-06-23
Harden automation storage ids

Hmbown:
## Goal

Harden durable automation storage path construction for the v0.8.65 lane.

## Changes

- Validate automation and automation-run ids before composing JSON/run storage paths.
- Route automation read, write, update/delete, run listing, and run writes through the same storage-id guard.
- Add regression coverage for path-like ids on automation lookup, automation save, run listing, and run save.

## Verification

- cargo fmt --all -- --check
- cargo test -p codewhale-tui --bin codewhale-tui --locked automation_manager
- git diff --check

## Risks

Low. CodeWhale-generated automation and run ids are UUID strings, which remain accepted. The new guard rejects only ids that are not safe single path segments.

### DIFF
--- a/crates/tui/src/automation_manager.rs
+++ b/crates/tui/src/automation_manager.rs
@@ -319,17 +319,21 @@ impl AutomationManager {
         Self::open(default_automations_dir())
     }
 
-    fn automation_path(&self, id: &str) -> PathBuf {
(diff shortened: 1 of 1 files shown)


Hmbown:
Closing this narrower branch because the same automation storage-id hardening is already covered in the broader v0.8.65 local-state hardening PR #3433.

PR #3444, opened 2026-06-23
Enable state store foreign key cascades

Hmbown:
## Goal

Make thread deletion honor the state schema existing cascading child-row cleanup.

## Changes

- Enable SQLite foreign key enforcement on every StateStore connection.
- Add a regression test that persists child rows for messages, dynamic tools, checkpoints, and thread goals, deletes the parent thread, and verifies the child rows are gone.

## Verification

- cargo fmt --all -- --check
- cargo test -p codewhale-state --locked delete_thread_cascades_child_rows
- cargo test -p codewhale-state --locked thread_goal_requires_existing_thread
- cargo test -p codewhale-state --locked
- git diff --check

## Risks

Low. The schema already declares the cascade relationships; this makes SQLite enforce them consistently for each connection.

### DIFF
--- a/crates/state/src/lib.rs
+++ b/crates/state/src/lib.rs
@@ -305,8 +305,11 @@ impl StateStore {
     }
 
     fn conn(&self) -> Result<Connection> {
-        Connection::open(&self.db_path)
-            .with_context(|| format!("failed to open state db {}", self.db_path.display()))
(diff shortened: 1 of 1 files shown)


Hmbown:
Closing this narrower branch because the same state-store foreign-key enforcement is already covered in the broader v0.8.65 local-state hardening PR #3433.

PR #3455, opened 2026-06-23
fix: prevent ordinary prompt text from being interpreted as a mode switch

### LINKED ISSUE #3387: Prompt text can be misinterpreted as a mode switch
## Problem

Sometimes when a user submits an ordinary prompt, CodeWhale appears to take part of the prompt text and interpret it as a mode-switch instruction. The result is that the session enters another mode even though the user did not explicitly request a mode change.

This should not happen for natural-language task text. Mode changes should only come from explicit mode-changing inputs such as the Tab cycle, `/mode ...`, a command-palette action, startup/config defaults, or another deliberately typed command surface.

## Why this matters

* The user prompt may be partially consumed, truncated, or reinterpreted instead of being sent as the task.
* The agent can end up operating under the wrong behavioral posture.
* If the mistaken switch affects Plan/Agent/YOLO, it can change approval, sandbox, shell, or trust behavior.
* The bug is intermittent and easy to miss because it can look like the model decided to change mode, when the harness should be enforcing the boundary.

## Observed behavior

Reported on 2026-06-22: while giving a normal prompt, CodeWhale sometimes takes a fragment of that prompt and enters another mode that is not appropriate or correct.

We do not yet have a minimal repro string, but likely trigger candidates include prompt text containing words or phrases such as `mode`, `plan`, `agent`, `yolo`, `enter ... mode`, or numbered aliases that overlap with `/mode` parsing.

## Expected behavior

Natural-language prompt content must be treated as prompt content unless it is submitted through an explicit command/mode-switch path.

In particular:

* `plan`, `agent`, `yolo`, and numeric aliases should only switch mode when parsed as explicit `/mode` arguments or through the official mode UI path.
* Free-form prompt text should never be scanned broadly for mode names and converted into a mode transition.
* If a prompt begins with text that resembles a command but is not an exact command invocation, it should remain user text.
* The full submitted prompt should reach the model unchanged except for existing intentional composer normalization.

## Areas to inspect

* TUI composer submit path and slash-command detection.
* `/mode` command parsing and aliases.
* Any command-palette, shortcut, or natural-language intent handling that can call `set_mode`.
* Turn setup and mode-policy resolution around `AppMode`, especially alongside Hmbown/CodeWhale#3386.
* Tests that simulate prompt submission text which contains mode-like words but is not an explicit command.

## Acceptance criteria

* Add regression coverage for prompt strings containing mode-like words that must not switch modes.
* Add coverage for exact command strings that should still switch modes, e.g. `/mode plan`, `/mode agent`, `/mode yolo`, and supported numeric aliases.
* Confirm that prompt submission and command dispatch have a clear boundary: only explicit command syntax reaches mode-command parsing.
* Log or trace the source of mode transitions in a way that distinguishes user UI action, slash command, config/default startup, and runtime/internal requests.
* Manual repro check: submit several natural-language prompts containing `mode`, `plan`, `agent`, `yolo`, and `enter mode`; CodeWhale should remain in the current mode unless the prompt is an explicit mode command.

## Related

* Hmbown/CodeWhale#3386 covers the broader Plan/Agent/YOLO policy split. This issue is specifically about preventing ordinary prompt text from being interpreted as a mode-switch command.

mvanhorn:
## Summary

Mode changes were able to leak from ordinary prompt text. On the runtime turn-start path, `parse_mode` resolved a mode string with a broad catch-all (`_ => AppMode::Agent`), so any unrecognized value silently became a valid mode instead of being rejected. Because `start_turn` prefers `StartTurnRequest.mode` over the thread's persisted mode, a stray prompt-like override (e.g. `enter yolo mode`) that should be ignored was coerced into `Agent`, silently changing a Plan/YOLO thread's mode without an explicit request (#3387).

This draws a clear boundary between prompt-submission text and mode-command parsing:

- Adds a strict `parse_mode_opt(mode) -> Option<AppMode>` that resolves **only** the explicit mode tokens, matching the `/mode` command's own accepted set: `agent`/`plan`/`yolo` plus the numeric aliases `1`/`2`/`3`. Anything else returns `None`.
- `parse_mode` stays an infallible wrapper (`parse_mode_opt(..).unwrap_or(AppMode::Agent)`), so the persisted/default mode path is unchanged (`AppMode::as_setting()` always yields a recognized token).
- At the `start_turn` call site, an unrecognized per-turn override now falls back to the thread's persisted mode rather than coercing to `Agent`, so an invalid override never crosses the mode boundary.

Mode transitions continue to originate only from explicit sources: the Tab cycle, `/mode`, the mode picker, and config/startup defaults. The composer/submit path (`looks_like_slash_command_input`, which requires a leading `/`) was already strict and is left untouched.

## Why this matters

Root cause: the lone runtime mode resolver treated every non-token string as a valid mode, and the turn-start call preferred the request override unconditionally, so free-form prompt text could silently change approval/sandbox/trust posture.

The failing case now fixed: starting a turn on a Plan or YOLO thread with a request mode such as `enter yolo mode` (or any non-token fragment) no longer flips the thread to `Agent`; the unrecognized override is ignored and the persisted mode is preserved. Regression tests assert that prompt fragments (`plan a trip to Tokyo`, `switch the agent on`, `enter yolo mode`, `agent of chaos`, `mode`) never resolve to a mode, while exact tokens and numeric aliases still do.

## Testing

- [x] `cargo fmt` clean on the changed crate
- [x] `cargo test -p codewhale-tui parse_mode` — all `parse_mode` / `parse_mode_opt` regression tests pass

(Note: the full `cargo test --workspace` build currently fails before reaching these tests due to a pre-existing non-exhaustive match in `crates/tui/src/main.rs` `doctor_api_key_source_label` — `ApiKeySource::Command` / `ApiKeySource::Secret` are uncovered on `main`, unrelated to this change. The `parse_mode` tests were verified passing under a local workaround for that build error.)

Fixes #3387

## Checklist

- [x] No UI changes (runtime mode-resolver logic only); no manual TUI verification needed

### DIFF
--- a/crates/tui/src/runtime_threads.rs
+++ b/crates/tui/src/runtime_threads.rs
@@ -2026,7 +2026,15 @@ impl RuntimeThreadManager {
             touch_lru(&mut active.lru, thread_id);
         }
 
-        let mode = parse_mode(req.mode.as_deref().unwrap_or(&thread.mode));
+        // A requested mode override only takes effect when it is an explicit,
+        // recognized mode token. An unrecognized override (e.g. a stray prompt
+        // fragment) must NOT silently change the mode: fall back to the
+        // thread's persisted mode rather than coercing to Agent (#3387).
+        let mode = req
+            .mode
+            .as_deref()
+            .and_then(parse_mode_opt)
+            .unwrap_or_else(|| parse_mode(&thread.mode));
         let requested_model = req.model.unwrap_or_else(|| thread.model.clone());
         let auto_model = requested_model.trim().eq_ignore_ascii_case("auto");
         let (provider, model, reasoning_effort) = if auto_model {
--- a/crates/tui/src/runtime_threads.rs
+++ b/crates/tui/src/runtime_threads.rs
@@ -3740,12 +3740,22 @@ fn enforce_lru_capacity(
     evicted
 }
 
-fn parse_mode(mode: &str) -> AppMode {
+/// Resolves only explicit mode tokens to an app mode. Free-form prompt text is
+/// never a valid mode token: `parse_mode_opt` returns `None` unless the input is
+/// exactly `agent`/`plan`/`yolo` or the numeric aliases `1`/`2`/`3`. Mode
+/// changes originate from the Tab cycle, `/mode`, the mode picker, or
+/// config/startup defaults, not from submitted natural-language prompt text.
+fn parse_mode_opt(mode: &str) -> Option<AppMode> {
     match mode.trim().to_ascii_lowercase().as_str() {
-        "plan" => AppMode::Plan,
-        "yolo" => AppMode::Yolo,
-        _ => AppMode::Agent,
+        "agent" | "1" => Some(AppMode::Agent),
+        "plan" | "2" => Some(AppMode::Plan),
+        "yolo" | "3" => Some(AppMode::Yolo),
+        _ => None,
     }
+}
+
+fn parse_mode(mode: &str) -> AppMode {
+    parse_mode_opt(mode).unwrap_or(AppMode::Agent)
 }
 
 fn tool_kind_for_name(name: &str) -> TurnItemKind {
--- a/crates/tui/src/runtime_threads/tests.rs
+++ b/crates/tui/src/runtime_threads/tests.rs
@@ -2466,6 +2466,38 @@ fn parse_mode_defaults_to_agent() {
     assert_eq!(parse_mode("plan"), AppMode::Plan);
 }
 
+#[test]
+fn parse_mode_opt_resolves_explicit_tokens_and_aliases() {
+    assert_eq!(parse_mode_opt("agent"), Some(AppMode::Agent));
+    assert_eq!(parse_mode_opt("1"), Some(AppMode::Agent));
+    assert_eq!(parse_mode_opt("plan"), Some(AppMode::Plan));
+    assert_eq!(parse_mode_opt("2"), Some(AppMode::Plan));
+    assert_eq!(parse_mode_opt("yolo"), Some(AppMode::Yolo));
+    assert_eq!(parse_mode_opt("3"), Some(AppMode::Yolo));
+    assert_eq!(parse_mode_opt(" PLAN "), Some(AppMode::Plan));
+}
+
+#[test]
+fn parse_mode_opt_rejects_prompt_fragments() {
+    for input in [
+        "plan a trip to Tokyo",
+        "switch the agent on",
+        "enter yolo mode",
+        "agent of chaos",
+        "mode",
+    ] {
+        assert_eq!(parse_mode_opt(input), None);
+    }
+}
+
+#[test]
+fn parse_mode_wrapper_defaults_and_resolves_numeric_aliases() {
+    assert_eq!(parse_mode("plan a trip to Tokyo"), AppMode::Agent);
+    assert_eq!(parse_mode("1"), AppMode::Agent);
+    assert_eq!(parse_mode("2"), AppMode::Plan);
+    assert_eq!(parse_mode("3"), AppMode::Yolo);
+}
+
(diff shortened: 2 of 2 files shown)


Hmbown:
> ## Summary
> 
> Mode changes were able to leak from ordinary prompt text. On the runtime turn-start path, `parse_mode` resolved a mode string with a broad catch-all (`_ => AppMode::Agent`), so any unrecognized value silently became a valid mode instead of being rejected. Because `start_turn` prefers `StartTurnRequest.mode` over the thread's persisted mode, a stray prompt-like override (e.g. `enter yolo mode`) that should be ignored was coerced into `Agent`, silently changing a Plan/YOLO thread's mode without an explicit request (#3387).
> 
> This draws a clear boundary between prompt-submission text and mode-command parsing:
> 
> - Adds a strict `parse_mode_opt(mode) -> Option<AppMode>` that resolves **only** the explicit mode tokens, matching the `/mode` command's own accepted set: `agent`/`plan`/`yolo` plus the numeric aliases `1`/`2`/`3`. Anything else returns `None`.
> - `parse_mode` stays an infallible wrapper (`parse_mode_opt(..).unwrap_or(AppMode::Agent)`), so the persisted/default mode path is unchanged (`AppMode::as_setting()` always yields a recognized token).
> - At the `start_turn` call site, an unrecognized per-turn override now falls back to the thread's persisted mode rather than coercing to `Agent`, so an invalid override never crosses the mode boundary.
> 
> Mode transitions continue to originate only from explicit sources: the Tab cycle, `/mode`, the mode picker, and config/startup defaults. The composer/submit path (`looks_like_slash_command_input`, which requires a leading `/`) was already strict and is left untouched.
> 
> ## Why this matters
> 
> Root cause: the lone runtime mode resolver treated every non-token string as a valid mode, and the turn-start call preferred the request override unconditionally, so free-form prompt text could silently change approval/sandbox/trust posture.
> 
> The failing case now fixed: starting a turn on a Plan or YOLO thread with a request mode such as `enter yolo mode` (or any non-token fragment) no longer flips the thread to `Agent`; the unrecognized override is ignored and the persisted mode is preserved. Regression tests assert that prompt fragments (`plan a trip to Tokyo`, `switch the agent on`, `enter yolo mode`, `agent of chaos`, `mode`) never resolve to a mode, while exact tokens and numeric aliases still do.
> 
> ## Testing
> 
> - [x] `cargo fmt` clean on the changed crate
> - [x] `cargo test -p codewhale-tui parse_mode` — all `parse_mode` / `parse_mode_opt` regression tests pass
> 
> (Note: the full `cargo test --workspace` build currently fails before reaching these tests due to a pre-existing non-exhaustive match in `crates/tui/src/main.rs` `doctor_api_key_source_label` — `ApiKeySource::Command` / `ApiKeySource::Secret` are uncovered on `main`, unrelated to this change. The `parse_mode` tests were verified passing under a local workaround for that build error.)
> 
> Fixes #3387
> 
> ## Checklist
> 
> - [x] Updated docs or comments as needed
> - [x] Added or updated tests where relevant
> - [x] No UI changes (runtime mode-resolver logic only); no manual TUI verification needed
> - [x] Harvested/co-authored credit uses a GitHub numeric noreply address
> 

This is awesome - thanks so much Matt!! Been frustrated with this behavior myself and have other attempts at solving this so will fold this in if merging directly doesn't make sense

Hmbown:
Thanks @mvanhorn — the fix is clean and correct; CI here was red only because the branch base predates a fix already on `main` (the diff itself applies cleanly). To get it landed against the live tree I cherry-picked both your commits (authorship preserved, plus a `Co-authored-by` trailer) onto current `main` in **#3491**, which is green-pending CI and closes #3387.

Closing this in favor of #3491 so the credit-preserving, rebased version can merge. Really appreciate the catch — it directly matches the guardrail that mode/route changes must come from explicit user choice, never from interpreting prompt text.

PR #3460, opened 2026-06-23
docs(config): document saved ask rules from approvals

greyfreedom:
## Summary

Document the current persistence behavior for ask rules created from approval cards.

## Scope

- Replace the stale statement that approval UI persistence is unsupported.
- Document that `S` is available only in `exec_shell` approval cards.
- Clarify that `S` approves once and persists an ask rule containing the approved command.
- Document that saved command rules use the existing arity-aware prefix matching.
- Clarify that file-path ask rules can be authored in `permissions.toml` and matched at runtime, but cannot yet be saved through the approval UI.

## Not in this slice

- Typed allow/deny rules.
- Glob expansion.
- File-rule creation from the approval UI.
- Changes to runtime approval or persistence behavior.

## Builds on

- The landed ask-only `permissions.toml` schema and runtime wiring.
- The landed approval-card persistence for `exec_shell` ask rules.

## Issues

Refs #1186 (partial)  
Refs #2242 (partial)

## Validation

- `git diff --check`

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1011,8 +1011,13 @@ If you are upgrading from older releases:
   `[[rules]]` entries with `tool` plus optional `command` or `path` fields.
   Loaded rules feed the execution policy engine and force approval in approval
   modes that can ask; under `approval_policy = "never"`, matching ask rules are
-  rejected because no prompt can be shown. This intentionally does not accept
-  typed allow/deny records, glob expansion, or approval UI persistence yet.
+  rejected because no prompt can be shown. In an `exec_shell` approval card,
+  press `S` to approve the request once and save an ask rule containing that
+  command to this file. Only `exec_shell` cards support this shortcut; saved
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you!! Just a heads up - I am going to revert "yolo" mode back to having complete freedom - if there's a certain middle ground you're hoping to get between agent and yolo mode, I'll do my best to figure out how to configure that.

greyfreedom:
> Thank you!! Just a heads up - I am going to revert "yolo" mode back to having complete freedom - if there's a certain middle ground you're hoping to get between agent and yolo mode, I'll do my best to figure out how to configure that.

Thanks for the heads up. I agree that `yolo` should keep its existing “full freedom” semantics and should not be constrained by typed `ask` rules or approval prompts. I’ll treat the permission-rule work as applying to the safer approval modes only.
If there is a useful middle ground between `agent` and `yolo`, I think it should be introduced as an explicit configuration or mode rather than changing what `yolo` means. I’m happy to follow your lead on the naming and semantics if/when we slice that out.

PR #3497, opened 2026-06-23
feat(config): add Models.dev catalog route foundation

Hmbown:
## Summary
- add a network-free `codewhale_config::models_dev` parser for Models.dev combined catalog JSON
- preserve provider-agnostic model facts separately from provider-scoped wire offerings
- emit chat-capable `ProviderModelOffering` rows from Models.dev provider data
- let `RouteResolver` consume injected offering rows and prefer catalog defaults for `auto` / no-model resolution

## Why
Models.dev gives CodeWhale the provider/model/pricing/capability taxonomy we need without growing DeepSeek-shaped or provider-specific match tables. This keeps namespace-prefixed wire IDs scoped to the provider that serves them and avoids treating prefixes as global model ownership proof.

## Verification
- `cargo fmt`
- `cargo test -p codewhale-config --locked models_dev`
- `cargo test -p codewhale-config --locked route`

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod auth_source;
+pub mod models_dev;
 pub mod provider;
 pub mod route;
 
--- a/crates/config/src/models_dev.rs
+++ b/crates/config/src/models_dev.rs
@@ -1,0 +1,503 @@
+//! Models.dev catalog schema and helpers.
+//!
+//! Models.dev is the upstream taxonomy CodeWhale should use for model facts,
+//! provider offerings, pricing, limits, and capabilities. This module is
+//! intentionally network-free: callers provide JSON from a bundled snapshot,
+//! live refresh, or tests. Runtime fetch/cache policy belongs above this layer.
+//!
+//! The important boundary is the same one Models.dev uses:
+//! - `models` are provider-agnostic model facts.
+//! - `providers.*.models` are provider-scoped wire offerings.
+//!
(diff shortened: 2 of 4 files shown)


Hmbown:
Addressed the review feedback in 32c1a64:

- **Empty-`modalities` false-negative (the flagged bug):** `supports_text_chat` now treats an empty `input`/`output` list the same as absent metadata, so a row with `"modalities": {}` stays chat-capable (matching the `None` branch). Only an explicitly populated, non-text list excludes a row.
- **Dedup:** provider-id resolution extracted into `ModelsDevProvider::effective_id`, used by both `provider_offering` and `provider_offerings`.
- **Canonical-model arm:** documented why `selector_matches_other_provider_offering` keeps it (guards bare / hand-authored canonical ids that the `wire_model_id` and `namespace_hint` checks would miss).
- **Tests added:** `empty_modalities_struct_is_chat_capable`, `provider_offerings_keep_rows_with_empty_modalities_object`, and `resolver_auto_falls_back_to_descriptor_default_without_catalog_default`.

Local verification: `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test -p codewhale-config` → **219 passed**.

For the record: the earlier Windows red was `tools::verifier::tests::run_verifiers_background_starts_shell_jobs_and_returns_task_ids` — the known-flaky `codewhale-tui` test (passes in isolation; documented in `AGENTS.md` as flaky under full-suite parallelism). It is unrelated to this PR, which only touches `crates/config`. CI is re-running on the new commit.

---

PR #3498, opened 2026-06-23
feat(config): Models.dev-backed catalog snapshot + secret-free live cache (#3385)

Hmbown:
## Summary

Adds a **network-free** `codewhale_config::catalog` layer on top of the Models.dev parser merged in #3497. This is the #3385 data layer that feeds EPIC #2608 / #3383 — the reusable provider-catalog types + secret-free cache, kept out of any UI surface.

- **`CatalogOffering`** — routing identity (provider + wire id + optional canonical model + endpoint) **plus** offering-owned Models.dev facts (family, limits, cost, `reasoning`, `reasoning_options`) with explicit provenance (`Bundled` / `Live` / `UserOverride`). Projects to `route::ProviderModelOffering` via `to_offering()`, so catalog rows reach routing only through the existing offering seam.
- **`bundled_offerings_from_models_dev()`** — hydrate chat offerings from a parsed `ModelsDevCatalog`, preserving prefixed wire ids verbatim and joining a canonical model **only** from an explicit `base_model` (never inferred from a prefix). Audio/TTS rows are excluded from routing offerings.
- **`ProviderCatalogCache`** — secret-free cache keyed by **provider + base-URL fingerprint**, with TTL/staleness, failure-fallback (prior rows survive a failed refresh; `record_failure` never drops them), and explicit `Fresh` / `Stale` / `Failed{reason}` / `Unknown` status.
- **`CatalogCompiler`** — layered merge with precedence `bundled < live < user overrides`, deduplicated by `(provider, wire id)`.
- **`base_url_fingerprint()`** — normalized (host case / trailing slash / default port folded), dependency-free FNV-1a scope key.

## Invariants preserved (from #2608 / #3497)
- A catalog row is **not** an executable route — rows still compile through `RouteResolver` → `ReadyRouteCandidate`.
- `wire_model_id` stays separate from `canonical_model`; a prefix never proves canonical ownership.
- Unknown / custom / local rows are supported with explicit provenance and a `None` canonical model.

## Deliberately out of this slice (remaining #3385 work)
- The HTTP `/models` fetch layer (this module ingests `ProviderCatalogDelta`s; it performs no I/O).
- TUI surface wiring: `/models`, `/model` picker, slash completions, and the hard-coded `ModelInventory` router (`crates/tui`). Those consume this layer and are tracked as follow-ups so they don't collide with concurrent TUI work.

## Verification
- `cargo fmt --all -- --check`
- `cargo clippy -p codewhale-config --all-targets` (clean)
[...]

Issue #3085, opened 2026-06-11
v0.8.65: Provider/offering usage and PricingSku engine with provenance


Hmbown:
## Goal

Build provider/offering-scoped usage and pricing metadata for the many-provider CodeWhale world.

This must not remain DeepSeek-shaped. Some routes have token SKU pricing, some have subscription/quota usage percentages, some have account-tier limits, some are local/resource-only, and some are unknown. CodeWhale should show the right kind of usage/cost signal for the route instead of forcing everything into DeepSeek-style per-token pricing.

## Architecture Contract

Use separate but related concepts:

- `PricingSku`: provider/offering/route-owned token or unit pricing.
- `UsageMeter`: provider/offering/account-owned quota, subscription, or usage-limit telemetry.
- `UsageNormalization`: provider response usage fields mapped into canonical token/unit classes.
- `CostDisplay`: UI projection that chooses price, quota percentage, unknown, local resource, or mixed display based on route metadata.

Pricing belongs mostly to provider model offerings because the same canonical model can have different prices through direct providers, hosted aggregators, cloud routes, local runtimes, custom endpoints, or user overrides.

Usage meter examples:

- DeepSeek native: token pricing plus cache-hit/cache-miss usage.
- OpenRouter/aggregators: provider-specific pricing and optional account/credit usage when exposed.
- ChatGPT/Codex OAuth routes: usage percentage/quota/reset information when available, not fake per-token pricing.
- Local runtimes: pricing unknown/not applicable; optionally show local resource usage if available.
- Custom endpoints: unknown unless user provides pricing or usage override.

## Suggested Shapes

```rust
pub struct PricingSku {
    pub provider: ProviderId,
    pub model: ModelId,
    pub route_id: Option<RouteId>,
    pub currency: Currency,
    pub input_per_million: Option<Decimal>,
    pub output_per_million: Option<Decimal>,
    pub cache_read_per_million: Option<Decimal>,
    pub cache_write_per_million: Option<Decimal>,
    pub unit: PricingUnit,
    pub provenance: PricingProvenance,
    pub effective_at: Option<DateTime<Utc>>,
}

pub enum UsageMeter {
    TokenPricing(PricingSku),
    SubscriptionQuota {
        used_percent: Option<f64>,
        remaining_percent: Option<f64>,
        reset_at: Option<DateTime<Utc>>,
        provenance: UsageProvenance,
    },
    AccountCredits {
        currency: Option<Currency>,
[...]

PR #3501, opened 2026-06-23
feat(config): offering pricing projection with provenance (#3085)

Hmbown:
## Summary

Network-free `codewhale_config::pricing` — the config-layer slice of #3085. Maps the catalog offerings from #3498 into provider/offering-scoped pricing rows with explicit provenance, so pricing stops being DeepSeek-shaped and a route's price is owned by its **offering**, not its model string.

- **`OfferingPricing`** — per-million `input` / `output` / `cache_read` / `cache_write` prices in a `Currency` (Usd / Cny / Other), with `PricingProvenance` (`ModelsDevBundled` / `ProviderLive` / `ProviderDocs` / `UserOverride` / `Unknown`) and an `effective_at` timestamp.
- **`from_catalog_offering()`** — seeds USD pricing from Models.dev `cost`; provenance and `effective_at` follow the offering's `CatalogSource`. Absent cost — or a cost object with no concrete price — yields `None` (**unknown, not free**).
- **`estimate_cost(TokenUsage)`** — pure per-class arithmetic that returns `None` if any *used* class has an unknown price, so a cost is never silently under-reported; all-zero usage → `Some(0.0)`.
- **`route_pricing_sku()` / `to_route_sku()`** — bridge to the existing `route::PricingSku`, mapping unpriced routes to `UnknownOrStale` rather than a fabricated zero price.
- **`is_stale()`** — age staleness for live (timestamped) rows only; bundled/override prices carry no fetch clock.

## Honesty invariants (from #2608 / #3085)
- Pricing lookup is keyed by provider + offering, not a bare model string.
- Unknown / local / custom / subscription routes are represented honestly — never implied free, never a fabricated token price.
- The coarse `SubscriptionQuota` / `AccountCredits` / `LocalOrNotApplicable` meter shapes already live in `route::PricingSku`; this layer feeds the token-pricing case and leaves the others to the route/UI layer.

## Deliberately out of this slice (separate surfaces)
- Provider usage-payload normalization in the client parsers (Chat Completions / Responses / Anthropic).
- `CostDisplay` UI wiring: `/provider`, `/model`, sidebar cost/usage, session metadata, `doctor --json`. (Left for whoever owns the TUI surface, to avoid collision.)

## Verification
- `cargo clippy -p codewhale-config --all-targets` (clean)
- `cargo test -p codewhale-config` → **245 passed** (11 new `pricing` tests: Models.dev→USD mapping + provenance, live provenance/effective-at, unknown/empty-cost, cache-aware cost estimation, unpriced-class → `None`, honest `route_pricing_sku`, currency round-trips, user-override + no-secrets guard, staleness).

Builds on #3497 + #3498 (both merged); rebased onto `main` so the diff is the pricing layer only.

https://claude.ai/code/session_01991fnUqBbWSgiUFw33L8XX

---
_Generated by [Claude Code](https://claude.ai/code/session_01991fnUqBbWSgiUFw33L8XX)_

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod auth_source;
 pub mod catalog;
 pub mod models_dev;
+pub mod pricing;
 pub mod provider;
 pub mod route;
 
--- a/crates/config/src/pricing.rs
+++ b/crates/config/src/pricing.rs
@@ -1,0 +1,222 @@
+//! Provider/offering-scoped pricing projection with provenance (#3085).
+//!
+//! Network-free. Maps Models.dev offering `cost` (and live / user-override
+//! rows) into pricing rows that carry explicit **provenance**, **currency**, and
+//! **effective-at** metadata, plus a pure cost estimator over normalized token
+//! usage. UI display (`CostDisplay`) and provider usage-payload parsing live
+//! above this layer and are out of scope here.
+//!
+//! Boundary with the route layer: this models *pricing* — offering-owned,
+//! per-token unit prices. The coarse route-facing meter shape already exists as
+//! [`crate::route::PricingSku`]
+//! (`Token` / `SubscriptionQuota` / `AccountCredits` / `LocalOrNotApplicable` /
+//! `UnknownOrStale`); [`OfferingPricing::to_route_sku`] and
+//! [`route_pricing_sku`] bridge to it.
+//!
+//! Honesty rule (#2608 / #3085): pricing is never assumed. A route with no
+//! sourced price yields `None` here and `UnknownOrStale` at the route layer —
+//! never a fabricated token price, and never an implicit "free" for
+//! local/custom/subscription routes.
+
+use serde::{Deserialize, Serialize};
+
+use crate::catalog::{CatalogOffering, CatalogSource};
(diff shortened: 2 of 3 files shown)


Hmbown:
Addressed the review in `6c19ad0`:

1. **Cache-only `Token{None,None}` (issue 2)** — `to_route_sku` now returns `UnknownOrStale` when neither input nor output rate is known, so a cache-only offering no longer projects to a rate-less `Token` badge. The cache rates stay usable via `estimate_cost`.
2. **Redundant `has_any_price()` guard (issue 1)** — dropped; `route_pricing_sku` is now `from_catalog_offering(..).map_or(UnknownOrStale, to_route_sku)`. The unpriced case is handled by `from_catalog_offering → None`, the cache-only case by `to_route_sku`.
3. **Docs** — noted `ProviderDocs` is never produced by `from_catalog_offering`, and that the `tokens as f64` cast is exact for per-turn counts.
4. **Tests** — added cache-only → `UnknownOrStale` (with `estimate_cost` still computing the cache cost), `UserOverride` provenance via `from_catalog_offering`, and an inclusive `>=` TTL boundary assertion.

`cargo fmt --check`, `cargo clippy --all-targets`, `cargo test -p codewhale-config pricing` → **14 passed**.

---

Issue #3084, opened 2026-06-11
v0.8.65: Provider descriptors, wire protocol adapters, and conformance tests


Hmbown:
## Goal

Introduce provider descriptor and adapter contracts so provider behavior is data-backed, testable, and consumed by route resolution instead of scattered across config, picker, doctor, routing, and client code.

This is the provider execution side of #2608. Adapters should execute and validate resolved routes; they should not decide which model or provider to use.

## Architecture Contract

Provider facts describe how to talk to a provider. Model facts describe what a model can do. Provider model offerings describe which provider serves which model under which wire id. Route resolution is the only layer that combines those facts with user config, auth, endpoint/base URL, aliases, and validation into an executable route.

Provider adapters must consume `ReadyRouteCandidate` from #3384. They must not accept naked provider/model strings as the execution boundary.

Use **wire protocol** to mean the selected endpoint's request/response shape: OpenAI Chat Completions, OpenAI Responses, Anthropic Messages, Ollama chat, local runtime APIs, or custom pass-through. This is not prompt-content-based mode switching.

Suggested split:

- `ProviderDescriptor`: id, display name, provider kind, endpoints, default endpoint, auth schemes, catalog policy, model-id policy, live catalog support, flags.
- `EndpointDescriptor`: endpoint key, wire protocol, default base URL, streaming support, endpoint quirks.
- `ProviderAdapter`: provider-owned live catalog fetch and route validation/probing.
- `WireProtocolAdapter`: request/response/stream encoding for OpenAI Chat Completions, OpenAI Responses, Anthropic Messages, Ollama/local, and custom pass-through protocols.

Provider ids and model ids should move toward string newtypes so custom providers, local endpoints, aggregators, and future providers do not require rewiring enum match tables everywhere.

## Current Problem

Provider-specific logic is spread across the codebase:

- `crates/tui/src/config.rs::provider_capability` hand-computes capability fields and request payload mode.
- `crates/tui/src/config.rs::model_completion_names_for_provider`, `normalize_model_name_for_provider`, and `wire_model_for_provider` are separate provider/model match tables.
- `crates/tui/src/model_routing.rs` encodes provider-specific big/cheap choices.
- `crates/tui/src/tui/provider_picker.rs` has separate hard-coded provider hints/env vars.
- `crates/tui/src/main.rs::provider_capability_report` builds doctor output from static helpers.
- Request wire behavior lives in separate client modules without a shared conformance surface.
[...]

PR #3502, opened 2026-06-23
test(config): provider descriptor conformance over all ProviderKinds (#3084)

Hmbown:
## Summary

The network-free, config-layer slice of #3084: a conformance suite that pins provider-descriptor behavior for **every** shipped `ProviderKind` (all 27, data-driven over `ProviderKind::all()`), so adding a provider without wiring its descriptor/resolver behavior fails CI here instead of at runtime.

For each `ProviderKind`:
- **Descriptor is well-formed** — non-empty id that agrees with `ProviderId::from_kind` (catches a provider added to one table but not the other), a default wire model, a default base URL, non-empty env-var names, and a `protocol()` accessor that doesn't panic.
- **Default route resolves** — a `None` selector resolves through the canonical `RouteResolver` to that provider's descriptor default wire model, on the right provider id.
- **`auto` resolves** — the `auto` selector stays the auto sentinel and resolves to a non-empty wire model.

This directly addresses the #3084 acceptance item *"Conformance tests cover all shipped providers,"* using the route descriptor/resolver foundation already in `codewhale_config`.

## Scope
- **Tests only** — no runtime code changes; purely additive.
- The provider **adapter** contracts (live-catalog fetch, wire-protocol request/response encoding, reasoning-label mapping) live in `crates/tui`/client code and are **not** in this slice — they're tracked as the remaining #3084 work to avoid colliding with concurrent TUI work.

## Verification
- `cargo clippy -p codewhale-config --all-targets` (clean)
- `cargo test -p codewhale-config conformance` → **3 tests green across all 27 provider kinds**

https://claude.ai/code/session_01991fnUqBbWSgiUFw33L8XX

---
_Generated by [Claude Code](https://claude.ai/code/session_01991fnUqBbWSgiUFw33L8XX)_

### DIFF
--- a/crates/config/src/route/conformance_tests.rs
+++ b/crates/config/src/route/conformance_tests.rs
@@ -6,6 +6,7 @@
 //! runtime. They are intentionally data-driven over [`ProviderKind::all`] and
 //! network-free; provider execution/adapter behavior is exercised elsewhere.
 
+use super::bundled_offerings;
 use super::descriptor::ProviderDescriptor;
 use super::ids::{LogicalModelRef, ProviderId};
 use super::resolver::{RouteRequest, RouteResolver};
@@ -63,6 +64,7 @@ fn every_provider_kind_has_a_wellformed_descriptor() {
 #[test]
 fn every_provider_kind_resolves_its_default_route() {
     let resolver = RouteResolver::new();
+    let bundled = bundled_offerings();
     for &kind in ProviderKind::all() {
         let descriptor = ProviderDescriptor::for_kind(kind);
         let candidate = resolver.resolve(&none_request(kind)).unwrap_or_else(|err| {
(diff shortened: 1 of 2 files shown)


Hmbown:
Addressed the review in `7dd2a45`:

- **Finding #1 (default-route assertion):** `every_provider_kind_resolves_its_default_route` now computes the expected wire model the way the resolver does — the bundled default offering's wire id when one exists, otherwise the descriptor default. So the deepseek/together/openrouter cases assert the real contract instead of passing only because the two strings coincide; future `OFFERING_SEEDS` / `default_model()` drift now fails honestly.
- **#2 (`env_vars()` non-empty for auth providers):** skipped intentionally — I can't safely enumerate which of the 27 kinds are legitimately keyless (e.g. OAuth routes like Codex may carry no env var), so a blanket or exclusion-list assertion risks a false failure. The per-entry non-empty check stays.
- **#3 (`none_request` reuse):** left as-is (style only).

`cargo fmt --check`, `cargo clippy --all-targets`, `cargo test -p codewhale-config conformance` → **3 passed** across all 27 kinds.

---

Issue #3086, opened 2026-06-11
v0.8.65: Resolved-route context budget service for windows, output caps, compaction, and UI pressure


Hmbown:
## Goal

Add a unified context-budget service so context window, output caps, reasoning tokens, tool-result truncation, compaction thresholds, prompt facts, and UI pressure all agree for the selected route.

This should consume provider/model/offering facts from #2608 and #3365. It should not be a raw `context_window_for_model(model)` string helper.

## Problem

Context behavior is a product-wide budget system. Input window, max output, reasoning tokens, tool-result truncation, compaction threshold, provider tokenizer assumptions, and UI pressure indicators all need to agree.

Right now those facts are spread across model heuristics, provider capability reports, hard-coded request caps, compaction defaults, prompt substitutions, and UI context-pressure logic. That makes context behavior hard to reason about and easy to break when adding Codex/ChatGPT OAuth, Claude, Kimi, Qwen, GLM, OpenRouter models, or local Ollama/vLLM tags.

## Architecture Contract

The service should answer budget questions for a resolved route and session, not for a naked model string.

Inputs should include:

- intrinsic `ModelProfile` context/output facts;
- provider model offering overrides;
- selected route wire model and endpoint protocol;
- user/session overrides;
- Fleet role/loadout policy when a worker is involved;
- tokenizer/counting confidence;
- long-context route requirements from #3205.

Outputs should be structured and provenance-aware:

- model context window and provenance;
- max output cap and effective request cap;
- reserved tokens for system prompt, tools, reasoning replay, server tools, and safety margin;
- compaction threshold and emergency compaction threshold;
- tokenizer/counting strategy and confidence level;
- whether reasoning tokens count against output, input, or separate budget for the selected route;
- user-facing pressure bands: safe/warm/hot/critical;
- fallback behavior when metadata is unknown.

## Expected Consumers

- Compaction thresholds and auto-compact decisions.
- Prompt model facts / constitution substitutions.
- `/model` and `/provider` metadata badges.
- `doctor --json` capability/context output.
- Sidebar/context pressure UI.
- Tool-result truncation and large-output routing budgets.
- Fleet/subagent fork and continuation budgeting.
- Request builders that choose max output tokens.

## Acceptance Criteria

- A single service/API returns the effective context budget for selected route + session settings.
- Compaction, prompt facts, doctor output, and UI pressure consume that service instead of independent helpers.
- Unknown metadata has explicit conservative behavior and visible provenance/reason.
[...]

PR #3523, opened 2026-06-24
[codex] feat(tui): feed route limits into context budgets

Hmbown:
## Summary
- Adds a `route_budget` adapter so resolved `RouteLimits` can drive context windows, output caps, compaction thresholds, prompt facts, and visible pressure readouts.
- Threads active route limits through `App`, `EngineConfig`, `SetModel`, provider switches, fallback switches, and `/model` changes.
- Keeps provider/model fallback behavior for startup, unknown, custom, and local routes where the resolver has no concrete offering limits yet.

Refs #3086.

## Verification
- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked context_budget -- --nocapture`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked route_context_budget -- --nocapture`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked effective_max_output_tokens_for_route -- --nocapture`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked context_usage_snapshot -- --nocapture`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked model_change_update_syncs_engine_model_before_compaction -- --nocapture`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked commands::groups::core::core -- --nocapture`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked route_runtime -- --nocapture`

### DIFF
--- a/crates/tui/src/commands/groups/config/status.rs
+++ b/crates/tui/src/commands/groups/config/status.rs
@@ -5,7 +5,6 @@ use std::path::Path;
 
 use super::CommandResult;
 use crate::compaction::estimate_input_tokens_conservative;
-use crate::config::provider_capability;
 use crate::tui::app::App;
 use crate::utils::{display_path, estimate_message_chars};
 
@@ -166,7 +165,11 @@ fn footer_items(app: &App) -> String {
 }
 
 fn context_usage(app: &App) -> (usize, u32, f64) {
-    let max = provider_capability(app.api_provider, &app.model).context_window;
+    let max = crate::route_budget::route_context_window_tokens(
+        app.api_provider,
+        app.effective_model_for_budget(),
+        app.active_route_limits,
+    );
     let estimated =
         estimate_input_tokens_conservative(&app.api_messages, app.system_prompt.as_ref());
     let total_chars = estimate_message_chars(&app.api_messages);
--- a/crates/tui/src/commands/groups/core/core.rs
+++ b/crates/tui/src/commands/groups/core/core.rs
@@ -129,6 +129,7 @@ pub fn model(app: &mut App, model_name: Option<&str>) -> CommandResult {
             app.last_effective_model = None;
             app.reasoning_effort = ReasoningEffort::Auto;
             app.last_effective_reasoning_effort = None;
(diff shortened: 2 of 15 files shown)


Hmbown:
Follow-up commit `f839a0e4` lands the review nits from both passes. To restate, none of these were correctness bugs today — they close hidden invariants and a test gap before merge:

1. **Window recomputation (nit #1):** `route_context_budget` now takes `window_tokens: u32` directly. The caller (`route_context_budget_for_route`) already computes the window to derive the output cap, so it passes the same value in, removing the invisible "pass the same `(provider, model, route_limits)` triple" invariant.
2. **Output reservation clamp (nit #2):** `route_output_reservation_for_window` now clamps a route output cap to `effective_max_output_tokens(model)` as well as `TURN_MAX_OUTPUT_TOKENS`, matching what `effective_max_output_tokens_for_route` actually sends. A small-window model whose route advertises a generous output cap no longer over-reserves output and reports an unnecessarily tight `available_input_tokens`.
3. **Blank line (nit #3):** added between `effective_max_output_tokens_for_route` and the next doc comment.
4. **Helper tests (nit #4):** added direct tests for `known_route_limits`, `compaction_threshold_for_route_at_percent` (percent clamping floor/ceiling + window scaling), `auto_compact_default_for_route` (threshold split + model fallback), and the now-window-param `route_context_budget`.
5. **Test comment (nit #5):** clarified that the `deepseek/deepseek-v4-pro` string is the Openrouter wire id, not prefix-based inference.

Nit #6 (the `#[cfg(test)]` `context_input_budget_for_provider` wrapper) is left as-is for now — removing it would churn the existing engine tests without behavior payoff, and it reads fine as a test-only delegate.

Verified locally: `cargo fmt --all -- --check`, `git diff --check`, and the full PR verification list (route_budget, route_context_budget, effective_max_output_tokens, context_usage_snapshot, model_change_update_syncs…, commands::groups::core::core, route_runtime) all green. Will let CI re-run and merge once it is back to CLEAN.

Issue #3461, opened 2026-06-23
v0.8.65: MCP duplicate server instance lifecycle and doctor coverage


stream2stream:
## Description

CodeWhale spawns **two MCP server processes** from a single `mcp.json` entry. One works correctly; the other is an orphan that wastes ~4MB RAM, shares a stdio pipe so that killing either process kills both, and never produces any output.

## Steps to reproduce

1. Create a single MCP server entry in `~/.codewhale/mcp.json`:
   ```json
   {
     "servers": {
       "codewhale-web": {
         "command": ".venv\\Scripts\\python.exe",
         "args": ["server/mcp_server.py"]
       }
     }
   }
   ```
2. Start CodeWhale (launches both `codewhale.exe` and `codewhale-tui.exe`)
3. Call any MCP tool (servers are lazy-loaded — they spawn on first use)
4. Observe that two `mcp_server.py` processes appear instead of one

## Expected behavior

Only one MCP server process should be spawned per `mcp.json` entry. `codewhale-tui.exe` should delegate MCP tool calls through `codewhale.exe` rather than spawning its own server instances.

## Actual behavior

Two processes appear at the same start time:

```
PS> (Get-Process -Name python* | Where-Object { $_.CommandLine -match "mcp_server" }).Count
2

PS> Get-Process -Name python* | Where-Object { $_.CommandLine -match "mcp_server" } |
      Select-Object Id, @{N="WS(KB)";E={$_.WorkingSet64/1KB}}, StartTime

Id     WS(KB)  StartTime
--     ------  ---------
25268   3,948  02:58:10   ← orphan (stuck at Python startup, no log output)
34720  37,764  02:58:11   ← real server (working, handles MCP requests)
```

Key observations:

- **Orphan writes no logs** — even with bootstrap logging added to `mcp_server.py`, the orphan never produces a log file. It likely fails to resolve the relative script path `server/mcp_server.py` due to a different working directory.
- **Shared pipe** — killing the orphan via `Stop-Process -Id <orphan_pid> -Force` kills the working server too. The engine treats both as sharing a single pipe.
- **Lazy loading** — no MCP processes exist until the first MCP tool call. Both spawn simultaneously on first use.

## Impact

- **~4MB memory waste** per duplicate MCP server entry per session
- **Cascading failures** — can't clean up the orphan without breaking the working server
- **Compounds with multiple servers** — N configured MCP servers → 2N processes (N working + N orphans)
- Happens every session, on every first tool call

[...]

PR #3524, opened 2026-06-24
fix(tui): make MCP connection drops explicit

cyq1017:
Refs #3461

## Summary
- centralize MCP connection removal through lifecycle helpers that log why stale connections are dropped
- use the helpers for config reloads, reconnects, stale-session retries, and disconnect_all
- add a reload regression test that asserts live transports are dropped when config content changes

## Verification
- cargo test -p codewhale-tui --bin codewhale-tui --locked reload_if_config_changed
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/tui/src/mcp.rs
+++ b/crates/tui/src/mcp.rs
@@ -1949,6 +1949,31 @@ impl McpPool {
         self
     }
 
+    fn drop_connection(&mut self, server_name: &str, reason: &str) {
+        if self.connections.remove(server_name).is_some() {
+            tracing::debug!(
+                target: "mcp",
+                server = %server_name,
+                reason = %reason,
+                "dropped MCP connection"
+            );
+        }
+    }
+
+    fn drop_all_connections(&mut self, reason: &str) {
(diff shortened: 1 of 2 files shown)


Hmbown:
awesome. will fold this in to oauth work on mcp that's coming in 0.8.65 
@claude can we figure out how to add this to the MCP PRs open?

Hmbown:
Thank you @cyq1017 — clean, well-tested fix. Naming each MCP connection drop with its reason (config reload / reconnect / stale-session retry / disconnect-all) makes the pool lifecycle much easier to debug, and the `reload_if_config_changed_drops_live_connections` test (via the Drop-counting transport) pins down that a content-changed config really does drop *and* `Drop` the stale transport.

Harvested into the v0.8.65 lane in #3529 with your authorship preserved — it'll auto-close this PR with credit once it reaches `main`. Scoped it as the connection-lifecycle-visibility slice; the broader duplicate-spawn root cause in #3461 stays tracked separately for v0.8.67.

PR #3565, opened 2026-06-25
fix(tui): catch_unwind in engine event loop to survive UTF-8 byte-boundary panics

Nikita-Tikhomirov:
## Problem

When model or tool output contains multi-byte UTF-8 characters (Cyrillic, CJK, etc.), a byte-boundary panic in the TUI text processing pipeline kills the engine event loop. The UI freezes — the terminal appears hung and requires a restart.

Crash dump excerpt:
\\\
Task: engine-event-loop
Location: crates/tui/src/core/engine.rs:2009:9
Panic: start byte index 26592 is not a char boundary; it is inside 'т' (bytes 26591..26593)
\\\

The \spawn_supervised\ function already catches the panic, but once caught the engine task is dead and the UI sits frozen waiting for events that will never arrive.

## Fix

Wraps each operation in the engine's \
un()\ loop in \AssertUnwindSafe + catch_unwind\. When a panic occurs:
- The error is logged
- A recoverable error event is sent to the UI
- The loop continues — the engine stays alive and processes subsequent messages

## Verification

- [ ] Builds cleanly
- [ ] Russian/Cyrillic prompts no longer freeze the terminal
- [ ] Normal English prompts continue to work unchanged

### DIFF
--- a/docs/V0865_RELEASE_LEDGER.md
+++ b/docs/V0865_RELEASE_LEDGER.md
@@ -1,58 +1,114 @@
 # v0.8.65 Release Ledger
 
-Updated: 2026-06-23 10:35 America/Los_Angeles.
+Updated: 2026-06-24, America/Los_Angeles.
 
-This ledger tracks completion by concrete outcome: merged, replaced by a clean PR, absorbed with evidence, closed after implementation evidence, or blocked only by Hunter approval.
+This ledger tracks the live v0.8.65 milestone by concrete outcome: merged on
+`main`, replaced by a clean PR, closed with evidence, or deliberately moved out
+of the milestone. It is not a public release record. Tags, package publishing,
+GitHub Releases, and website deploys remain separate maintainer actions.
 
-## Current Baseline
+## Current Verdict
 
-| Item | Track | Status | Completion evidence required | Next action |
-| --- | --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


Hmbown:
> ## Problem
> 
> When model or tool output contains multi-byte UTF-8 characters (Cyrillic, CJK, etc.), a byte-boundary panic in the TUI text processing pipeline kills the engine event loop. The UI freezes — the terminal appears hung and requires a restart.
> 
> Crash dump excerpt:
> \\\
> Task: engine-event-loop
> Location: crates/tui/src/core/engine.rs:2009:9
> Panic: start byte index 26592 is not a char boundary; it is inside 'т' (bytes 26591..26593)
> \\\
> 
> The \spawn_supervised\ function already catches the panic, but once caught the engine task is dead and the UI sits frozen waiting for events that will never arrive.
> 
> ## Fix
> 
> Wraps each operation in the engine's \
> un()\ loop in \AssertUnwindSafe + catch_unwind\. When a panic occurs:
> - The error is logged
> - A recoverable error event is sent to the UI
> - The loop continues — the engine stays alive and processes subsequent messages
> 
> ## Verification
> 
> - [ ] Builds cleanly
> - [ ] Russian/Cyrillic prompts no longer freeze the terminal
> - [ ] Normal English prompts continue to work unchanged

Ah thank you so much. I think I have an issue for adding Russian as well, so this can be folded into more substantial work in the future too!!

Issue #889, opened 2026-05-06
Can the ACP protocol be accepted to adapt to Paseo?


Cesditarllas:
As the title suggests, paseo is this open-source project: https://github.com/getpaseo/paseo  
I hope it can integrate with paseo like mainstream CLI tools, so that I can remotely issue programming tasks even when I am not at my computer.

Hmbown:
Thanks, this is a useful integration direction.

The first ACP slice has landed on `main` and is included in the v0.8.15 prep branch: you can run DeepSeek TUI as a local stdio ACP server with:

```bash
deepseek serve --acp
```

For editors such as Zed, the custom agent config points at `deepseek` with args `serve --acp`. I am going to keep this issue open for Paseo-specific compatibility, because remote task dispatch may need more than the baseline ACP prompt/session flow.

If you have a Paseo config snippet or an error log from trying `deepseek serve --acp`, please add it here. That will let us keep the follow-up small and testable instead of turning it into a broad integration rewrite.

boudra:
<img width="807" height="110" alt="Image" src="https://github.com/user-attachments/assets/8781e99e-edcd-416f-9a1a-a7e4d49f9421" />

Hi all, this is integrated in Paseo now.

I noticed the project has rebranded so i'll update that soon.

Another thing i noticed is that the ACP only allows DeepSeek (no models are exposed), since the TUI allows choosing models it'd be great if the ACP could expose that too for full parity

Hmbown:
Closing as completed for the original Paseo compatibility request. CodeWhale already exposes the ACP stdio adapter via , and a later comment reports that Paseo has integrated it. The remaining request in that comment — exposing model choices/parity through ACP — is narrower than Paseo compatibility and should be tracked separately from this integration issue.

PR #3576, opened 2026-06-25
feat: expose provider and model selection over the ACP stdio adapter

### LINKED ISSUE #3546: Extend ACP support to expose provider and model selection
## Problem

CodeWhale already supports the ACP stdio adapter and can be integrated with Paseo. However, provider and model selection do not appear to be exposed or controllable through ACP.

This means that Paseo can use the integration, but it cannot fully mirror the provider/model choices that are available in CodeWhale directly. Users who rely on Paseo therefore cannot easily discover, select, or switch models/providers through the ACP-based workflow.

## Proposed solution

Extend CodeWhale’s ACP support so that ACP clients such as Paseo can access provider and model selection capabilities.

Ideally, ACP should allow clients to:

* discover the available providers
* discover the available models for each provider
* read the currently selected provider and model
* select or switch the active provider
* select or switch the active model

This could be implemented as additional ACP capabilities, commands, or configuration/state exposed through the existing ACP stdio adapter.

## Use case

I would like to use CodeWhale through Paseo with the same flexibility that is available when using CodeWhale directly.

For example, when working from Paseo, I want to be able to choose whether CodeWhale should use a specific provider/model combination without having to leave Paseo or manually change CodeWhale-side configuration outside of the ACP workflow.

The current workflow is enough for basic Paseo compatibility, but not for full provider/model parity.

## Alternatives considered

The main workaround is to configure the provider and model outside of Paseo, for example directly in CodeWhale configuration or through whatever default CodeWhale currently uses.

However, this is less convenient and does not provide a complete ACP-based integration. It also prevents Paseo from presenting or managing CodeWhale’s available model choices in a consistent way.

## Impact

This would significantly improve the Paseo integration for users who work with multiple providers or models.

I would use this whenever working with CodeWhale through Paseo, especially in workflows where switching between different models is important for cost, latency, quality, or capability reasons.

It would make the ACP integration feel complete rather than only partially compatible.

## Additional context

This is a follow-up to #889.

That issue covered the original Paseo compatibility request and was closed as completed because CodeWhale already exposes the ACP stdio adapter and Paseo has integrated it.

The remaining request is narrower: exposing provider/model choice and parity through ACP so that Paseo can fully control the same model selection workflow available in CodeWhale.

mvanhorn:
## Summary

Expose provider and model selection over the existing ACP stdio adapter (`crates/tui/src/acp_server.rs`) so an external ACP client (e.g. Paseo) can discover available providers, read the active provider/model, and switch them for the running session. This is the narrowed follow-up to #889 requested in #3546.

Three JSON-RPC methods are added to the ACP request dispatcher, and a `modelSelection` capability flag is advertised in the `initialize` response so clients can feature-detect it:

- `session/listProviders` returns the provider set (id, displayName, defaultModel) from CodeWhale's existing `ApiProvider::sorted_for_display()`, plus any user-defined `[providers.<name>]` custom entries.
- `session/currentModel` returns the session's active provider/model (defaults to the values `AcpServer` was constructed with).
- `session/selectModel` switches the active model and optional provider for the running session, validating the provider against the built-in registry and configured custom tables.

The change reuses CodeWhale's existing config/provider surface and introduces no new auth, credential, sandbox, or release plumbing. Custom provider keys are preserved (not canonicalized to `custom`) so a `[providers.<name>]` entry round-trips correctly through selection, current-selection reads, and the provider listing.

Fixes #3546

## Testing

- [x] `cargo fmt --all -- --check` (formatted the touched file with `rustfmt --edition 2024`)
- [x] `cargo clippy -p codewhale-tui --bins` (clean for touched code)
- [x] `cargo test -p codewhale-tui acp_server` (14 passed, including 6 new tests)

New colocated `#[cfg(test)]` tests cover capability advertisement, provider listing, current-selection default, a successful switch, unknown-provider rejection, and missing-model (invalid-params) rejection.

## Checklist

- [x] Verified TUI behavior manually if UI changes (N/A: ACP stdio adapter, no TUI change)
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address (N/A: original work)

AI was used for assistance.

### DIFF
--- a/crates/tui/src/acp_server.rs
+++ b/crates/tui/src/acp_server.rs
@@ -193,10 +193,23 @@ impl AcpServer {
             let provider_name = provider_value
                 .as_str()
                 .ok_or_else(|| AcpError::invalid_params("provider must be a string"))?;
-            let provider = ApiProvider::parse(provider_name).ok_or_else(|| {
-                AcpError::invalid_params(format!("unknown provider: {provider_name}"))
-            })?;
-            self.config.provider = Some(provider.as_str().to_string());
+            // Accept either a built-in provider id/alias or a user-defined
+            // custom provider name that has a `[providers.<name>]` table. For
+            // custom providers, preserve the raw key so routing can still find
+            // the configured base URL / auth / model (#1519); canonicalizing to
+            // "custom" would lose that table key.
+            let is_custom = self
+                .config
+                .providers
+                .as_ref()
+                .and_then(|providers| providers.custom_provider_config(provider_name))
+                .is_some();
+            if !is_custom && ApiProvider::parse(provider_name).is_none() {
+                return Err(AcpError::invalid_params(format!(
+                    "unknown provider: {provider_name}"
+                )));
+            }
+            self.config.provider = Some(provider_name.to_string());
         }
 
         self.model = model;
--- a/crates/tui/src/acp_server.rs
+++ b/crates/tui/src/acp_server.rs
@@ -161,7 +161,7 @@ impl AcpServer {
     }
 
     fn list_providers(&self) -> Value {
-        let providers = ApiProvider::sorted_for_display()
+        let mut providers = ApiProvider::sorted_for_display()
             .into_iter()
             .map(|provider| {
                 json!({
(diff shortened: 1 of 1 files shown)


Hmbown:
Thank you so much!!! This is such a huge help for the project.

bjspi:
Yes, thanks indeed :) How to find out when exactly this will be released in whcih version? :)

PR #3577, opened 2026-06-25
fix(tools): make js_execution fetch honor proxy env vars (#3273)

### LINKED ISSUE #3273: js_execution Node fetch does not honor proxy config/env on Windows
﻿## Summary

On Windows, CodeWhale shell tools can reach the internet through a local VPN/proxy, but the built-in `js_execution` tool cannot fetch the same URL. It times out with Undici connect timeout even when CodeWhale config and proxy environment variables are set.

## Environment

- CodeWhale: `0.8.61 (14ac0319e8f4)`
- codewhale-tui: `0.8.61`
- OS: Windows x86_64
- Node.js: `v22.11.0`
- npm: `11.14.1`
- Local proxy/VPN endpoint: `http://127.0.0.1:20499`
- `~/.codewhale/config.toml` includes:

```toml
http_proxy = "http://127.0.0.1:20499"
https_proxy = "http://127.0.0.1:20499"
```

The launching shell also has these env vars set:

```text
HTTP_PROXY=http://127.0.0.1:20499
HTTPS_PROXY=http://127.0.0.1:20499
ALL_PROXY=http://127.0.0.1:20499
http_proxy=http://127.0.0.1:20499
https_proxy=http://127.0.0.1:20499
all_proxy=http://127.0.0.1:20499
NO_PROXY=localhost,127.0.0.1,::1,192.168.*
```

## Reproduction

Run CodeWhale with proxy env vars set, then ask an agent to use `js_execution` to fetch Google:

```text
Use js_execution only. In JavaScript, print whether process.env.HTTPS_PROXY is set. Then fetch https://www.google.com and report status or exact error/cause code.
```

## Expected

`js_execution` should either:

1. honor CodeWhale's `http_proxy` / `https_proxy` config for Node fetches, or
2. honor `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` for Node fetches, or
3. provide a documented way to configure proxy behavior for `js_execution`.

## Actual

Inside CodeWhale agent:

- Shell tools work:
  - `curl.exe -L https://www.google.com` returns HTTP 200 through the proxy.
  - PowerShell `Invoke-WebRequest https://www.google.com` returns HTTP 200 through the proxy.
- `js_execution` sees proxy env vars such as `HTTPS_PROXY`, `HTTP_PROXY`, and `ALL_PROXY`, but `fetch("https://www.google.com")` fails:

```text
FETCH ERROR: cause=UND_ERR_CONNECT_TIMEOUT | message=fetch failed
```

## Additional debugging

Plain Node's built-in `fetch` on Node 22 does not automatically honor proxy env vars. A local test succeeds if an Undici `ProxyAgent` is installed with `setGlobalDispatcher()` before fetch:

```js
const { ProxyAgent, setGlobalDispatcher } = require("undici");
setGlobalDispatcher(new ProxyAgent(process.env.HTTPS_PROXY));
```

When launched normally with this preload, Node fetch returns HTTP 200. However, CodeWhale `js_execution` appears not to inherit or preserve `NODE_OPTIONS` / `NODE_PATH`; inside `js_execution`, both are reported as not set even when the parent CodeWhale launch environment sets them.

A blunt workaround of putting a `node.cmd` wrapper earlier in PATH was not safe: it caused a CodeWhale `js_execution` test to hang, so I backed that out.

## Request

Please make `js_execution` proxy-aware, ideally by applying CodeWhale's configured proxy or inherited proxy env vars to Undici's global dispatcher before executing user JavaScript. Alternatively, please document the supported way to configure proxy behavior for `js_execution` on Windows.

LeoLin990405:
## Problem

Fixes #3273 — on a host with a local proxy/VPN, `js_execution` can't reach the network even though CodeWhale config and the shell both have `HTTP_PROXY`/`HTTPS_PROXY` set. `fetch()` times out with an Undici connect error, while the shell tools reach the same URL fine.

## Root cause

Node's built-in `fetch` (undici) **ignores** `HTTP_PROXY`/`HTTPS_PROXY` environment variables by default. It only honors them when `NODE_USE_ENV_PROXY` is set (equivalent to `--use-env-proxy`, Node ≥ 24). `js_execution` spawns `node script.js` inheriting CodeWhale's environment — so the proxy variables *are* present in the child, but Node's `fetch` never consults them.

Verified locally (Node 26):

```
# proxy env honored only with the flag → connection goes to the (dead) proxy port
$ HTTPS_PROXY=http://127.0.0.1:9 NODE_USE_ENV_PROXY=1 node -e "fetch('https://example.com').catch(e=>console.log(e.cause.code, e.cause.port))"
ECONNREFUSED 9
# without it → fetch ignores the proxy and connects directly
$ HTTPS_PROXY=http://127.0.0.1:9 node -e "fetch('https://example.com').then(r=>console.log(r.status))"
200
```

## Fix

Default `NODE_USE_ENV_PROXY=1` on the `js_execution` Node child so its `fetch()` uses the same `HTTP(S)_PROXY` (and honors `NO_PROXY`) as the rest of CodeWhale. Only defaulted when the variable is unset, so an explicit `NODE_USE_ENV_PROXY=0` opt-out still wins. It's a no-op on Node < 24 (the variable is simply ignored), so it's safe across versions — the proper fix for the reporter's Node 22 would be upgrading to Node ≥ 24, but this removes the CodeWhale-side gap for every Node that supports it.

## Testing

- New test `execute_js_enables_env_proxy_so_fetch_honors_proxy_vars` spawns the Node child and asserts it sees `NODE_USE_ENV_PROXY=1` (skipped when Node is absent or the caller already set the variable).
- `cargo clippy -p codewhale-tui --all-features` clean; `cargo test -p codewhale-tui --all-features js_execution` green.

### DIFF
--- a/crates/tui/src/tools/js_execution.rs
+++ b/crates/tui/src/tools/js_execution.rs
@@ -89,6 +89,17 @@ pub async fn execute_js_execution_tool(
     })?;
     cmd.arg(&script_path).current_dir(workspace);
 
+    // #3273: Node's built-in `fetch` (undici) ignores HTTP(S)_PROXY env vars
+    // unless `NODE_USE_ENV_PROXY` is set (Node >= 24). This child already
+    // inherits CodeWhale's proxy environment, so enabling the flag lets
+    // `js_execution`'s `fetch()` reach the network through the same proxy/VPN
+    // as the rest of the app and honor `NO_PROXY`. Only default it on when the
+    // user hasn't chosen a value, so an explicit opt-out (`NODE_USE_ENV_PROXY=0`)
+    // still wins. No-op on Node < 24, which ignores the unknown variable.
+    if std::env::var_os("NODE_USE_ENV_PROXY").is_none() {
+        cmd.env("NODE_USE_ENV_PROXY", "1");
+    }
+
     let output = tokio::time::timeout(Duration::from_secs(120), cmd.output())
         .await
         .map_err(|_| ToolError::Timeout { seconds: 120 })
@@ -183,6 +194,31 @@ mod tests {
         assert!(
             result.content.contains("intentional fail"),
             "stderr payload must surface the error message; got {}",
+            result.content
+        );
+    }
+
+    #[tokio::test]
+    async fn execute_js_enables_env_proxy_so_fetch_honors_proxy_vars() {
+        if !node_present() {
+            return;
+        }
+        // The tool defers to an explicit caller choice; only assert the
+        // default-on behavior when the surrounding env hasn't set it.
+        if std::env::var_os("NODE_USE_ENV_PROXY").is_some() {
+            return;
+        }
+        let tmp = tempdir().expect("tempdir");
+        let result = execute_js_execution_tool(
+            &json!({ "code": "process.stdout.write(String(process.env.NODE_USE_ENV_PROXY))" }),
+            tmp.path(),
+        )
+        .await
+        .expect("execute");
+        assert!(
+            result.content.contains("\"stdout\":\"1\""),
+            "#3273: js_execution must default NODE_USE_ENV_PROXY=1 so Node's fetch \
+             routes through HTTP(S)_PROXY; got {}",
             result.content
         );
     }


Hmbown:
> ## Problem
> 
> Fixes #3273 — on a host with a local proxy/VPN, `js_execution` can't reach the network even though CodeWhale config and the shell both have `HTTP_PROXY`/`HTTPS_PROXY` set. `fetch()` times out with an Undici connect error, while the shell tools reach the same URL fine.
> 
> ## Root cause
> 
> Node's built-in `fetch` (undici) **ignores** `HTTP_PROXY`/`HTTPS_PROXY` environment variables by default. It only honors them when `NODE_USE_ENV_PROXY` is set (equivalent to `--use-env-proxy`, Node ≥ 24). `js_execution` spawns `node script.js` inheriting CodeWhale's environment — so the proxy variables *are* present in the child, but Node's `fetch` never consults them.
> 
> Verified locally (Node 26):
> 
> ```
> # proxy env honored only with the flag → connection goes to the (dead) proxy port
> $ HTTPS_PROXY=http://127.0.0.1:9 NODE_USE_ENV_PROXY=1 node -e "fetch('https://example.com').catch(e=>console.log(e.cause.code, e.cause.port))"
> ECONNREFUSED 9
> # without it → fetch ignores the proxy and connects directly
> $ HTTPS_PROXY=http://127.0.0.1:9 node -e "fetch('https://example.com').then(r=>console.log(r.status))"
> 200
> ```
> 
> ## Fix
> 
> Default `NODE_USE_ENV_PROXY=1` on the `js_execution` Node child so its `fetch()` uses the same `HTTP(S)_PROXY` (and honors `NO_PROXY`) as the rest of CodeWhale. Only defaulted when the variable is unset, so an explicit `NODE_USE_ENV_PROXY=0` opt-out still wins. It's a no-op on Node < 24 (the variable is simply ignored), so it's safe across versions — the proper fix for the reporter's Node 22 would be upgrading to Node ≥ 24, but this removes the CodeWhale-side gap for every Node that supports it.
> 
> ## Testing
> 
> - New test `execute_js_enables_env_proxy_so_fetch_honors_proxy_vars` spawns the Node child and asserts it sees `NODE_USE_ENV_PROXY=1` (skipped when Node is absent or the caller already set the variable).
> - `cargo clippy -p codewhale-tui --all-features` clean; `cargo test -p codewhale-tui --all-features js_execution` green.
> 

This is awesome - thank you so much. I'm working on adding Tailscale comparability so things like this are increasing in value for the project in real time!!

PR #3549, opened 2026-06-24
feat(i18n): extract Chinese translations into dedicated JSON file

cy2311:
## Summary

Extract Simplified Chinese translations from the hard-coded localization.rs (5385 lines) into a dedicated locales/zh-Hans.json file. This is Step 1 of the i18n refactoring proposed in issue #3537.

### Changes

- **locales/zh-Hans.json** (new) — 408 Chinese translation entries in standard JSON format
- **localization.rs** — Reduced by 541 lines. The chinese_simplified() function now reads from zh-Hans.json at runtime

### Next steps

- Step 2: Extract remaining languages (ja, pt-BR, es-419, vi, zh-Hant)
- Step 3: Consider adopting a formal i18n library (fluent-rs, rust-i18n)
- Step 4: Add CI checks for missing translations

Issue #3537, opened 2026-06-24
Replace hard-coded localization file with a dedicated i18n library


hongqitai:
## Problem

The localization strings in the current project are centralized in `crates/tui/src/localization.rs`. This file has exceeded 5,000 lines and is entirely hard-coded, which may lead to low maintainability, affect compilation speed, and create difficulties in using translation toolchains.

## Proposed solution

Consider introducing a mature Rust i18n library, such as `fluent-rs` or `rust-i18n`, and moving the localization translations to dedicated localization files.

## Impact

- **Increased dependencies**: Adding 2–3 new crates, but they are mature and well-maintained. This will slightly increase build dependencies, yet the benefits far outweigh the costs.
- **Performance overhead**: Most i18n libraries have been thoroughly optimized, so the performance difference compared to hard‑coded strings should be negligible.
- **Migration effort**: The workload is relatively large and may introduce logic changes, so a detailed refactoring plan needs to be discussed.

cy2311:
Started implementing this with Step 1: extracted Chinese Simplified translations into a dedicated JSON file.

PR: https://github.com/Hmbown/CodeWhale/pull/3549

- locales/zh-Hans.json (new) — 408 Chinese translations
- localization.rs reduced by 541 lines

Next steps:
- Extract remaining languages
- Evaluate fluent-rs or rust-i18n
- Add CI checks for coverage

Implemented by @cy2311

Hmbown:
thank you all so much!! this is so helpful.

Hmbown:
Closing as completed by #3583, merged to `main` as `8c8f63f9efa5713590cf1216efbc00f9283716f4`.

What landed:
- `crates/tui/src/localization.rs` was reduced from the large hard-coded translation table to the `MessageId`/locale registry plus a `rust-i18n` lookup path.
- Dedicated locale files now live under `crates/tui/locales/` for English, Simplified Chinese, Traditional Chinese, Japanese, Vietnamese, Brazilian Portuguese, and Latin American Spanish.
- `rust-i18n` is wired into the TUI binary, and the high-visibility call sites now consume the localized `Cow` values safely.

Verification before merge:
- GitHub CI for #3583 was green across lint, ubuntu, macOS, Windows, mobile runtime smoke, npm wrapper smoke, version drift, Claude review, and GitGuardian.
- On the current-main merge result, I ran `cargo fmt --all -- --check`.
[...]

PR #3583, opened 2026-06-25
refactor(localization): extract hardcoded localization texts into JSON and load via rust-i18n(#3537)

hongqitai:
## Summary

Ref #3537
Follow-up of #3549, #3559

- Extract hardcoded localization texts from `crates/tui/src/localization.rs` to `crates/tui/locales`
- Add crate `rust-i18n` to load localization texts
- Next step: use `rust-i18n::t!` macro to replace manual string concatenation
- If you think `rust-i18n` is a wrong way, merge `crates/tui/locales` only

## Testing

- [ x ] `cargo fmt --all -- --check`
- [ x ] `cargo clippy --workspace --all-targets --all-features`
- [ x ] `cargo test --workspace --all-features`

## Checklist

- [ x ] Updated docs or comments as needed
- [ x ] Added or updated tests where relevant
- [ x ] Verified TUI behavior manually if UI changes
- [ x ] Harvested/co-authored credit uses a GitHub numeric noreply address


Hmbown:
Thanks for taking on the localization extraction. This is a broad refactor plus a new runtime dependency, so I’m going to keep it out of the late 0.8.65 release lane until the checks are fully green and we have more room to review the i18n shape.

Current blocker: macOS fails in `commands::plugins::tests::test_plugins_empty_directory` at `crates/tui/src/commands/plugins.rs:215` because the emitted message no longer contains the canonical empty plugin directory path. Windows was cancelled after that failure. The likely fix is to preserve the path interpolation in the localized/plugin empty-directory message, then rerun the TUI test job.

PR #3593, opened 2026-06-25
fix(tui): honor shell-only exec tool surface

### LINKED ISSUE #2954: v0.8.56: Make benchmark shell-only mode actually shell-only
## Goal

Make the Terminal-Bench / Harbor benchmark profile expose a Codex-like shell-focused tool surface so token comparisons are not distorted by extra native tool schemas or extra tool-call habits.

## Current evidence

The current benchmark runner sets `CODEWHALE_TOOL_SURFACE=shell-only` and injects a benchmark config with shell tools in `scripts/benchmark/harbor_agents/codewhale.py`.

However, the CodeWhale stream for `cancel-async-tasks` still shows non-shell tool calls including:

- `git_status`
- `list_dir`
- `read_file`
- `checklist_write`
- `write_file`
- `exec_shell`

That means either the shell-only profile is not being applied, the tool registry is still advertising non-shell schemas, or the profile means something different than the benchmark needs.

## Why this matters

Native structured tool schemas and structured tool outputs can increase both input and output token counts. A Codex CLI comparison should not mix shell-only Codex behavior with a broader CodeWhale tool surface unless the summary explicitly calls that out.

## Scope

- Trace how `CODEWHALE_TOOL_SURFACE=shell-only` is read and applied.
- Ensure the benchmark profile can hide non-shell tools from the model, not just prefer shell tools.
- Decide the minimal benchmark surface: likely `exec_shell` plus wait/interact variants if needed.
- Keep normal interactive CodeWhale behavior unchanged unless explicitly opted into the shell-only profile.
- Add regression coverage that the shell-only prompt/tool registry does not advertise `read_file`, `write_file`, `list_dir`, `git_status`, or checklist tools.

## Acceptance criteria

- In a shell-only benchmark run, the model-visible tool list is shell-only.
- A stream fixture or test proves non-shell tools are unavailable under the benchmark profile.
- Re-running at least `cancel-async-tasks` records whether input/output tokens move closer to Codex after the tool surface is reduced.
- If CodeWhale intentionally keeps non-shell tools available, the comparison harness labels that run as `native-tools`, not `shell-only`.

Hmbown:
## Summary

- maps `CODEWHALE_TOOL_SURFACE=shell-only` in `codewhale exec` to a shell-only `allowed_tools` gate when `--allowed-tools` is not passed
- keeps explicit `--allowed-tools` precedence and leaves `full`/`native-tools` surfaces unchanged
- adds regression coverage that the shell-only catalog hides `read_file`, `write_file`, `list_dir`, `git_status`, and `checklist_write`

Refs #2954.

## Verification

- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked exec_shell_only_tool_surface_env_sets_shell_allowlist`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked exec_explicit_allowed_tools_override_shell_only_env`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked exec_full_tool_surface_env_leaves_allowlist_unset`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked tool_catalog_shell_only_benchmark_surface_hides_native_tools`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked exec_parses_tool_gate_and_hardening_flags`
- `cargo check -p codewhale-tui --locked`

## Note

I have not rerun the Harbor `cancel-async-tasks` benchmark/token comparison in this PR, so this is intentionally `Refs #2954` rather than `Fixes #2954`.

### DIFF
--- a/crates/tui/src/core/engine/tests.rs
+++ b/crates/tui/src/core/engine/tests.rs
@@ -215,6 +215,33 @@ fn tool_catalog_filter_applies_allow_and_deny_gates() {
 }
 
 #[test]
+fn tool_catalog_shell_only_benchmark_surface_hides_native_tools() {
+    let mut catalog = vec![
+        catalog_tool("exec_shell"),
+        catalog_tool("exec_shell_wait"),
+        catalog_tool("exec_shell_interact"),
+        catalog_tool("read_file"),
+        catalog_tool("write_file"),
+        catalog_tool("list_dir"),
+        catalog_tool("git_status"),
+        catalog_tool("checklist_write"),
+    ];
+    let shell_only = [
+        "exec_shell".to_string(),
+        "exec_shell_wait".to_string(),
+        "exec_shell_interact".to_string(),
+    ];
+
+    filter_tool_catalog_for_gates(&mut catalog, Some(&shell_only), None);
+
+    let names: Vec<&str> = catalog.iter().map(|t| t.name.as_str()).collect();
+    assert_eq!(
+        names,
+        ["exec_shell", "exec_shell_wait", "exec_shell_interact"]
+    );
+}
+
+#[test]
 fn tool_catalog_filter_is_inert_without_gates() {
     let mut catalog = vec![catalog_tool("read_file"), catalog_tool("exec_shell")];
     filter_tool_catalog_for_gates(&mut catalog, None, None);
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -371,6 +371,56 @@ enum ExecOutputFormat {
     StreamJson,
 }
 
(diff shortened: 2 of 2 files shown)


Hmbown:
Follow-up pushed in `6c7d5258b` to fix the CI lint failure (clippy `manual_map` in `resolve_exec_allowed_tools`). Verified locally with:\n\n- `cargo fmt --all -- --check`\n- `git diff --check`\n- `cargo test -p codewhale-tui --bin codewhale-tui --locked exec_shell_only_tool_surface_env_sets_shell_allowlist`\n- `cargo clippy --workspace --all-features --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::assertions_on_constants`

Issue #3256, opened 2026-06-15
Tool-call output: collapse happy path by default, expand failures, make verbosity configurable


Hmbown:
## Summary

Today tool-call output shows the full command string + all stdout sections regardless of outcome, so successful runs are ~80% scaffolding noise. We should make the **collapsed, intent-first** view the default for everyone, and make expansion a config option.

## Default behavior (proposed)

Success and failure should get opposite verbosity:

- **On success** — collapse each call to two lines:
  - intent (one line: *why* the call was made)
  - result (the single fact/verdict it produced)
- **On failure (non-zero exit / error)** — auto-expand to full command + full stderr/stdout.
- **Batched related calls** — show a rollup (e.g. `ground truth: HEAD ✓, binary ✓, smoke script ✓`) instead of N separate blocks.

Mental test for the default: *if every call in a turn succeeds, can the user still follow what happened on one screen?* Right now: no.

## What to drop from the default (success path)
- Full command string (reconstructable, rarely read)
- `echo "=== X ==="` section scaffolding
- `ls -la` perms/bytes/timestamps unless that's the thing being checked
- Successful intermediate steps in a chain

## Config
- Add a setting to control tool-call verbosity / expand tool calls (e.g. `collapsed` (default) | `expanded`).
- Full command + raw output should remain available on demand.

## Remove broken `Alt+V`
- The current `Alt+V` ("Alt+V for details") affordance does not actually work — remove it.
- Replace the "expand" path with the config option above (and/or a working keybinding).

## Acceptance
- [ ] Collapsed intent+result is the default tool-call rendering
- [ ] Failures auto-expand with full command + output
- [ ] Verbosity is configurable
- [ ] `Alt+V` affordance removed (no dead "Alt+V for details" hint)

Hmbown:
Landed #3695 (`7a70455`) to accept `collapsed`/`collapse` as aliases for the canonical `compact` tool-collapse mode, with help text and round-trip settings coverage.

I am leaving this issue open because the original request also discusses batched-call rollups / broader display behavior, but the naming/config gap is now covered. Thanks @findshan for the precise follow-through here.

PR #3639, opened 2026-06-26
fix(tui): expand failed tool output live

cyq1017:
## Summary
- expand failed generic tool output in the live transcript instead of hiding lines behind the omission marker
- show the failed tool input summary so the failing command/path remains visible
- keep successful generic tool collapse and exec shell output behavior unchanged

## Verification
- cargo fmt --check
- cargo test -p codewhale-tui --bin codewhale-tui tui::history::tests::
- cargo test -p codewhale-tui --bin codewhale-tui chat_widget_
- git diff --check

Part of #3256.

### DIFF
--- a/crates/tui/src/tui/history.rs
+++ b/crates/tui/src/tui/history.rs
@@ -1357,7 +1357,8 @@ impl GenericToolCell {
                 ));
             }
         } else {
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you @cyq1017. Expanding failed tool output while it is live gives users much better debugging visibility when something goes wrong. Appreciate the focused runtime UX improvement.

Issue #3488, opened 2026-06-23
v0.8.68: Unicode, CJK, and terminal-width rendering QA


Hmbown:
## Problem

Several reports point at text rendering problems that are broader than a single widget: CJK text, mixed-width Unicode, wrapped output, selector rows, and terminal-width changes can break readability or make output look incomplete.

This should be handled as an explicit QA lane after the v0.8.66 visual matrix starts landing, rather than as isolated fixes per report.

Related reports and themes:

- garbled Chinese / CJK real-time output reports
- incomplete displayed output reports
- selector and settings surfaces with width/contrast issues
- terminal-native copy including visual line breaks from wrapped output (#1853)
- narrow Windows/macOS terminal rendering differences

## Scope

Add Unicode and terminal-width QA coverage for CodeWhale TUI surfaces.

Cover:

- CJK task titles and transcript output
- mixed English/CJK rows in sidebars, modals, and selectors
- wide glyphs, combining marks, emoji, and ambiguous-width punctuation where practical
- copy/export behavior for wrapped output
- terminal resize/narrow-width behavior
- statusline and modal title truncation

## Suggested Implementation

- Add fixture strings with CJK, mixed-width Unicode, and long branch/objective names.
- Test width calculations through the same helpers used by production rendering.
- Add snapshot/golden-state cases for narrow and medium terminals.
- Ensure truncation preserves semantic prefixes/suffixes and never splits grapheme clusters.
- Audit copy behavior separately from visual wrapping.

## Acceptance Criteria

- CJK and mixed-width rows do not corrupt borders or hide key state.
- Truncation/wrapping respects display width, not byte length.
- Copy/export can preserve logical text without visual wrap artifacts where promised.
- At least one macOS Terminal-like and one Windows/PowerShell-like rendering scenario is represented by fixtures or documented manual QA.
- Known related issues are linked or moved into this lane as they are reproduced.

Hmbown:
Additional screenshot-derived Unicode/CJK case from 2026-06-23 dogfood:

The sub-agent overlay includes a running agent named `抹香鲸` next to English metadata:

- `抹香鲸  agent_e0b2dcf1  running  10+  124838ms`
- `role: implementer`
- `git: branch codex/issue-3439-zhipu-glm-fixture @ issue-3439`
- `objective: QUESTION: Add Zhipu GLM as a first-class provider-scoped ...`

[...]

PR #3641, opened 2026-06-26
fix(tui): bound agent card text width

cyq1017:
Part of #3488.

Problem:
- Delegate/fanout agent cards ignored the render width, and action truncation used character count instead of terminal display width. Mixed CJK/ASCII text could push status/detail rows past the available columns.

Change:
- Reuse the shared display-width truncation helper for agent-card headers and action/summary rows.
- Add a CJK regression fixture based on the sub-agent dogfood case.

Verification:
- cargo test -p codewhale-tui --bin codewhale-tui tui::widgets::agent_card::tests::
- cargo fmt --check
- git diff --check


Hmbown:
Thank you @cyq1017. The agent card wrapping and CJK width handling make the TUI feel much more respectful of real-world terminals and languages. Really appreciate this kind of detail work.

Issue #2967, opened 2026-06-09
telegram-bridge: streaming & resilience hardening — progress-edit, typing, MarkdownV2, chunking, offset/replay safety, backoff


Hmbown:
## Summary

The Telegram bridge (`integrations/telegram-bridge/src/index.mjs`, `src/lib.mjs`) is a deliberately minimal MVP. This issue folds the production behaviors a long-running phone bridge needs into one incremental workstream (one issue, sliced PRs — avoids issue spam). Depends on the dispatch-concurrency bug fix (#2966) landing first.

This picks up the **deferred/partial rows** of the "Telegram hardening checklist" in `docs/REMOTE_SETUP_DESIGN.md` (the doc's own note says "Revisit after `remote-setup` lands" — this is that revisit; update the table statuses as PRs land):

| Checklist row | Status in doc | Picked up here |
|---|---|---|
| Text batching + progress-edit (edit one msg vs spam) | deferred | Yes — items 1 and 4 |
| MarkdownV2 escaping + table rendering | deferred | Yes — item 3 |
| Network/connect-timeout retry | partial (generic 3s backoff) | Yes — item 6 |
| 409 polling conflict | done (flat 10s forever) | Item 6 adds escalating ladder + fatal taxonomy |
| 429 retry_after | done (poll path only) | Item 6 extends to the send path |
| Webhook mode | out of scope | Still out of scope (non-goal) |

### 1. Progress-edit streaming (one message, edited in place)

- **Gap:** deltas accumulate silently; a flush requires BOTH >3500 chars AND >15s (`src/index.mjs:577-585`), so any response under ~3500 chars is total silence between "Started turn" and the final dump (`src/index.mjs:636`). Every flush is a brand-new `sendMessage` (`src/index.mjs:802-815`); the only edit helper today is `editMessageReplyMarkup` (`src/index.mjs:825-831`).
- **Design:** first flush sends the progress message and captures its `message_id`; subsequent flushes `editMessageText` that same message with a trailing cursor marker (e.g. `▌`); flush when ~0.8s elapsed or a small content threshold accrues; on flood/rate errors back off adaptively (doubling toward ~10s), and after repeated edit failures degrade gracefully to today's append-only sends; the final edit removes the cursor and becomes the answer.

### 2. Typing indicator cadence

- **Gap:** the bridge never calls `sendChatAction` — zero feedback during the (common) silent phase of a turn.
[...]

PR #3648, opened 2026-06-26
fix(bridge): keep Telegram chunks readable

cyq1017:
Refs #2967

Problem:
- Long bridge replies were split at fixed character boundaries, which could split words and leave fenced code blocks unreadable across Telegram chunks.

Change:
- Prefer newline/whitespace split points before hard splitting.
- Close and reopen fenced code blocks across chunks while keeping each rendered chunk within the configured limit.
- Keep surrogate-pair handling covered.

Verification:
- node --test test/*.test.mjs (integrations/bridge-core)
- node --test test/*.test.mjs (integrations/telegram-bridge)
- node --test test/*.test.mjs (integrations/feishu-bridge)
- node --test test/*.test.mjs (integrations/wecom-bridge)
- node --test test/*.test.mjs (integrations/weixin-bridge)
- npm run check (integrations/bridge-core, telegram-bridge, feishu-bridge, wecom-bridge, weixin-bridge)
- git diff --check

### DIFF
--- a/integrations/bridge-core/src/lib.mjs
+++ b/integrations/bridge-core/src/lib.mjs
@@ -291,15 +291,76 @@ export function preservedChatStateFields(state = {}, fields = ["model"]) {
 
 export function splitMessage(text, maxChars = 3500) {
   const value = String(text || "");
+  const limit = Math.max(1, Math.floor(Number(maxChars) || 3500));
   const chars = Array.from(value);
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you @cyq1017. Making split bridge chunks readable is one of those fixes users feel immediately, even if it looks small in the diff. Appreciate the careful bridge UX polish.

PR #3649, opened 2026-06-26
ci: add DCO (Developer Certificate of Origin) check

pkeging:
- Add dco.yml workflow to verify Signed-off-by on all PR commits
- Add DCO section to CONTRIBUTING.md with usage instructions
- Uses pull_request_target for safe contributor PR handling

Ref: developer-certificate-of-origin (DCO)

## Testing

## Checklist

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -1,0 +1,46 @@
+name: DCO
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+  statuses: write
+
+jobs:
+  dco:
+    name: Check Signed-off-by
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check commits for Signed-off-by
+        shell: bash
+        run: |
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks @pkeging. I’m going to keep this out of the 0.8.65 release lane and treat it as post-release CI policy work.

Two reasons:

- This turns DCO into an enforcing contributor gate. Per our current stewardship rule, new contributor gates should stay dry-run unless Hunter explicitly approves enforcement.
- The workflow uses `pull_request_target` with a default `actions/checkout`, which normally checks out the base repository/ref. That means `git rev-list origin/main..HEAD` can end up validating the base branch checkout rather than the PR commit range, so the check may pass/fail for the wrong reasons.

The shape is useful, but I’d want this reworked as either a dry-run/advisory check first, or as a carefully scoped implementation that explicitly fetches the PR head safely and validates the intended commit range without running untrusted PR code.

Hmbown:
Closing as superseded by #3655, which landed the DCO workflow in advisory-only form with pull_request, PR-head checkout, merge-base commit range handling, and fork-safe checkout repository support. This avoids the enforcing/pull_request_target concerns from this first version. Thanks for starting the thread.

PR #3654, opened 2026-06-26
fix(telegram): retry transient send failures

cyq1017:
Refs #2967

Problem:
- Telegram sendMessage calls throw immediately on 429 or transient network failures.
- A failed turn update can stop the bridge stream even though the runtime turn is still active.

Change:
- Add bounded retry delays for sendMessage on 429 retry_after and short transient network failures.
- Keep non-send Telegram methods unchanged.
- Log turn update send failures after retries instead of ending the stream.

Verification:
- node --test test/*.test.mjs (integrations/telegram-bridge)
- npm run check (integrations/telegram-bridge)
- git diff --check

### DIFF
--- a/integrations/telegram-bridge/src/index.mjs
+++ b/integrations/telegram-bridge/src/index.mjs
@@ -21,7 +21,8 @@ import {
   stripGroupPrefix,
   threadListKeyboard,
   telegramIdentity,
-  telegramRetryDelayMs
+  telegramRetryDelayMs,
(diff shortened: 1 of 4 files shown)


Hmbown:
Thank you @cyq1017. The retry path is a quiet but important reliability improvement for real Telegram use, especially when transient send failures happen mid-turn. Appreciate you continuing to harden this bridge.

PR #3656, opened 2026-06-26
fix(telegram): keep typing indicator alive during turns

cyq1017:
Problem
- Telegram users do not see a typing indicator while long-running turns are streaming.
- Approval prompts should pause the indicator while the bridge waits for a user decision.

Change
- Send a bounded, non-blocking sendChatAction typing signal while turn events are streamed.
- Pause typing while an approval prompt is pending and stop it on terminal turn events.
- Thread an optional AbortSignal through Telegram API calls for the typing timeout.

Verification
- node --test test/*.test.mjs
- npm run check
- git diff --check

### DIFF
--- a/integrations/telegram-bridge/src/index.mjs
+++ b/integrations/telegram-bridge/src/index.mjs
@@ -26,6 +26,9 @@ import {
 } from "./lib.mjs";
 import { ThreadStore as CoreThreadStore } from "../../bridge-core/src/lib.mjs";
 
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you @cyq1017. This is exactly the kind of polish that makes the bridge feel alive during real use: bounded, tested, and careful around approval waits. Really appreciate the steady Telegram bridge work.

PR #3658, opened 2026-06-26
fix(telegram): persist handled update offsets

cyq1017:
## Summary
- persist the Telegram getUpdates offset in bridge state and restore it on startup
- advance the offset only after an update is handled successfully
- ignore replayed callback updates by update_id and consume resume action tokens before dispatch

Part of #2967.

## Verification
- npm test (integrations/bridge-core)
- npm run check (integrations/bridge-core)
- npm test (integrations/telegram-bridge)
- npm run check (integrations/telegram-bridge)
- git diff --check upstream/main...HEAD

### DIFF
--- a/integrations/bridge-core/src/lib.mjs
+++ b/integrations/bridge-core/src/lib.mjs
@@ -3,6 +3,14 @@ import path from "node:path";
 
 const DEFAULT_ACTION_TTL_MS = 24 * 60 * 60 * 1000;
 
+function normalizeCursorValue(value, fallback = 0) {
+  const number = Number(value);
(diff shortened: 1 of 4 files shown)


Hmbown:
Thank you @cyq1017. This is a strong correctness slice for the Telegram bridge: persisted offsets, advance-after-success handling, replay protection, and one-shot resume actions are exactly the kind of reliability work that keeps remote use from feeling spooky after restarts. Really appreciate the careful tests here.

cyq1017:
Thanks Hunter. I’ll keep the remaining Telegram bridge hardening slices in #2967 narrow and tested, and will continue to watch for overlap before opening follow-ups.

PR #3665, opened 2026-06-26
fix(telegram): debounce turn sequence writes

cyq1017:
Part of #2967.

## Summary
- debounce Telegram turn-stream `lastSeq` writes instead of saving on every SSE event
- force a final `lastSeq` flush when the stream exits so reattach can resume from the latest observed event
- cover the stream state-write behavior in the dispatch concurrency regression test

## Verification
- `npm test`
- `npm run check`
- `git diff --check upstream/main...HEAD`


Hmbown:
Thanks @cyq1017, I inspected the diff. Debouncing `lastSeq` writes with a forced final flush is a clean shape for this Telegram stream hot path, and the regression test covers the important behavior. I’m leaving it unmerged while it is still marked draft and the slow checks are running, but this looks promising once you mark it ready and the matrix is green.

PR #3679, opened 2026-06-27
ci: retry OHOS dependency graph probe

Hmbown:
## Summary
- retry the OHOS `cargo tree` dependency graph probe before failing the release drift check
- preserve the final nonzero `cargo tree` status for real dependency or package errors
- document the CI hardening in the root and TUI changelogs

## Why
The Version drift workflow failed once after the version check passed because Cargo hit a transient registry EOF while resolving `serde` for the OHOS dependency graph. Local rerun of the same script passed, so this keeps the useful drift gate while making it less brittle to one-off registry/network drops.

## Verification
- `bash -n scripts/release/check-ohos-deps.sh`
- `./scripts/release/check-ohos-deps.sh`
- forced retry path with `CODEWHALE_OHOS_DEP_PACKAGE=definitely-not-a-real-package CODEWHALE_OHOS_DEP_RETRIES=2 CODEWHALE_OHOS_DEP_RETRY_DELAY_SECONDS=0 ./scripts/release/check-ohos-deps.sh`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Hardened the OHOS dependency drift check against transient Cargo registry EOFs
+  by retrying the dependency graph probe before failing CI.
 - Updated the `/links` provider fallback to the current CodeWhale docs URL and
   added a Baidu Qianfan docs link. Harvested from #3621 by @noaft.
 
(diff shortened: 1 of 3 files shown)


Hmbown:
Thank you so much for catching this — wow. I feel a little silly for missing the zero-retry edge, and I really appreciate you taking the time to make the review this concrete. I tightened it up so `CODEWHALE_OHOS_DEP_RETRIES=0` now fails loudly instead of vacuously passing.

PR #3689, opened 2026-06-27
fix(telegram): bound polling conflict retries

cyq1017:
Summary
- replace the flat 10s Telegram getUpdates conflict retry with a bounded 15/25/35/45/55s ladder
- fail with a clear operator message when another bridge keeps owning the bot token
- add tests for the conflict delay helper and poll loop behavior

Verification
- node --test integrations/telegram-bridge/test/lib.test.mjs integrations/telegram-bridge/test/dispatch-concurrency.test.mjs
- npm test --prefix integrations/telegram-bridge
- npm run check --prefix integrations/telegram-bridge
- git diff --check

Refs #2967

### DIFF
--- a/integrations/telegram-bridge/src/index.mjs
+++ b/integrations/telegram-bridge/src/index.mjs
@@ -21,6 +21,7 @@ import {
   stripGroupPrefix,
   threadListKeyboard,
   telegramIdentity,
+  telegramPollingConflictDelayMs,
   telegramRetryDelayMs,
   telegramSendRetryDelayMs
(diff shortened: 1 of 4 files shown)


Hmbown:
Thank you @cyq1017 — this is a really thoughtful operator-experience fix. The bounded retry ladder plus a clear conflict message is much kinder than an endless flat poll loop, and I appreciate the focused tests around the helper and poll behavior. Checks are green; since it is still draft, I am leaving it unmerged for you to mark ready when comfortable.

Issue #3354, opened 2026-06-21
中文环境下，可以提供并加载中文skill，更省token


fan-wenjie:

Hmbown:
smart! thank you @fan-wenjie

Hmbown:
Fixed by #3690 (0fb32b9f): skills can now provide locale-specific `description_<tag>` frontmatter, and the session prompt uses the matching locale description with the normal English fallback. The merged tests cover parsing localized descriptions, exact/primary fallback, Simplified Chinese fallback, Traditional Chinese not borrowing Simplified `zh`, and prompt rendering by locale.

Thank you @fan-wenjie for the original Chinese skill-token request, and thank you @findshan for the implementation plus the quick zh-Hant follow-up. Please reopen or file a follow-up if a real Chinese session still receives the wrong skill description.

PR #3690, opened 2026-06-27
feat(skills): locale-aware skill descriptions to save tokens (#3354)

findshan:
## Summary

`Refs #3354` (Chinese environment: provide/load Chinese skills to save tokens).

Skills inject their **name + description** into every session's system prompt
(progressive disclosure). In a non-English session the model is instructed to
respond in e.g. Chinese, yet every skill description stays English — that wastes
prompt tokens and hands the model mixed-language context.

This lets a `SKILL.md` ship native-language descriptions via locale-suffixed
frontmatter keys, selected by the session locale with an English fallback:

```yaml
---
name: demo
description: A demo skill            # default / fallback (unchanged behavior)
description_zh: 一个演示技能          # used when the session locale is zh*
description_zh-Hant: 一個示範技能
---
```

### How it works
- `Skill` gains `localized_descriptions: HashMap<lowercased-tag, String>`,
  populated from `description_<tag>:` keys. The frontmatter parser already
  collects every key into a map, so this is a small extraction — no parser
  rewrite.
- `Skill::description_for_locale(tag)`: exact (lowercased) match → primary
  language subtag (`zh-cn` → `zh`) → default `description`. It deliberately does
  **not** fold `zh-hant` into `zh` (Traditional/Simplified are authored
  separately; each is keyed by the tag the author means).
- `render_skills_block` selects per the session `locale_tag`, threaded from
  `PromptSessionContext`. The no-locale public render wrappers pass `"en"`.

**Fully backward compatible:** skills without localized variants render exactly
as today.

## Testing

- [x] `cargo test -p codewhale-tui --bin codewhale-tui` — **5481 passed, 0 failed**.
- [x] New unit tests: `parse_skill_collects_localized_description_frontmatter`,
  `description_for_locale_matches_exact_then_primary_then_falls_back`,
  `description_for_locale_uses_default_when_no_localized_variants`,
  `render_skills_block_selects_description_by_locale`.
- [x] `prompts::` suite green (touched the skills-block call site).
- [x] `cargo fmt --all -- --check` clean; `cargo clippy` — no new warnings in touched files.

## Checklist

- [ ] Verified TUI behavior manually if UI changes — n/a (prompt-text/data change; unit-tested at the render-helper level)
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address — n/a (single author)

Note: the `description_<tag>` frontmatter convention is a proposal — happy to
adjust to whatever shape you prefer for the v0.8.67 skills work.

### DIFF
--- a/crates/tui/src/prompts.rs
+++ b/crates/tui/src/prompts.rs
@@ -1130,11 +1130,13 @@ pub fn system_prompt_for_mode_with_context_skills_session_and_approval(
                 workspace,
                 dir,
                 skill_discovery_mode,
+                session_context.locale_tag,
             )
         }
         None => crate::skills::render_available_skills_context_for_workspace_with_mode(
             workspace,
             skill_discovery_mode,
+            session_context.locale_tag,
         ),
     };
     if let Some(block) = skills_block {
--- a/crates/tui/src/runtime_api/tests.rs
+++ b/crates/tui/src/runtime_api/tests.rs
@@ -3650,12 +3650,14 @@ fn skill_entry_is_bundled_requires_configured_bundle_path() {
     let bundled_skill = crate::skills::Skill {
         name: "delegate".to_string(),
         description: String::new(),
+        localized_descriptions: std::collections::HashMap::new(),
         body: String::new(),
         path: bundled_skill_path,
     };
     let override_skill = crate::skills::Skill {
         name: "delegate".to_string(),
         description: String::new(),
+        localized_descriptions: std::collections::HashMap::new(),
         body: String::new(),
         path: override_skill_path,
     };
--- a/crates/tui/src/skills/mod.rs
+++ b/crates/tui/src/skills/mod.rs
@@ -87,10 +87,15 @@ impl Skill {
     /// Pick the best description for a session `locale_tag`, falling back to the
     /// default `description` when no localized variant matches.
     ///
-    /// Matching is conservative on purpose: an exact (lowercased) tag match
-    /// first, then the primary language subtag (e.g. `zh-cn` → `zh`). We do not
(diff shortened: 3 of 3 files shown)


Hmbown:
Thank you for this, @findshan — this is a strong direction and I really like the token-saving angle for Chinese/non-English sessions. I did one code-level pass while the remaining checks are still running and found one thing I think we should tighten before landing: the PR text says zh-Hant should not fold into zh, but description_for_locale currently falls back from any hyphenated locale to its primary subtag, so zh-Hant will use description_zh when no exact description_zh-hant exists. Could you either adjust the helper/tests so script-specific Chinese tags such as zh-Hant and zh-Hans fall back to default unless explicitly authored, or update the documented behavior if generic zh fallback is intentional? Once that is aligned, this looks very close.

findshan:
Good catch — you're right, the helper contradicted the PR text. Fixed in `fab049b2`.

The primary-subtag fallback now refuses to cross the Simplified/Traditional written-form boundary: a Traditional tag (`zh-Hant`, or the Traditional regions `zh-TW` / `zh-HK` / `zh-MO`) no longer borrows a Simplified `description_zh` — it matches only an exact `description_zh-hant`-style key, otherwise the default. Simplified tags (`zh`, `zh-Hans`, `zh-CN`) still fold to `description_zh`, and generic region fallback (`en-US`→`en`, `pt-BR`→`pt`) is unchanged.

Added tests: `zh-Hant` / `zh-TW` / `zh-HK` fall back to the default when no Traditional key is authored, and use the exact `zh-hant` key when it is. fmt + clippy (`--workspace --all-features --locked -Dwarnings`) clean; skills suite green. Thanks for the careful read.

Issue #3638, opened 2026-06-26
exposed main prompt for broader use cases


DracheTek:
## Problem

我正在考虑将codewhale tui的使用例拓展至更多目的：文学创作，背景围读等类，为这些目的使用软件工程所用的提示词是不合用的。在修改AGENTS.md的基础上可能还需要修改personality乃至constiute的一部分。

## Proposed solution

将hard load的文本文件改为指向config目录的同样文本文件的链接。考虑两种做法：不留hard load释放二进制大小或保留hard load作为删去了/未配置config时的fallback

## Use case

见上

## Alternatives considered

目前的主要workaround是直接修改AGENTS.md 但是缺乏某种意义上的可扩展性和自由度。

## Impact

本人经常在网页版deepseek进行上述创作目的的调用，但是网页版如今缺乏最强思考/工具调用/多轮思考等增强手段，质量一般。

Hmbown:
Implemented by #3696 (`674c8768`): CodeWhale now supports an explicit config-directory base prompt override at `~/.codewhale/prompts/constitution.md` / `$CODEWHALE_HOME/prompts/constitution.md`, gated by `CODEWHALE_ALLOW_BASE_PROMPT_OVERRIDE=1` so a stray file cannot silently replace the bundled Constitution. Missing or empty files still fall back to the shipped prompt.

Thanks @DracheTek for the broader-use-case request around writing/review workflows, and thank you @findshan for the careful implementation and docs. One caveat: we are still working toward a more first-class customizable prompt/config story, so this escape hatch may be renamed, moved, or replaced later; for now it gives advanced users a deliberate opt-in path without removing the runtime-owned safety layers.

PR #3696, opened 2026-06-27
feat(prompts): allow overriding the base prompt from the config dir (#3638)

findshan:
## What

Closes the core of #3638: let a user repurpose the TUI for non-software use cases (long-form writing, document review, etc.) by **swapping the base/constitutional system prompt from a config-directory file**, without editing in-tree files or building a custom embedder.

The override *hooks* already exist for embedders (`set_base_prompt_override` + the `OnceLock` cells in `prompts.rs`), but there was no user-facing source feeding them. This PR adds that source.

## How

- At startup (once, right after CLI parse, before any engine/subcommand spawns — the override cells are first-call-wins), look for `~/.codewhale/prompts/constitution.md` (under `$CODEWHALE_HOME` when set).
- If it exists and is non-empty, install it via the existing `set_base_prompt_override` path; otherwise no-op → the bundled constant is used. **Fully backward compatible.**
- Empty/whitespace-only files are ignored, so a stray file can't silently blank the system prompt.

## Safety / trust-boundary scope (deliberately narrow)

I know the prompt surface is a trust boundary, so the scope is intentionally minimal: **only the byte-stable base prompt segment is overridable.** Mode deltas, the approval policy, the tool taxonomy, Context Management, and the Compaction Relay stay owned by CodeWhale's runtime assembly (per the existing `StaticPromptCtx` contract), so an override **cannot strip safety-relevant guidance** (sandbox/approvals) — it only swaps the task/voice framing. Byte-stability of the composed prompt when no override is set is unchanged (the existing byte-stable tests still pass).

Happy to gate this behind a feature flag or adjust the path/semantics if you'd prefer — flagging the trust-boundary explicitly for sign-off.

## Files

- `crates/tui/src/prompts.rs` — pure, unit-tested resolver `read_prompt_override_file` + `load_config_dir_prompt_overrides` / `load_prompt_overrides_from_config_home`.
- `crates/tui/src/main.rs` — one wiring call.
- `docs/CONFIGURATION.md` — documents the file, its scope, and that it can't remove safety layers.

## Tests

3 new unit tests (present / absent / empty-file resolution). The global-install path isn't unit-tested by design — `set_base_prompt_override` writes a process-wide `OnceLock` that would leak across the test binary (same reason the existing `prompt_override_storage_reports_duplicate_sets` uses a local cell). Verified locally:

```
cargo test -p codewhale-tui --bin codewhale-tui --locked prompts::   # 89 passed (incl. byte-stability)
cargo fmt --all -- --check
cargo clippy -p codewhale-tui --all-features --locked --bin codewhale-tui   # clean (CI flags)
```

## Follow-up (not in this PR)

#3638 also mentions personality overlays. Those don't have override hooks yet; happy to add `prompts/personalities/<name>.md` overrides as a follow-up if you want the same treatment.

Refs #3638

### DIFF
--- a/crates/tui/src/main.rs
+++ b/crates/tui/src/main.rs
@@ -1183,6 +1183,12 @@ async fn main() -> Result<()> {
     let cli = Cli::parse();
     logging::set_verbose(cli.verbose || logging::env_requests_verbose_logging());
 
+    // Install any user prompt overrides from the config directory before an
+    // engine can compose a system prompt. The override cells are
+    // first-call-wins; doing this once here keeps every downstream turn
+    // consistent. Missing files are a no-op (bundled defaults). See #3638.
+    crate::prompts::load_prompt_overrides_from_config_home();
+
     // Handle subcommands first
     if let Some(command) = cli.command.clone() {
         return match command {
--- a/crates/tui/src/prompts.rs
+++ b/crates/tui/src/prompts.rs
@@ -445,10 +445,32 @@ pub fn set_static_prompt_composer_override(
 // `StaticPromptCtx`), so an override cannot strip safety-relevant guidance.
 // A missing or empty file is a no-op — the bundled constant is used — so this
 // is fully backward compatible.
+//
+// Because replacing the base prompt is a trust-boundary action (per maintainer
+// review on #3638), the override file alone is NOT sufficient: the user must
+// also set an explicit opt-in flag (`CODEWHALE_ALLOW_BASE_PROMPT_OVERRIDE`).
+// This keeps replacing the global Constitution a deliberate, auditable act
+// rather than something a stray file can do.
 
 /// Relative path, under the config directory, of the optional base-prompt
 /// (constitution) override file.
 pub const CONSTITUTION_OVERRIDE_FILE: &str = "prompts/constitution.md";
+
+/// Env flag that must be set (`1`/`true`/`on`/`yes`) to enable config-dir base
+/// prompt overrides. Required in addition to the override file so the global
+/// base prompt can never be replaced by file presence alone.
+pub const BASE_PROMPT_OVERRIDE_OPT_IN_ENV: &str = "CODEWHALE_ALLOW_BASE_PROMPT_OVERRIDE";
+
+/// Whether the user has explicitly opted in to base-prompt overrides.
+fn base_prompt_override_opt_in() -> bool {
+    match std::env::var(BASE_PROMPT_OVERRIDE_OPT_IN_ENV) {
+        Ok(v) => matches!(
+            v.trim().to_ascii_lowercase().as_str(),
+            "1" | "true" | "on" | "yes"
+        ),
+        Err(_) => false,
+    }
+}
 
 /// Read an optional prompt-override file rooted at `config_dir`.
 ///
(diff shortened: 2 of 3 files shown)


Hmbown:
Thank you so much for this, @findshan — this is thoughtful and very clearly scoped, and thanks @DracheTek for the broader-use-case request behind #3638. The implementation shape is crisp and the checks are green.

I am going to hold this one for explicit maintainer sign-off rather than merge it automatically, because it changes a user-facing prompt trust boundary: config-dir base prompt overrides effectively make the global Constitution replaceable at runtime. Current repo guidance still treats the shipped constitution as the sole base prompt and avoids runtime prompt/tag injection.

That does not mean the idea is bad — it may be exactly the right direction for broader writing/review workflows — but I think it needs an intentional product/security call, probably around whether this should be behind an explicit opt-in flag/name/path and how we describe what safety layers remain owned by the runtime. Really appreciate the care you put into making the slice narrow and reversible.

findshan:
Completely agree it's a trust-boundary call that's yours to make — thanks for framing it that way rather than just bouncing it.

I pushed one change that I think de-risks the security half regardless of the product decision, and left the product decision to you:

**Explicit opt-in flag.** The override file alone is no longer sufficient. The user must also set `CODEWHALE_ALLOW_BASE_PROMPT_OVERRIDE=1`; otherwise the file is ignored and the bundled Constitution stays in place (with a log line naming the flag). So replacing the global base prompt is now a deliberate, two-step, auditable act — a stray or copied-in `constitution.md` can't silently swap it. (Env flag for the narrow slice; happy to promote it to a `[prompt]` config field / different name if you'd prefer config over env.)

**On "what safety layers remain owned by the runtime":** the override only replaces the byte-stable base segment via the existing `set_base_prompt_override` hook. Everything the runtime composes *after* it is untouched — mode deltas, the approval policy, the tool taxonomy, Context Management, and the Compaction Relay (the `StaticPromptCtx` contract already scopes embedder overrides to "base/personality segment only"). So an override can change task framing/voice but cannot remove sandbox/approval guidance. I've stated that explicitly in the docs section.

Totally fine to hold for your product/security call — just wanted to make the "yes" version as conservative as possible. Open questions I'd defer to you: env flag vs `[prompt]` config field, whether the path should be a more clearly-intentional name than `constitution.md`, and whether you'd want personalities overridable too (separate follow-up). No rush.

Hmbown:
Thanks again @findshan — I re-reviewed the follow-up commit and the explicit `CODEWHALE_ALLOW_BASE_PROMPT_OVERRIDE=1` gate is exactly the kind of de-risking this needed. The implementation is now a much cleaner two-step opt-in instead of file-presence behavior, and I really appreciate how carefully you scoped the safety boundary in both code and docs.

I am still leaving it unmerged for the moment because this is now a product/security sign-off question rather than a CI/readiness question: do we want user-configurable replacement of the base Constitution at all, even behind an explicit flag, and is `prompts/constitution.md` the right long-term surface/name? Once Hunter signs off on that direction, this looks technically ready to take.

Hmbown:
Approved for merge. Thanks again @findshan — this is the conservative version we wanted: config-directory prompt customization is explicit, documented, and gated behind `CODEWHALE_ALLOW_BASE_PROMPT_OVERRIDE=1`, so it does not silently change the bundled Constitution just because a file exists.

One product caveat for future readers: we are working toward a broader, first-class customizable prompt/config story. This file/env-flag surface gives advanced users a useful escape hatch now, but we may rename, move, or replace it later when the durable config UX lands. For now, the docs make the opt-in and scope clear, and the runtime still owns the safety layers after the base segment.

Issue #3730, opened 2026-06-28
Auto mode: read-only commands (e.g. `codewhale --version`) flagged DESTRUCTIVE and require approval; policy copy references YOLO not Auto


Hmbown:
## Summary

In **Auto** mode, a trivially read-only command (`codewhale --version`) is classified **DESTRUCTIVE** and forced through an approval prompt. Auto mode does not behave as described — safe/read-only shell calls should not require approval, and `codewhale --version` is not destructive by any definition.

## Screenshot
Mode: Auto. Approval takeover shown:

```
┌ DESTRUCTIVE approval — exec_shell ─────────
  [DESTRUCTIVE]  exec_shell   Type: Bash
  About:  Review the Bash command before it runs.
  Impact: Executes a Bash command in your workspace.
  Policy: YOLO skips ordinary approvals; this prompt means a review rule
          or explicit ask-rule is active.
  Command:
      codewhale --version
  Save: 1 ask rule  (tool=exec_shell command=codewhale --version)
  ▶ [1 / y] Approve once
```

## What's wrong

1. **Misclassification:** `codewhale --version` is read-only and is labeled `DESTRUCTIVE`. The destructive/risk classifier is far too aggressive (appears to treat any `exec_shell`/Bash as destructive regardless of the command).
2. **Auto mode not behaving as described:** Auto is supposed to auto-approve ordinary/safe actions and only stop for genuinely risky ones. Here it stops for a version check.
3. **Confusing policy copy:** The explanation talks about **YOLO** ("YOLO skips ordinary approvals…") while the user is in **Auto** mode. The approval rationale should describe the *current* mode's policy, not YOLO's.

## Expected

- Read-only / clearly-safe commands (`--version`, `--help`, `git status`, `ls`, etc.) should not be classified DESTRUCTIVE.
- In Auto mode, ordinary safe shell actions run without approval; only genuinely destructive/irreversible actions prompt.
- The approval prompt's policy explanation should reference the active mode (Auto/Ask/YOLO/Plan), not always YOLO.

## Environment
- codewhale version: 0.8.x (deepseek-v4-pro), TUI, Auto mode
- Surface: approval takeover / exec_shell risk classification

Hmbown:
## Root cause (decision: defer Auto to 0.8.67)

Per maintainer call, **cut Auto mode from 0.8.66** and ship a corrected version in 0.8.67. Diagnosis of *why* Auto is broken, for the 0.8.67 fix:

[...]

PR #3746, opened 2026-06-28
fix(tui): skip approval for readonly auto shell

cyq1017:
## Problem
Auto mode still prompts for direct `!` read-only shell shortcuts because the composer shell path uses the static tool approval requirement.

## Change
Use the shell spec's input-aware approval requirement for direct user shell commands, and add a regression test for read-only shell in Auto mode.

Part of #3730.

## Verification
- `cargo test -p codewhale-tui --bin codewhale-tui run_shell_command_op --locked`
- `cargo test -p codewhale-tui --bin codewhale-tui auto_review_policy --locked`
- `cargo test -p codewhale-tui --bin codewhale-tui shell_readonly --locked`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -1091,7 +1091,8 @@ impl Engine {
                 "Tool 'exec_shell' is disabled by feature flag".to_string(),
             ))
         } else if let Some(spec) = registry.get(&tool_name) {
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you so much for this, @cyq1017. This is a really clean catch: wiring the direct shell path to the input-aware shell approval requirement is exactly the small fix #3730 needed, and the regression test makes the behavior much easier to trust.

Issue #3386, opened 2026-06-22
v0.8.65: Untangle Plan/Agent/YOLO mode cycling from permission policy


Hmbown:
## Problem

The current tabbed mode system is carrying too many responsibilities in one enum and a few scattered branches. `AppMode` currently drives the visible Tab cycle, prompt profile, sandbox shape, shell availability, approval behavior, trust mode, tool catalog shape, and Plan/YOLO save-restore behavior. That makes simple mode cycling harder to reason about and easy to regress.

Observed pressure points in the current codebase:

- `crates/tui/src/tui/app.rs`
  - `AppMode` is both the visible UI state and the source of permission side effects.
  - `App::set_mode` mutates `allow_shell`, `trust_mode`, `approval_mode`, `yolo`, `plan_prompt_pending`, and save/restore state.
  - `cycle_mode` / `cycle_mode_reverse` hard-code the order separately from `/mode` parsing and docs.
- `crates/tui/src/core/engine.rs`
  - `effective_input_policy` performs additional authority downgrades based on input provenance and review-only intent.
  - Turn setup separately carries `mode`, `allow_shell`, `trust_mode`, `auto_approve`, and `approval_mode`, making it hard to see the final effective policy.
- `crates/tui/src/core/engine/tool_setup.rs`
  - sandbox, shell, and tool registry behavior are derived directly from `AppMode` plus session booleans.
- `crates/tui/src/commands/groups/config/config.rs`, `crates/tui/src/settings.rs`, `docs/MODES.md`, and `docs/KEYBINDINGS.md`
  - valid modes, numeric aliases, labels, and cycle descriptions are repeated in multiple places.

The result is that tabbing between Plan, Agent, and YOLO feels fragile: permissions can appear to stick, restore behavior is hard to audit, and contributors have to inspect several files to know what a mode actually does.

## Goal

Keep the existing visible modes exactly as they are: `Plan`, `Agent`, and `YOLO`.

Refactor the mode system so the visible mode cycle is separate from the effective permission/tool policy used for a turn. This should make Tab behavior predictable and make permission behavior table-driven enough that future changes do not require scattered conditionals.

## Proposed Shape

Introduce a single mode-policy resolver for the existing modes:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
enum AppMode {
    Plan,
    Agent,
    Yolo,
}

#[derive(Debug, Clone, Copy, PartialEq, Eq)]
enum PermissionProfile {
[...]

PR #3756, opened 2026-06-29
fix(tui): default interactive Agent shell to approval-gated on

Hmbown:
## Summary
- expose shell tools by default for interactive Agent-mode TUI sessions so approvals can gate shell commands
- keep explicit allow_shell=false respected for Agent sessions
- keep GHSA-safe durable task defaults: omitted task fields still do not grant shell or auto-approve
- stamp approved tool results with model-visible approval text plus structured metadata, so agents do not infer that shell ran ungated
- update docs/MODES.md to describe the interactive default versus durable-task behavior

## Verification
- cargo fmt --all --check
- git diff --check
- ./scripts/release/check-versions.sh
- cargo test -p codewhale-tui --bin codewhale-tui --locked approval_stamp
- cargo test -p codewhale-tui --bin codewhale-tui --locked interactive_tui_defaults_agent_shell_to_approval_gated_on
- cargo test -p codewhale-tui --bin codewhale-tui --locked allow_shell_defaults_to_false_when_unset
- cargo test -p codewhale-tui --bin codewhale-tui --locked add_task_without_optional_fields_does_not_grant_shell_or_auto_approve
- cargo test -p codewhale-tui --bin codewhale-tui --locked build_tool_context_uses_typed_shell_policy_per_mode
- cargo test -p codewhale-tui --bin codewhale-tui --locked agent_tools_with_allow_shell

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -3640,11 +3640,26 @@ impl Config {
         self.context.project_pack.unwrap_or(true)
     }
 
-    /// Return whether shell execution is allowed. Defaults to `false`: shell
+    /// Return whether shell execution is allowed for noninteractive and
+    /// durable-task profiles. Defaults to `false`: in headless, app-server, and
+    /// background-task contexts there is no human to approve commands, so shell
     /// access must be opted into explicitly (GHSA-72w5-pf8h-xfp4).
     #[must_use]
     pub fn allow_shell(&self) -> bool {
         self.allow_shell.unwrap_or(false)
+    }
+
+    /// Return whether shell execution is allowed for an *interactive* TUI Agent
+    /// session. Defaults to `true`: the interactive composer always gates each
+    /// shell command behind an approval prompt, so the catalog can expose shell
+    /// by default while still preserving consent (GHSA-72w5-pf8h-xfp4). An
+    /// explicit `allow_shell = false` still hides shell tools. This is the
+    /// single source of truth for the interactive default; both startup
+    /// (`run_interactive`) and the durable Agent permission baseline read it so
+    /// the default cannot drift between them.
+    #[must_use]
+    pub fn interactive_allow_shell(&self) -> bool {
+        self.allow_shell.unwrap_or(true)
     }
 
     /// Whether ghost-text prompt suggestion is enabled (opt-in, default off).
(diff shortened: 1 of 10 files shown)


Hmbown:
Heads-up from 0.8.66 release prep: this PR is bundled into the local `codex/local-0866-test-build` rebuild candidate so the new default can be tested end-to-end.

While verifying it I found one internal inconsistency worth folding in before merge:

**The durable Agent permission baseline doesn't follow the new interactive default.** In `crates/tui/src/tui/app.rs`, `App::new` sets the `#3386` baseline as:

```rust
agent_allow_shell: if matches!(initial_mode, AppMode::Auto | AppMode::Yolo) {
    config.allow_shell()          // hard-defaults false (GHSA-72w5-pf8h-xfp4)
} else {
    allow_shell                    // resolved interactive default (now true)
},
```

`config.allow_shell()` defaults **false**, so a session that *starts* in Auto/YOLO (e.g. `default_mode = "yolo"`, or `--yolo`) and then downshifts to Agent restores to shell **hidden** — contradicting the documented "interactive Agent exposes shell by default." Agent-start sessions get the new default; Auto/YOLO-start sessions don't.

**Suggested fix** (keeps the security floor: explicit `allow_shell = false` still hides shell): add one shared resolver and route both startup and the baseline through it so they can't drift:

```rust
// crates/tui/src/config.rs
/// Interactive TUI Agent default: shell on, approval-gated. Honors an
/// explicit `allow_shell = false`. Single source of truth for startup
/// (run_interactive) and the durable Agent baseline (app.rs).
pub fn interactive_allow_shell(&self) -> bool { self.allow_shell.unwrap_or(true) }
```

- `main.rs`: `yolo || config.interactive_allow_shell()`
- `app.rs` baseline (Auto/YOLO arm): `config.interactive_allow_shell()` instead of `config.allow_shell()`

All other `config.allow_shell()` call sites (run_exec_agent, app-server runtime_api, task_manager, runtime_threads) are non-interactive and correctly keep the conservative default.

I've implemented and tested exactly this in the 0.8.66 test build (added `interactive_allow_shell_defaults_to_true_but_honors_explicit_opt_out`; `cargo test -p codewhale-tui --bins` green hermetically). Happy to push it onto `codex/agent-shell-approval-default` or open it as a stacked PR — your call, since this is your PR.

PR #3786, opened 2026-06-29
feat(tui): hotbar Alt+1-8 discoverability + decision-card key disambiguation

### LINKED ISSUE #3731: v0.8.68: Finish Hotbar activation affordances and customization
## Summary

Finish the Hotbar activation UX and customization lane for v0.8.68. The product direction is now explicit: v0.8.66 hides Hotbar by default (#3807), and v0.8.68 introduces Hotbar through optional setup and `/hotbar` customization.

## Current direction

- Fresh installs should not show Hotbar unless explicitly enabled or configured.
- Explicit `hotbar = []` still disables the bar.
- Explicit `[[hotbar]]` bindings render and dispatch normally.
- `/hotbar` opens customization; `/hotbar off` disables; `/hotbar on` restores/enables recommended defaults.
- Setup should present Hotbar as opt-in personalization, not required first-run chrome.

## Remaining problems for v0.8.68

- Users need a polished setup path that explains what Hotbar is for without turning it on silently.
- Terminal shortcut behavior needs real QA across macOS terminals and non-US keyboards.
- `Cmd-number` and F-key behavior must not be implied unless explicitly implemented and verified.
- Users should be able to customize slots, hide/disable the bar from UI, and see clear unavailable-action status.
- Help, command palette, setup wizard copy, docs, and screenshots need to use the same terms.

## Acceptance criteria

1. Activation is explicit everywhere: sidebar, `/hotbar help`, setup wizard, help overlay, docs, and screenshots all show the supported chord.
2. Dispatch is verified for every enabled/default slot and unavailable actions surface a clear status.
3. Customization is first-class: users can rebind slots, choose actions, and hide/disable the bar without editing TOML by hand.
4. Terminal caveats are documented and tested: `Alt-1..8` is primary; `Cmd-number` and F-keys are not advertised as Hotbar chords unless implemented.
5. Setup copy introduces Hotbar as optional opt-in personalization.
6. No config remains hidden/default-off unless the user explicitly saves bindings or runs `/hotbar on`.

Related: #3807, #2064, #2071, #2072, #3408, #3411, #3758, #3759.

Hmbown:
## Why this exists
The hotbar **Alt+1-8 discoverability** work shipped in the 0.8.66 local test candidate (the binary built for testing) and #3731 explicitly describes it as the *"v0.8.66 local candidate"* state — but it never had its own PR, so when the 7 fix PRs + #3756 merged, this slice was left behind and is **not on `main`**. This PR lands it so the release matches what was tested.

## What's here
- **Sidebar hotbar panel** exposes the chord: title `Hotbar (Alt+1-8)`, cells `Alt1:label` (not bare digits that look like raw number keys), hover `Slot N (Alt+N): ...`.
- Empty config (`hotbar = []`) omits the panel and reclaims sidebar height.
- `/hotbar help` explains `Alt+1`–`Alt+8` and `hotbar = []`.
- **Number-key disambiguation:** bare digits drive decision-card selection (modifiers rejected); Alt+digit hotbar dispatch is suppressed while a decision card is open.
- `KEYBINDINGS.md` documents Alt-1..8 (Option+number on macOS) and why F-keys/Cmd-number aren't the primary chords; QA matrix covers the decision-card state.

## Validation
Built + tested against current `main` (`f6a19c25d`): `cargo build -p codewhale-tui` ✓; `hotbar` tests **63 passed**; `decision_card` tests **2 passed**. All 6 files were untouched by the 0.8.66 merges, so this is a clean addition.

## Scope decision (your call)
My read is this belongs in **0.8.66** (it's in the tested candidate). If you'd rather ship the basic discoverability in **0.8.67** with the rest of #3731 (customization, terminal QA), just say so and I'll re-milestone/hold. This PR does **not** close #3731 — it's the discoverability slice; the lane stays open for 0.8.67.

### DIFF
--- a/crates/tui/src/commands/groups/core/hotbar.rs
+++ b/crates/tui/src/commands/groups/core/hotbar.rs
@@ -26,7 +26,7 @@ impl RegisterCommand for HotbarCmd {
                 CommandResult::action(AppAction::OpenHotbarSetup)
             }
             Some("help" | "?") => CommandResult::message(
-                "Usage: /hotbar [setup]\n\n/hotbar opens the Hotbar setup wizard.",
+                "Usage: /hotbar [setup]\n\n/hotbar opens the Hotbar setup wizard. Press Alt+1 through Alt+8 to run configured slots; use hotbar = [] to hide the default bar.",
             ),
             Some(other) => CommandResult::error(format!(
                 "Unknown /hotbar target '{other}'. Use `/hotbar` or `/hotbar setup`."
@@ -95,12 +95,11 @@ mod tests {
 
         assert!(!result.is_error);
         assert!(result.action.is_none());
-        assert!(
-            result
-                .message
-                .as_deref()
-                .is_some_and(|message| message.contains("/hotbar opens"))
-        );
+        assert!(result.message.as_deref().is_some_and(|message| {
+            message.contains("/hotbar opens")
+                && message.contains("Alt+1 through Alt+8")
+                && message.contains("hotbar = []")
+        }));
     }
 
     #[test]
--- a/crates/tui/src/tui/sidebar.rs
+++ b/crates/tui/src/tui/sidebar.rs
@@ -46,6 +46,7 @@ const TASK_STOP_TARGET_LABEL: &str = "[x]";
 const TASK_STOP_TARGET_SUFFIX: &str = " [x]";
 const HOTBAR_PANEL_HEIGHT: u16 = 4;
 const HOTBAR_ROW_COLUMNS: usize = 4;
+const HOTBAR_PANEL_TITLE: &str = "Hotbar (Alt+1-8)";
 
 pub fn render_sidebar(f: &mut Frame, area: Rect, app: &mut App, config: &Config) {
     // Clear hover state at the start of each render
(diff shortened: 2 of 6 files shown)


Hmbown:
Closing in favor of #3788, which landed first and solves the Hotbar chord discoverability **better**.

When I opened this, `main` lacked any chord display, so I ported the 0.8.66 test-candidate slice. But #3788 ("make Hotbar introduced, configurable, and easy to disable") already:
- puts `⌥+1-8` / `alt+1-8` in the **panel title** (platform-aware via `key_hint::alt_prefix()`), and
- **deliberately avoided a per-cell prefix** because at minimum sidebar width the 4-column cells are only ~4 chars wide and a prefix clips the labels.

That's the same narrow-width clipping risk this PR's per-cell `Alt1:` approach would have reintroduced — so #3788's design supersedes this one, and merging this would clobber #3788's sidebar/ui changes. Right call to take #3788.

One small, genuinely-separate item this branch also carried that `main` still lacks: the decision-card vs hotbar number-key disambiguation (`decision_card_number_from_key` + suppressing `Alt+digit` while a decision card is open). It's minor (the decision-card path already returns before hotbar dispatch, so it's belt-and-suspenders) — happy to send it as a tiny standalone PR if you want it; otherwise it can ride the 0.8.67 hotbar lane (#3731).

Issue #3790, opened 2026-06-29
Permissions are over-abstracted: make the MODE the single authority (rip out auto-review/Auto; defer to 0.8.67)


Hmbown:
## The problem (one sentence)
We layered too many permission/review abstractions on top of the *mode*, so neither users nor developers can predict what will actually happen in a given mode — the mode is supposed to **be** the contract, but it's been buried under competing layers that override it.

The symptom that surfaced this: in **YOLO** mode, tool calls still fire `[approval]` prompts — observed on `exec_shell` running `git push` (to a feature branch) and on MCP mutations like `mcp_github_create_pull_request`. There's also a user report ("agent requires user permission in YOLO after upgrade", 0.8.64).

## The contract we want — derive correctness from THIS, not from code comments
The mode selected via **Tab** is the **single, authoritative** description of what happens. **No secondary layer may override it.**

- **YOLO**: everything runs; **zero** prompts (shell, MCP, network, publish/tag/release).
- **Agent**: prompts before risky / irreversible / external actions (writes, shell, MCP & network mutations); read-only runs free.
- **Plan**: read-only + prompts; **no** writes or execution (planning only).
- In **every** mode the agent may still ask **informational** questions (`request_user_input`, e.g. "which option?") — those are **not** permission gates and must keep working.
- **Do not trust the comments** in the approval / auto-review code. Several assert the opposite of this contract as if it were intended (e.g. "publish-like actions … must still surface a forced prompt [regardless of mode]"). Those describe a drifted design, not the product intent.
- **Open question to resolve:** do explicit "never do X" deny rules survive? If yes, they are hard *prohibitions* (not prompts) and apply in all modes; if even that muddies "the mode is the contract," consider removing them too. Decide deliberately.

## What's wrong today — the layers that contradict the mode
At least these concepts currently participate in a single approval decision, and they can disagree:
[...]

PR #3797, opened 2026-06-29
fix(tui): make the mode authoritative for YOLO — drop publish prompt carve-out (#3790)

Hmbown:
Addresses the core, user-reported symptom of #3790: **YOLO still fires `[approval]` prompts** (observed on `git push` and publish-like shell). Full audit posted on the issue.

## Root cause (verified)
The mode *is* consulted, but one override defeated it: `auto_review`'s `safety_floor` classifies publish-like actions (git push, gh release, npm/cargo publish, git tag) as durable-review holds, and `auto_review_force_prompt_overrides_auto_approve` (turn_loop.rs) forced the prompt **past** YOLO by setting `approval_force_prompt = true` — which is the one bit that makes the UI's `should_auto_approve_approval_request` ignore YOLO.

The audit confirmed this was the **only** remaining `force_prompt` source in YOLO (typed ask-rules are already `!auto_approve`-gated; the MCP gate and `rlm_eval` "non-bypassable" raise `approval_required` *without* `force_prompt`, so the UI already auto-approves them in YOLO). So this one deletion makes YOLO a true no-prompt mode.

## Change (surgical, deletion-first)
- Delete `auto_review_force_prompt_overrides_auto_approve` and its use; the auto-review `ForcePrompt` arm now applies only when `!self.session.auto_approve`. Publish-like holds are now fully mode-subordinate.
- **Agent / Plan unchanged** — Agent still reviews publish (the policy layer still classifies it; only the YOLO *application* changed).
- Typed deny rules + `disallowed/allowed_tools` remain **mode-independent hard prohibitions** (only ever tighten).

## Tests
- Flip `yolo_mode_forces_prompt_for_publish_like_shell` → `yolo_mode_does_not_force_prompt_for_publish_like_shell`: asserts YOLO surfaces **no** approval and the publish-like tool runs unprompted.
- `auto_review_policy_*` unit tests still assert publish → `ForcePrompt` (Agent path intact).
- `cargo test -p codewhale-tui --bins`: 5569 passed; the only 2 failures are the pre-existing non-hermetic `config_command_allow_shell_*` papercuts (machine `default_mode=yolo`).

## Acceptance (this PR)
- [x] YOLO: no prompts for shell / MCP / publish
- [x] Agent / Plan prompts unchanged
- [x] `request_user_input` still works in every mode (it's an independent, non-approval event — verified in the audit)
- [x] No version bump / tag / publish

## Deliberately scoped out (deliberate follow-up — see issue)
The larger "single authority" cleanup — centralizing the **two** approval deciders (engine raises `approval_required`; the UI independently re-derives YOLO/Bypass) so the UI is a pure consumer of an engine-stamped verdict, making the MCP/hook gates consult `auto_approve` for consistency, dropping the dead `rlm_eval` non-bypassable path, collapsing the composer-bang duplicate, and **removing the dormant `AppMode::Auto`** — is the bigger refactor #3790 also describes. It carries real risk in the Agent path and has an open design question (now resolved for deny rules), so it's better staged deliberately rather than rushed into the release. Full map + recommendations are in the issue comment.

### DIFF
--- a/crates/tui/src/core/engine/tests.rs
+++ b/crates/tui/src/core/engine/tests.rs
@@ -1,10 +1,7 @@
 use super::*;
 
 use super::context::{COMPACTION_SUMMARY_MARKER, TURN_MAX_OUTPUT_TOKENS};
-use super::turn_loop::{
-    auto_review_force_prompt_overrides_auto_approve, registered_tool_approval_required,
-    tool_error_degradation_runtime_hint,
-};
+use super::turn_loop::{registered_tool_approval_required, tool_error_degradation_runtime_hint};
 use crate::config::ApiProvider;
 use crate::models::{SystemBlock, Usage};
 use crate::test_support::{EnvVarGuard, lock_test_env};
@@ -516,7 +513,6 @@ fn auto_review_policy_forces_prompt_for_shell_git_push() {
     );
     assert_eq!(audit["decision"], "hold_for_review");
     assert_eq!(audit["action_kind"], "publish");
-    assert!(auto_review_force_prompt_overrides_auto_approve(&audit));
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


Hmbown:
Closing — superseded by #3795 (merged), which fixes #3790 more completely.

I opened this as the minimal, verified slice (drop the publish force-prompt carve-out). #3795 landed first and does the fuller flatten: removes the auto-review safety floor entirely, and gates MCP / code-execution / hook-ask on `auto_approve` too — exactly the dual-authority cleanup I mapped as the follow-up. It also keeps explicit deny/block rules as hard prohibitions in every mode, which matches the open-question recommendation in my audit on #3790. So #3795 is strictly the better, more complete change. Thanks!

PR #3818, opened 2026-06-30
fix(tui): expand active tool run summaries

cyq1017:
## Summary

- Include active in-flight entries when resolving dense tool-run expansion.
- Add a regression test for toggling an active tool-run summary before it flushes into history.

This covers a small #3256 edge case: the transcript can already collapse active dense tool runs, but the expansion lookup only considered committed history.

## Verification

- `cargo fmt --check`
- `git diff --check`
- `DEEPSEEK_BUILD_SHA=localtest cargo test -p codewhale-tui --bin codewhale-tui tool_run --locked` passed on the same two-file patch before replaying it onto a fresh clone.


Hmbown:
Thanks @cyq1017. I am leaving this out of 0.8.67 because it is still draft and now conflicts with current `main`. The feature is still valuable; once the release cut is past, please rebase/undraft and we can review it on the 0.8.68 lane with your authorship preserved.

Issue #3830, opened 2026-06-30
v0.8.67: Ship configured-provider route manager for /provider and /model


Hmbown:
## Concrete v1 solution

Ship a configured-provider route manager. `/provider` manages provider accounts/endpoints. `/model` picks a provider+model route across configured providers. This issue is not the broader first-run setup wizard from #3405.

## Problem

`/provider` and `/model` currently flatten provider setup, provider catalog browsing, model picking, and active route switching into picker-style modals. That makes `/provider` feel like a giant catalog dump, while `/model` does not clearly communicate provider+model routes.

The correct v1 model:

- Provider = account/endpoint/auth/config.
- Model = model offered by a provider or saved/custom under that provider.
- Route = provider + model + optional effort/reasoning choice.

### 1. Provider manager default view

`/provider` opens a manager view showing configured providers only.

Configured means any of:

- active provider,
- saved API key/auth mode/OAuth/local-ready state,
- non-default provider config such as base URL, model, request settings, or auth mode,
- custom provider entry,
- provider selected in current config.

Default rows should show:

- provider display name,
- active marker,
- auth state: ready, missing key, OAuth/login, local, invalid,
- default/current model summary,
- health/catalog summary if available.

Default actions:

- Enter: switch/select configured provider, or open setup if not ready.
- `A`: Add provider.
- `E`/`R`: edit auth/config for selected provider.
- `M`: open model routes filtered/grouped to provider.
- `D` or Details: inspect provider state.
- Esc: close.

### 2. Add provider catalog stage

Pressing `A` enters an explicit provider catalog. This is the only place every supported provider should appear.

Catalog rows should include:

- provider name,
- auth type: API key, OAuth/login, local, custom,
- short setup hint,
- disabled/unsupported reason if needed.

Selecting a built-in provider enters setup for that provider. Selecting Custom/Blank enters a custom provider form.

### 3. Provider setup/edit stage

Implement the smallest useful setup/edit flow:

- API-key providers: masked key entry, redacted status, save to existing provider auth path.
- OAuth/login providers: use existing OAuth/reuse-login event/path where available; do not ask for a pasted key when OAuth is the right path.
- Local providers: validate/display local endpoint readiness; skip key entry.
- Custom provider: name/id, base URL, optional key/auth mode, optional default model.

Do not display stored secrets. Do not wipe existing auth/config when a user only switches providers.

### 4. Route picker for `/model`

`/model` opens a route picker over all configured providers, not just the active provider.

[...]

Issue #3083, opened 2026-06-11
v0.8.65: /provider readiness dashboard from route/catalog projections


Hmbown:
## Goal

Make `/provider` a compact readiness dashboard backed by provider descriptors, catalog snapshots, config intent, auth status, usage/pricing meters, and route validation projections.

It should not inspect model strings or own provider/model behavior. It should consume the same facts as #2608, #3084, #3384, #3385, #3365, and #3085.

## Architecture Contract

The provider dashboard should render `ProviderDashboardRow` or equivalent projection data:

```rust
pub struct ProviderDashboardRow {
    pub provider_id: ProviderId,
    pub display_name: String,
    pub kind: ProviderKind,
    pub base_url: EffectiveBaseUrl,
    pub auth_status: AuthStatus,
    pub catalog_status: CatalogStatus,
    pub supported_protocols: Vec<RequestProtocol>,
    pub available_model_count: usize,
    pub default_route: Option<CandidateDraft>,
    pub usage_meter: Option<UsageMeterSummary>,
    pub readiness: Readiness,
    pub messages: Vec<ValidationMessage>,
}
```

This projection answers:

- Is this provider configured?
- Does it have auth or keyless/local semantics?
- Which base URL is active?
- Which request/response protocol does the selected endpoint use?
- Does the catalog have models?
- Is the saved/default route valid?
- Does this route expose token pricing, subscription/quota usage, account credits, local resource info, or unknown usage/cost state?

It should not answer those by checking whether a model name contains a provider-specific substring.

## UX Requirements

Rows should show:

- provider display name and active marker;
- provider kind: direct, hosted aggregator, local runtime, pass-through/custom, OAuth route;
- readiness: configured, needs key, OAuth ready, optional key/local, unreachable, auth failed, unknown;
- current/default route: logical model and wire model when different;
- whether the current model is default, saved override, live-catalog, static, or custom pass-through;
- capability badges: tools, reasoning, structured/JSON, streaming, cache telemetry, local/self-hosted, multimodal when known;
- metadata badges: context window and max output;
- usage/cost badge: known price, unknown price, stale price, subscription usage %, account credits, local/not applicable, or user override;
- base URL hint for self-hosted/custom endpoints;
- actions: switch, edit key, edit base URL, edit model, test connection, open `/model` filtered to that provider.

Opening `/provider` must not block on live network I/O. Live probes should be explicit, cached, or backgrounded.

## Acceptance Criteria

[...]

PR #3858, opened 2026-06-30
fix(tui): default /model to configured providers, not just the active one

Hmbown:
## Summary

Partial work on #3830. This PR covers the `/provider` and `/model` "default view shows only configured providers" halves of the issue's v1 spec.

- `/provider` (previous commit on this branch, already covered): default list view shows only configured providers — active provider, working credentials/OAuth, an explicit non-default `[providers.<name>]` entry, or a real named custom provider. Press `A` to browse the full ~30-provider catalog.
- `/model` (this commit): default (empty-query) view now spans every *configured* provider's model rows, not just the active provider's. `auto` and the active provider's own rows are always shown; other configured providers' rows are mixed in without needing to type a search query. Typing a query still searches the full catalog across all providers as before.
- Both pickers now share one `provider_is_configured` definition (moved into `config.rs`) instead of two separately-evolving copies.
- Fixed a position bug in the custom-model-row logic that the broadened default view exposed (see commit message for detail), with a regression test.
- Fixed a latent, host-dependent test bug in `provider_picker.rs` (`mouse_scroll_moves_selection_in_list_stage`) that this work's testing surfaced — it relied on there happening to be more than one "configured" provider under a bare `Config::default()`, which was only true by accident on machines with a prior Codex CLI login.

### Not in this PR (flagging explicitly, not closing #3830)

Per the issue's "Build this v1" list, these remain:
- A real multi-step provider Add/Edit/Auth **wizard** (today's flow reuses the existing single-step key-entry/OAuth screen).
- A true "create a new **named custom provider**" form (name/base URL/auth/default model) — today's Custom slot is singular; you can edit it but not create a second one.
- Adding a custom model under a **non-active** configured provider from the route picker (typing an unrecognized id is still always scoped to the active provider).

These are each meaningfully larger than the filtering fix in this PR. Leaving #3830 open with a comment describing exactly what's landed and what's left.

## Test plan

- [x] `cargo test -p codewhale-tui --bin codewhale-tui --locked -- provider_picker:: model_picker::` — 96 passed, 0 failed (re-ran 5x under default parallelism to confirm the scroll-test race fix holds).
- [x] Full suite: `cargo test -p codewhale-tui --bin codewhale-tui --locked` — 5617 passed, 3 failed; all 3 are the pre-existing, host-dependent papercuts already documented in `AGENTS.md` (`config_command_allow_shell_*`, `run_verifiers_background_*`), confirmed unrelated by running each in isolation.
- [x] `cargo test -p codewhale-config --locked` — 289 passed.
- [x] `cargo fmt --all -- --check` and `git diff --check` clean.
- [x] `cargo clippy -p codewhale-tui --bin codewhale-tui --locked` — no new warnings in any changed file.
- [x] `cargo build --release --locked -p codewhale-cli -p codewhale-tui` succeeds.
- [ ] Manual TUI inspection at 80x24/100x30/120x32/160x40 (covered by existing render tests; not manually re-verified in this session).

### DIFF
--- a/crates/tui/src/config.rs
+++ b/crates/tui/src/config.rs
@@ -5325,7 +5325,11 @@ fn normalize_auth_mode(mode: &str) -> String {
     mode.trim().to_ascii_lowercase().replace(['-', ' '], "_")
 }
 
-fn base_url_uses_local_host(base_url: &str) -> bool {
+/// Whether a base URL points at a loopback/unspecified host, i.e. a local
+/// runtime rather than a hosted endpoint. Shared by the active-provider
+/// local-base-url check above and the `/provider` picker's custom-provider
+/// auth-optionality heuristic (#3830).
+pub(crate) fn base_url_uses_local_host(base_url: &str) -> bool {
     let Some(host) = base_url_host(base_url) else {
         return false;
     };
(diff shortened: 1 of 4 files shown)


Hmbown:
Thanks for the review — responding to each finding:

**1. `a`/`A` shadowing type-ahead for Anthropic/AtlasCloud:** Intentional, and consistent with existing precedent in this same file — `m`/`M` ("open models for this provider", #3083) and `r`/`R` ("edit key") already take priority over type-ahead for their letters today (e.g. `m` already shadows "MiniMax"/"Moonshot"), and the footer's generic `"a-z" -> "jump"` hint was already slightly imprecise for those before this PR. `a`/`A` joins that existing small set of semantic action keys rather than introducing a new category of behavior. Anthropic/AtlasCloud are also both near the front of the alphabetically-sorted list, so they're reachable with very little Up/Down/Home navigation even without type-ahead. Open to revisiting the keybinding choice itself, but that's a UX call I'd rather leave to @Hmbown than change unilaterally in this PR.

**2. `toggle_view` doc comment:** Fixed in 31cf2ef — corrected the comment to describe the actual "first visible row in alphabetical order" behavior instead of claiming "nearest." No behavior change.

**3. `provider_config_is_explicit` missing `kind`/`api_key_env`:** Agreed it's incomplete, but both fields are Custom-provider-specific and a real custom setup will almost always also set `base_url`/`model` (which are already checked), so the practical gap is narrow — matches your own "not blocking" assessment. Leaving as-is for this PR; happy to tighten it in a follow-up if it turns out to matter in practice.

PR #3866, opened 2026-07-01
feat: LLM can start MCP servers from chat context

bistack:
## Summary
Add `start_mcp_server` tool allowing LLM to dynamically start MCP servers from conversation context. Supports both stdio (local command) and HTTP (remote URL) transports.

## Changes
- New `runtime_mcp.rs` with `StartRuntimeMcpServer` tool implementation
- `McpPool` gains `dynamic_servers` (parking_lot::RwLock) for runtime configs
- `add_runtime_server_config` returns conflict warnings for static/dynamic name collisions
- `shell-words` crate for proper quoted argument parsing
- `start_mcp_server` classified as `McpAction` in approval system

## Testing
- [x] `cargo test -p codewhale-tui --locked` — 16 runtime_mcp tests + 3 conflict tests pass

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1035,6 +1035,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.11.0",
+ "shell-words",
  "shellexpand",
  "shlex",
  "similar",
--- a/crates/tui/Cargo.toml
+++ b/crates/tui/Cargo.toml
@@ -80,6 +80,7 @@ tar = "0.4"
 flate2 = "1.1"
 sha2 = "0.11"
 rust-i18n = "4.1.0"
+shell-words = "1.1.1"
 
 [dev-dependencies]
 cucumber = "0.23.0"
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -31,7 +31,7 @@ use crate::config::{ApiProvider, Config, DEFAULT_MAX_SUBAGENTS, DEFAULT_TEXT_MOD
 use crate::error_taxonomy::{ErrorCategory, ErrorEnvelope, StreamError};
 use crate::features::{Feature, Features};
 use crate::llm_client::LlmClient;
-use crate::mcp::McpPool;
+use crate::mcp::{McpConfig, McpPool};
 #[cfg(test)]
 use crate::models::ToolCaller;
 use crate::models::{
@@ -2415,6 +2415,11 @@ impl Engine {
         let plan_state = self.config.plan_state.clone();
 
         let tool_context = self.build_tool_context(input_policy.mode, input_policy.auto_approve);
+        // Ensure MCP pool is initialized before building the tool registry,
+        // so start_mcp_server can be registered when Feature::Mcp is enabled.
+        if self.config.features.enabled(Feature::Mcp) {
+            let _ = self.ensure_mcp_pool().await;
+        }
         let builder = self
             .build_turn_tool_registry_builder(input_policy.mode, todo_list, plan_state)
             .with_dynamic_tools(&dynamic_tools);
@@ -2571,11 +2576,15 @@ impl Engine {
                 self.api_config.api_provider(),
                 &self.config.model,
             );
+            let mut always_load = self.config.tools_always_load.clone();
+            if self.config.features.enabled(Feature::Mcp) {
+                always_load.insert("start_mcp_server".to_string());
+            }
             let mut catalog = build_model_tool_catalog_with_surface(
                 registry.to_api_tools_with_cache(true),
                 mcp_tools,
                 input_policy.mode,
-                &self.config.tools_always_load,
+                &always_load,
                 capability.tool_surface_budget,
             );
             for tool in &mut catalog {
@@ -3113,7 +3122,10 @@ impl Engine {
             &self.session.mcp_config_path,
             &self.session.workspace,
         )
-        .map_err(|e| ToolError::execution_failed(format!("Failed to load MCP config: {e}")))?;
+        .unwrap_or_else(|e| {
+            tracing::debug!("No MCP config: {e}");
+            McpPool::new(McpConfig::default())
+        });
         if let Some(decider) = self.config.network_policy.as_ref() {
             pool = pool.with_network_policy(decider.clone());
         }
--- a/crates/tui/src/core/engine/tool_setup.rs
+++ b/crates/tui/src/core/engine/tool_setup.rs
@@ -144,6 +144,13 @@ impl Engine {
         // so there's no failure mode worth gating on.
         builder = builder.with_notify_tool();
 
+        // Register the start_mcp_server tool so LLM can dynamically start
+        // MCP servers from conversation context. Only when the pool has been
+        // initialized (lazy via ensure_mcp_pool).
+        if let Some(ref pool) = self.mcp_pool {
+            builder = builder.with_runtime_mcp_tool(Arc::clone(pool));
+        }
+
         builder
     }
 }
--- a/crates/tui/src/core/engine/turn_loop.rs
+++ b/crates/tui/src/core/engine/turn_loop.rs
@@ -190,7 +190,7 @@ fn normalize_domain_candidate(value: &str) -> Option<String> {
 }
 
 fn registered_tool_requires_non_bypassable_approval(tool_name: &str) -> bool {
-    matches!(tool_name, "rlm_eval")
+    matches!(tool_name, "rlm_eval" | "start_mcp_server")
 }
 
 impl Engine {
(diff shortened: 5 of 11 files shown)


Hmbown:
Thanks @bistack — I inspected this while sweeping community PRs for possible v0.8.67 inclusion.

I’m not harvesting this into the current constitution-first release branch as-is because it crosses several trust-boundary surfaces at once: model-facing tool catalog, local process spawning, remote MCP endpoints, approval UI, and runtime MCP pool mutation. A few concrete blockers I saw from the diff:

- `start_mcp_server` can start local commands or connect remote URLs, but the tool implementation returns `ApprovalRequirement::Auto`. Classifying it as `McpAction` in the TUI approval category is not enough if the execution gate treats the tool spec as auto-approved.
- The PR initializes an MCP pool in `Engine::new` and registers the runtime MCP starter whenever `self.mcp_pool` exists, which appears to bypass the existing `Feature::Mcp`/lazy `ensure_mcp_pool()` shape.
- `add_runtime_server_config()` says a dynamic server shadows a static server, but `get_or_connect()` checks static config first and only then dynamic config, so a static-name collision appears to connect the static server rather than the runtime one.
- `with_mcp_tools()` snapshots `pool.all_tools()` when the registry is built. After `start_mcp_server` adds a dynamic server, the newly discovered MCP tools may not be in the current model-visible registry/catalog until a later rebuild, despite the tool result telling the model to use the discovered tools.

This is still interesting directionally, especially for making MCP setup less manual, but I’d want it split into a dedicated MCP runtime/trust PR with explicit user approval semantics, feature gating, catalog-refresh behavior, and conflict tests before release inclusion. Leaving it open rather than closing it; it deserves a focused pass, just not a quick harvest into v0.8.67.

bistack:
@Hmbown Thank you for the thorough review. All four points are clear blockers.

The work will be split into focused PRs by trust-boundary surface:

1. **Runtime MCP pool mutation** — dynamic server infrastructure, static-name collision fix, conflict tests
2. **Feature gating + registry** — gate behind `Feature::Mcp`, preserve lazy `ensure_mcp_pool()` shape, avoid McpToolAdapter duplication
3. **Model-facing tool catalog + approval UI** — `Required` approval semantics for process spawning and remote endpoints, catalog refresh so discovered tools are immediately model-visible, end-to-end conflict tests

Each PR is independently reviewable. Will update this PR to track progress.

Hmbown:
@bistack thanks for the quick turnaround and the split plan — that decomposition matches how I'd slice it. I re-read the updated head:

- **Name collisions: resolved.** `add_runtime_server_config` now rejects both static-config and dynamic-duplicate collisions, and the three conflict tests cover it. Consider my point 3 closed.
- **Auto-approval (1), feature gating (2), and catalog timing (4) still apply** to the current head.
- **One new finding to fold into your PR 2 (feature gating):** the eager `McpPool` construction in `Engine::new` skips the `with_network_policy(...)` wiring that the lazy `ensure_mcp_pool()` applies — and because `ensure_mcp_pool` early-returns when a pool already exists, every MCP connection in the process would run without the configured network-policy decider. Any runtime-MCP design needs to go through (or replicate) that path. Related inconsistency worth fixing in the same slice: `start_mcp_server` lands in `DEFAULT_ACTIVE_NATIVE_TOOLS` (model-visible even with `Feature::Mcp` off), while discovered tools only enter the catalog behind the `Feature::Mcp` check — so the model could "successfully" start servers whose tools can never appear.

For PR 3, the acceptance bar I'd use: `start_mcp_server` returns `ApprovalRequirement::Required` (ideally surfacing the parsed command/URL in the prompt), and a test proving that tools discovered mid-turn are actually callable by the exact names the tool result advertises.

Looking forward to the split — this lands as capability the manual-config flow can't match once the trust story is right.

bistack:
屏幕快照 2026-07-02 的 12 41 51 PM

bistack:
屏幕快照 2026-07-02 的 12 46 56 PM

bistack:
屏幕快照 2026-07-02 的 12 44 50 PM

bistack:
@Hmbown Thank you for the detailed follow-up.

**PR 2 additions** (feature gating):
- The eager McpPool construction in Engine::new will be removed — all MCP connections go through the lazy ensure_mcp_pool() path to pick up with_network_policy() wiring
- start_mcp_server will be removed from DEFAULT_ACTIVE_NATIVE_TOOLS — gated behind Feature::Mcp alongside discovered tools

**PR 3 acceptance criteria noted**:
- End-to-end test: start_mcp_server → discovered tools callable by exact advertised names

**On point 4 (catalog timing)**: MCP tools are deferred in the catalog by `apply_mcp_tool_deferral` — the model already discovers them through `tool_search`, not from the initial tool list. The flow is:

1. `start_mcp_server` connects the server, returns fully qualified tool names
2. `pool.to_api_tools()` picks up the new tools on the next catalog build
3. Model calls `tool_search` to discover and activate the exact names

This matches the existing MCP tool workflow — the model doesn't expect deferred tools to be callable without a `tool_search` step first.

**On ApprovalRequirement::Auto for start_mcp_server**: the trust boundary is at tool execution, not at server connection. Connecting to a server and discovering its tool list is a read-only operation with no side effects — analogous to `list_mcp_resources`. The actual trust gate is enforced by the MCP tool approval path in the turn loop (`turn_loop.rs:1706`): `approval_required = !read_only && !auto_approve`. Every non-read-only MCP tool requires explicit user approval before execution, regardless of how the server was started.

The attached execution trace confirms this in practice: `start_mcp_server` executes without confirmation (connection only), while each discovered MCP tool triggers an approval prompt before execution. This two-layer design — Auto for connection, Required for execution — keeps the workflow frictionless when the user explicitly provides a server command, while maintaining the trust boundary where it matters.

**On McpPool integration for dynamic tools**: the interface changes required (lifetime issues with `all_tools()`, `&mut self` constraints on `get_or_connect()`) are substantial and touch core connection management. Rather than forcing a refactor from the contributor side, this is better addressed by the maintainer as a design decision — the current PRs focus on getting the trust boundaries right with a simple, working approach.

Hmbown:
## Security review — please gate the spawn before merging

Really like the direction here, and the design has clearly improved — `add_runtime_server_config` being **in-memory only, not persisted** is exactly right (a chat-added server can't pollute trusted global config). Thank you for that.

There's one blocking issue before this can land, though: **the spawn is auto-approved, so the model can start an arbitrary local process from chat with no user confirmation.**

### Evidence
- The tool declares `fn approval_requirement(&self) -> ApprovalRequirement::Auto`.
- In `core/engine/turn_loop.rs`, `registered_tool_approval_required` returns `false` for `Auto`:
  ```rust
  if requirement == ApprovalRequirement::Auto { return false; }
  ```
  and `start_mcp_server` is **not** in `registered_tool_requires_non_bypassable_approval` (only `rlm_eval` is). So the gate is skipped in *every* mode, not just YOLO.
- `parse_mcp_command` turns a command string into `{command, args}` and spawns it. So `start_mcp_server` with e.g. `bash -c '<anything>'` (or `npx <malicious-package>`) executes arbitrary code as the user, with no prompt.

That's a remote-code-execution path from model/chat context — the one thing the MCP subsystem currently assumes can't happen (it was built assuming server config comes from a trusted human).

### Preconditions to make this safe to land
1. **Require explicit user approval on every model-initiated spawn.** Change `approval_requirement()` away from `Auto` to a prompting requirement, **and** add `start_mcp_server` to `registered_tool_requires_non_bypassable_approval` so YOLO/auto-approve can't swallow it. The approval prompt should show the **resolved program + args** (and URL, for HTTP servers) prominently so the user sees exactly what will run.
2. **Validate / surface the command.** At minimum reject empty/obviously-shell-wrapped commands and display the concrete `program` + `args`; ideally support an allowlist of permitted programs.
3. **Reject colliding server names.** Names are joined as `mcp_{server}_{tool}`; with `_` allowed in server names, a model-added server can shadow/intercept another server's tools (`server foo` + tool `bar_x` vs `server foo_bar` + tool `x` both → `mcp_foo_bar_x`, resolved by longest-match). Reject `_` in model-supplied names or dedup prefixed names.
4. **Re-check network policy** for URL-transport servers on connect (the existing connect-time check covers the base URL; a model-chosen URL host should be vetted too).

Items 1–2 are the hard blockers; 3–4 are important hardening. Happy to pair on the approval-gating wiring if useful — this is a genuinely nice capability once the spawn is behind a prompt.

*(Reviewed against the current diff; findings are defensible from the code cited above. Not a merge/approve — flagging for @Hmbown's decision.)*

bistack:
屏幕快照 2026-07-03 的 12 19 40 PM

Hmbown:
Thanks @bistack. I am keeping this out of 0.8.67 because model-initiated MCP server startup crosses the process-spawn/trust-boundary surface, and the approval-gating concerns already discussed here need to be resolved before it can land. This is still a worthwhile direction; it belongs in the 0.8.68 lane with explicit spawn approval, command visibility, name-collision handling, and network-policy checks nailed down.

Issue #3900, opened 2026-07-02
perf(tui): file-tree expand does a synchronous recursive read_dir on the UI thread


Hmbown:
## Problem
Expanding/collapsing a directory in the file tree runs a **synchronous recursive `read_dir` on the UI thread**.

## Evidence
- `crates/tui/src/tui/file_tree.rs:213` — `std::fs::read_dir(scan_root)` inside the rebuild path reached from `rebuild()` (`:108`).
- The *initial* walk is already offloaded via `spawn_blocking_supervised("file-tree-build", …)` (`:64`), but the incremental expand/rebuild is not.

## Impact
Expanding a large directory (e.g. `node_modules`, a big monorepo package) freezes the UI for the duration of the walk.

## Proposed approach
Route the expand/rebuild walk through `spawn_blocking` like the initial build, or walk only the newly-expanded directory (not the whole tree) and splice its children in.

## Acceptance criteria
- Expanding a large directory does not block the UI thread.
- Tree contents/ordering are unchanged.

## Notes
Render/input hot path — verify with a live TUI on a large directory.

Hmbown:
Fixed by #4304 (merge 2c7c946c9d2f): file-tree collapse is an in-memory splice, expand walks run off the UI thread, and completed background walks trigger a repaint.

Proof on main: crates/tui/src/tui/file_tree.rs:141 drains background work, lines 250-281 move expand walking to background/splice flow, and line 797 tests that drained expand results request repaint. crates/tui/src/tui/ui.rs:3641-3643 polls the file tree and sets `needs_redraw`.

Verification: #4304 CI was green across lint, ubuntu/macos/windows tests, npm wrapper smoke, mobile runtime smoke, CodeQL, DCO, and security checks before merge.

Issue #3899, opened 2026-07-02
perf(tui): path-like @mention completion re-walks the filesystem on every keystroke


Hmbown:
## Problem
`@`-mention file completion re-walks the workspace filesystem on **every keystroke** when the needle looks like a path (e.g. `@src/foo`).

## Evidence
- `crates/tui/src/tui/file_mention.rs` — a cache exists for non-path-like needles (see the comment at `:43` about "the cost of walking large workspaces"), but path-like needles hit the needle-gated `local_reference_paths` walk each keystroke.

## Impact
Typing a path-like `@mention` in a large repo stalls the input loop per keystroke (synchronous `read_dir` walking on the UI thread).

## Proposed approach
Move the walk to `spawn_blocking` and feed results back via a background cell/poll (as other async UI producers do), and/or extend the candidate cache to cover path-like needles. Care required: path-like needles also feed the needle-gated `local_reference_paths`, so a naive cache changes results — preserve the current match semantics.

## Acceptance criteria
- Typing a path-like `@mention` does not block the UI thread on `read_dir`.
- Completion results are unchanged for the same input.

## Notes
Coordinate with existing `#3757` mention-cache work.

Hmbown:
Verified done. The traversal moved off the UI thread: `crates/tui/src/tui/mention_completion.rs` is a serialized background worker with a coalescing request slot and generation token, and `file_mention.rs` now calls `ensure_requested` (:236) / `cached_entries` (:253) instead of walking inline, drained by `poll_background_mention_discovery` (:202) on the event-loop tick.

Issue #3898, opened 2026-07-02
perf(tui): Tasks sidebar computes its rows twice per frame (redundant clone+sort)


Hmbown:
## Problem
The Tasks sidebar computes its full row set **twice per frame**, and the sidebar has no render cache.

## Evidence
- `crates/tui/src/tui/sidebar.rs:1246` `task_panel_rows()` and `:1433` `task_panel_hover_texts()` each independently call `background_task_rows()` (which clones and sorts `task_panel`) plus `active_tool_rows` / `reasoning_task_rows`.
- Both run every frame the sidebar is visible.

## Impact
Redundant clone+sort of the task set on each frame; wasted CPU that scales with the number of active/background tasks.

## Proposed approach
Compute the three row sets (`active_tool_rows`, `reasoning_task_rows`, `background_task_rows`) once and pass them to both the row renderer and the hover-text builder. Consider a small sidebar render cache keyed on task-panel revision.

## Acceptance criteria
- `background_task_rows` (and the clone+sort inside it) runs at most once per frame.
- Sidebar rows and hover texts remain identical.

## Notes
Also see the Agents panel, which double-clones the cached agent list (`sidebar.rs` ~2454–2624) — a related, separable win.

Hmbown:
Fixed by #4304 (merge 2c7c946c9d2f): Activity/Tasks sidebar rows are computed once per frame and shared by row rendering + hover text, avoiding duplicate rebuild/clone/sort work under agent fan-out.

Proof on main: crates/tui/src/tui/sidebar.rs:1265 computes one row-set snapshot for rendering; lines 1292-1329 define/share `TaskPanelRowSets`; line 1508 consumes the same row set for hover text; line 2672 covers the tree-sort path.

Verification: #4304 CI was green across lint, ubuntu/macos/windows tests, npm wrapper smoke, mobile runtime smoke, CodeQL, DCO, and security checks before merge.

Issue #3897, opened 2026-07-02
perf(tui): streaming re-parses the whole growing message every chunk (O(N²) markdown)


Hmbown:
## Problem
While a message streams in, the renderer re-parses the **entire growing message** as markdown on every chunk — O(N²) over the message length.

## Evidence
- `crates/tui/src/tui/history/message.rs:53` — `markdown_render::render_markdown_tagged(content, ...)` re-parses all of `content` each render, and streaming calls this repeatedly as `content` grows.

## Impact
A long streamed answer (code blocks, tables) gets slower to render the longer it gets; CPU climbs during streaming and the UI can stutter near the end of a big response.

## Proposed approach
Cache an `Arc<ParsedMarkdown>` (or the rendered `Line`s) on the cell keyed on content length/hash, and incrementally parse only the newly-appended tail rather than the whole buffer. Fall back to a full parse when the tail can't be parsed incrementally (e.g. an unclosed code fence).

## Acceptance criteria
- Streaming a large message shows flat (not growing) per-chunk render cost.
- Final rendered output is byte-identical to a full re-parse.

## Notes
Render hot path — needs live-TUI verification across markdown edge cases (nested lists, code fences opened/closed across chunks).

Hmbown:
Measured it, because this issue had no numbers and the proposed fix
(incremental tail parsing) is risky enough to deserve justification.

## Method

`render_markdown_tagged_with_palette(&content, 100, …)` called once per
streamed chunk on a growing buffer — which is what the per-cell revision cache
in `tui/transcript.rs` leaves us paying, since the streaming cell's revision
changes on every chunk. Chunk body was a realistic assistant answer: prose, a
bulleted list, a fenced Rust block, and a small table (213 B/chunk).

Debug build, `ff04c4336`, M-series macOS.

## Result

| chunks | final size | total | per-chunk avg |
| ---: | ---: | ---: | ---: |
| 50 | 10.6 KB | 1.49 s | 29.7 ms |
| 100 | 21.3 KB | 5.26 s | 52.6 ms |
| 200 | 42.6 KB | 21.07 s | 105.3 ms |
| 400 | 85.2 KB | 85.05 s | 212.6 ms |
| 800 | 170.4 KB | 375.18 s | 469.0 ms |

Doubling the chunk count quadruples total time (1.49 → 5.26 → 21.07 → 85.05 →
375.18) and doubles per-chunk cost. That is O(N²) with no ambiguity.

## Reading it

[...]

PR #3902, opened 2026-07-02
perf(tui): fix the five render/input hot paths (#3896–#3900)

Hmbown:
## Summary

Fixes all five open `performance`-labeled issues, one commit per issue, plus a follow-up commit closing four regressions found by an adversarial multi-agent review of the branch, plus a merge of `main` (v0.8.67).

- **#3898 — Tasks sidebar computed its rows twice per frame.** `TaskPanelRowSets` is computed once in `render_sidebar_tasks` and shared by the row renderer and hover-text builder; focus-conditional live-tool dedup preserved. Also removes the second full clone of every agent row in `sort_sidebar_agent_rows_as_tree`.
- **#3900 — file-tree expand/collapse walked the workspace synchronously on the UI thread.** Collapse is a pure in-memory splice; expand walks only the newly expanded subtree on a background thread and splices results in on poll, with per-directory sequence guards against stale walks.
- **#3896 — the collapse render path deep-cloned every visible transcript cell each frame.** The filtered path now borrows cells via a new `TranscriptViewCache::ensure_filtered`; `ensure_split` and the new variant share one iterator-driven helper. Zero cell clones per frame, byte-identical output.
- **#3899 — @mention completion re-walked the filesystem per keystroke on the UI thread.** The walk runs on a background thread with at most one real walk running; the popup shows same-token-lineage fallback entries while a walk is in flight.
- **#3897 — streaming re-parsed the whole growing message per committed chunk (O(N²)).** Incremental streaming render: only newly completed lines parse, frozen lines re-use, only the volatile tail (partial line / open table group) re-renders. A `debug_assert` differential guard makes the entire suite double as a byte-identity test against the full re-parse.

**Adversarial review round** (34 reviewer/verifier agents over the five commits — sidebar/transcript/markdown came back clean): confirmed and fixed four async-window regressions in commit `80e058a` — file-tree expand results now trigger a redraw when they land; Enter can no longer submit a half-typed mention during the first walk's in-flight window; the stale popup fallback is bounded to the same token lineage (no wrong-file splice); superseded walks are serialized instead of stacking concurrently.

**Merge with v0.8.67 (`4504153`):** main's #3757 candidate cache (one needle-independent walk serves every keystroke of a non-path-like mention token, ranked in memory) is kept as the fast path, adapted to the key-based cache form; this branch's #3899 background walk now covers exactly the needles that cache bypasses (path-like and browser). The two fixes compose: non-path needles rarely walk at all, and when a walk is needed it never blocks the UI thread.

## Testing

- [x] `cargo test -p codewhale-tui --locked` (post-merge) — 5715 passed / 1 failed: `sandbox::tests::test_parity_linux_landlock_available`, which asserts the host kernel exposes Landlock and fails in this container on any branch (environment-only, untouched by this PR)

Targeted suites post-merge: `mention` 50, `sidebar::` 80, `file_tree` 8, `transcript::` 31, `widgets::` 188, `markdown_render` 38, `history` 148 — all green. Each commit message describes its new tests.

## Checklist

- [ ] Harvested/co-authored credit uses a GitHub numeric noreply address (n/a — no harvested work)

https://claude.ai/code/session_01Ffi9Tm4TfXhK5ngpA68xw3


Hmbown:
## Review — perf hot paths (verifying output-equivalence, not just speed)

Read-only review focused on the thing that matters for a perf change to render code: does each optimized path produce **byte-identical output**, and does every new cache **invalidate correctly**? Traced each fix against its issue.

**Overall: merge-ready, with two tracked follow-ups.** 4/5 fixes are correct with strong equivalence tests; the one real gap is bounded and has an Esc escape hatch.

### Per-fix verdicts
- **#3896 transcript deep-clone → CORRECT.** Slow path now borrows `&HistoryCell` and pre-materializes only the synthetic tool-run summaries; the per-cell reuse key (revision + index + layout/folded change) is unchanged, so collapse/expand/stream invalidation is unaffected. New `.expect("summary cell materialized")` push sites are provably safe (only called when `run.start == idx && collapsed_run_starts.contains(idx)`). Equivalence + reuse + frame-stability tests present.
- **#3897 streaming O(N²) markdown → CORRECT (highest blast radius).** Single shared `parse_line_into` classifier ⇒ classification identical by construction; only cross-line state is `in_code_block`; the tail parses against a **scratch** fence state; `stable_end` backs up over trailing table groups so column widths always see the whole group. A `debug_assert!(rendered_lines_eq(full_reparse, streaming))` differential guard + width×chunk sweep tests pin it. **Caveat:** the guard is `#[cfg(debug_assertions)]` (compiled out in release), and this rewrites the render path for *every streaming message* — the most-repainted surface — so any future parser change that breaks prefix-stability would be a silent release-only visual regression. Keep the guard debug-gated and scrutinize future edits here. (Also: parse is now O(delta), but the frozen-lines `clone()` per committed chunk keeps an O(N) memcpy — still a big win, not fully linear.)
- **#3898 sidebar rows twice/frame → CORRECT.** `task_panel_row_sets` computes once; rows + hover-texts share it, so they can't disagree across a mid-frame TTL boundary. Dedup/`recent`-gating preserved verbatim; the tree-sort move-materialization takes each slot exactly once (tested incl. a 2-node cycle).
- **#3899 path-like @mention fs walk → RISKY (bounded).** The core anti-race property holds: a completed walk is stored under **its own** needle key and only returned when it equals the live needle, so a slow old walk can't masquerade as the current one. **But** `stale_mention_fallback_entries` keeps the previous keystroke's entries visible for prefix-lineage tokens with the menu "open", so a fast **Enter before the new walk lands** can apply a completion that doesn't match the live token. Repro: `@docs/a` (shows `docs/apple.md`) → type `b` → Enter immediately → `docs/apple.md` inserted against `@docs/ab`. Pre-async, Enter always saw exact-needle results. Bounded (always a real file, prefix-lineage only) but a genuine equivalence gap. Minor secondary: if the walk thread panics, `mention_completion_pending` never clears and Enter is swallowed for that token until Esc.
- **#3900 file-tree sync read_dir → CORRECT.** Expand walk offloaded to `spawn_blocking_supervised`; per-dir `seq` drops superseded results; ancestor-collapsed-mid-walk and double-splice are guarded; cursor/scroll math adjusted on splice in/out. `incremental_expand_matches_full_rebuild` asserts entry-by-entry parity. Only visible change: ▼ shows for one poll cycle before children appear (progressive reveal) — imperceptible on local FS.

### Suggested follow-ups (neither blocks the perf win)
1. **#3899 (worth doing before merge if you want strict equivalence):** filter the stale fallback to entries still matching the live needle, **or** don't let Enter *apply* the fallback while `mention_completion_pending` is set (hold it like the empty-menu case).
2. **#3899 robustness:** clear `mention_completion_pending` if a walk fails to produce a result, so a panicked walk can't soft-lock Enter for that token.

Test coverage is genuinely strong (not asserted-by-comment) — the one untested window is the #3899 fast-Enter-applies-stale case above.

*(Detailed read; happy to open a small follow-up PR for the #3899 items if useful.)*

Hmbown:
Both #3899 follow-ups are in as `ebf2313` — no separate PR needed:

1. **Fast-Enter-applies-stale (your `docs/apple.md` vs `@docs/ab` repro):** took the stricter of your two options — the stale fallback now filters entries to those still matching the **live** needle (case-insensitive substring, same rule the walk itself uses), so anything shown or applied is a subset of what the fresh walk would return. When the filter empties the list, the popup reads as pending and the existing Enter-hold takes over. The exact repro is pinned by `mention_fallback_drops_entries_that_no_longer_match_the_live_needle`.
2. **Panicked-walk soft-lock:** the walker now routes its result through a completion guard whose `Drop` writes an empty result on unwind, so `mention_completion_pending` always drains and Enter releases. Pinned by `mention_walk_completion_fills_cell_even_on_unwind`.

On the #3897 caveat: agreed and leaving the differential guard debug-gated as you suggest — every `cargo test` run exercises it, and the two O(N) residuals you noted (frozen-lines clone per chunk, plus the message-layer decoration pass) are called out in the commit message as the known follow-up if profiling ever warrants it.

Post-push suite: 5717 passed / 1 failed (`test_parity_linux_landlock_available`, container-environment only).

---

Hmbown:
## v0.8.68 triage note

Draft + **DIRTY** vs main. Touches many hot-path files (`app.rs`, `ui.rs`, `sidebar.rs`, `transcript.rs`, …) that have moved substantially since the branch diverged.

Related perf work already landed separately (#4152 map lookups, #4149 parking_lot, etc.). Recommend:

1. Convert remaining unfixed hot-path items into focused per-issue PRs against current main, or
2. Full rebase + re-validation of the adversarial-review follow-ups before undrafting.

Not merging in the current form for v0.8.68 closeout.

Hmbown:
## Rebased partial harvest

Landed the clean subset on current `main` as **#4304** (`codex/v0868-4014-perf-hotpaths`):

- #3896 transcript collapse borrow path
- #3898 Activity panel row-sets once per frame
- #3900 file-tree expand off UI thread + `poll_background` drain

Still deferred from this draft (conflicts):
- #3899 @mention background walk (needs recompose with #3757 candidate cache)
- #3897 incremental streaming markdown

This draft can stay open as the source of the remaining two, or close once #4304 merges if we track the rest as follow-ups.

Hmbown:
Partial harvest landed on main via #4304 (merge 2c7c946c9d2f):

- #3896 transcript collapse borrow path
- #3898 Activity/Tasks row sets once per frame
- #3900 file-tree expand off UI thread + repaint drain

#3896/#3898/#3900 are now closed with proof comments. Leaving this draft open as source context for the remaining conflicting slices: #3899 mention background walk and #3897 incremental streaming markdown.

Hmbown:
Closing this draft as superseded for v0.8.68. The clean, current-main subset (#3896, #3898, #3900) was harvested into merged PR #4304 (`2c7c946c9d2f72ac32cdfbb05ce7821c37196402`), with contributor/source credit preserved there. The conflicting #3897/#3899 slices remain intentionally deferred rather than merged without a fresh rebase and validation; their follow-up issues stay open for future work.

Thanks to the original author and reviewers for the performance investigation and adversarial review that made the safe harvest possible.

PR #3967, opened 2026-07-03
perf(tui): avoid redundant composer input wrapping per frame

### LINKED ISSUE #3909: perf(tui): composer input is re-wrapped up to five times per frame
## Problem
One frame wraps the same composer input up to five times: desired-height calc, render, cursor positioning, scroll mapping, and mouse mapping each redo the width-aware wrap.

## Evidence
In `render()` (`crates/tui/src/tui/ui.rs`):
1. `:8555-8563` — a throwaway `ComposerWidget` computes `desired_height` via `composer_height` (`widgets/mod.rs:1127-1136`);
2. `:8845-8852` — a second `ComposerWidget` renders the composer;
3. `:8853` — `cursor_pos` re-runs `layout_input` (`widgets/mod.rs:1138-1163`);
4. `:8872-8885` — a third `ComposerWidget` is constructed just for `active_menu_reserved_rows`, then `layout_input_with_scroll` wraps again;
5. `:8898` — `wrap_input_lines_for_mouse` wraps once more (its own comment calls it "an approximation matching the render path").

## Impact
Runs on every keystroke and every streaming frame. Negligible for one-liners; with a multi-KB pasted draft the same wrap is redone five times per frame, hurting typing latency exactly when the input is big. Also a maintainability trap: five call sites must agree on budget math.

## Proposed approach
Compute one `ComposerLayout` per frame (wrapped lines, cursor row/col, scroll offset, menu reserved rows) from a single wrap pass and hand it to all five consumers.

## Acceptance criteria
- Input text is wrapped at most once per frame per width (twice if desired-height legitimately uses a different width than the final rect).
- Composer layout tests pass; cursor position and mouse mapping unchanged.

reidliu41:
## Summary

 Fix performance issue where composer input text was being wrapped up to five times per render frame. The redundant wrapping occurred  because `layout_input_with_scroll` and the selection rendering code each called their own wrap functions.

  ## Fixes
  Fixes #3909

## Testing

## Checklist

### DIFF
--- a/crates/tui/src/tui/widgets/mod.rs
+++ b/crates/tui/src/tui/widgets/mod.rs
@@ -3225,9 +3225,7 @@ fn wrap_input_lines_internal(input: &str, width: usize) -> (Vec<String>, Vec<(us
     let mut char_idx = 0usize;
 
     if input.is_empty() {
-        if width == 0 {
-            lines_with_indices.push((0, String::new()));
-        }
+        lines_with_indices.push((0, String::new()));
         return (lines, lines_with_indices);
     }
 
@@ -3378,7 +3376,8 @@ mod tests {
         apply_send_flash, build_empty_state_lines, composer_height, composer_max_height,
         composer_min_input_rows, composer_top_padding, cursor_row_col, empty_composer_visual_rows,
         layout_input, pad_lines_to_bottom, placeholder_visual_lines, push_command_entry,
-        should_render_empty_state, slash_completion_hints, wrap_input_lines, wrap_text,
+        should_render_empty_state, slash_completion_hints, wrap_input_lines,
+        wrap_input_lines_for_mouse, wrap_text,
     };
     use crate::config::{ApiProvider, Config};
     use crate::localization::Locale;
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @reidliu41. This is a good, focused perf slice and I would like to take it, but I want one small correctness/test follow-up first: please pin the empty-input behavior for `wrap_input_lines_for_mouse("", width)` and make sure the refactor does not change the empty composer/mouse-selection path.

The PR is clean and green, and it merged into the local scratch train with the other 0.8.67 PRs. If you can add that small guard quickly, this can still fit before I cut 0.8.67; otherwise I will route it to 0.8.68 rather than lose the work.

Hmbown (approved):
Good dedupe — routing both `wrap_input_lines` and `wrap_input_lines_for_mouse` (and the composer selection render) through one `wrap_input_lines_internal` removes the redundant re-wrap per frame. The `width == 0` branch is only reachable via direct tests; the composer render always passes `content_width = inner_area.width.max(1)`, so the mouse-path change is inert in production. CI green across all three OSes. Thanks @reidliu41.

---

PR #3968, opened 2026-07-03
feat(engine): expose context_input_budget_for_route for external budget reuse

h3c-hexin:
## Problem

`context_input_budget_for_route` — the internal input-budget computation (`window − window-dependent output reservation − headroom`) that is the single source of truth for the preflight / emergency-recovery / capacity-trim decisions — is `pub(super)` to `core::engine`.

External hosts / wrappers that embed the engine and derive their own compaction or pressure trigger line have no way to reuse it, so they end up **mirroring the constants**: the `≥500K` large-window threshold, `TURN_MAX_OUTPUT_TOKENS`, the output-reservation split in `route_output_reservation_for_window`, and the headroom. When the engine later changes any of these, the mirrored copy still compiles but **silently diverges** — the host computes a different budget than the engine actually enforces. (In our case the host's "compact now" line drifted above the engine's emergency line, so the graceful summarize path stopped firing and every long session fell through to emergency recovery.)

## Change

Make `context_input_budget_for_route` `pub` and re-export it from `core::engine`, so a host can compute the **exact** same budget the engine uses. Pass `input_tokens = 0` to get the full route budget.

Purely a visibility change:
- No behavior change; the function body is untouched.
- All parameters and the return type are already public (`ApiProvider`, `RouteLimits`, `usize`).

This lets downstream embedders reuse the engine's budget math instead of re-deriving (and drifting from) its internal constants.

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -3905,10 +3905,13 @@ mod approval;
 mod context;
 mod handle;
 pub(crate) use context::compact_tool_result_for_context;
+/// Public so external hosts/wrappers can reuse the engine's input-budget math
+/// (see `context_input_budget_for_route`'s doc) instead of re-deriving it.
+pub use context::context_input_budget_for_route;
 #[cfg(test)]
 use context::route_context_budget_for_provider;
 use context::{
-    MAX_CONTEXT_RECOVERY_ATTEMPTS, MIN_RECENT_MESSAGES_TO_KEEP, context_input_budget_for_route,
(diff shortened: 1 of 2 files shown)


Hmbown:
Merged into `main` for the 0.8.67 finalization train as #3968 (`25cce4074096b22dc9b531a7810f67dc0391e3e6`). Thanks @h3c-hexin. I treated this as behavior-neutral API reuse surface, and it compiled cleanly in the local scratch train with the other accepted PRs.

PR #4045, opened 2026-07-06
[codex] fix edit_file UTF-8 fuzzy cursor panic

### LINKED ISSUE #3971: `edit_file` panics on non-ASCII text after punctuation-normalized fuzzy match
## Environment

- CodeWhale: `0.8.66 (54eb3d3cb9cc)`
- Platform: macOS
- Reproduced on: 2026-07-03
- Affected area: TUI file tool, `edit_file`

## Agent-assisted note

This report was prepared with Codex assistance from a local, minimal reproduction. The repro uses synthetic text only (`数据 "x"`) and intentionally omits private project paths, user file contents, credentials, and local logs that are not needed to reproduce the panic. A human should review any patch or PR before merge.

## Summary

`edit_file` can panic when the search string only matches after punctuation normalization and the matched text starts with a multi-byte UTF-8 character.

The panic is reproducible with a tiny file containing Chinese text plus ASCII quotes, when the `search` argument uses smart quotes:

```text
数据 "x"
```

Search:

```text
数据 “x”
```

Replace:

```text
数据 y
```

Observed panic:

```text
thread 'main' panicked at crates/tui/src/tools/file.rs:845:44:
start byte index 1 is not a char boundary; it is inside '数' (bytes 0..3 of string)
```

I also saw the same panic in normal TUI sessions with larger Chinese documents:

```text
Process panicked at crates/tui/src/tools/file.rs:845:44:
start byte index 6510 is not a char boundary; it is inside '数'

Process panicked at crates/tui/src/tools/file.rs:845:44:
start byte index 4394 is not a char boundary; it is inside '这'
```

## Related upstream PRs checked

I checked the nearby PR range `#2154` through `#2168`.

- `#2154` (`donglovejava`): `fix(edit_file): always retry fuzzy fallback on first search miss` changed `crates/tui/src/tools/file.rs` so fuzzy fallback runs automatically when exact `search` misses. This is related to the affected path, but it did not address UTF-8 cursor advancement.
- `#2167` (`donglovejava`): `fix(runtime): truncate assistant summary on UTF-8 char boundary to prevent CJK panic` fixed a similar UTF-8 boundary issue in `crates/tui/src/runtime_threads.rs` by replacing byte slicing with `crate::utils::truncate_with_ellipsis(...)`. This confirms the same class of CJK byte-boundary panic has been fixed elsewhere, but it does not cover `edit_file`.

As of the current `main` branch checked on 2026-07-03, both fuzzy match helpers in `crates/tui/src/tools/file.rs` still advance the cursor with `norm_start.saturating_add(1)`.

## Focused checks performed

- Reproduced the panic through CodeWhale's own stdio MCP file tools with a synthetic one-line file.
- Confirmed the current `main` branch still contains two `cursor = norm_start.saturating_add(1)` sites in `crates/tui/src/tools/file.rs`.
- Checked related PRs `#2154` through `#2168`; `#2167` fixes the same class of CJK byte-boundary panic in runtime summary truncation, but not in `edit_file`.
- Did not run `cargo test` locally because this environment does not currently have the Rust toolchain installed.

## Minimal repro path

Create a file:

```text
数据 "x"
```

Then invoke `edit_file` after a prior `read_file`, with:

```json
{
  "path": "codewhale_utf8_repro.md",
  "search": "数据 “x”",
  "replace": "数据 y"
}
```

This reaches the punctuation-normalized fuzzy matching path because exact match fails but smart quotes normalize to ASCII quotes.

## Direct MCP repro

If using CodeWhale's stdio MCP server for a direct tool-level repro, expose the file tools:

```toml
[server]
expose_tools = ["file_read", "file_edit"]
require_approval = false
```

Then call:

```json
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"utf8-repro","version":"0"}}}
{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"file_read","arguments":{"path":"codewhale_utf8_repro.md"}}}
{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"file_edit","arguments":{"path":"codewhale_utf8_repro.md","search":"数据 “x”","replace":"数据 y"}}}
```

The process exits with code `101`.

## Suspected root cause

In `crates/tui/src/tools/file.rs`, both fuzzy match helpers advance `cursor` by one byte after each match:

```rust
cursor = norm_start.saturating_add(1);
```

This is unsafe for UTF-8 strings. If `norm_start` points to the start of a multi-byte character, `norm_start + 1` is not a valid char boundary. The next iteration slices with:

```rust
norm_contents[cursor..]
```

and Rust panics.

Confirmed locations in v0.8.66:

- `leading_whitespace_fuzzy_matches`: around `file.rs:791`
- `punctuation_normalized_matches`: around `file.rs:845` / `file.rs:853`

The existing smart-quote test starts with ASCII text (`let s = ...`), so `norm_start + 1` remains a valid UTF-8 boundary and does not catch this case.

## Suggested fix direction

Advance `cursor` to the next UTF-8 character boundary instead of adding one byte. For example, add a helper:

```rust
fn next_char_boundary(input: &str, idx: usize) -> usize {
    if idx >= input.len() {
        return input.len();
    }
    let mut next = idx.saturating_add(1);
    while next < input.len() && !input.is_char_boundary(next) {
        next += 1;
    }
    next
}
```

Then replace:

```rust
cursor = norm_start.saturating_add(1);
```

with:

```rust
cursor = next_char_boundary(&norm_contents, norm_start);
```

in both fuzzy match functions.

## Suggested regression test

Add a test next to `test_edit_file_fuzz_tolerates_smart_quote_substitution`:

```rust
#[tokio::test]
async fn test_edit_file_fuzz_tolerates_smart_quote_after_multibyte_start() {
    let tmp = tempdir().expect("tempdir");
    let ctx = ToolContext::new(tmp.path().to_path_buf());

    let test_file = tmp.path().join("smart_cjk.md");
    fs::write(&test_file, "数据 \"x\"\n").expect("write");
    read_before_edit(&ctx, "smart_cjk.md").await;

    let tool = EditFileTool;
    let result = tool
        .execute(
            json!({
                "path": "smart_cjk.md",
                "search": "数据 \u{201C}x\u{201D}",
                "replace": "数据 y",
                "fuzz": true
            }),
            &ctx,
        )
        .await
        .expect("execute");

    assert!(result.success);
    let edited = fs::read_to_string(&test_file).expect("read");
    assert_eq!(edited, "数据 y\n");
}
```

It may also be worth adding a similar regression test for `leading_whitespace_fuzzy_matches`, since it has the same cursor advancement pattern.

## User-visible impact

The panic aborts the current CodeWhale task and surfaces as an internal engine error. In interactive TUI use, this may also leave the terminal state feeling unstable afterwards, especially while editing Chinese-language documents.

nightt5879:
## Summary

Fixes `edit_file` fuzzy matching when a normalized match starts on a multibyte UTF-8 character. The previous cursor advance used `norm_start + 1`, so the next search iteration could slice from the middle of a CJK character and panic.

This patch advances the normalized cursor to the next UTF-8 character boundary for both fuzzy paths:

- leading-whitespace fuzzy matching
- punctuation-normalized fuzzy matching

It also adds CJK-starting regression tests for both paths.

Fixes #3971.

Credit: reported with repro and suggested fix direction by @taixinguo.

## Validation

- `cargo fmt`
- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui --locked test_edit_file_fuzz_tolerates` (5 passed)
- `cargo test -p codewhale-tui --bin codewhale-tui --locked test_edit_file` (12 passed)
- `python scripts/check-coauthor-trailers.py --range HEAD~1..HEAD`

Full-suite note: `cargo test -p codewhale-tui --bin codewhale-tui --locked` was attempted on this branch and failed only in unrelated full-suite tests. A clean `upstream/main` worktree also fails under the same local Windows environment in full-suite mode, including `core::engine::tests::edit_last_turn_preserves_current_mode` timing out with `Elapsed(())`, so the targeted `edit_file` validation above is the authoritative coverage for this change.

### DIFF
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -759,6 +759,18 @@ fn line_start_before(input: &str, idx: usize) -> usize {
         .map_or(0, |newline| newline.saturating_add(1))
 }
 
+fn next_char_boundary(input: &str, idx: usize) -> usize {
+    if idx >= input.len() {
+        return input.len();
+    }
+
+    let mut next = idx.saturating_add(1);
+    while next < input.len() && !input.is_char_boundary(next) {
+        next = next.saturating_add(1);
+    }
+    next
+}
+
 fn leading_whitespace_fuzzy_matches(contents: &str, search: &str) -> Vec<(usize, usize)> {
     let (normalized_contents, byte_map) = strip_line_leading_whitespace_with_map(contents);
     let (normalized_search, _) = strip_line_leading_whitespace_with_map(search);
@@ -788,7 +800,7 @@ fn leading_whitespace_fuzzy_matches(contents: &str, search: &str) -> Vec<(usize,
             };
         let original_end = byte_map.get(norm_end).copied().unwrap_or(contents.len());
         matches.push((original_start, original_end));
-        cursor = norm_start.saturating_add(1);
+        cursor = next_char_boundary(&normalized_contents, norm_start);
     }
     matches
 }
@@ -850,7 +862,7 @@ fn punctuation_normalized_matches(contents: &str, search: &str) -> Vec<(usize, u
         };
         let original_end = byte_map.get(norm_end).copied().unwrap_or(contents.len());
         matches.push((original_start, original_end));
-        cursor = norm_start.saturating_add(1);
+        cursor = next_char_boundary(&norm_contents, norm_start);
     }
     matches
 }
@@ -1792,6 +1804,35 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_edit_file_fuzz_tolerates_leading_whitespace_after_multibyte_start() {
+        let tmp = tempdir().expect("tempdir");
+        let ctx = ToolContext::new(tmp.path().to_path_buf());
+
+        let test_file = tmp.path().join("fuzzy_cjk.txt");
+        fs::write(&test_file, "数据\n").expect("write");
+        read_before_edit(&ctx, "fuzzy_cjk.txt").await;
+
+        let tool = EditFileTool;
+        let result = tool
+            .execute(
+                json!({
+                    "path": "fuzzy_cjk.txt",
+                    "search": "    数据",
+                    "replace": "记录",
+                    "fuzz": true
+                }),
+                &ctx,
+            )
+            .await
+            .expect("execute");
+
+        assert!(result.success, "{}", result.content);
+        assert!(result.content.contains("fuzzy indentation match"));
+        let edited = fs::read_to_string(&test_file).expect("read");
+        assert_eq!(edited, "记录\n");
+    }
+
+    #[tokio::test]
     async fn test_edit_file_fuzz_tolerates_smart_quote_substitution() {
         // The file on disk has ASCII quotes. The search comes from a
         // browser paste with curly quotes. Exact match fails; the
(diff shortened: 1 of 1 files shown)


Hmbown (approved):
Reviewed for the maintenance sweep: small scoped change, green checks, and safe to land on main. Preserving contributor authorship via merge commit.

PR #4181, opened 2026-07-07
fix(tui): Fleet setup role/profile roster editor (#4093)

### LINKED ISSUE #4093: Fleet setup modal is provider-scoped instead of a role/profile roster editor
## Release blocker

Target: v0.8.68

Fleet setup is currently shaped like a provider-scoped model picker, but it needs to behave like a **role/profile roster editor**. This blocks reliable Fleet setup because users cannot confidently assign model routes to worker profiles across configured providers, and the saved profile is ambiguous.

## Product intent

The Fleet setup modal should let a user assemble and edit the worker roster:

- roles/profiles such as Scout, Reviewer, Builder, Verifier, or custom profiles;
- editable role metadata: display name, role hint/type, instructions, tool restrictions, trust/sandbox posture, budgets/concurrency defaults where supported;
- one or more model route options per profile, selected from **all configured providers**, not only the parent/current provider;
- a clear default/primary route and optional fallbacks where supported;
- validation that the selected provider/model route is configured and usable;
- a save/preview flow that makes the final profile and route binding obvious.

The modal should feel like: `Fleet Setup -> Profile roster -> Edit role behavior -> Add/choose model routes -> Validate -> Save`, not `current provider -> pick one local model id`.

## Observed behavior

- In Fleet profile setup, changing a profile model only exposes models from the **current provider**.
- Intended behavior was that Fleet profile model selection should expose models from **all configured providers**, because Fleet workers can be assigned routes independent of the parent/current provider.
- Saving a DeepSeek profile produced:

```toml
description = "A fast reconnaissance worker that explores codebases, APIs, or data sources, gathers relevant facts, and returns structured summaries to the parent agent."
display_name = "Scout"
id = "scout-deepseek-v4-flash"
model = "deepseek-v4-flash"
role_hint = "scout"
```

from:

```text
/Users/hunter/Desktop/Harnesses/CW/.codewhale/agents/scout-deepseek-v4-flash.toml
```

This persisted profile stores only `model = "deepseek-v4-flash"` and does not record provider/route identity, so it appears ambiguous unless resolution remains scoped to the active provider. That is the opposite of the expected Fleet behavior.

## Draft preview UI issue

The ratification preview compounds the problem:

- Header says `fleet profile - draft by deepseek-v4-pro (g ratifies)`.
- Footer simultaneously says `g/G top/bottom`.
- In the scrollable preview, pressing `g` goes to top instead of ratifying. The user had to press `Esc` and then `g` to save.

This makes the save path look broken and makes it unclear whether the selected model actually became the Fleet profile route.

## Expected behavior

- Fleet profile model selection lists/searches all configured provider routes, not only the current provider.
- Each profile can attach/edit model route options independently of the parent session route.
- Persisted Fleet profiles store enough route identity to be unambiguous, for example provider + model or a resolved route/profile reference, not only a provider-local model id.
- Role/profile behavior can be edited in the same setup flow without forcing users into raw TOML for normal changes.
- The draft preview has no conflicting `g` binding. Ratify should either work in the preview or use a distinct/non-conflicting command shown consistently in header and footer.
- The post-save message should state the provider/route actually bound to the saved profile.

## Acceptance criteria

- A user can create/edit a Fleet profile for a model from any configured provider while the parent session is on a different provider.
- The modal shows role/profile roster state and supports editing role behavior and model route options in a coherent flow.
- The saved profile can be inspected and clearly identifies the intended provider/model route.
- The profile can be used by Fleet without depending on the current provider at launch time.
- Route validation catches missing credentials/provider config before save or before launch with a clear error.
- Ratifying the preview has a discoverable, non-conflicting key path.

## Impact

This is a release blocker for v0.8.68 because Fleet profile setup is a headline workflow. The current behavior can save an ambiguous or wrong profile, hides cross-provider model options, and makes the final confirmation feel broken.

Hmbown:
Refs #4093 (headline cross-provider fix landed for the `codewhale fleet run` CLI path; the interactive TUI in-process remainder is tracked in #4193, so this does not auto-close #4093)

## Summary
- Align Fleet setup modal with role/profile roster editing instead of provider-scoped model picker
- Expose model routes from all configured providers when editing Fleet profiles
- Persist unambiguous provider+model route identity in saved profiles
- Fix draft preview ratify key binding conflict (`g` scroll vs ratify)

## Acceptance criteria
- [ ] User can create/edit a Fleet profile for a model from any configured provider while parent session uses a different provider
- [ ] Modal shows role/profile roster state with coherent edit flow for role behavior and model routes
- [ ] Saved profile clearly identifies intended provider/model route on inspection
- [ ] Profile usable by Fleet without depending on current provider at launch time
- [ ] Route validation catches missing credentials before save/launch
- [ ] Ratifying preview has discoverable, non-conflicting key path

Issue: https://github.com/Hmbown/CodeWhale/issues/4093

## Test plan
- [ ] `cargo test -p codewhale-tui fleet_setup` (or focused fleet setup tests)
- [ ] Manual: create Scout profile with DeepSeek route while parent on different provider
- [ ] Manual: verify saved TOML includes provider/route identity, not bare model id
- [ ] Manual: ratify draft preview without `g` scroll conflict
- [ ] CI green

Signed-off-by: Hunter <hunter@codewhale.dev>

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fleet setup is a role/profile roster editor, not a provider-scoped model
+  picker: the Model step lists routes from every configured provider (not
+  only the active one), a picked route's provider is persisted explicitly in
+  the saved profile TOML (`provider = "..."`, never inferred from the model
+  id), and the loader/route resolver read that field back out verbatim. The
+  draft-preview ratify keypress no longer competes with a separate pager's
+  `g`/`G` scroll bindings — the exact TOML preview now renders inline on the
+  same Review step that ratifies it (#4093).
 - Workflow correctness: completion polling fails closed instead of
   fabricating success when a sub-agent reports no terminal status; cancel
   interrupts the JS VM (cancel handle + abort) and blocks further spawns;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,14 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   draft-preview ratify keypress no longer competes with a separate pager's
   `g`/`G` scroll bindings — the exact TOML preview now renders inline on the
   same Review step that ratifies it (#4093).
-- Fleet workers now actually launch on their profile-pinned route, not just
-  record it on the receipt: `codewhale exec` gains a non-secret `--provider`
-  flag, and a worker whose profile pins provider B is dispatched with
-  `--provider B --model <B's model>` even when the parent session is on
-  provider A (credentials still resolve from the worker's own environment;
-  provider is never inferred from the model id). Workers with no
-  profile-bound provider are unchanged — no `--provider`, run-level model
-  (#4093).
+- The headless `codewhale fleet run` CLI now launches workers on their profile-pinned route, not just records it on the receipt: `codewhale exec` gains a non-secret `--provider` flag, and a worker whose profile pins provider B is dispatched with `--provider B --model <B's model>` even when the parent session is on provider A (credentials still resolve from the worker's own environment; provider is never inferred from the model id). Workers with no profile-bound provider are unchanged — no `--provider`, run-level model. The interactive TUI spawns roster members in-process and does not yet honor the pinned provider (it uses the session provider); that remainder is tracked in #4193 (#4093).
 - The Fleet setup `m` model-assisted redraft no longer drops a picked
   cross-provider route: the provider/model the operator chose are re-pinned
   onto the drafted profile (a model draft is always `provider: None`), so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   draft-preview ratify keypress no longer competes with a separate pager's
   `g`/`G` scroll bindings — the exact TOML preview now renders inline on the
   same Review step that ratifies it (#4093).
+- Fleet workers now actually launch on their profile-pinned route, not just
+  record it on the receipt: `codewhale exec` gains a non-secret `--provider`
+  flag, and a worker whose profile pins provider B is dispatched with
+  `--provider B --model <B's model>` even when the parent session is on
+  provider A (credentials still resolve from the worker's own environment;
+  provider is never inferred from the model id). Workers with no
+  profile-bound provider are unchanged — no `--provider`, run-level model
+  (#4093).
+- The Fleet setup `m` model-assisted redraft no longer drops a picked
+  cross-provider route: the provider/model the operator chose are re-pinned
+  onto the drafted profile (a model draft is always `provider: None`), so
+  ratifying it keeps the explicit route instead of persisting an ambiguous,
+  provider-scoped profile (#4093).
+- Ratifying a Fleet profile now fails with a clear message when it pins a
+  provider that has no configured credentials, using the same
+  configured-provider check the model picker uses (#4093).
 - Workflow correctness: completion polling fails closed instead of
   fabricating success when a sub-agent reports no terminal status; cancel
   interrupts the JS VM (cancel handle + abort) and blocks further spawns;
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1146,6 +1146,20 @@ pub struct FleetProfile {
     /// validates the executable provider/model/wire-model decision.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub model: Option<String>,
+    /// Optional explicit provider id for this profile's model (#4093).
+    ///
+    /// Present only when the profile was created against a specific,
+    /// credential-checked provider (e.g. via the Fleet setup model picker),
+    /// so a worker can be pinned to a route independent of the parent/current
+    /// session provider. `None` means "no route pin" (inherit), matching
+    /// `model: None`; a profile must never carry `provider` without `model`.
+    ///
+    /// EPIC #2608 explicit-config-only mandate: this field is the ONLY
+    /// authority for the profile's provider. It is never inferred by sniffing
+    /// a substring/prefix out of `model` — callers that need the provider for
+    /// this profile must read this field, not guess from the model id.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub provider: Option<String>,
     /// Permission defaults requested by the profile.
     #[serde(default)]
     pub permissions: FleetProfilePermissions,
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -39,6 +39,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fleet setup is a role/profile roster editor, not a provider-scoped model
+  picker: the Model step lists routes from every configured provider (not
+  only the active one), a picked route's provider is persisted explicitly in
+  the saved profile TOML (`provider = "..."`, never inferred from the model
+  id), and the loader/route resolver read that field back out verbatim. The
+  draft-preview ratify keypress no longer competes with a separate pager's
+  `g`/`G` scroll bindings — the exact TOML preview now renders inline on the
+  same Review step that ratifies it (#4093).
 - Workflow correctness: completion polling fails closed instead of
   fabricating success when a sub-agent reports no terminal status; cancel
   interrupts the JS VM (cancel handle + abort) and blocks further spawns;
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -47,14 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   draft-preview ratify keypress no longer competes with a separate pager's
   `g`/`G` scroll bindings — the exact TOML preview now renders inline on the
   same Review step that ratifies it (#4093).
-- Fleet workers now actually launch on their profile-pinned route, not just
-  record it on the receipt: `codewhale exec` gains a non-secret `--provider`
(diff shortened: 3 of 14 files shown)


Hmbown:
Pushed `569d74b` on top of the existing `31006ed`. Status: **not ready to merge yet** — reporting progress + a precise remaining gap rather than merging on a hunch, per the release-blocker guardrail for #4093.

**What this commit does (verified locally: `cargo fmt --check`, `cargo clippy --workspace --all-features` with the repo's standard allow-list, `cargo test -p codewhale-tui` for the whole `fleet::` module = 131 passed, and `cargo test -p codewhale-tui fleet_setup` = 14 passed, all green):**

1. Fixes the compile mismatch the earlier reviews flagged (already fixed on `31006ed`, reconfirmed).
2. Threads an explicit `provider` field end to end: `FleetProfileDraft` → `render_toml()` → the TOML loader (`AgentProfileToml`) → `FleetProfile` (in `codewhale-config`). The field is validated against the real `ApiProvider` vocabulary at load time and is never inferred from the model id (EPIC #2608). The Fleet setup Model step now stores the canonical provider id (not just a display label), so a cross-provider pick round-trips correctly into the saved profile TOML as `provider = "..."` alongside `model = "..."`.
3. Fixes `resolve_fleet_route` in `worker_runtime.rs` to resolve within the profile's own explicit provider scope when it has one, instead of a hardcoded DeepSeek default — so the receipt/route-resolution surface (what `/fleet status` and task receipts show) reflects a pinned cross-provider route rather than silently defaulting.
4. Fixes the review screen's Model section, which was showing the parent/active session's provider instead of the actually-picked route's provider (a real bug the review bot flagged).
5. Fixes the `g`/ratify vs pager `g`/`G`-scroll conflict: the draft preview (both the deterministic starter draft and the model-drafted flow) now renders inline on the same Review step that ratifies it, instead of opening a separate `PagerView` on top that silently swallowed a lone `g` keypress (vim-style `gg`-to-top arming) and forced an Esc-then-`g` round trip to actually save.
6. New tests: a cross-provider save → reload-from-disk → route-resolution test and a direct provider-precedence test in `worker_runtime.rs`; a provider round-trip test and an unrecognized-provider-name rejection test in `profile.rs` (replacing the now-obsolete "provider is always rejected" test, since provider is now an intentionally accepted, validated field); inline-preview coverage in `fleet_setup.rs`.

**What is NOT fixed here — the real remaining gap:**

The actual Fleet worker dispatch path (`fleet/executor.rs`'s `build_worker_exec_command_with_profiles`, invoked from `fleet/manager.rs`'s `start_leased_workers`) runs every Fleet worker as a separate `codewhale exec --model <model> ...` **subprocess**. Today:
- There is no `--provider` CLI flag on `exec` at all.
- `start_leased_workers` only ever passes the **run-level session model** to that subprocess command — it does not resolve or pass the per-task/profile pinned model either, let alone a provider.

So while the profile's provider is now correctly persisted, validated, and honored by the resolver/receipt surface this PR touches, a Fleet worker's *actual spawned process* does not yet reliably launch on the profile's own provider — it inherits whatever provider the freshly-started subprocess's own on-disk config resolves to, with only an inventory-based model→provider match (in `model_routing::resolve_explicit_route_with_inventory`) as a fallback, which only works when the pinned model id is unambiguous across the user's configured providers. This is a materially larger, pre-existing architectural gap (adding a `--provider` flag to `ExecArgs` in `main.rs`, plus rewiring `start_leased_workers` to resolve and pass the per-task model+provider) that I did not attempt in this pass, to avoid destabilizing the primary CLI entrypoint under release-blocker time pressure without dedicated review.

`cargo test --workspace` was still running in the background when I had to hand off; I did not wait to confirm it before pushing. CI has not been polled yet on this push.

**Recommendation:** do not merge until (a) `cargo test --workspace` is confirmed green, (b) CI is green on all platforms, and (c) a follow-up either closes the subprocess-dispatch gap above or the issue/PR scope is explicitly narrowed to acknowledge it (i.e., decide whether #4093 is "done" once profiles round-trip an explicit, validated provider, or whether it must also guarantee the real spawned worker process honors it).

---

Hmbown:
Follow-up: `cargo test --workspace --locked` finished — 5976 passed, 1 failed, 2 ignored. The one failure is `sandbox::tests::test_parity_linux_landlock_available`, unrelated to this change (a Linux landlock kernel-parity check that fails in this sandboxed dev environment, not touched by this diff). Every fleet/profile/config test is green. Still holding off on merge for the reason in my previous comment (the subprocess-dispatch gap) plus CI not yet polled on `569d74b`.

---

Hmbown:
**Status update (commit `a63a5d6` on top of `569d74b`) — not merging yet, precise remaining gap below**

## CI fixes in this commit

Both CI failures on `569d74b` were environment/formatting issues, not logic bugs:
- `Lint` (`cargo fmt --all -- --check`): the `write_workspace_profile(...)` call added to `fleet/roster.rs`'s `broken_workspace_dir_degrades_to_built_ins_and_config` test (for the new #4093 provider-validation coverage) exceeded the line-width limit. Reformatted; no behavior change.
- `Version drift` (`scripts/sync-changelog.sh --check`): `crates/tui/CHANGELOG.md` (the packaged slice embedded in the binary) had drifted from the root `CHANGELOG.md`'s new `### Fixed` entry for #4093. Re-ran `./scripts/sync-changelog.sh`.

## Re-verified locally after the fix, before pushing

- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p codewhale-tui --all-features --locked -- -D warnings` (repo's standard allow-list) — clean.
- `cargo test -p codewhale-tui fleet` — **175 passed, 0 failed** (covers `fleet_setup` 14/14, the full `fleet::` module 131/131 including the two new cross-provider tests, plus `roster`/`profile`/`worker_runtime`).
- `./scripts/sync-changelog.sh --check` — clean.

## What's genuinely solid as of `569d74b` (verified by reading the code, not just trusting it compiles)

- `FleetProfileDraft`/`AgentProfile` carry an explicit, validated `provider` field (never inferred from a model-id substring — EPIC #2608), persisted as `provider = "..."` in the saved TOML alongside `model`, and read back verbatim by the loader (`validate_agent_profile_provider` in `profile.rs`).
- `resolve_fleet_route` (`worker_runtime.rs:133`) resolves within the profile's own explicit provider via `effective_fleet_provider` when it has one — not a hardcoded default, not the parent/session provider. Covered by `resolve_fleet_route_honors_explicit_profile_provider_not_the_default` and the required cross-provider save→disk→reload→resolve test `cross_provider_profile_saves_reloads_and_resolves_to_its_own_provider`.
- `render_review()` in `fleet_setup.rs` shows the picked route's own provider, not the active session's.
- The `g`/ratify vs pager-scroll conflict is genuinely fixed: the draft preview renders inline on the Review step (`render_scrollable_text`) and `g` is bound to ratify only there (`handle_key`, `fleet_setup.rs:572`); scrolling uses Up/Down/PageUp/PageDown/Home instead, so there's no more overlap. Confirmed by reading `handle_key` directly.
- This closes the ledger/receipt-visible surface (what `resolve_task_route`/task receipts report) and the setup-flow UX end-to-end — real, user-visible progress.

## What is NOT yet fixed — why I'm not merging a release blocker on this alone

Neither place that actually **launches** a worker consults the profile's `provider` field yet, so acceptance criterion "profile usable by Fleet without depending on the current provider at launch time" isn't true end-to-end:

1. **`crates/tui/src/fleet/executor.rs::build_worker_exec_command_from_prompt`** (the headless Fleet-run subprocess path) only ever adds `--model <model>` to the `codewhale exec …` argv passed to `FleetWorkerCommand`; there's no `--provider`. `start_leased_workers` (`manager.rs:809`) passes the flat run-level `model`, not the per-task `resolve_task_route(...)` result, into the builder. The CLI wrapper (`crates/cli/src/lib.rs`) already has a global `--provider` flag that would work here — it must precede `exec` (`codewhale --provider openrouter exec ...`, enforced by `reject_exec_global_flags`) — so wiring `resolved_route.provider_id` through `start_leased_workers` → `build_worker_exec_command_with_profiles` → the argv (inserted before `"exec"`) would close this specific gap. This looks tractable.
2. **`crates/tui/src/tools/subagent/mod.rs::spawn_subagent_from_input`** (the in-process `agent`-tool spawn path Fleet profiles also flow through via `apply_spawn_profile`) never reads `member.profile.provider` at all — zero references to it in the file. `child_runtime = runtime.background_runtime()` clones the **parent's own `DeepSeekClient`** and only overrides `.model`/`.reasoning_effort`; the client that actually issues the request stays bound to the parent's provider. This is the bigger gap: `SubAgentRuntime`/`ToolContext` carry no `Config` reference to rebuild a client from, so honoring a profile's provider pin here means new credential-aware client-construction plumbing, not just threading a string.

Net effect: a saved profile pinning `provider = "openrouter", model = "deepseek-v4-flash"` while the parent session is on native DeepSeek will show correctly in the ledger/receipt (fixed here), but the worker that actually runs still executes under the parent/local provider via path (2) with certainty, and via path (1) unless the spawned subprocess's own local config/keyring happens to also route that model to the intended provider. That's the "launches-on-parent-provider-instead-of-its-own" bug the issue calls out, and it isn't fully closed by this branch.

## Recommendation

Per the release-blocker guardrail, I'm not merging this as a complete fix of #4093. Suggest one of:
- (a) land this now (compile fix, key-binding fix, and full provider persistence/validation/resolver-level correctness with real tests are genuine progress) and open a fast-follow issue for the two dispatch-layer gaps above before calling #4093 fully closed, or
- (b) hold this PR for a follow-up commit closing (1) and ideally (2).

I didn't attempt either gap in this pass: (1) touches the primary CLI entrypoint's exec dispatch and (2) needs new credential/client-construction plumbing with no existing reusable path — both felt like the wrong thing to rush under release-blocker time pressure without dedicated review.

CI on `a63a5d6`: Lint, Version drift, Claude review, CodeQL/analyze (all languages), Check Signed-off-by, and the Linux test/npm-wrapper jobs are green. `Test (macos-latest)`, `Test (windows-latest)`, and `Mobile runtime smoke` were still compiling as of this comment (long but not stuck — verified each is mid `cargo test`/smoke-test step, not hung).

---

PR #4243, opened 2026-07-08
perf(tui): migrate runtime_threads maps to parking_lot::Mutex (#4149)

### LINKED ISSUE #4149: v0.9.0 6.1: finish parking_lot migration at hot lock sites
Parent: Hmbown/CodeWhale#4113  <br>Source: CODEWHALE_0_8_68.md Section 6.1 / B1.2

Finish migrating hot `std::sync` lock sites to `parking_lot` where appropriate.

Open sites:

* `app-server/src/lib.rs` runtime locks.
* `runtime_threads` map.

Already partial:

* Subagent lease path is done.

Acceptance:

* Hot sites use `parking_lot` where safe.
* No lock guards cross `.await`.
* Existing app-server and TUI tests pass.

wuisabel-gif:
Closes #4149 (v0.8.68 §6.1 / B1.2 — finish parking_lot migration at hot lock sites).

I claimed this on the issue before starting — see [my comment](https://github.com/Hmbown/CodeWhale/issues/4149#issuecomment-4917698353).

## What

`RuntimeThreadManager` held four synchronous maps behind `std::sync::Mutex`:

- `task_manager`
- `automations`
- `pending_approvals`
- `pending_dynamic_tools`

All four are locked in **short, non-async critical sections** (attach / register / cancel / deliver / count / snapshot-clone), so they're safe `parking_lot::Mutex` sites. `parking_lot` guards are `!Send` and cannot cross `.await`; I verified none of these sites hold a guard across an await.

Because `parking_lot::Mutex` doesn't poison, `lock()` returns the guard directly — which removes the `if let Ok(..)` / `match ... Err` boilerplate at every call site (net **−21 lines**).

## Behavior note

The one intentional behavior change: after a panic *while holding* one of these locks, callers now continue with the recovered map instead of silently skipping via the poisoned-`Err` branch. That's the standard, intended `parking_lot` tradeoff, and for these best-effort registries (pending sender maps, manager slots) continuing is the safer default.

## Deliberately left as-is

- **`state` / `active`** — `tokio::sync::Mutex`, held across `.await` on purpose; not parking_lot candidates.
- **`config`** — already `parking_lot::RwLock`.
- **app-server `runtime locks`** (also named in the issue) — those are `tokio::sync::{Mutex, RwLock}` async locks held across `.await`; migrating them to `parking_lot` would be incorrect, so they're out of scope.

## Acceptance

- [x] Hot sites use `parking_lot` where safe
- [x] No lock guards cross `.await`
- [x] Existing tests pass — `cargo check -p codewhale-tui` clean; **178** approval/pending tests pass, 0 failed

### DIFF
--- a/crates/tui/src/runtime_threads.rs
+++ b/crates/tui/src/runtime_threads.rs
@@ -14,7 +14,7 @@ use std::collections::{HashMap, HashSet, VecDeque};
 use std::fs::{self, File, OpenOptions};
 use std::io::{BufRead, BufReader, Write};
 use std::path::{Component, Path, PathBuf};
-use std::sync::{Arc, Mutex as StdMutex};
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::{Context, Result, anyhow, bail};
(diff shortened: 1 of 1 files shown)


Hmbown:
thank you so much for contributing!!

wuisabel-gif:
> thank you so much for contributing!!

you're very welcome~

PR #4311, opened 2026-07-09
feat(models): add GPT-5.6 and Muse Spark routes

Hmbown:
## What

Lands the two model/provider commits from the local v0.8.68 lane, unchanged and with Hunter's authorship preserved:

- `feat(models): add GPT-5.6 and Muse Spark routes` — OpenAI GPT-5.6 family (`gpt-5.6` alias, `-sol`, `-terra`, `-luna`) and Meta Model API's `muse-spark-1.1` across provider metadata, configuration, secrets, CLI selection, catalogs, model limits, pricing, reasoning payloads, and docs.
- `fix(web): keep provider facts inventory complete` — adds Meta Model API and the existing xAI route to both provider-facts maps so the website inventory and drift path can't silently omit them.

## Fact verification (first-party where possible)

**GPT-5.6 — fully first-party verified** against [developers.openai.com/api/docs/models](https://developers.openai.com/api/docs/models):
- IDs `gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna`; `gpt-5.6` aliases to Sol ✓ (matches registry seeding)
- 1.05M context, 128K max output ✓
- Pricing $5/$30, $2.50/$15, $1/$6 per MTok with 90% cache-read discount ($0.50/$0.25/$0.10) ✓ (matches `pricing.rs` exactly)
- Reasoning efforts none/low/medium/high/xhigh/max ✓

**Muse Spark 1.1** ([Meta announcement](https://ai.meta.com/blog/introducing-muse-spark-meta-model-api/)):
- Name/version, 1M context, Meta Model API public preview, OpenAI-compatible: **first-party** (blog) ✓
- Endpoint `https://api.meta.ai/v1`: **live-probed** — returns OpenAI-style `invalid_api_key` JSON on unauthenticated `/models` ✓
- Wire ID `muse-spark-1.1`, pricing $1.25/$4.25: corroborated by Bloomberg, CNBC, The Decoder (all dated 2026-07-09) and SDK guidance in coverage; portal docs are behind the preview waitlist.
- ⚠️ 32K max output and the `none`/`minimal` ends of the effort range are **corroborated but not first-party-confirmed** (secondary coverage confirms `minimal`→`xhigh`). Flagging honestly; worth re-checking once Meta Model API docs are accessible. No live paid call was made.

## Verification

- `cargo test -p codewhale-config --locked` → 52 passed
- `cargo test -p codewhale-agent --locked` → 366 passed
[...]

Issue #3227, opened 2026-06-14
v0.8.61: Isolate provider/model selection per TUI session and make route changes atomic


Hmbown:
## Problem

Live QA found a severe provider/model isolation bug across multiple CodeWhale terminals. Hunter had one CodeWhale session open with GLM 5.2 / Z.ai active, then in another terminal repeatedly changed provider/model back to DeepSeek / `deepseek-v4-pro`. On send, the second terminal showed an impossible route: the visible model chip said `deepseek-v4-pro`, the provider badge still said `Z.ai`, and the request failed with:

```text
Invalid request (400): Unknown Model, please check the model code.
```

That means CodeWhale can currently combine a model from one provider with another provider route, probably due to shared/global provider-model persistence and/or non-atomic in-memory route updates. This breaks parallel QA, multi-terminal work, Fleet workers, and any user running different projects/models at once.

## Why this matters

Changing a model in one terminal must not mutate the active provider/model route of another already-running terminal. A running TUI session should own its route until the user changes it in that session. Global config should only affect new sessions, or explicit `save as default` actions.

## Current code pointers to inspect

- `crates/tui/src/commands/groups/core/core.rs`
  - `/model` updates `app.provider_models` and calls `Settings::persist_provider_model_selection(app.api_provider, model)`
  - that persistence also writes `default_provider`, so a model change can become a global provider/default change
- `crates/tui/src/settings.rs`
  - `default_provider`, `default_model`, and `provider_models` are shared settings
  - `set_provider_model_selection` writes both `default_provider` and provider-scoped model
- `crates/tui/src/tui/ui.rs`
  - model picker persists provider-scoped model changes with `settings.save()`
  - provider switch persists root `provider` plus shared `default_provider`
  - route changes need to update `app.api_provider`, `app.model`, `config.provider`, provider model override, engine config, compaction budget, and header/status together

## Desired behavior

1. **Session-local by default.** `/model`, model picker, and `/provider` should change the current TUI session route only unless the user explicitly chooses to persist/save as default.

2. **Atomic provider/model route.** CodeWhale must not send `deepseek-v4-pro` to Z.ai, GLM to DeepSeek, or any provider/model mismatch. Provider, model, base URL, auth mode, reasoning effort, compaction budget, and engine client should switch as one validated route.

[...]

Issue #3216, opened 2026-06-14
v0.8.61: Make multi-agent fanout durable and nonblocking before shipping swarm UX


Hmbown:
## Problem

A live v0.8.60 session froze/stalled again when the agent tried to fan out across six sub-agents. The children are headless, but the parent turn can still block because the in-process sub-agent path deliberately holds the parent open while children are running.

Concrete code path to inspect:

- `crates/tui/src/core/engine/turn_loop.rs`
  - child completions are drained from `rx_subagent_completion`
  - the parent waits while `mgr.running_count() > 0`
  - `should_hold_turn_for_subagents(queued_completions, running_children)` currently returns true for any queued completion or running child
- `crates/tui/src/tools/subagent*` and `crates/tui/src/core/engine.rs`
  - `agent_open` / `agent_eval` use the in-process `SubAgentRuntime` / `SubAgentManager` path
- `crates/tui/src/fleet/*`
  - Fleet already has durable worker, lease, heartbeat, retry, receipt, and inspection concepts, but normal fanout still does not consistently use those semantics

## Desired behavior

The parent agent must be able to launch multiple workers, keep working, accept user steering, run independent shell/verifier work, and consume child results as they arrive. A child model call, timeout, or long tool run must not make the parent feel frozen.

## Concrete implementation plan

1. **Split child launch from child join.** Change the parent turn-loop contract so launching children is not the same as waiting for them. `should_hold_turn_for_subagents` should not hold solely because `running_children > 0`.

2. **Surface completions opportunistically.** Keep the existing completion sentinel path, but only resume the parent when there are queued completions or when the parent explicitly requested a final join. Running children alone should become background work visible in the sidebar/status surface.

3. **Add an explicit wait/finalize operation.** If the model really needs to join all workers, require a deliberate tool/action such as `agent_eval(..., block:true)` or a fleet wait/inspect command. The default path should be nonblocking.

4. **Move high-fanout onto Fleet semantics.** For N-way work, prefer durable Fleet-backed workers over in-process sub-agent sessions: worker ids, ledger state, heartbeats, retry policy, receipts, logs, and inspect/restart controls. The user-facing word can still be “sub-agent,” but the runtime should behave like Fleet.

5. **Keep user steering live.** While workers run, Cmd+Enter/Ctrl+S/queued steering must be accepted and delivered to the parent or next continuation. The parent should not sit in a wait loop that makes steering feel ignored.

[...]

Issue #3203, opened 2026-06-13
v0.8.61: Make queued steering reliable and add Ctrl+S send


Hmbown:
## Problem

During live v0.8.60 testing, steering the running agent from the composer was unreliable. Cmd+Enter sometimes did not submit the typed steering message while the model or child workers were busy. Hunter also asked for Ctrl+S to send a queued steering message.

This is especially painful when workers/sub-agents are waiting on model responses: the user can type guidance, but the TUI may appear to ignore it until the current turn finally advances.

## Desired behavior

- Cmd+Enter reliably submits a steering/queued message from the composer in all busy states where steering is allowed.
- Ctrl+S is added as an alternate send binding for queued steering.
- The input is acknowledged immediately: clear/preserve the composer intentionally, show the queued/pending state, and make it obvious whether the message will steer the parent turn, the active worker group, or the next safe continuation boundary.
- Steering delivery should be scheduler-aware: it should not disappear behind a blocking child-agent/fleet wait.

## Acceptance criteria

- Add keybinding coverage for Cmd+Enter and Ctrl+S in the relevant composer/input mode(s).
- Add a PTY or app-state regression that simulates a busy/model-waiting turn, sends queued steering, and verifies the message is accepted without waiting for child workers to finish.
- The Work/Agents/status surface distinguishes queued user steering from ordinary pending tool/model work.
- Existing Enter/newline behavior remains unchanged.
- Manual QA: start a long-running worker or background verifier, type a steering message, send it with Cmd+Enter and Ctrl+S, and verify both are acknowledged and eventually delivered exactly once.

## Notes

This came from Hunter live-testing CodeWhale inside the v0.8.60 branch on 2026-06-13. It is related to the broader headless worker/Fleet migration in #3096 and non-blocking work in #3200, but the keyboard/input path is concrete enough to track separately.

Hmbown:
Retargeting to v0.8.61 rather than closing. v0.8.60 landed the first concrete slice in 5ca618d70: `agent_eval` is nonblocking by default, follow-up input can be delivered without waiting for child completion, and Ctrl+S sends the next queued/draft message into the active turn or dispatches it when idle.

The issue acceptance criteria also call out Cmd+Enter reliability, busy/model-wait app-state regression coverage, and clearer queued-steering status. Those are still worth finishing as a focused v0.8.61 polish/reliability item.

Hmbown:
Shipped in v0.8.61 (main is now at c9193e2e). Closing as completed.

PR #4316, opened 2026-07-09
test: isolate hotbar skill cache and automate live TUI QA targets

Hmbown:
## What

Two test-hardening commits from the local v0.8.68 lane, authorship preserved:

- `test(hotbar): isolate configured skill cache` — the hotbar skill activation test now sets `Config.skills_dir` explicitly, matching production startup semantics and preventing a developer's global `~/.agents/skills` from shadowing the temporary fixture.
- `test(release): automate live TUI QA targets` — new `crates/tui/tests/release_runtime_qa.rs` drives the shipped TUI through real pseudo-terminals with sealed homes and loopback providers: six-worker fanout liveness + Esc cancellation (#3216/#2211), concurrent Muse/GPT route isolation (#3227), and queued Ctrl+S steering (#3203). Wired into the release-sweep skill as an explicit gate.

## Why base = codex/v0868-models

The route-isolation scenario drives `CODEWHALE_PROVIDER=meta` / `muse-spark-1.1` and `gpt-5.6-terra`, which only exist with #4311's provider metadata. This PR retargets to main automatically when #4311 merges.

## Verification

- `cargo test -p codewhale-tui --bin codewhale-tui --locked -- hotbar` → 86 passed
- Release QA suite verified green in the source worktree before split (release_runtime_qa, three PTY scenarios); will re-verify in CI and in the final release sweep.

Part of the v0.8.68 milestone.

### DIFF
--- a/crates/tui/src/runtime_threads/tests.rs
+++ b/crates/tui/src/runtime_threads/tests.rs
@@ -1604,7 +1604,10 @@ async fn approval_required_awaits_external_decision_allow() -> Result<()> {
         })
         .await?;
 
-    let deadline = Instant::now() + Duration::from_secs(2);
+    // A busy or constrained runner can be quiet for more than one 200 ms poll
+    // even though the engine is still making progress. Keep polling until the
+    // actual deadline instead of treating the first quiet interval as failure.
+    let deadline = Instant::now() + Duration::from_secs(5);
     while Instant::now() < deadline && manager.pending_approvals_count() == 0 {
         sleep(Duration::from_millis(20)).await;
     }
@@ -1943,7 +1946,8 @@ async fn thinking_delta_emits_agent_reasoning_item() -> Result<()> {
                     completed_seen = true;
                 }
             }
-            _ => break,
+            Ok(Err(_)) => break,
+            Err(_) => continue,
         }
     }
     assert!(delta_seen, "expected item.delta with kind=agent_reasoning");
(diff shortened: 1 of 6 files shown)


Hmbown:
Fresh maintainer rerun on macOS at head f8d6e7fc89134bec1bf05bf496f211d2ae65bc8e:\n\n- 32 children live: 1.825s\n- sidebar visible: 2.521s\n- typing echo: 92ms, 719ms, 851ms\n- Esc cancellation: 1.394s\n- RSS: 61,904 KiB idle -> 187,040 KiB storm -> 326,368 KiB immediately after cancel\n\nCommand:\n\n    cargo test -p codewhale-tui --test release_runtime_qa --locked -- --ignored bench_thirty_two --nocapture --test-threads=1\n\nResult: PASS (1 passed, 9.35s). The interaction acceptance is reproducible; #4326 separately owns the retained-RSS plateau investigation so this release benchmark does not mislabel allocator retention as a proven leak.

Hmbown:
CI hardening follow-up found while validating #4323 against main:

- Windows failed only runtime_threads::tests::thinking_delta_emits_agent_reasoning_item.
- This is the timing flaw previously documented in #2223: the test declares a multi-second deadline but broke on its first 200 ms quiet poll.
- The test now continues polling through quiet intervals until its real 5-second deadline, while still failing immediately if the event channel closes.
- Net diff is confined to that reasoning-event test.

Local verification at head c238f4544db01731c31f32b6320e1fdf2d7cadf7:
- runtime_threads suite: 57/57 passed with 16 test threads.
- focused reasoning test: passed.
- cargo fmt --all --check: passed.

No provider calls were made.

PR #4370, opened 2026-07-15
feat: add TelecomJS provider support with configuration and catalog i…

baendlorel:
## Problem
After registering the TelecomJS (Telecom JiangSu) in the `custom` provider, the model picker only shows 1 model (`deepseek-v4-pro`) instead of all models returned by the `/v1/models` endpoint.

## Root Cause
`refresh_catalog_cache`/`fetch_catalog_delta` is never invoked in production — it only exists in test code. At startup, only the Models.dev catalog is refreshed, which has no entries for telecomjs. The picker therefore falls back to `model_completion_names_for_provider(Telecomjs)`, which returns only the default model.

## Fix
1. **`provider_lake.rs`** — Add `merge_live_offerings()` that merges new offerings into the existing live snapshot by `(provider, wire_model_id)` identity rather than replacing it. This allows Models.dev rows and per-provider rows to coexist.

2. **`client.rs`** — Add `spawn_active_provider_catalog_refresh()` associated function. When the active provider is Telecomjs, it spawns a background task that calls `fetch_catalog_delta()` (which already inherits metadata from the bundled catalog via case-insensitive matching) and merges all 52 models into the provider lake.

3. **`main.rs`** — Call the new function at startup, after the existing `models_dev_live::spawn_background_refresh()`.

4. **`config.rs`** — Expand the `model_completion_names_for_provider(Telecomjs)` fallback list from 1 to 11 models, so the picker shows useful options even before the background refresh completes.

## Testing
- `cargo check --workspace` — pass
- `cargo test -p codewhale-config` — 360 tests passed
- `cargo test -p codewhale-tui` (provider_lake + client catalog) — 14 tests passed
- `cargo build --release` — success

## Final Result

codewhale增加江苏tokenhub

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -298,6 +298,16 @@ pub struct ProvidersToml {
         alias = "grok"
     )]
     pub xai: ProviderConfigToml,
+    /// Jiangsu Telecom TokenHub — OpenAI-compatible AI gateway.
+    #[serde(
+        default,
+        skip_serializing_if = "ProviderConfigToml::is_empty",
+        alias = "telecom-js",
+        alias = "telecom_js",
+        alias = "telecomjs-cn",
+        alias = "tokenhub"
+    )]
+    pub telecomjs: ProviderConfigToml,
     /// Catch-all table for the dynamic OpenAI-compatible custom provider
     /// identity (#1519). Arbitrary `[providers.<name>]` tables are handled by
     /// the tui-side flatten map; this named slot keeps the canonical
@@ -408,6 +418,7 @@ impl ProvidersToml {
             ProviderKind::OpencodeGo => &self.opencode_go,
             ProviderKind::Meta => &self.meta,
             ProviderKind::Xai => &self.xai,
+            ProviderKind::Telecomjs => &self.telecomjs,
             ProviderKind::Custom => &self.custom,
         }
     }
@@ -448,6 +459,7 @@ impl ProvidersToml {
             ProviderKind::OpencodeGo => &mut self.opencode_go,
             ProviderKind::Meta => &mut self.meta,
             ProviderKind::Xai => &mut self.xai,
+            ProviderKind::Telecomjs => &mut self.telecomjs,
             ProviderKind::Custom => &mut self.custom,
         }
     }
@@ -2303,6 +2315,7 @@ impl ConfigToml {
                 ProviderKind::OpencodeGo => DEFAULT_OPENCODE_GO_BASE_URL.to_string(),
                 ProviderKind::Meta => DEFAULT_META_BASE_URL.to_string(),
                 ProviderKind::Xai => DEFAULT_XAI_BASE_URL.to_string(),
+                ProviderKind::Telecomjs => DEFAULT_TELECOMJS_BASE_URL.to_string(),
                 // The custom provider has no built-in endpoint; fall back to its
                 // descriptor placeholder so the lookup is total. Real custom
                 // routes always supply a configured base_url before this point.
@@ -2939,6 +2952,7 @@ fn default_model_for_provider(provider: ProviderKind) -> &'static str {
         ProviderKind::OpencodeGo => DEFAULT_OPENCODE_GO_MODEL,
         ProviderKind::Meta => DEFAULT_META_MODEL,
         ProviderKind::Xai => DEFAULT_XAI_MODEL,
+        ProviderKind::Telecomjs => DEFAULT_TELECOMJS_MODEL,
         // No built-in default model; the registry placeholder keeps this total.
         ProviderKind::Custom => provider.provider().default_model(),
     }
@@ -2980,6 +2994,7 @@ fn default_base_url_for_provider(provider: ProviderKind) -> &'static str {
         ProviderKind::OpencodeGo => DEFAULT_OPENCODE_GO_BASE_URL,
         ProviderKind::Meta => DEFAULT_META_BASE_URL,
         ProviderKind::Xai => DEFAULT_XAI_BASE_URL,
+        ProviderKind::Telecomjs => DEFAULT_TELECOMJS_BASE_URL,
         // No built-in default base URL; the registry placeholder keeps this total.
         ProviderKind::Custom => provider.provider().default_base_url(),
     }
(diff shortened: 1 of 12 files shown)


Hmbown:
Thanks @baendlorel—this is a real v0.9.1 gap, and the PR is now in the v0.9.1 queue. Current `main` already supports TokenHub as an OpenAI-compatible custom route, but it does not make the provider's live `/v1/models` inventory available in the picker, so the problem you traced is valid.

I reviewed the current patch against `main` at `07b6422d`. Before it is merge-ready, please rebase and address these two correctness boundaries:

1. `provider_lake` still has a single replaceable live snapshot. `spawn_background_refresh()` and the new provider refresh race: if the TelecomJS merge completes first, a later Models.dev `set_live_snapshot()` erases those rows. Please make live data source-scoped or otherwise prove both completion orders preserve both sets; add deterministic tests for Models.dev→TelecomJS and TelecomJS→Models.dev.
2. Matching the first bundled row solely by case-insensitive `wire_model_id` can copy capabilities and cost from an unrelated provider/version. Preserve unknown metadata unless there is an explicit canonical mapping, and test ambiguous IDs. A gateway exposing `DeepSeek-R1` does not by itself prove identical limits, prices, reasoning controls, or tool support.

Please also:

- amend the commit with a `Signed-off-by` trailer and force-push with lease;
- rerun the current full CI after rebasing;
- add a failure/race regression showing a catalog refresh never deletes previously published rows;
- verify the reasoning request fields against actual TelecomJS behavior rather than assuming every gateway model accepts the same `reasoning_effort` plus `thinking` combination.

The provider wiring and production refresh direction are worth preserving. If you no longer have time to refresh the branch, say so here and we can harvest the commit onto current main while retaining your authorship and credit.

Hmbown:
The TelecomJS live-model gap is still real, but this branch is not safe to merge as written. It races the single replaceable live snapshot, so a later Models.dev refresh can erase TelecomJS rows; it also borrows metadata from the first case-insensitive model-ID match across unrelated providers. Please rebuild on current main with source-scoped snapshots, both refresh-order tests, conservative unknown metadata, actual TelecomJS request-shape proof, DCO, and current CI. Contributor credit will be preserved if harvested.

Hmbown:
Thanks for digging into this — your root-cause analysis is confirmed correct: `fetch_catalog_delta`/`refresh_catalog_cache` (`crates/tui/src/client.rs:1464`, `:1559`) have no production call sites, so custom providers never load their live catalogs. That underlying defect is now tracked separately as #4462 with credit to you.

This PR can't merge in its current form, and we're retargeting it out of v0.9.1 to v0.9.2, for four concrete reasons:

1. **Single global snapshot.** `merge_live_offerings()` merges into the one global `LIVE_SNAPSHOT` (`crates/tui/src/provider_lake.rs:28`). Once production actually calls per-provider refresh, Models.dev and provider-specific refreshes can overwrite each other's rows. The fix needs source-layered catalog state (Models.dev layer + scoped per-provider layers, merged at read time by `(provider, wire_model_id)`) — specified in #4462.
2. **Fabricated metadata.** The PR copies price/capability facts from a home provider onto TelecomJS models whose IDs match. TelecomJS never supplied those facts; provider capability must stay route-effective and truthful. Custom-provider rows must carry only what that provider's own `/v1/models` returned, with unknown fields left unknown.
3. **Base URL and stale-catalog scope** need review against the same contract.
4. **Authorship/DCO.** The commit author email does not line up with a DCO sign-off we can accept as-is; please add a `Signed-off-by` trailer matching your GitHub identity (see CONTRIBUTING / the `Check Signed-off-by` CI job).

Path forward: the catalog-state rewrite lands via #4462 first; then this PR can be rebased down to just the TelecomJS provider preset (endpoints, default models, honest unknown metadata) on top of it. If you'd like to take the preset slice after #4462 lands, it's yours — say so and we'll assign it to you.

baendlorel:
> Thanks for digging into this — your root-cause analysis is confirmed correct: `fetch_catalog_delta`/`refresh_catalog_cache` (`crates/tui/src/client.rs:1464`, `:1559`) have no production call sites, so custom providers never load their live catalogs. That underlying defect is now tracked separately as #4462 with credit to you.
> 
> This PR can't merge in its current form, and we're retargeting it out of v0.9.1 to v0.9.2, for four concrete reasons:
> 
>     1. **Single global snapshot.** `merge_live_offerings()` merges into the one global `LIVE_SNAPSHOT` (`crates/tui/src/provider_lake.rs:28`). Once production actually calls per-provider refresh, Models.dev and provider-specific refreshes can overwrite each other's rows. The fix needs source-layered catalog state (Models.dev layer + scoped per-provider layers, merged at read time by `(provider, wire_model_id)`) — specified in [Custom provider live catalog is never refreshed in production (scoped catalog state rewrite) #4462](https://github.com/Hmbown/CodeWhale/issues/4462).
> 
>     2. **Fabricated metadata.** The PR copies price/capability facts from a home provider onto TelecomJS models whose IDs match. TelecomJS never supplied those facts; provider capability must stay route-effective and truthful. Custom-provider rows must carry only what that provider's own `/v1/models` returned, with unknown fields left unknown.
> 
>     3. **Base URL and stale-catalog scope** need review against the same contract.
> 
>     4. **Authorship/DCO.** The commit author email does not line up with a DCO sign-off we can accept as-is; please add a `Signed-off-by` trailer matching your GitHub identity (see CONTRIBUTING / the `Check Signed-off-by` CI job).
> 
> 
> Path forward: the catalog-state rewrite lands via #4462 first; then this PR can be rebased down to just the TelecomJS provider preset (endpoints, default models, honest unknown metadata) on top of it. If you'd like to take the preset slice after #4462 lands, it's yours — say so and we'll assign it to you.
Thanks for the detailed feedback and for tracking the underlying defect in #4462.

I've worked through the changes requested in your first review and pushed them to this branch (commit 85f674c). Summary:

Source-scoped live snapshots: provider_lake.rs now partitions the live snapshot by source (ModelsDev and per_provider), so a later Models.dev refresh no longer overwrites TelecomJS rows. Added deterministic tests for both completion orders (Models.dev→TelecomJS and TelecomJS→Models.dev).
Conservative cross-provider metadata: fetch_catalog_delta now only inherits family (a name-derived safe field) for cross-provider wire_model_id matches; all other unknown fields stay None instead of being copied from an unrelated provider. Added tests for ambiguous and same-provider matches.
Catalog refresh regression: added a test verifying a catalog refresh never deletes previously published rows.
Reasoning fields: verified against the TelecomJS API docs — /v1/chat/completions does not accept reasoning_effort or thinking (the latter is Anthropic Messages-only). TelecomJS is now excluded from apply_reasoning_effort.
DCO: commit includes Signed-off-by.
Rebased onto 07b6422d and force-pushed with lease. GitGuardian check has passed.
Hope these changes help. Happy to rebase further into a preset-only slice once #4462 lands — just let me know.

PR #4372, opened 2026-07-15
fix(skills): preserve inline task text

### LINKED ISSUE #3915: ux(skills): `$skill <task>` and `/<skill> <task>` silently discard the task text
## Problem
The natural Claude-Code-style invocation — `$debug why does auth fail` — activates the skill but throws away "why does auth fail". The skill is armed for the NEXT message and the user must retype their request. `/skill <name> <args>` is worse: the args are treated as part of the name and the lookup fails.

## Evidence
- `crates/tui/src/commands/mod.rs:119-127` — the `$name rest` arg is extracted and passed to `run_skill_by_name`, whose signature ignores it (`_arg`, `groups/skills/skills.rs:288-299`); same for the unified `/<skill> args` path (`mod.rs:188`).
- `activate_skill` only stores the body in `app.active_skill` (`skills.rs:343`), consumed by the next queued message (`tui/ui.rs:5984-5987`); `looks_like_slash_command_input` routes the whole `$…` line into dispatch (`app.rs:102-120`), so the trailing prose never becomes a message.
- `skills.rs:313-324` — `/skill deep-research fix auth` fails with "Skill 'deep-research fix auth' not found".
- Contradicts the shipped design doc: `docs/SKILL_INVOCATION_DESIGN.md` §2 ("the rest of the line is the user's request, passed through").

## Impact
The single most-used flow of a skill system loses the user's request text every time.

## Proposed approach
When a non-empty arg accompanies the skill name, immediately queue it as the user message with the skill instruction attached (reuse `QueuedMessage::new(input, skill_instruction)` from `ui.rs:5986`); keep defer-to-next-message only for bare `$name`/`/name`. Make `/skill <name> <args>` split name from args.

## Acceptance criteria
- `$foo do X`, `/foo do X`, and `/skill foo do X` each send one turn with skill foo active and user text "do X".
- Bare `$foo` still arms the skill for the next message.
- Tests cover args-with-skill and a skill literally named `install` (subcommand collision).

nightt5879:
## Summary

- dispatch trailing task text in the same turn for `$<skill> do X`, `/<skill> do X`, and `/skill <skill> do X`
- keep bare `$<skill>` activation armed for the next user message
- preserve `/skill install` as the management command while allowing a literal `install` skill through `$install ...` and `/install ...`
- carry the active skill through immediate, queued, and steer message paths exactly once

## Root cause

The unified skill command path accepted trailing arguments but discarded them before activation. The `/skill <name> <task>` path also treated the whole raw argument string as the skill name, so it could not both select the skill and dispatch the task.

## Impact

Inline skill invocations now behave as a single request instead of silently dropping the user's task. Existing bare activation and skill-management behavior remain compatible.

## Validation

- `cargo test -p codewhale-tui --bin codewhale-tui --locked commands::groups::skills::skills::tests -- --test-threads=1` (27 passed)
- `cargo test -p codewhale-tui --bin codewhale-tui --locked inline_skill_request_keeps_instruction_when_busy_queueing -- --test-threads=1` (1 passed)
- CI-equivalent `cargo clippy --workspace --all-features --locked -- -D warnings ...` passed
- `cargo test --workspace --all-features --locked` reached 6667 passed / 3 parallel-state failures; all three failures passed individually with `--test-threads=1`
- `cargo build --release --locked -p codewhale-cli -p codewhale-tui`; both binaries start and report version 0.8.68
- offline eval harness succeeded (6 steps, 0 tool errors)
- version drift, provider registry, changelog sync, diff, Cargo.lock, and co-author credit checks passed

Closes #3915

### DIFF
--- a/.github/AUTHOR_MAP
+++ b/.github/AUTHOR_MAP
@@ -50,6 +50,7 @@ zlh124 = zlh124 <56312993+zlh124@users.noreply.github.com>
 LeoLin990405 = LeoLin990405 <101193087+LeoLin990405@users.noreply.github.com>
 THINKER-ONLY = THINKER-ONLY <181556007+THINKER-ONLY@users.noreply.github.com>
 nightt5879 = nightt5879 <87569709+nightt5879@users.noreply.github.com>
+CCChisato = Fushimi Rio <158128433+CCChisato@users.noreply.github.com>
 aznikline = aznikline <27564626+aznikline@users.noreply.github.com>
 Aznable = aznikline <27564626+aznikline@users.noreply.github.com>
 anikline@gmail.com = aznikline <27564626+aznikline@users.noreply.github.com>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Preserve the trailing task in inline skill invocations: `$<skill> do X`,
+  `/<skill> do X`, and `/skill <skill> do X` now dispatch `do X` immediately
+  with the selected skill instructions attached, while a bare `$<skill>` still
+  arms the next message. A skill literally named `install` remains callable
+  through `$install` and `/install` without shadowing the existing `/skill
+  install` management path (#3915, with fix direction from @CCChisato).
 - Make offline `scorecard` pricing provider-aware: `turn_end` records carry the
   effective route and a non-secret billing surface, runtime exports and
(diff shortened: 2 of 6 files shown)


nightt5879:
@codex review

Please review final head 66a6edf5, including inline task dispatch, bare activation compatibility, the exact-name install skill collision, and queued skill delivery.

Hmbown:
Thank you, @nightt5879 and @CCChisato — merged for v0.8.68 with the original commit and co-author credit preserved. The release candidate keeps the additional queue-delivery regression and records both contributors in the release notes.

PR #4379, opened 2026-07-16
feat(mcp): add cancellable OAuth login API

### LINKED ISSUE #4380: MCP OAuth login has no cancellation boundary for host applications
## Problem

`perform_oauth_login_for_server` owns a local callback listener until the browser authorization flow completes. A host application cannot stop that flow or reliably wait for the listener to close before starting a replacement flow.

## Proposal

Add an opt-in cancellation-token variant that drops the in-flight login future before returning. The existing API and its behavior remain unchanged.

## Scope

This is an API-level cancellation boundary only; it does not alter OAuth token handling, callback validation, or default authorization behavior.

h3c-hexin:
## Summary

Fixes #4380. Add an opt-in `perform_oauth_login_for_server_with_cancel` API while preserving the existing login API. Cancelling waits for the in-flight future to drop, which closes its callback listener before a replacement flow begins.

## Testing

- `cargo fmt --check`
- `cargo test -p codewhale-tui --bin codewhale-tui mcp::oauth::tests::cancellable_oauth_drops_in_flight_flow_before_returning`

## Checklist

- [x] Added a focused unit test
- [x] Documented the public API

### DIFF
--- a/crates/tui/src/mcp/oauth.rs
+++ b/crates/tui/src/mcp/oauth.rs
@@ -16,6 +16,7 @@ use sha2::{Digest, Sha256};
 use tiny_http::{Response, Server};
 use tokio::sync::{Mutex, oneshot};
 use tokio::time::timeout;
+use tokio_util::sync::CancellationToken;
 use urlencoding::decode;
 
 use super::McpServerConfig;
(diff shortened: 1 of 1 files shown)


Hmbown:
Maintainer current-main check: commit 3b3861b cherry-picks cleanly onto origin/main 17ac2681. On that exact base, cargo fmt --all -- --check and git diff --check pass, and mcp::oauth::tests::cancellable_oauth_drops_in_flight_flow_before_returning passes (1/1). The existing Linux, macOS, and Windows CI is also green. I updated the body to Fixes #4380 because the patch matches that issue scope exactly.

Hmbown (approved):
Reviewed the rebased cancellation path at 68c3ab279c1f1a029d3d4763e918e1b90e2e563c. The callback listener is guarded so dropping the OAuth future unblocks the server, the existing API remains compatible, and the full macOS/Linux/Windows matrix plus lint, DCO, version drift, and npm smoke are green.

PR #4381, opened 2026-07-16
fix(automation): keep hourly schedules anchored

h3c-hexin:
## Summary

Allow `HOURLY` automations to opt into `BYHOUR` and/or `BYMINUTE` anchors. Future slots are calculated against the automation creation date, so retries, resumes, and day changes do not shift the configured phase. Existing hourly rules without an anchor retain their current behavior.

## Testing

- `cargo fmt --check`
- `cargo test -p codewhale-tui --bin codewhale-tui automation_manager::tests::hourly_rrule`

## Checklist

- [x] Added focused regression tests
- [x] No migration is needed: existing RRULEs remain valid

### DIFF
--- a/crates/tui/src/automation_manager.rs
+++ b/crates/tui/src/automation_manager.rs
@@ -169,6 +169,8 @@ pub enum AutomationSchedule {
     Hourly {
         interval_hours: u32,
         byday: Option<Vec<Weekday>>,
+        anchor_hour: Option<u32>,
+        anchor_minute: Option<u32>,
     },
     Weekly {
         byday: Vec<Weekday>,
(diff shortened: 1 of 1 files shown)


Hmbown:
Roadmap triage: this PR is now in v0.9.1 for AutomationManager reliability and is an explicit dependency of #4398's later /loop UX. The anchored schedule behavior should remain the source of truth for loop next-run previews.

Hmbown:
Current-main review found one release-blocking recurrence bug before branch refresh: the anchored path resolves a local anchor and then advances it with elapsed `Duration::seconds`. For `INTERVAL=24`, that preserves 24 elapsed hours rather than the requested wall clock, so an 08:30 automation becomes 09:30 after the spring DST transition and 07:30 after the fall transition. The fix should step in local naive/calendar time, then resolve each candidate through the timezone (with explicit nonexistent/ambiguous-time behavior). Please add spring-forward, fall-back, resume/restart, and BYDAY coverage; once that lands I can rerun it on exact current main. The patch otherwise cherry-picks cleanly onto origin/main 17ac2681.

Hmbown:
Changes needed before v0.9.1 merge: advancing an anchored schedule by elapsed seconds shifts wall-clock schedules across DST (for example, daily 08:30 becomes 09:30 or 07:30). Please compute the next recurrence in the schedule's calendar/time-zone semantics, define nonexistent and ambiguous local-time behavior, and add spring-forward, fall-back, restart/resume, and BYDAY regressions. Existing green CI is on a 163-behind base.

Hmbown:
Closing in favor of the credited replacement #4458, which has now merged into `main`.

Your calendar-anchored hourly recurrence fix landed with full authorship preserved: your commit was harvested into #4458 with you as commit author plus `Co-authored-by` and `Signed-off-by` receipts, and you are credited in `docs/CONTRIBUTORS.md` and the v0.9.1 changelog entries (`#4381 by @h3c-hexin`).

The replacement was needed to rebase onto current main and add maintainer documentation, but the fix itself — BYHOUR/BYMINUTE as the initial wall-clock phase, calendar-local interval advancement, DST gaps skipped, folds using the first occurrence, and the persisted `created_at` anchor across create/update/resume/restart — is yours. Thank you for the contribution!

PR #4383, opened 2026-07-16
fix(shell): avoid blocked reader joins after Windows kill

h3c-hexin:
## Summary

When a background shell is killed on Windows, do not synchronously join reader threads that can remain blocked on inherited pipe handles. Other terminal states and non-Windows platforms retain the existing join behavior so final output is still collected.

A Windows-only regression test holds a reader thread deliberately and verifies that cleanup returns promptly after `Killed`.

## Testing

- `cargo fmt --check`
- `cargo test -p codewhale-tui --bin codewhale-tui tools::shell::tests::test_orphaned_subprocess_does_not_block_collect_output`
- Windows-only regression test added; this host has no Windows Rust target, so it is left for CI.

## Checklist

- [x] Added a focused regression test
- [x] No behavior change outside Windows cancellation

### DIFF
--- a/crates/tui/src/tools/shell.rs
+++ b/crates/tui/src/tools/shell.rs
@@ -608,10 +608,10 @@ impl BackgroundShell {
         #[cfg(windows)]
         terminate_and_close_windows_job(self.windows_job.take());
         if let Some(handle) = self.stdout_thread.take() {
-            let _ = handle.join();
+            finish_background_reader(handle, &self.status);
         }
         if let Some(handle) = self.stderr_thread.take() {
-            let _ = handle.join();
+            finish_background_reader(handle, &self.status);
         }
         self.stdin = None;
         self.child = None;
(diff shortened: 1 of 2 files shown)


Hmbown:
Maintainer current-main check: commit 37b4c4a cherry-picks cleanly onto origin/main 17ac2681. On that exact base, cargo fmt --all -- --check and git diff --check pass. The added regression is correctly Windows-gated (0 tests selected on macOS), while this PR current head already passed GitHub Windows CI. The remaining release-lane gate is a fresh Windows run after the branch is updated to current main.

Hmbown (approved):
Replayed onto exact current main and reviewed the Windows-only cancellation path. The targeted local checks are clean, and the full Ubuntu/macOS/Windows matrix plus lint, DCO, version drift, npm smoke, and automated review are green.

PR #4425, opened 2026-07-17
🧪 Add test coverage for ToolError::missing_field display format

Hmbown:
🎯 **What:** Improved the testing suite by enhancing the test for `ToolError::missing_field` constructor in `crates/tools/src/lib.rs`.
📊 **Coverage:** The test now verifies not only the state-based construction (`matches!`) but also the exact Display implementation output (using `err.to_string()`). Refactored variable destructuring to prevent partial moves, allowing proper string comparisons.
✨ **Result:** Better error message validation prevents accidental changes to `ToolError` formatting.

---
*PR created automatically by Jules for task [2590320779045522146](https://jules.google.com/task/2590320779045522146) started by @Hmbown*

### DIFF
--- a/crates/tools/src/lib.rs
+++ b/crates/tools/src/lib.rs
@@ -599,7 +599,11 @@ mod tests {
     #[test]
     fn tool_error_missing_field_constructor() {
         let err = ToolError::missing_field("my_field");
-        assert!(matches!(err, ToolError::MissingField { field } if field == "my_field"));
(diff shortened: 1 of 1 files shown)


Hmbown:
Closing this intake PR because the exact ToolError::missing_field display contract is already covered by tool_error_display_matches_legacy_text in the same module. The constructor-shape assertion remains covered separately, so this change would duplicate the contract without increasing behavior coverage. Thanks, Jules.

Issue #3495, opened 2026-06-23
v0.8.66: Adopt Moraine as CodeWhale's memory backend (ingestion adapter + MCP recall)


Hmbown:
## Goal

Adopt **Moraine** (https://github.com/eric-tramel/moraine, Apache-2.0) as CodeWhale's long-term
agent-memory backend: Moraine ingests CodeWhale's persisted sessions losslessly and exposes them as
searchable MCP recall tools (`search`, `search_conversations`, `list_sessions`, `get_session`,
`open`). This **supersedes** the in-repo push/inject memory path (`crates/tui/src/memory.rs`, which
prepends a `<user_memory>` block into the system prompt) in favor of **pull/recall via Moraine MCP**.

Local-first, no prompt injection (MCP is read-only retrieval). CodeWhale keeps persisting sessions;
Moraine indexes them. Supersedes community memory PRs **#3381** (@pkeging, memory tags) and
**#2933** (@cy2311, hippocampal v2 — glossary/namespaces/rollback/auto-inject) — their ideas inform
the recall layer but the implementation is Moraine-backed, not in-repo.

This is scoped for **v0.8.66**, not 0.8.65.

## Research already done (design is concrete)

Moraine fork: `Hmbown/moraine` (upstream `eric-tramel/moraine`, on `main` @ v0.6.2).

### Key finding — the `session_json` ingest format already fits CodeWhale
`crates/moraine-ingest-core/src/dispatch.rs::process_session_json_file` reads a whole-file JSON
snapshot and extracts `session_doc["messages"]` — **exactly CodeWhale's shape**
(`~/.codewhale/sessions/{id}.json` → `{ schema_version, metadata, messages[], artifacts, system_prompt }`).
It builds synthetic `session_meta` + per-`session_message` records and checkpoints by message index
(incremental ingest), then dispatches each to the harness adapter's `normalize()`.

### The fork is already half-wired
`crates/moraine-config/src/lib.rs::default_sources()` **already contains** the CodeWhale entry:
`harness: "codewhale"`, `glob: "~/.codewhale/sessions/*.json"`, `format: SOURCE_FORMAT_SESSION_JSON`.

## Work items (Moraine fork)

1. **Add adapter** `crates/moraine-ingest-core/src/sources/codewhale.rs` implementing `IngestSource`
   (`harness()="codewhale"`, `format()=SessionJson`), mirroring the Hermes session handlers
[...]

PR #4442, opened 2026-07-17
Remove legacy memory composition block in refresh_system_prompt

Hmbown:
Removes the legacy memory block composition logic from `refresh_system_prompt` and related system prompt setup paths (`Engine::new`, `build_app_system_prompt_with_goal`). The old memory retrieval block is replaced by setting `user_memory_block: None` in `PromptSessionContext` since the new Moraine recall method is stable. Also removed the associated TODO comment in `tool_setup.rs`.

---
*PR created automatically by Jules for task [3913683453046725750](https://jules.google.com/task/3913683453046725750) started by @Hmbown*

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -1011,10 +1011,6 @@ impl Engine {
         // Set up stable system prompt with project context (default to agent mode).
         // Per-turn working-set metadata is injected into the latest user
         // message at request time so file churn does not rewrite this prefix.
-        let user_memory_block = crate::memory::compose_block(
-            config.memory_enabled && !config.moraine_fallback, // TODO(v0.8.71): remove when Moraine recall stable; see #3490, #3495
-            &config.memory_path,
-        );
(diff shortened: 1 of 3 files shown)


Hmbown:
Moved to v0.9.4 draft. #3495 is still open and requires the Moraine adapter, MCP recall wiring, and end-to-end proof before the legacy prompt injection is config-gated/deprecated. This slice alone makes documented memory writes unreadable while leaving config, tools, commands, and docs active.

Hmbown:
Closing as superseded. This draft targeted the legacy memory push/inject path, which no longer has the shape this change assumed — the memory surface was reworked after this was opened. Re-deriving the intent against current `main` is cheaper than reconciling this diff (see AGENTS.md, "Intent is the artifact"). Nothing here is lost: the memory work is tracked in the v0.9.2 memory pass (smartgrok M-items), and the do-not-delete guardrail still lists `tui/src/memory.rs` as actively imported.

PR #4447, opened 2026-07-17
chore: remove legacy quick-add functionality

Hmbown:
This PR removes the legacy `# foo` memory quick-add functionality from the composer's event loop, as Moraine recall is now considered stable. Inputs starting with `#` will now fall through to normal turn submission instead of being intercepted. 

Changes include:
- Removing the interception block in `run_event_loop`.
- Deleting the `is_memory_quick_add`, `should_intercept_memory_quick_add`, and `handle_memory_quick_add` helper functions.
- Deleting the dedicated `memory_quick_add_tests` test module.

Fixes #3490, #3495.

---
*PR created automatically by Jules for task [6682938610938053130](https://jules.google.com/task/6682938610938053130) started by @Hmbown*

PR #4446, opened 2026-07-17
Remove remember tool registration when Moraine recall stable

Hmbown:
This pull request removes the `remember` tool registration logic from the agent's tool setup since Moraine recall is now considered stable. 

- Removes `should_register_remember_tool` function and its tests.
- Removes `with_remember_tool` block from `ToolRegistryBuilder`.
- Removes the `memory_tool_enabled` configuration logic passed to the surface options.
- Deletes the dead `crates/tui/src/tools/remember.rs` file.
- Cleans up catalog, core engine, and subagent tests that referenced the deprecated tool.

---
*PR created automatically by Jules for task [15733292684462658497](https://jules.google.com/task/15733292684462658497) started by @Hmbown*

PR #4444, opened 2026-07-17
Remove legacy memory push/inject in build_headless_context_report

Hmbown:
Removed the block handling legacy memory push/inject in `build_headless_context_report` within `crates/tui/src/context_report.rs` (marked by a v0.8.71 TODO for Moraine recall stability), as well as the related unit test `headless_context_report_omits_legacy_memory_when_moraine_fallback_enabled`.

---
*PR created automatically by Jules for task [1440907162443058655](https://jules.google.com/task/1440907162443058655) started by @Hmbown*

PR #4455, opened 2026-07-17
Remove legacy memory push/inject in report building

Hmbown:
This PR removes the legacy memory push/inject blocks and related properties that were guarded by a TODO indicating their planned removal. The `UserMemory` and `ConfigEnabled` tokens have also been cleaned up along with unused properties like `App::moraine_fallback`.

---
*PR created automatically by Jules for task [3120618009560725208](https://jules.google.com/task/3120618009560725208) started by @Hmbown*

### DIFF
--- a/crates/tui/src/context_report.rs
+++ b/crates/tui/src/context_report.rs
@@ -140,7 +140,6 @@ pub enum SourceKind {
     CompactionRelayTemplate,
     RuntimePolicy,
     EnvironmentBlock,
-    UserMemory,
     SessionGoal,
     HandoffRelay,
     ToolSchemas,
@@ -155,7 +154,6 @@ pub enum SourceKind {
 pub enum ActivationReason {
     AlwaysOn,
     FilePresent,
-    ConfigEnabled,
     RuntimeState,
     PerRequest,
     Omitted,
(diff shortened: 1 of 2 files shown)


Hmbown:
Thanks — this belongs with the Moraine migration in v0.9.4, not the v0.9.1 release train, and I have moved it there and marked it draft.

This cannot land as an isolated cleanup yet. It removes `UserMemory` reporting and the `App::moraine_fallback` projection while legacy memory is still active in the engine, quick-add interception, remember-tool registration, runtime-thread configuration, and CLI surface. That would make the context report stop disclosing memory that can still be injected, while #3495 still requires the Codewhale Moraine adapter, MCP wiring, and end-to-end recall proof before the legacy path is deprecated.

Please hold this as part of one coherent #3495 cutover. The eventual patch needs to remove or migrate every legacy consumer together, preserve truthful context/runtime reporting during the transition, and prove Codewhale session ingest plus MCP recall end to end. It also overlaps the held Moraine cleanup PRs #4442, #4444, #4446, and #4447, so those should be consolidated rather than merged independently.

Hmbown:
Closing as superseded. This draft targeted the legacy memory push/inject path, which no longer has the shape this change assumed — the memory surface was reworked after this was opened. Re-deriving the intent against current `main` is cheaper than reconciling this diff (see AGENTS.md, "Intent is the artifact"). Nothing here is lost: the memory work is tracked in the v0.9.2 memory pass (smartgrok M-items), and the do-not-delete guardrail still lists `tui/src/memory.rs` as actively imported.

PR #4475, opened 2026-07-17
fix(tools): retain read-before-edit snapshots across turns

Angel-Hair:
## Summary

After compaction, `edit_file` rejects edits on previously-read files with "not been read in this session". The root cause is not in `edit_file` — it correctly requires a prior `read_file` invocation. The problem is in compaction: it deduplicates `read_file` results into `<TOOL_RESULT_REF>` references, telling the model "the content is still available", but the in-memory `file_read_tracker` is not updated. The model sees the content and assumes it can edit, but the tracker sees no record and rejects the edit. The two mechanisms give the model contradictory signals.

## Root cause

`compact_tool_result_for_wire` classifies `read_file` as dedup-eligible and collapses identical large results into `<TOOL_RESULT_REF>`. The dedup logic saves tokens but skips the tool invocation — so `note_file_read` (which populates `file_read_tracker`) is never called. Meanwhile, `edit_file`'s `require_fresh_file_read` checks `file_read_tracker` and rejects when no snapshot is found.

## Fix

Exclude `read_file` from the dedup_eligible predicate so that after compaction, file contents are naturally absent from context. This makes the model's context accurately reflect what it can act on: file contents are gone, `file_read_tracker` is empty, everything is consistent. The model will naturally re-read files before editing.

## Testing

- [x] `cargo fmt -- --check`
- [x] `cargo clippy -p codewhale-tui` (zero warnings)
- [x] `cargo test -p codewhale-tui -- client::chat::stream_decoder` (34/34 passed)

## Checklist

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -1471,7 +1471,8 @@ fn compact_tool_result_for_wire(
     // the model to chase a reference, and there's no retrieval burden to
     // satisfy, so both predicates are false.
     let persist_eligible = original_chars >= TOOL_RESULT_DEDUP_MIN_CHARS;
-    let dedup_eligible = persist_eligible && !is_mutation_tool(tool_name);
+    let dedup_eligible =
+        persist_eligible && !is_mutation_tool(tool_name) && tool_name != "read_file";
 
     if dedup_eligible && let Some(previous) = seen_tool_results.get(&sha) {
         // Re-check persistence before emitting a ref. If the file is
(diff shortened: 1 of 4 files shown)


Hmbown:
Maintainer amendment pushed at `f6d6516d6` while preserving @Angel-Hair's original commit and authorship.

The reported failure is real, but wire compaction does not skip `read_file` execution: it only serializes already-executed tool results for a later model request. The actual mismatch was that `Engine::build_tool_context` created a fresh `FileReadTracker` for every turn even though the contract calls it session-scoped.

The amended PR now:

- keeps one read tracker on the engine and shares it with each transient turn context;
- preserves identical `read_file` result dedup and the existing retrieval contract;
- preserves stale-file rejection after an observed file changes;
- adds a cross-turn regression proving read -> later context -> edit authority works.

Local focused gates passed:

- exact cross-turn tracker regression: 1 passed;
- medium identical-result dedup regression: 1 passed;
- large mutation-result inline/persistence regressions: 2 passed;
- `cargo fmt --all --check` and `git diff --check`.

Fresh exact-head CI is now running. Thank you for surfacing this — the report led directly to the lifetime fix.

PR #4476, opened 2026-07-17
fix(tools): make apply_patch replacement explicit and compatible

Angel-Hair:
## Summary

`apply_patch` accepts a `changes` parameter that performs full file replacement, not patching. The name is easily confused with the `patch` parameter (which applies a unified diff). When both are passed simultaneously, `changes` silently takes precedence and `patch` is discarded — no error, no warning, and the file is overwritten with only the `changes` content.

## Root cause

In `execute()`, the `changes` branch returns immediately if present, making the `patch` branch unreachable. The JSON schema declares `oneOf` mutual exclusivity, but this is advisory — the Rust handler enforces precedence, not rejection.

## Fix

- Rename `changes` to `replace` across JSON schema, code, function names, enum variants, error messages, and tests
- Add explicit conflict detection: if both `patch` and `replace` are present, return an error telling the model to choose one

## Testing

- [x] `cargo fmt -- --check`
- [x] `cargo clippy -p codewhale-tui` (zero warnings)
- [x] `cargo test -p codewhale-tui -- tools::apply_patch` (26/26 passed)

## Checklist

### DIFF
--- a/crates/tui/src/client/responses.rs
+++ b/crates/tui/src/client/responses.rs
@@ -1140,11 +1140,11 @@ mod tests {
                 "type": "object",
                 "properties": {
                     "patch": {"type": "string"},
-                    "changes": {"type": "array"}
+                    "replace": {"type": "array"}
                 },
                 "oneOf": [
                     {"required": ["patch"]},
-                    {"required": ["changes"]}
+                    {"required": ["replace"]}
                 ]
             }),
             allowed_callers: None,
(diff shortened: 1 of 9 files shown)


Hmbown:
Maintainer compatibility amendment is now on the contributor branch at `e23cd2b9f`.

The original report and contributor commit remain intact. The follow-up keeps `replace` as the canonical field while accepting deprecated `changes`, routes execution, preflight, repo-law, approval, diagnostics, and UI previews through one parser, and rejects all three ambiguous pairs before any file write (`patch` + `replace`, `patch` + `changes`, and `replace` + `changes`). It also corrects the Responses schema constraint note that failed the first macOS/Windows run.

Verification on the identical final source tree: 29 apply-patch tests; Responses and schema-sanitizer regressions; repo-law, approval-key, approval-preview, edited-path, and tool-routing coverage; `cargo check -p codewhale-tui`; formatting and diff checks. Fresh exact-head CI is now the final merge gate.

Thank you @Angel-Hair for finding and fixing the silent precedence bug.

PR #4505, opened 2026-07-17
fix(auth): isolate xAI device login from Tokio

### LINKED ISSUE #4410: Restore xAI device-code OAuth login and surface endpoint errors
## Summary

`/auth xai-device` currently fails immediately with `Failed to parse xAI device-code response`.

## Root cause

CodeWhale posts to a hard-coded device authorization path:

```
https://auth.x.ai/oauth2/device/code
```

The installed official Grok CLI (0.2.101) uses `/oauth2/device`. CodeWhale then calls `response.json()` before checking the HTTP status or content type, so a stale-path 404/HTML/non-JSON response is reported only as a JSON parse failure.

Relevant implementation: `crates/tui/src/xai_oauth.rs` (`request_device_code`).

## Expected

- Discover `device_authorization_endpoint` from `https://auth.x.ai/.well-known/openid-configuration` instead of hard-coding the path; retain a documented fallback only if necessary.
- Check status and content type first; return bounded/redacted diagnostic text such as HTTP status and expected JSON rather than an opaque parse error.
- Verify current scopes against the official Grok CLI flow (including whether `team:read` remains accepted).

## Workaround

Until fixed, use the official `grok login --device-auth` flow and configure `[providers.xai]
 auth_mode = "oauth"`, or use an `XAI_API_KEY`. Do not paste credentials into issue comments.

## Tests

Cover current device endpoint, legacy/non-JSON failure handling, discovery fallback, and a structured `invalid_scope` response.

Hmbown:
## Summary

- run the synchronous reqwest device discovery, request, browser handoff, polling, and persistence flow on Tokio's blocking pool behind one async xAI login entry point
- make both `codewhale auth xai-device` and the TUI provider flow await that shared boundary
- add a multi-thread Tokio regression that reaches loopback OIDC discovery and the advertised device endpoint, then fails before browser/polling/persistence
- extend the existing v0.9.1 #4410 changelog entry without replacing its provenance

## Root cause

reqwest 0.13's blocking client owns a private Tokio runtime. The CLI previously built and dropped that client directly inside Codewhale's `#[tokio::main]` runtime, so the live command panicked before it could display a device code. The TUI had a separate `block_in_place` workaround. A single `spawn_blocking` boundary now owns the complete synchronous workflow for both callers.

## Exact-head verification

Head: `31b148878b9600c78cd988c76cc6afd5b6966aaf`
Base: `8ab67a0bc03ac78406677339f3e4e3714e138885`

- `cargo test --workspace --all-features --locked`
- terminal cancellation regression stress — 100/100 passed
- `cargo test -p codewhale-tui --bin codewhale-tui --locked xai_oauth -- --nocapture` — 26 passed
- `cargo clippy -p codewhale-tui --all-targets --locked -- -D warnings`
- `./scripts/release/check-ohos-deps.sh`
- `python3 scripts/check-coauthor-trailers.py --range origin/main..HEAD --check-authors`
- actual `cargo run -p codewhale-tui --bin codewhale-tui --locked -- auth xai-device` against a process-local OIDC server: discovery GET and advertised device POST completed; the CLI returned the mock HTTP 400 `invalid_scope` error with no panic

The exact head also hardens the pre-existing macOS PTY cancellation race exposed by CI: cancellation confirmation is retried within the existing two-second contract and fails closed if the shell never acknowledges it. This was also observed in CI runs `29604840054` and `29544532601`.

No real xAI login or endpoint was used. The dogfood server refused the device request before any browser launch, token polling, or auth-file write.

Fixes #4410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,10 @@ site's draft, feed, login, and content-watch boundaries.
 - Restore xAI/Grok device-code OAuth login against the live xAI OIDC
   contract: discovery with issuer/endpoint validation and documented
   fallbacks, user-principal scope set, RFC 8628 `slow_down` backoff capped at
-  code expiry, and bounded, sanitized error reporting for denial, expiry, and
-  malformed responses (#4410).
+  code expiry, bounded, sanitized error reporting for denial, expiry, and
+  malformed responses, and a shared blocking-worker boundary for both CLI and
+  TUI login so reqwest's blocking client never creates or drops its private
+  runtime inside Codewhale's Tokio runtime (#4410).
 - Anchor `FREQ=HOURLY` automations with `BYHOUR`/`BYMINUTE` to persisted
   local-calendar slots so intervals keep their wall-clock phase across DST,
   restart, resume, RRULE updates, duplicate-slot recovery, and post-run
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -54,8 +54,10 @@ site's draft, feed, login, and content-watch boundaries.
 - Restore xAI/Grok device-code OAuth login against the live xAI OIDC
   contract: discovery with issuer/endpoint validation and documented
   fallbacks, user-principal scope set, RFC 8628 `slow_down` backoff capped at
-  code expiry, and bounded, sanitized error reporting for denial, expiry, and
-  malformed responses (#4410).
+  code expiry, bounded, sanitized error reporting for denial, expiry, and
+  malformed responses, and a shared blocking-worker boundary for both CLI and
+  TUI login so reqwest's blocking client never creates or drops its private
+  runtime inside Codewhale's Tokio runtime (#4410).
 - Anchor `FREQ=HOURLY` automations with `BYHOUR`/`BYMINUTE` to persisted
   local-calendar slots so intervals keep their wall-clock phase across DST,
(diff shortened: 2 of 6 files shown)


Hmbown:
Pushed `31b148878b9600c78cd988c76cc6afd5b6966aaf` to resolve the macOS CI failure.

Root cause: this is a pre-existing `terminal/cancel` race, not an xAI regression or runner-only failure. Canonical PTYs can flush the single completion-sentinel resend while processing ETX; on a busy runner the foreground `sleep 10` then runs to completion. The same test also failed in CI runs `29604840054` and `29544532601`.

The production cancel path now retries confirmation within its existing 2-second contract and returns an error instead of claiming an unconfirmed cancellation. The regression test calls that production helper and verifies exit 130 plus subsequent session reuse.

Exact-head local evidence:
- targeted cancellation regression: passed
- direct targeted stress: 100/100 passed
- `cargo test --workspace --all-features --locked`: passed (including PTY/release-runtime suites)
- `git diff --check`

Fresh GitHub checks are now pending on this exact head.

Hmbown:
Maintainer integration note: the runtime-isolation and PTY-cancellation fixes are ready to land, but #4410 will remain open after merge. The separate #4507 consent boundary still needs Codewhale-owned xAI login storage, default-disabled external credential access, and live-account request verification where access permits. The closing keyword in this branch therefore does not represent full issue acceptance.

Issue #4479, opened 2026-07-17
BUG: TUI rendering glitch — missing / extra spaces in text, recovers on mouse selection


SparkofSpike:
## Description

Intermittent rendering glitches in the TUI: text occasionally loses characters or gains extra spaces. Selecting the affected text with the mouse immediately restores correct rendering.

## Environment
- OS: Windows 10/11
- Terminal: Windows Terminal
- codewhale version: v0.9.0 (commit d167c07c9628)
- Settings: `composer_vim_mode = "normal"`, `bracketed_paste = false`

## Steps to Reproduce
1. Use codewhale normally for ~15-30 minutes
2. At some point text in the transcript area will render incorrectly — missing characters or extra stray spaces between words
3. Click and drag to select the garbled text with the mouse → text immediately snaps back to correct rendering

## Expected Behavior
TUI rendering remains correct at all times.

## Actual Behavior
Occasional rendering artifacts (dropped glyphs, phantom spaces). The underlying data is intact — text recovers on any repaint trigger (e.g. mouse selection, resize). This suggests a terminal diff/buffer synchronization issue rather than data corruption.

## Possible Root Cause
The ratatui diff-based incremental rendering may produce an incorrect diff in some edge case, leaving stale cells in the terminal buffer. Mouse selection forces a full repaint that flushes the buffer.

Hmbown:
Thanks, @SparkofSpike. The fact that mouse selection or resize repairs the text is useful evidence that this is repaint/diff state rather than transcript corruption. I’ve added it to the v0.9.1 Windows/TUI workdown; we’ll first audit existing invalidation/full-redraw paths and add the narrowest defensible fix and verification. We’ll preserve your report credit if it lands.

Hmbown:
We audited the current repaint paths and do not yet have evidence for a safe code change; a timer/full-clear workaround would mask the fault and cause flicker. Could you capture one more A/B on the affected Windows machine?

1. Post Windows Terminal’s exact version (`Settings → About`) and whether hardware acceleration/software rendering is changed from default.
2. Run `/set synchronized_output off`, restart CodeWhale, and note whether the corruption still appears during another 15–30 minute session. This isolates DEC 2026 synchronized output from ratatui diff/cache behavior.
3. If it appears, press `Ctrl+L` once and confirm whether that repairs it exactly like selection/resize.
[...]

PR #4509, opened 2026-07-18
fix(tui): enable cjk feature for unicode-width to fix Ambiguous-width glyph overlap

SparkofSpike:
## Problem
ec97a8ff995cba2d3313606219c5d2dd

Circled numbers (①②③) and other Ambiguous-width Unicode characters were incorrectly measured as 1 column by `unicode-width` without the `cjk` Cargo feature. On Windows Terminal these characters render as 2 columns wide, causing Ratatui's layout engine to calculate wrong positions — subsequent glyphs overlap with the mis-measured character.

## Root Cause

`unicode-width = "0.2"` was declared without the `cjk` feature. Characters with East Asian Width `Ambiguous` are measured as 1 column without the feature and 2 columns with it.

Characters affected (east_asian_width = Ambiguous):
- Circled digits ①②③ (U+2460-U+2473)
- Circled numbers ❶❷❸ (U+2776-U+2778)
- Black square ■ (U+25A0)
- And other Ambiguous-width codepoints

## Fix

`crates/tui/Cargo.toml`: enable the `cjk` feature for `unicode-width`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,15 @@ Thank you to the contributors whose code, reports, and reviews shaped v0.9.1:
 - [@seanthefuturegorilla](https://github.com/seanthefuturegorilla) — the
   canonical OpenCode Go/Zen provider request and acceptance direction
   (#1481).
+- [@nightt5879](https://github.com/nightt5879) — the Solarized Light
+  background preservation fix (PR #4471).
+- [@AiurArtanis](https://github.com/AiurArtanis) — the Solarized Light
+  regression report and reproduction (#4457).
+- [@shenjackyuanjie](https://github.com/shenjackyuanjie) — the HarmonyOS
+  workflow-js bindgen, portable-pty gating, and SDK environment work
+  (PR #4470).
+- [@shenyongqing](https://github.com/shenyongqing) — the original HarmonyOS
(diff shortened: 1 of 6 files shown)


Hmbown:
Thank you, @SparkofSpike — the glyph example is useful, but this exact patch cannot change CodeWhale's measurements yet.

On current main, `cargo tree -p codewhale-tui -e features -i unicode-width@0.2.2` already resolves `unicode-width feature "cjk"` through its default feature. In unicode-width 0.2.2, enabling `cjk` only makes the separate `width_cjk()` methods available; ordinary `width()` remains explicitly non-CJK and continues treating Ambiguous characters as one column. CodeWhale/Ratatui's ordinary width calls therefore behave the same before and after this Cargo.toml edit.

The branch also contains stale unrelated changelog/UI/credit changes and now conflicts with main, so I cannot merge it as-is. Please keep the PR open while we audit the actual width call sites and the independent intermittent repaint report in #4479. A revised fix needs a target-aware measurement path plus a regression test that fails on main and passes with the change; it must not globally force two-column Ambiguous width for every terminal, because that would misalign non-CJK terminals that render those glyphs as one cell.

Your report and contributor credit are preserved even if the final fix lands through a narrower maintainer branch.

PR #4510, opened 2026-07-18
fix(tui): keep keycap and emoji rendering grapheme-safe (#4479)

### LINKED ISSUE #4479: BUG: TUI rendering glitch — missing / extra spaces in text, recovers on mouse selection

SparkofSpike:
## Problem

keycap rendering evidence
Windows TUI corruption evidence

@SparkofSpike reduced #4479 to a deterministic case: fully qualified keycap sequences such as `1️⃣` render as two terminal columns, while Codewhale helper paths that summed individual codepoint widths accounted for one. Slicing or wrapping inside the sequence could then leave the custom layout state out of alignment with Ratatui.

## Integrated fix

The final branch preserves the contributor fix and adds a maintainer follow-up that:

- measures rendered text at extended grapheme boundaries using the same string-level width contract as Ratatui;
- keeps keycaps, ZWJ emoji, modifiers, and combining sequences intact during transcript wrapping, slicing, truncation, and selection;
- preserves text-form and isolated combining-mark semantics instead of globally assigning U+20E3 a nonzero width;
- restores an unrelated CI workflow deletion that slipped into an amended contributor commit.

## Evidence

- shared text helper tests: 24 passed;
- Markdown rendering and wrapping tests: 40 passed;
- keycap selection regression: passed;
- Codewhale TUI all-target Clippy with `-D warnings`: passed;
- formatting and diff checks: passed.

Fresh exact-head Windows, macOS, Linux, CodeQL, DCO, and secret checks are required before merge. Physical confirmation on the reporters Windows Terminal remains useful release dogfood evidence.

Fixes #4479.


Hmbown:
Thank you for narrowing #4479 to the keycap grapheme, @SparkofSpike — the mismatch is real: string/grapheme width sees `1️⃣` as two columns while a per-codepoint sum sees one.

One production path still bypasses this patch: `truncate_line_to_width` iterates characters and calls `UnicodeWidthChar::width` directly at `crates/tui/src/tui/ui_text.rs:39` and `:53`. Its initial `UnicodeWidthStr::width` check sees two columns, but the truncation loop still budgets the keycap as one. The plain/code wrapping helper in `markdown_render.rs` also has a separate per-character width function.

Before merge, please add focused regressions covering at least:

- `text_display_width("1️⃣") == 2`
- slicing/selecting a row containing `1️⃣` without splitting or shifting it
- `truncate_line_to_width` always returning a string whose `UnicodeWidthStr::width()` is within the requested budget

The cleanest fix would make the relevant helpers share one grapheme-aware width contract (`unicode-segmentation` is already a TUI dependency), which also avoids special-casing future emoji sequences. A narrower patch is fine if every affected loop is routed through it and the tests prove the reproduced table row. I’ll keep this in the v0.9.1 release review and retest the exact head when updated.

Hmbown:
Thanks again, @SparkofSpike. I fetched the current exact head 4e369284abbe85739dbced90e48ab759394df71c.

One unrelated regression slipped into the latest squash: it deletes .github/workflows/sync-cnb.yml in full. That workflow must remain. The keycap fix itself is still useful evidence, and I am preparing a maintainer follow-up that keeps your authored commit while moving the shared width, slice, and wrap paths to whole-grapheme accounting and adding regression coverage.

Please avoid another force-push for the moment so I can land the follow-up against this exact head without racing you. No further work is needed from you right now; I will restore the workflow as part of the integration.

SparkofSpike:
Apologies for the extra noise, should have stepped away when I wasn't at my best. Appreciate your patience and for taking it from here. Happy to leave the PR open for your follow-up branch to land against.

Hmbown:
Merged as 74afd81f20426368053a3cb8395a95ee39b8339c after the exact-head macOS, Windows, Linux, lint, reviewer, DCO, npm, and secret checks passed.\n\nThank you, @SparkofSpike. Your deterministic keycap reduction and authored commit are preserved in the merge; the maintainer follow-up extended the same finding across wrapping, slicing, truncation, and selection so future grapheme sequences use one contract. We will still include a physical Windows Terminal raster check in packaged release dogfood, since CI proves the width logic and platform build but cannot judge a real terminal screen.

Issue #2970, opened 2026-06-10
HarmonyOS/OpenHarmony tier-2 target: CI cargo-check job + remaining sandbox/clipboard gating


Hmbown:
## Background

HarmonyOS/OpenHarmony support was contributed by @shenjackyuanjie in PR #2634 and harvested to main with credit (cfg-gating for Linux-specific surfaces, rustls/ring switch, `scripts/ohos-env.{sh,ps1}`, `check-ohos-deps.sh`, `execpolicy/parser_ohos.rs`, PTY/nix gating). He verified `cargo check --target aarch64-unknown-linux-ohos -p codewhale-tui` passes on the stewardship branch.

What's missing to call OHOS a real tier-2 target:

## Tasks

- [ ] **CI job**: a `cargo check --target aarch64-unknown-linux-ohos` job (with the OHOS SDK or a stub toolchain) so the target can't silently regress
- [ ] **Remaining sandbox gating**: audit `crates/tui/src/sandbox/` for Linux-isms still reachable on OHOS
- [ ] **Clipboard gating**: confirm clipboard integration degrades gracefully on OHOS
- [ ] Document the target's support tier in the README/docs build matrix

## Invitation

@shenjackyuanjie — you know this target best; the remaining gating work would be very welcome as small PRs against `main` (the same layered shape as your original contribution). You're on the contributor allowlist, so the gate won't slow you down.

Related: #2634, #2625

shenjackyuanjie:
I understand, but my final exam week is next week, so I won't be able to start working until at least June 18th

Hmbown:
Community PR #4384 covers the rquickjs bindgen slice for the OHOS target and is now placed in v0.9.1. Keep this issue open for the remaining CI, sandbox, clipboard, and support-matrix acceptance criteria.

Hmbown:
Consolidation note: #2625's original HarmonyOS port and dependency-workaround history is retained there as context. #2970 owns the remaining CI, sandbox, clipboard, and support-matrix work.

Hmbown:
Retargeting the HarmonyOS residual to v0.9.2: the workflow-js slice is in draft PR #4459 (v0.9.1); the remaining acceptance criteria — CI cargo-check job, sandbox-gating audit, clipboard graceful-degradation, support-matrix docs — are unstarted and beyond this patch window.

shenjackyuanjie:
I have send #4470 to fix some issue on building for ohos

[...]

PR #4566, opened 2026-07-19
update tui Cargo.toml for HarmonyOS build

shenyongqing:
1. move portable-pty crate from target_evn ohos to cfg("unix") in tui Cargo.toml.
2. the newest portable-pty 0.9 and nix transitived have fulfilled requirements of musl, so both of them lifed from the gate.

## Summary
Compiled codewhale on HarmonyOS PC and got the TUI running successfully again.

## Testing

## Checklist

### DIFF
--- a/crates/tui/Cargo.toml
+++ b/crates/tui/Cargo.toml
@@ -102,6 +102,7 @@ vt100 = "0.16"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
+portable-pty = "0.9"
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.61.2", features = ["Win32_Foundation", "Win32_Security", "Win32_Security_Authorization", "Win32_Storage_FileSystem", "Win32_System_Threading"] }
@@ -110,7 +111,6 @@ windows-sys = { version = "0.61.2", features = ["Win32_Foundation", "Win32_Secur
 arboard = "3.4"
 
 [target.'cfg(not(target_env = "ohos"))'.dependencies]
-portable-pty = "0.9"
 starlark = "0.13.0"
 
 [target.'cfg(target_os = "macos")'.dependencies]
--- a/scripts/release/check-ohos-deps.sh
+++ b/scripts/release/check-ohos-deps.sh
@@ -118,7 +118,7 @@ cargo_tree_with_retry() {
 tree="$(cargo_tree_with_retry "${package}" --all-features --no-dedupe)"
 
 disallowed="$(
-  grep -E '^(nix v0\.(28|29)\.|portable-pty v|starlark v|arboard v|keyring v)' <<<"${tree}" || true
+  grep -E '^(starlark v|arboard v|keyring v)' <<<"${tree}" || true
 )"
 
 if [[ -n "${disallowed}" ]]; then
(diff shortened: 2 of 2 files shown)


Hmbown (changes requested):
Thank you for testing on a HarmonyOS PC. I audited this exact commit against current main and cannot merge it as written yet.

Two release-blocking points:

1. Moving portable-pty into cfg(unix) does not enable or repair an OHOS code path. Current main still compiles the terminal_session module, shell PTY imports and variants, child PTY environment path, and persistent terminal tool registration out under cfg(target_env = "ohos"). The added dependency is therefore unused on the OHOS release build; ordinary non-PTY shell execution was already the supported path and has a prior real-SDK cargo-check receipt.

2. The patch weakens check-ohos-deps.sh while leaving docs/HarmonyOS.md explicitly saying portable-pty and nix 0.28/0.29 are excluded and terminal/* is unavailable. That would leave the guard, documentation, and shipped capability disagreeing.

A focused package check confirms portable-pty 0.9.0 itself resolves for aarch64-unknown-linux-ohos, which is useful evidence, but it does not show that this patch changes Codewhale. Please add the exact rustc -vV / target triple and before-versus-after Codewhale build commands and errors from the HarmonyOS machine. If the intended change is to enable persistent PTY tools, that needs the matching cfg removals, runtime tests, sandbox/clipboard audit, and documentation update under #2970; those residuals are explicitly v0.9.2 scope. If the intent is only to fix a current build, we need a reproducer that fails on current main and passes on this head.

I am leaving the PR and source branch intact so we can incorporate the real hardware evidence safely.

shenyongqing:
Hi Hunter,

It turns out that in the latest version @6cd2953, compilation on HarmonyOS PC succeeds without the need for a patch, and the behavior is expected in a short running. Here are the details regarding the HarmonyOS PC compilation environment what i got right now， the following is the output of `rustc -vV`:

/storage/Users/currentUser % rustc -vV
rustc 1.95.0 (59807616e 2026-04-14) (built from a source tarball)
binary: rustc
commit-hash: 59807616e1fa2540724bfbac14d7976d7e4a3860
commit-date: 2026-04-14
host: aarch64-unknown-linux-ohos
release: 1.95.0
LLVM version: 22.1.2

/storage/Users/currentUser % uname -a
HarmonyOS localhost HongMeng Kernel 1.12.0 #1 SMP Tue Jun 23 02:15:22 UTC 2026 aarch64 Toybox

So to close the PR is ok for me.

Regards,
Yongqing

Hmbown:
Closing per your own follow-up, @shenyongqing — thank you for coming back with receipts rather than leaving it open.

Your 2026-07-21 comment confirmed compilation on HarmonyOS PC succeeds at @6cd2953 without the patch (rustc 1.95.0, host aarch64-unknown-linux-ohos, HongMeng Kernel 1.12.0), and said closing was fine.

For the record on why we're keeping the current state: `crates/tui/Cargo.toml:116-118` still gates `portable-pty` and `starlark` behind `cfg(not(target_env = "ohos"))`, and `scripts/release/check-ohos-deps.sh:121` still enforces the full guard including `nix 0.28/0.29` and `portable-pty`. Narrowing that guard would leave the script, `docs/HarmonyOS.md`, and the shipped capability disagreeing, and would pull an unused dependency into the OHOS release graph while the PTY paths remain `cfg`'d out.

The real HarmonyOS work you care about is tracked in #2970 (tier-2 target: CI cargo-check job plus remaining sandbox/clipboard gating), which is in the v0.9.2 milestone. Your verification is useful evidence there.

PR #4574, opened 2026-07-19
feat(tools): make terminal outcomes explicit

Hmbown:
## Summary

- add a typed terminal outcome contract for succeeded, failed, denied, invalid arguments, cancelled, and timed out calls
- route cancellation and timeout sites through explicit typed errors while preserving legacy ToolResult/event/storage compatibility
- deterministically repair dangling, orphaned, duplicate, early, and too-late tool results at session load and provider request boundaries
- expose a bounded visible repair receipt in persisted history and terminal status in tool audit receipts
- keep the provider-bound repair receipt-free so Anthropic-style APIs never interpret diagnostics as an assistant prefill

## Safety and compatibility

- no journal, storage-schema, public event, or provider credential changes
- repair is deterministic and idempotent
- legacy result consumers remain supported through one conversion seam
- cancellation remains a legacy error-shaped result but is no longer classified as generic failure internally
- result/error state is private and constructor-enforced

## Exact-head verification

Head: `859f3b99c433ae28ea1f37b79e2507340afaaeee`
Base: `d53b8f97b1e6f9f8419e22111247a2a3d5b8d686`

- `cargo test -p codewhale-tools --locked` (23 unit + 3 parity passed)
- `cargo test -p codewhale-tui --locked` (7,512 unit passed, 0 failed, 3 ignored; all integration suites green, including QA PTY 17/17 with 1 documented ignore, release runtime 5/5 with 1 heavy benchmark ignored, tool lifecycle 3/3, workflow stream 3/3)
- direct provider-bound regression proves repair ends on the synthetic user tool result with no receipt prefill
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- gitleaks exact two-commit scan: no leaks
- added-line private-path and credential-value scans: clean

Only the pre-existing macOS compact-unwind linker warning appeared during tests.

### DIFF
--- a/crates/tools/src/lib.rs
+++ b/crates/tools/src/lib.rs
@@ -10,9 +10,11 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use tokio::sync::{OwnedRwLockReadGuard, OwnedRwLockWriteGuard, RwLock};
 
+mod outcome;
 mod prepared;
 mod resources;
 
+pub use outcome::{ToolExecutionOutcome, ToolTerminalStatus};
 pub use prepared::PreparedToolCall;
 pub use resources::{ResourceClaim, schedule_non_conflicting};
 
@@ -62,6 +64,8 @@ pub enum ToolError {
     ExecutionFailed { message: String },
     #[error("Failed to execute tool: operation timed out after {seconds}s")]
     Timeout { seconds: u64 },
+    #[error("Tool execution cancelled: {message}")]
+    Cancelled { message: String },
     #[error("Failed to locate tool: {message}")]
     NotAvailable { message: String },
     #[error("Failed to authorize tool execution: {message}")]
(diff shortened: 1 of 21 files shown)


Hmbown:
Fixed the Claude correctness finding in `859f3b99c`.

- provider-bound repair now fixes call/result pairing without appending the session-visible assistant receipt, so Anthropic-style payloads end on the synthetic user tool result instead of an assistant prefill
- persisted session repair still appends the visible `[tool_history_repair]` receipt
- added direct provider-bound regression coverage
- made `ToolExecutionOutcome.result/error` private so the exactly-one invariant can only be constructed through the safe constructors

Post-fix local gates: focused outcome tests, 8 repair/history tests, provider-bound boundary test, fmt/diff, and strict all-workspace/all-target/all-feature clippy all green. Hosted checks are rerunning at exact head `859f3b99c`.

PR #4575, opened 2026-07-19
feat(web): unify search backend receipts

Hmbown:
## Summary

- add a provider-neutral SearchBackend contract covering all nine configured search providers
- route both web_search and web.run text search through the same adapter and normalization seam
- add ranked/domain-aware results plus receipts for capabilities, honored knobs, degradation, latency, and cache state
- preserve legacy query/source/count/message/results fields and existing SearXNG/custom DuckDuckGo behavior

## Verification

- cargo fmt --all -- --check
- cargo test -p codewhale-tui --all-features (7,534 passed; 3 ignored)
- cargo test -p codewhale-tui web_search -- --nocapture (49 passed)
- cargo test -p codewhale-tui tools::web:: -- --nocapture (46 passed; 1 ignored)
- cargo test -p codewhale-tui text_search_uses_configured_shared_backend_and_exposes_receipt -- --nocapture
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- gitleaks protect --staged --redact --no-banner

## Release boundary

This is W2 only. It does not tag, publish, create release assets, or deploy.

### DIFF
--- a/crates/tui/src/tools/web/backend.rs
+++ b/crates/tui/src/tools/web/backend.rs
@@ -1,0 +1,153 @@
+//! Search backend selection and the shared async adapter contract.
+
+use std::time::Instant;
+
+use async_trait::async_trait;
+
+use super::contract::{BackendId, BackendSearch, QueryCapabilities, SearchQuery};
+use crate::config::SearchProvider;
+use crate::tools::spec::{ToolContext, ToolError};
+
+#[async_trait]
+pub(crate) trait SearchBackend {
+    fn id(&self) -> BackendId;
+    fn capabilities(&self) -> QueryCapabilities;
+    async fn search(
+        &self,
+        query: &SearchQuery,
+        deadline: Instant,
+    ) -> Result<BackendSearch, ToolError>;
+}
+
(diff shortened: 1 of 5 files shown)


Hmbown:
Addressed both review notes in a631cbef7:

- centralized the result cap as MAX_SEARCH_RESULTS in the shared contract and reused it from both web_search and web.run
- added direct JSON assertions for ChallengeDetected and ScrapeFallback backend transitions plus their human-visible warning

Focused contract tests (4), web_search tests (49), strict workspace clippy, fmt, and staged gitleaks are green on the updated head.

Hmbown:
Follow-up `532791beb` closes the remaining review gap: the DuckDuckGo challenge → Bing success case now runs through the production scrape fallback path and shared response finalizer with mocked endpoints, asserting the serialized `challenge_detected` and `scrape_fallback` receipt entries. It also types `MAX_SEARCH_RESULTS` as `u8`, so the constructor fallback cannot diverge from the shared cap. Local receipts: focused fallback test 1/1, `web_search` 50/50, and workspace all-target/all-feature clippy with `-D warnings` green.

Issue #3736, opened 2026-06-28
v0.8.67: Separate work mode from approval policy before any Auto loop


Hmbown:
## Summary

The mode/permission model carries **four overlapping knobs** in `EffectiveModePolicy` (`crates/tui/src/tui/app.rs:1110-1119`): `allow_shell`, `approval_mode`, `trust_mode`, `auto_approve`. Three of them co-vary in every mode, which is the structural source of the "UI says one mode but a different flag decides" bugs (#3733 Auto, #3735 YOLO-publish).

## Current state (3 surviving modes after Auto deferral)

| Mode | allow_shell | approval_mode | trust_mode | auto_approve |
|---|---|---|---|---|
| Plan | false | Suggest | false | false |
| Agent | prefs | prefs | prefs | false |
| YOLO | true | **Bypass** | **true** | **true** |

YOLO sets **three** flags that all mean "approve everything."

## What each knob really does
- **`approval_mode`** (`approval.rs:41-51`): `Suggest`/`Auto`/`Bypass`/`Never` — the one genuinely meaningful control.
- **`trust_mode`** (bool): enables sandbox **workspace writes**, skips the **untrusted-workspace** safety hold, **and** contributes to auto-approval (`auto_approve || trust_mode`, `runtime_threads.rs:3296`).
- **`auto_approve`** (bool): skip prompts; fully covered by `approval_mode == Bypass`.

## Proposal — collapse 4 → 2
1. **`approval_mode`** stays the single approve/ask/bypass/never control. Derive auto-approval from `approval_mode == Bypass` and **delete `auto_approve`** (remove the separate flag at `app.rs:1117-1118` and its mirrors; replace `auto_approve || trust_mode` with `approval_mode == Bypass || trust_mode`).
2. **`trust_mode`** keeps its *one* distinct job — workspace-write capability + untrusted-workspace handling — and stops doubling as an approval flag.
3. `allow_shell` can likely be derived too (Plan = no shell; Agent/YOLO = shell), but keep for now if prefs need per-user shell toggle in Agent.

## Why
Removing `auto_approve` as an independent flag eliminates the layer split that lets `safety_floor`'s `HoldForReview` get silently bypassed (#3735) and that makes mode label and behavior disagree (#3733). One control = one source of truth.

## Verification
- Existing mode/approval tests still pass after deriving auto-approval from `approval_mode`.
[...]

PR #4582, opened 2026-07-19
fix: bypass MCP tool deferral when trust_mode or Bypass approval is active

Angel-Hair:
## Summary

When working in Full Access Agent mode (trust_mode=true, auto_approve=true), MCP tools are still marked defer_loading=true and do not appear in the model's available tool list. Although the user is functionally equivalent to YOLO mode for approval purposes, MCP tool visibility does not receive the same bypass treatment. This forces every session to manually discover MCP tools via tool_search (e.g. mcp_playwright_browser_navigate), which itself depends on the MCP server having completed its handshake — for slow-starting servers like Firefox-driven Playwright, the first tool_search call may return zero results, creating a false impression that "MCP is unavailable."

## Root Cause

apply_mcp_tool_deferral in crates/tui/src/core/engine/tool_catalog.rs only checks mode != AppMode::Yolo to decide whether to defer. The break in the pipeline is in engine.rs: build_model_tool_catalog_with_surface receives only input_policy.mode (i.e. AppMode::Agent), discarding the trust_mode / auto_approve / approval_mode bypass semantics already present on TurnAuthority. TurnAuthority carries the full set of fields, and EffectiveModePolicy populates them correctly, but build_model_tool_catalog_with_surface's signature accepts only AppMode — the bypass semantics are lost mid-pipeline.

## Fix

Map the bypass condition to AppMode::Yolo inline at the call site in engine.rs, with zero function signature changes. #3736 landed in v0.8.67: the auto_approve || trust_mode bypass semantics were replaced with approval_mode == Bypass. The auto_approve field remains as a compat shim, so both must be checked:

```rust
let bypass = input_policy.auto_approve
    || input_policy.approval_mode == ApprovalMode::Bypass;
// …
if bypass { AppMode::Yolo } else { input_policy.mode },
```

## Version

Reproducible on v0.9.0.

## Testing

- [x] cargo check -p codewhale-tui
- [x] cargo test -p codewhale-tui (27 defer/MCP-related tests pass; 34 pre-existing Windows TUI UI failures unrelated)
- [x] cargo clippy --workspace --all-targets --all-features (1 pre-existing warning in mcp.rs:261, unrelated)

## Checklist

- [x] Updated docs or comments as needed (single-site logic fix, no API surface change)
- [ ] Added or updated tests where relevant (existing defer tests cover this path)

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -3005,10 +3005,16 @@ impl Engine {
             if self.config.features.enabled(Feature::Mcp) {
                 always_load.insert("start_mcp_server".to_string());
             }
+            let bypass = input_policy.auto_approve
+                || input_policy.approval_mode == crate::tui::approval::ApprovalMode::Bypass;
             let mut catalog = build_model_tool_catalog_with_surface(
                 registry.to_api_tools_with_cache(true),
                 mcp_tools,
-                input_policy.mode,
+                if bypass {
+                    AppMode::Yolo
+                } else {
+                    input_policy.mode
+                },
                 &always_load,
                 capability.tool_surface_budget,
             );


Hmbown (approved):
Reviewed against the current TurnAuthority and catalog contracts. The change only controls model-visible MCP deferral; execution still passes through the existing approval, trust, and tool-policy gates. Provenance narrowing removes standing Bypass/auto-approval before this call site, and the full cross-platform CI matrix is green. Approved for the 0.9.1 lane; contributor branch and credit will be preserved.

PR #4584, opened 2026-07-19
perf(tui): keep debt gate out of system prefix

Hmbown:
## Summary

- remove the changing SlopLedger completion gate from the fingerprinted system prompt
- attach it once to each top-level external user-turn tail, between user text and final `<turn_meta>`
- let mid-turn steers reuse that already-present transcript block instead of duplicating debt text
- leave runtime-origin user messages untouched so continuations do not reinject token debt
- keep the existing modified-time cache and completion-gate semantics

## Verification

- `cargo test -p codewhale-tui slop_gate_ --all-features` (2 passed)
- `cargo test -p codewhale-tui refresh_system_prompt --all-features` (3 passed)
- `cargo test -p codewhale-tui --test cache_guard --all-features` (9 passed)
- `cargo clippy -p codewhale-tui --all-targets --all-features -- -D warnings`
- `gitleaks git --log-opts=origin/main..HEAD --no-banner`

No publication, deploy, release, registry, or branch-deletion action is included.

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -2414,7 +2414,11 @@ impl Engine {
         )
     }
 
-    fn with_slop_ledger_gate_for_external_turn(
+    /// Snapshot the mutable completion gate once for a new top-level user
+    /// turn. Mid-turn steers stay inside the same `handle_deepseek_turn` and
+    /// reuse this already-present block; reinjecting it on every steer would
+    /// duplicate debt text and erase the token-economy benefit of this seam.
+    fn with_slop_ledger_gate_for_initial_user_turn(
         &mut self,
         message: Message,
         provenance: UserInputProvenance,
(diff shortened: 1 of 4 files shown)


Hmbown:
Review disposition: the steer behavior is intentional, not a missing gate path. A steer remains inside the same `handle_deepseek_turn`; the gate attached to the initial top-level external message is already in the model-visible transcript. Reattaching the same mutable block to every steer would duplicate debt text and undermine this token-economy slice. A fresh top-level user turn snapshots a fresh gate; runtime-origin continuations and mid-turn steers do not reinject it. Commit `HEAD` now makes that boundary explicit in the helper name/comments and adds a regression assertion for the plain steer constructor. The focused two-test gate suite remains green.

Hmbown:
Re-review disposition: valid compaction gap fixed in `813cbd30f`. The exact gate-bearing initial user message now lives in `TurnContext` and is merged into authoritative compaction pins for both the normal auto-compaction path and context-overflow recovery. Matching gate text from older turns is not pinned, and mid-turn steers still do not duplicate the gate. New async regression moves the initial gate outside the recent-message tail, proves it would otherwise be summarized, forces real compaction through the injected model boundary, and asserts the exact message and marker survive. Local gates: 3/3 slop-gate tests, 3/3 system-prompt refresh tests, cache_guard 9/9, strict all-target/all-feature clippy, fmt/diff, and three-commit gitleaks all green.

PR #4609, opened 2026-07-20
fix(tui): respect umask for workspace atomic writes

### LINKED ISSUE #4606: write_file ignores system umask, creates files with 600 in shared dev environments
## Description

`write_file` creates files with mode `600` (`rw-------`), ignoring the shell's `umask`. In shared development environments (e.g., Laravel + PHP-FPM), files need to be readable by the `www` user, but `write_file` leaves them inaccessible to anyone except the agent's user. The `exec_shell` tool correctly respects `umask`, but `write_file` is a native Rust tool that hardcodes the permission bits.

This causes a cycle of: agent writes → file is `600` → PHP-FPM gets Permission denied → user must manually `chmod` → repeat on every edit.

## Steps to reproduce

1. Set `umask 002` in `.zshrc` (so new files default to `664`)
2. Use `write_file` to create a file inside a setgid directory (group `www`)
3. Check the file permissions

```bash
# Verify umask
$ umask
002

# exec_shell respects umask → 664 (correct)
$ echo "test" > /project/foo.txt && ls -la /project/foo.txt
-rw-rw-r-- 1 agent www ... /project/foo.txt

# write_file ignores umask → 600 (broken)
# (create file via write_file tool)
$ ls -la /project/bar.txt
-rw------- 1 agent www ... /project/bar.txt
```

## Expected behavior

`write_file` should create files with the same default permissions as `exec_shell` — i.e., respect the process `umask` (`0o666 & ~umask`). In a `umask 002` environment, new files should be `664` (`rw-rw-r--`).

## Actual behavior

`write_file` always creates files with mode `600` (`rw-------`), regardless of `umask`.

## Impact

Every file edited by the agent in a shared server environment (Laravel, WordPress, any PHP project with PHP-FPM running as a separate user) is unreadable by the web server until the human manually fixes permissions. This turns every edit session into a permission-chasing exercise and undermines trust in the agent's autonomy.

## Workaround

After every `write_file`, manually run:
```bash
chmod 640 /path/to/file   # or chmod g+r
```

Or set up a cron/watchdog that auto-fixes permissions on the project tree.

## Environment

- OS: Linux 6.8.0-136-generic x86_64
- codewhale version: 0.9.0 (d167c07c9628)
- Install method: unknown
- Model/provider: deepseek-v4-pro
- Terminal app: codewhale TUI
- Shell: zsh

FluxForgeLab:
## Summary

Fixes #4606 by separating atomic-write permission policy for user workspace
files from CodeWhale's private internal persistence.

Workspace tools (`write_file`, `edit_file`, `apply_patch`, FIM) previously used
the shared `write_atomic()` helper. That helper creates a `NamedTempFile` whose
Unix default mode is `0600`, then atomically renames that inode over the
destination. Because umask can only remove permission bits, `umask 002` cannot
turn the tempfile's `0600` into the expected `0664`. Atomic replacement also
replaced the mode of existing files, so editing an existing `0664` or `0755`
file changed it to `0600`.

## Changes

- Added `write_atomic_workspace()` with a shared `write_atomic_with_permissions()`
  implementation; kept `write_atomic()` private (tempfile default `0600`).
- New Unix workspace files request mode `0666`; the kernel applies the inherited
  process umask (same candidate mode as ordinary `std::fs::write`).
- Existing workspace files retain ordinary `rwx` bits (`mode & 0o777`), including
  executable bits. setuid/setgid/sticky are intentionally not restored.
- Migrated workspace call sites:
  - `WriteFileTool` / `EditFileTool` (`tools/file.rs`)
  - `apply_patch` success path and rollback (`tools/apply_patch.rs`)
  - FIM write-back (`tools/fim.rs`)
- Kept internal persistence on private `write_atomic()`:
  session, history, stash, trust, task, truncate/spillover, terminal session,
  MCP config, session export, artifacts, slop ledger, runtime preset restore, etc.
- Symlink-safe mode capture: use `symlink_metadata` (do not follow links). Symlink
  directory entries are treated as “no mode to preserve” so inaccessible or
  dangling targets do not abort the write; rename still replaces the directory
  entry (pre-#4606 semantics).
- Preserved Windows persist retries and existing file/parent `sync_all` behavior.

## Tests

- Utils: new workspace file mode matches same-directory `fs::write` control file.
- Utils: existing `0664` mode preserved on overwrite.
- Utils: existing `0755` executable mode preserved.
- Utils: special bits are not restored (`& 0o777`).
- Utils: private `write_atomic` new files do not gain group/other access.
- Utils: symlink replacement does not follow target; unreadable target still writes.
- Tool-level: `WriteFileTool` / `EditFileTool` permission regressions (Unix).
- Windows: existing atomic replace / contention retry tests still pass.

**Local verification (Windows):**
- `cargo fmt -p codewhale-tui -- --check` — pass
- `utils::atomic_write_tests` — pass (Unix-only permission/symlink tests cfg’d out)
- `tools::file::tests` — pass
- Full `cargo test -p codewhale-tui`: unrelated i18n/UI failures observed; none tied to this diff
- `cargo clippy -p codewhale-tui --all-targets --all-features -- -D warnings`: blocked by pre-existing `mcp.rs` `needless_return`, not this change

Unix permission/symlink coverage should be confirmed on Linux/macOS CI.

## Known limits

This change preserves ordinary Unix mode bits. It does not claim to preserve
arbitrary ACLs, xattrs, ownership, group ownership, or to redesign symlink
semantics beyond avoiding a follow-metadata abort regression.

Temporary workspace files created with `Builder::permissions(0666)` may be
group-visible from creation (consistent with final umask-applied mode).

## Test plan

- [ ] On Linux with `umask 002`, `write_file` creates a new file with mode `664`
- [ ] Overwrite an existing `664` file via `write_file` / `edit_file`; mode stays `664`
- [ ] Edit an existing `755` script via `edit_file`; mode stays `755`
- [ ] Confirm session/history/trust files remain owner-only (`mode & 077 == 0`)
- [ ] Confirm Windows atomic replace behavior has no regression
- [ ] CI green for `codewhale-tui` Unix permission tests

### DIFF
--- a/crates/tui/src/tools/apply_patch.rs
+++ b/crates/tui/src/tools/apply_patch.rs
@@ -1092,7 +1092,7 @@ fn apply_pending_writes(pending: &[PendingWrite]) -> Result<(), ToolError> {
             };
 
             parent_result.and_then(|()| {
-                crate::utils::write_atomic(&entry.path, content.as_bytes()).map_err(|e| {
+                crate::utils::write_atomic_workspace(&entry.path, content.as_bytes()).map_err(|e| {
                     ToolError::execution_failed(format!(
                         "Failed to write {}: {}",
                         entry.path.display(),
@@ -1127,7 +1127,7 @@ fn rollback_pending_writes(applied: &[PendingWrite]) {
     for entry in applied.iter().rev() {
         match entry.original.as_ref() {
             Some(content) => {
-                let _ = crate::utils::write_atomic(&entry.path, content.as_bytes());
+                let _ = crate::utils::write_atomic_workspace(&entry.path, content.as_bytes());
             }
             None => {
                 let _ = fs::remove_file(&entry.path);
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -734,7 +734,7 @@ impl ToolSpec for WriteFileTool {
             })?;
         }
 
-        crate::utils::write_atomic(&file_path, file_content.as_bytes()).map_err(|e| {
+        crate::utils::write_atomic_workspace(&file_path, file_content.as_bytes()).map_err(|e| {
             ToolError::execution_failed(format!("Failed to write {}: {}", file_path.display(), e))
         })?;
         context.note_file_read(&file_path);
@@ -891,7 +891,7 @@ impl ToolSpec for EditFileTool {
             (contents.replace(search, replace), count, None)
         };
 
-        crate::utils::write_atomic(&file_path, updated.as_bytes()).map_err(|e| {
+        crate::utils::write_atomic_workspace(&file_path, updated.as_bytes()).map_err(|e| {
             ToolError::execution_failed(format!("Failed to write {}: {}", file_path.display(), e))
         })?;
         context.note_file_read(&file_path);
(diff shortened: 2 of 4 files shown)


Hmbown (approved):
Independent exact-head audit approved at 6415e64a49ae5b77ba364cacbe679abdf1848d1e. The contributor commit preserves kernel umask behavior for new workspace files; the signed macOS/BSD correction masks mode() special bits without weakening the symlink-safe same-directory atomic rename path. Focused atomic-write tests: 12/12; file tool tests: 63/63; full locked TUI: 7,651 passed, 0 failed, 3 ignored; strict TUI clippy and formatting passed. No unresolved review threads. Merge remains gated only on the final hosted runner.

Issue #4520, opened 2026-07-18
feat(tui): add configurable session token breakdown (in / cch / out) to header bar


eugenicum:
## Request

PR #2411 compacted the `tokens` statusline chip to a single cumulative total (`tok 1.5M`), which is great as a default. But the old verbose format — showing the full breakdown of **input**, **cache hit**, and **output** tokens in one chip — was valuable for users who want to see all three numbers at a glance.

## Proposed solution

Add a config toggle:

```toml
[tui]
tokens_verbose = true   # default: false (compact)
```

When `true`, the `tokens` chip renders the old verbose format:

```
18K in · 12K cch · 6K out
```

When `false` (default), it stays as the current compact `tok 1.5M`.

## Motivation

The separate `cache` chip shows cache hit **rate** (percentage), but not the raw token counts. Users who reason about token economics need the raw numbers: "how many input tokens did this turn cost?", "how much did caching save?", "how large was the output?". The compact total hides this information; `codewhale metrics` is an extra step.

## Related

- #2309 — originally noted the tokens chip was "very verbose" and suggested a compact mode
- #2411 — implemented the compact mode (harvested from #2405)

eugenicum:
**Update:** This request is specifically for the **header** (top bar), not the footer. The header currently shows context utilization (`176K/1.0M 18%`) but has no way to display session token spending (`in · cch · out`).

The header should support configurable chips similar to how the footer has `status_items`, or at minimum allow adding the session token breakdown alongside the existing context percent display.

Currently the header rendering in `header.rs` only passes `total_conversation_tokens` (a single u32) — it has no access to the breakdown fields (`total_input_tokens`, `total_cache_hit_tokens`, `total_output_tokens`). Adding these to `HeaderData` and exposing a config toggle would solve this.

Hmbown:
Thanks @eugenicum — the clarification makes this a coherent header capability rather than a request to undo the compact footer.

I agree that context utilization and session token spending answer different questions. The compact default should remain, while an opt-in header treatment can expose canonical input, cache-hit, and output counts without fabricating data for providers that do not report a field.

I am preserving this for v0.9.2 rather than expanding the v0.9.1 cut at the last release gate. The implementation should cover:

- one canonical usage source carried into HeaderData;
- an extensible header configuration contract, with the current compact presentation unchanged by default;
- deterministic fallback or elision at narrow terminal widths;
- focused 40/60/80/100-column rendering tests and config round-trip coverage;
[...]

PR #4610, opened 2026-07-20
feat(tui): add configurable session token header

XhesicaFrost:
## Summary

  Refs #4520.

  Add an opt-in session token breakdown to the active underwater TUI header:

  - Introduce the extensible `tui.header_items` configuration contract.
  - Support `header_items = ["tokens"]` for cumulative input, cache-hit, and output token counts.
  - Preserve the existing compact header when the option is not configured.
  - Elide the optional token breakdown deterministically at narrow terminal widths.
  - Keep the built-in context-utilization meter ahead of optional header content under width pressure.
  - Reuse the existing compact token-count formatter.
  - Avoid adding another cache-rate or total-token presentation to the footer.

  Example configuration:

  ```toml
  [tui]
  header_items = ["tokens"]
 ```

  The active underwater header continues to use its existing App data path and reads the three cumulative session counters from app.session.
  This avoids introducing a second data-access pattern within the same renderer.

  Unknown header_items values are ignored so that the configuration can be extended without breaking older clients.

  Provider-specific missing-usage semantics are intentionally outside the scope of this change; the header treats the counters consistently
  with the existing session token totals.

  ## Testing

  - [x] cargo clippy --workspace --all-targets --all-features
    - PR head: `codewhale-tui` reported 7,639 passed, 15 failed, and 3 ignored.
    - Direct parent baseline: `codewhale-tui` reported 7,636 passed, 12 failed, and
    3 ignored under the same environment.
    - Eleven failures reproduced on the direct parent commit.
    - Four additional terminal-session failures appeared only during the parallel PR
    run; the complete terminal-session test group passed 8/8 when run single-
    threaded on both the PR head and its direct parent.
    - All six tests added by this PR passed.
    - `commands::tests::execute_links_and_aliases_return_links_message` fails on
    both revisions because it expects `https://codewhale.net/en/docs`.

  - [x] Focused header rendering tests at 40, 60, 80, and 100 columns
  - [x] Default opt-in behavior test
  - [x] Header configuration parsing and round-trip tests
  - [x] cargo build -p codewhale-tui --bin codewhale-tui --locked

The width tests verify deterministic rendering at the requested terminal widths:

  - At 40 columns, both the optional token breakdown and the built-in context meter
  are hidden by the existing compact layout.
  - At 60 columns, the optional token breakdown is elided while the built-in context
  meter remains visible.
  - At 80 columns, both the token breakdown and context meter are visible.
  - At 100 columns, both the token breakdown and context meter are visible.

  At every tested width, the mode and effective-permission indicators remain
  visible. When the optional token breakdown does not fit, only that configured item
  is removed; the existing context and version chrome retain their original
  behavior.

  ## Checklist

    - N/A — this is an original contribution and contains no harvested code.


Hmbown:
Thank you for picking up the header token-breakdown request and for documenting the width matrix and baseline failures carefully. The linked issue was explicitly held for v0.9.2 so we do not expand the frozen v0.9.1 candidate; I have applied that label to the PR and am leaving the source branch intact. This is a deferral, not a rejection. After the 0.9.1 cut, the maintainer pass should rebase onto current main, run the strict locked workspace gate, resolve the full-suite delta, and review the provider missing-usage semantics called out in the issue before merge.

Issue #2438, opened 2026-05-31
Kimi Coding Plan: tool schema rejected — type should be defined in anyOf items instead of parent


A-Corner:
## Description

When using the `moonshot` provider with a Kimi Coding Plan API key (prefix `sk-kimi-`), every request fails with HTTP 400:

```json
{"error":{"message":"tools.function.parameters is not a valid moonshot flavored json schema, details: <At path root: when using anyOf, type should be defined in anyOf items instead of the parent schema>","type":"invalid_request_error"}}
```

## Root Cause

CodeWhale generates tool function schemas where `type` is placed at the root level alongside `anyOf`:

```json
{
  "type": "object",
  "anyOf": [
    {"properties": {...}},
    {"properties": {...}}
  ]
}
```

Kimi's API (both `https://api.kimi.com/coding/v1` and the standard Moonshot endpoint) enforces stricter JSON Schema validation and requires `type` to be inside each `anyOf` item:

```json
{
  "anyOf": [
    {"type": "object", "properties": {...}},
    {"type": "object", "properties": {...}}
  ]
}
```

## Reproduction

1. Configure Kimi Coding Plan key:
```toml
[providers.moonshot]
api_key = "sk-kimi-..."
auth_mode = "kimi_oauth"
```

2. Switch to `moonshot` provider in TUI
3. Send any message (even "hi") — fails immediately because tool definitions are always sent with requests

The issue affects ALL providers when pointing at Kimi's API (tested with `moonshot`, `vllm`). Plain `curl` without tools works fine.

## Environment

- CodeWhale v0.8.47
- Kimi Coding Plan API: `https://api.kimi.com/coding/v1`
- Model: `kimi-for-coding` (Kimi-k2.6)

## Relevant Code

The tool schema generation logic likely lives in the tool registry or LLM client layer. The fix would need to move `type` declarations from the parent schema level into each `anyOf`/`oneOf` item for Kimi/Moonshot compatibility.

## Note

z.ai / GLM Coding Plan (via `openai` provider) works fine — they accept the current schema format. The issue is Kimi-specific strictness.

Hmbown:
Thanks @A-Corner — this is a very helpful repro because it names the exact provider path, key flavor, model, and schema validation error. I do not see this fixed by the recent model/provider merges, so I am leaving it open as a focused Kimi/Moonshot tool-schema compatibility bug.

[...]

PR #4613, opened 2026-07-20
fix(tui): sanitize Moonshot tool parameters per MFJS spec

bistack:
## Summary

Moonshot/Kimi validates tool `parameters` against MFJS (Moonshot Flavored JSON
Schema): the root MUST be `type:"object"`, and `anyOf`/`oneOf`/`allOf` are not
supported at root — only nested inside `properties`. Tools whose `input_schema`
uses root-level composition (`apply_patch`, `finance`,
`mcp_install_and_run_info`, …) were rejected with alternating 400s:

- root `type` + `anyOf` present: *"when using anyOf, type should be defined in
  anyOf items instead of the parent schema"*
- root `type` removed: *"parameters.type is required and must be 'object'"*

`sanitize_for_kimi_parameters` now:

1. delegates to `sanitize_for_responses` (same approach as xAI): merges root
   composition properties into `properties`, removes the composition keywords,
   ensures `type:"object"`;
2. keeps wrapping a bare `$ref` in `allOf` (no sibling keywords);
3. runs `sanitize_for_kimi` afterwards so nested `type`+`anyOf`/`oneOf` schemas
   still push `type` into items per the #2438 rule — MFJS supports composition
   only in nested positions.

Spec: https://github.com/MoonshotAI/walle/blob/main/docs/mfjs-spec.md

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features` (clean for this
  change; two pre-existing `into_owned` warnings in `codewhale-cli` tests
  remain, unrelated)
- [x] `cargo test --workspace --all-features` — 7652 passed / 16 failed; all 16
  reproduced on `main` (`de15b5c6`) in an isolated worktree and are
  environmental (macOS process-inspection `Operation not permitted`, macOS
  Vision OCR backend, Seatbelt availability, timing-sensitive tests). The
  changed module's focused suite passes: `cargo test -p codewhale-tui
  schema_sanitize` → 37/37.

## Checklist

- [x] Updated docs or comments as needed (MFJS spec URL in
  `schema_sanitize.rs` and `chat.rs`; no config-surface change, so
  `docs/CONFIGURATION.md` needs no update)
- [x] Added or updated tests where relevant (updated 4 tests; new regression
  test for nested `oneOf` push-down)
- [x] Verified TUI behavior manually if UI changes — N/A, no UI changes
- [ ] Harvested/co-authored credit uses a GitHub numeric noreply address — N/A

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -260,8 +260,10 @@ impl DeepSeekClient {
                 .iter()
                 .map(|tool| tool_to_chat_for_base_url(tool, &self.base_url))
                 .collect();
-            // Kimi / Moonshot enforces stricter JSON Schema: `type` must be
-            // inside `anyOf` / `oneOf` items, not on the parent (#2438).
+            // Kimi / Moonshot MFJS does not support anyOf/oneOf/allOf at
+            // parameters root. sanitize_for_kimi_parameters merges composition
+            // properties and removes the keywords. Same approach as xAI.
+            // Spec: https://github.com/MoonshotAI/walle/blob/main/docs/mfjs-spec.md
             if matches!(self.api_provider, crate::config::ApiProvider::Moonshot) {
                 for t in &mut chat_tools {
                     if let Some(fn_obj) = t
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you @bistack. We independently audited your MFJS fix and carried it into the v0.9.1 integration PR #4617 as preserved commit 0c8d55aace61cacd8a02add8969f745183e05bb0, retaining the exact `Harvested from PR #4613 by @bistack` line and your co-author trailer. The integration adds the final fail-closed MFJS limits plus real streaming, deferred-tool, runtime, and General-child catalog captures. Exact local gates and two independent audits are green; hosted CI is running now. We will use a normal merge so the credited commit history is retained, then close this source PR with the merge receipt.

Hmbown:
Landed through normal merge #4617 at 9aeeb4ba442fa5d9be3ead87eb5bf72a4a76c537. Your contribution remains in main as preserved commit 0c8d55aace61cacd8a02add8969f745183e05bb0 with the exact Harvested from PR #4613 by @bistack line and canonical co-author trailer. We kept this source branch intact; the source PR is closed only to avoid duplicate integration. Thank you again for catching the MFJS incompatibility.

PR #4656, opened 2026-07-21
fix(route): honor explicit limits for unknown local models

### LINKED ISSUE #4655: Self-hosted route limits are capped by the unknown-model 4K fallback
## Problem

A self-hosted route can resolve an explicit output limit for an arbitrary wire model alias, but `effective_max_output_tokens_for_route` first intersects the request with `provider_capability`. Because aliases outside the static model catalogue receive the compatibility fallback of 4096 tokens, a concrete route limit such as 24576 can never raise the effective cap above 4096.

## Reproduction

Use a vLLM route with:

- wire model: `arbitrary-local-wire-alias`
- `CODEWHALE_MAX_OUTPUT_TOKENS=65536`
- resolved route limits: `context_tokens=262144`, `output_tokens=24576`

The current function returns 4096. Without route metadata, 4096 is the correct conservative fallback; with a concrete route maximum, the expected result is 24576.

## Expected behavior

For self-hosted providers only, when the wire alias has no known model entry, a concrete route output limit should replace the compatibility fallback. Known model caps and hosted-provider caps should remain authoritative and continue to intersect with route limits.

## Impact

Self-hosted deployments that use deployment-specific wire aliases cannot apply their declared output capacity, and the request/compaction budget remains artificially constrained to 4K.

h3c-hexin:
## Summary

- let concrete route output limits replace the generic 4K compatibility fallback for unknown self-hosted wire aliases
- preserve known-model and hosted-provider caps
- add regression coverage for both the explicit-limit path and the conservative fallback

Fixes #4655.

## Scope

This changes only output-cap resolution for self-hosted providers when the wire model has no static catalogue entry. It does not change hosted-provider policy or known-model caps.

## Testing

- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui route_budget::tests:: -- --nocapture`

Local `cargo clippy -p codewhale-tui --bin codewhale-tui -- -D warnings` reaches an unrelated existing `clippy::collapsible_match` finding in `crates/tui/src/tui/ui.rs:5038` under Rust 1.95; this PR does not touch that file.

### DIFF
--- a/crates/tui/src/route_budget.rs
+++ b/crates/tui/src/route_budget.rs
@@ -78,9 +78,23 @@ pub(crate) fn effective_max_output_tokens_for_route(
     model: &str,
     route_limits: Option<RouteLimits>,
 ) -> u32 {
-    let cap =
-        effective_max_output_tokens(model).min(provider_capability(provider, model).max_output);
-    let cap = route_output_limit_tokens(route_limits).map_or(cap, |route_cap| cap.min(route_cap));
+    let requested_cap = effective_max_output_tokens(model);
+    let compatibility_cap = provider_capability(provider, model).max_output;
+    let route_cap = route_output_limit_tokens(route_limits);
+    // Arbitrary wire aliases on self-hosted engines cannot appear in the
+    // static model catalogue. When that is the only reason the compatibility
+    // cap fell back to 4K, a concrete route limit is the stronger fact. Known
+    // model and hosted-provider caps remain authoritative and are still
+    // intersected with any route maximum.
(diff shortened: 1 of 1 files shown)


Hmbown:
Thank you @h3c-hexin — this is integrated on the v0.9.1 release lane as `ac2cde40c` (author preserved).

Also closes the self-hosted/unknown-model 4K fallback path reported in #4655. Credit is in the 0.9.1 changelog when the lane ships to `main`.

Hmbown:
Landed on the v0.9.1 release lane as ac2cde40c. Thank you!

PR #4657, opened 2026-07-21
fix(streaming): report progress on idle timeouts

h3c-hexin:
## Summary

Include received-byte and timing telemetry in Chat Completions SSE idle-timeout errors. This distinguishes a prefill stall with zero response bytes from a stream that stalled after partial output, including truncated tool-call arguments.

The existing transport-error diagnostics already retain this state; this change makes the timeout path equally actionable without changing timeout or retry behavior.

## Testing

- `git diff --check`
- `cargo test -p codewhale-tui --bin codewhale-tui stream_diagnostics_tests:: -- --nocapture`

### DIFF
--- a/crates/tui/src/client/chat.rs
+++ b/crates/tui/src/client/chat.rs
@@ -48,6 +48,22 @@ fn stream_open_timeout_from_env(value: Option<&str>) -> Duration {
     Duration::from_secs(secs)
 }
 
+fn stream_idle_timeout_message(
+    idle: Duration,
+    bytes_received: usize,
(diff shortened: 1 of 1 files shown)


Hmbown:
Thank you @h3c-hexin — integrated on the v0.9.1 release lane as `b6bb0fb69` (author preserved). Idle-timeout progress is now reported instead of a silent hang.

Hmbown:
Landed on the v0.9.1 release lane as b6bb0fb69. Thank you!

PR #4673, opened 2026-07-21
fix(shell): default no-cwd shell commands to context.workspace

### LINKED ISSUE #4674: BashTool ignores context.workspace for default cwd — sub-agent worktrees run commands in parent checkout
## Problem

When a sub-agent is spawned with `worktree: true`, shell commands that don't specify an explicit `cwd` run in the parent checkout's workspace instead of the sub-agent's worktree directory.

The `BashTool`'s cwd resolution defaults to `None` when no `cwd` is provided, which falls through to the shared `ShellManager`'s `default_workspace` (set to the parent workspace). The foreground execution path (`execute_foreground_via_background`) also always passed `None` as the working directory, so even if the cwd were resolved differently upstream, it was ignored.

## Root Cause

Two issues in `crates/tui/src/tools/shell.rs`:

1. **No-op default**: `BashTool`'s `working_dir` resolution returned `None` when no `cwd` was given, deferring to `ShellManager.default_workspace` instead of `context.workspace`.
2. **Missing thread**: `execute_foreground_via_background` hardcoded `None` as the working directory, so it couldn't receive a resolved directory even if one were provided.

## Fix (PR #4673)

- Default no-`cwd` shell commands to `context.workspace` instead of `None`.
- Thread the resolved `working_dir` through `execute_foreground_via_background`.
- Added regression test `default_cwd_uses_context_workspace_not_shell_manager_default`.

## Expected Behavior

Shell commands without an explicit `cwd` should run in the sub-agent's worktree directory when `worktree: true` is used.

/cc @fleitz

fleitz:
When the BashTool receives no explicit `cwd`, the working directory now defaults to `context.workspace` (the sub-agent's worktree when `worktree: true` was used) instead of `None`, which fell through to the shared ShellManager's parent-workspace default.

This ensures sub-agents in isolated worktrees run shell commands in their worktree directory rather than the parent checkout.

The fix threads the resolved `working_dir` through `execute_foreground_via_background` (which previously always passed `None`) and adds a test that verifies the context.workspace priority over the ShellManager's default_workspace.

Closes #4674

### DIFF
--- a/crates/tui/src/tools/shell.rs
+++ b/crates/tui/src/tools/shell.rs
@@ -2459,6 +2459,7 @@ fn exec_shell_input_starts_detached(input: &serde_json::Value) -> bool {
 async fn execute_foreground_via_background(
     context: &ToolContext,
     command: &str,
+    working_dir: Option<String>,
     timeout_ms: u64,
     stdin_data: Option<&str>,
     tty: bool,
@@ -2474,7 +2475,7 @@ async fn execute_foreground_via_background(
         manager.clear_foreground_background_request();
         manager.execute_with_options_env_for_owner_and_work(
             command,
-            None,
+            working_dir.as_deref(),
             timeout_ms,
             true,
             stdin_data,
@@ -2797,7 +2798,10 @@ impl ToolSpec for BashTool {
                 let resolved = context.resolve_path(dir)?;
                 Some(resolved.to_string_lossy().to_string())
             }
-            None => None,
+            // Default to the tool context's workspace (which reflects the
+            // child agent's worktree when `worktree: true` was used), not the
+            // shared ShellManager's parent-workspace default_workspace.
+            None => Some(context.workspace.display().to_string()),
         };
 
         // #456 — collect env from any configured `shell_env` hooks. Runs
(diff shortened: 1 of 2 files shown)


Hmbown:
Maintainer follow-up added without rewriting Fred’s commit: `6d0ebc881a8bd2469c45b25f2a606fa63681e112` remains the exact parent of `64ddbdf22d580269e4b0889064f46130bfbb5b84`.

The newly authorized Windows matrix run confirmed the implementation itself and 7,694 other TUI tests, but its new regression compared PowerShell `pwd` output to a canonical Windows path spelling. The follow-up keeps the same behavioral assertion and selects the active directory through distinct marker files instead. Local exact test, `cargo fmt --all -- --check`, and `git diff --check` pass. CI is now rerunning against main with the additive Fred identity mapping from #4678.

Hmbown (approved):
Reviewed the exact contributor commit and the additive portability follow-up. The original Fred Leitz commit 6d0ebc881a8bd2469c45b25f2a606fa63681e112 remains unchanged; 64ddbdf22d580269e4b0889064f46130bfbb5b84 makes only the regression test path-independent. Focused local shell tests passed, and hosted lint, DCO, macOS, Windows, web, and offline eval checks are green.

Issue #1519, opened 2026-05-12
v0.8.65: Custom provider endpoints, models, and auth within provider-scoped routing


hhhaiai:
## Goal

Support custom endpoints, custom model ids, and custom auth configuration through the same provider-scoped routing architecture as built-in providers.

This issue is the broad custom-endpoint/user-config fixture under #2608, #3311, #3084, and #3384. It should no longer be read as one combined feature for endpoint routing plus unrelated fork/upstream repository sync. Repository auto-sync is out of scope for v0.8.65 provider routing and should be tracked separately if still desired.

## Original Request Summary

Reporter: @hhhaiai requested:

- configurable custom deployment endpoints;
- custom model selections;
- custom API keys;
- HTTP and HTTPS custom addresses;
- less manual environment-specific setup;
- enterprise/private deployment support.

Those endpoint/auth/model requirements are in scope here. Automatic upstream fork synchronization is not part of the provider/model route architecture.

## Architecture Contract

Custom endpoints are providers or provider overrides, not global model strings.

Use provider-scoped config:

```toml
[providers.custom_name]
kind = "openai-compatible"
base_url = "https://api.example.com/v1"
model = "custom-model-v1"
auth = { env = "EXAMPLE_API_KEY" }
```

Or explicit overrides on a known provider when the request protocol is intentionally reused:

```toml
[providers.openrouter]
base_url = "https://zenmux.ai/api/v1"
model = "deepseek/deepseek-v4-pro"
```

Custom model ids are preserved as provider-scoped wire ids. CodeWhale must not infer provider switching from prefixes such as `deepseek/`, `deepseek-ai/`, `anthropic/`, or `openai/`.

## Scope

1. Represent custom providers/endpoints as dynamic provider descriptors or provider user config overrides.
2. Preserve custom model ids exactly for pass-through/custom endpoints.
3. Support HTTP and HTTPS base URLs with explicit validation and clear local/insecure warnings where appropriate.
4. Support provider-scoped auth sources: env var, secret manager/external command when #3004 lands, OAuth where available, local/no-auth where appropriate.
5. Ensure custom routes resolve through #3384 before mutating app/config/engine state.
6. Expose custom provider readiness in `/provider` (#3083) and custom model rows in `/model` (#3383/#3075).

## Non-Goals

- No repository fork/upstream auto-sync in this issue.
- No storing secrets in prompts, logs, links, argv, or persisted event payloads.
[...]

PR #4686, opened 2026-07-22
feat(minimax): add China / Token Plan provider routes for minimaxi.com

ffaacceelee:
Add four new provider identifiers that target api.minimaxi.com (the Chinese / Token Plan sibling of api.minimax.io):

  - minimax-cn                   (OpenAI Chat Completions on api.minimaxi.com/v1)
  - minimax-anthropic-cn         (Anthropic Messages on api.minimaxi.com/anthropic)

The Anthropic-compatible variant is what Token Plan subscribers use, since config.example.toml already mentions api.minimaxi.com/anthropic in passing but no provider route actually pointed there.

Both providers share the same [providers.minimax] runtime config when keys land in the defaults (parallel to how SiliconflowCN reuses Siliconflow's fields), so no new env-vars are introduced; MINIMAX_API_KEY continues to cover pay-as-you-go and Token Plan keys.

Note: this is a partial plumbing change. Additional non-exhaustive match arms in crates/tui/src/config.rs and elsewhere may still need CN variants added to make cargo check green; follow-up patch welcome.

## Testing

## Checklist

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -21,7 +21,7 @@
 # `api_key` / `base_url` are
 # still read as DeepSeek defaults when `[providers.deepseek]` is absent
 # (backward compatibility).
-provider = "deepseek" # deepseek | deepseek-cn | deepseek-anthropic | nvidia-nim | openai | atlascloud | wanjie-ark | volcengine | openrouter | xiaomi-mimo | novita | fireworks | siliconflow | siliconflow-CN | arcee | moonshot | zai | stepfun | minimax | sglang | vllm | ollama | huggingface | together | qianfan | openai-codex | anthropic | openmodel | deepinfra | sakana | longcat | opencode-go | meta | xai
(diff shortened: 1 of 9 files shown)


Hmbown:
Thanks @ffaacceelee — closing with a pointer to a path that works today.

You were right that `main` lacks the CN variants: `crates/config/src/provider_defaults.rs:127-128` still pins `api.minimax.io` for both the OpenAI and Anthropic routes, and there are no `MinimaxCN`/`MinimaxAnthropicCN` enum variants.

The reason not to merge is the one you named yourself in the PR body — it is partial plumbing and would not build. Adding two `ProviderKind` variants requires updating every non-exhaustive match; there are 52 `ProviderKind::` match sites in `crates/tui/src/config.rs` alone.

The good news is you do not need new enum variants for this. `main` ships a generic custom-provider route: set `[providers.<your-id>]` with `kind = "openai-compatible"` and your own `base_url`, and it is validated at `crates/config/src/lib.rs:2013-2017`. That gives you minimaxi.com China / Token Plan endpoints without touching the enum, and it is how we intend region and plan variants to be configured generally (see #1519).

If that route turns out to be missing something specific for minimaxi.com — a header, an auth shape, a model-id mapping — please open an issue describing the gap. That is a change we can make narrowly.

PR #4742, opened 2026-07-24
fix(workflow): preserve hashes in fleet strings

### LINKED ISSUE #4732: workflow: named_fleet minimal TOML parser truncates fields containing '#'
## Problem

`parse_fleet_toml_minimal` (the fallback minimal TOML parser for named-fleet definitions) strips comments with a naive split that doesn't respect quoting:

```rust
// crates/workflow/src/named_fleet.rs:~64-66
for raw in text.lines() {
    let line = raw.split('#').next().unwrap_or("").trim();
    ...
}
```

Any `#` character inside a quoted string value (e.g. a `description = "...#4178..."` field referencing a GitHub issue number, which is this exact codebase's own convention) gets treated as a comment start and silently truncates the rest of the line.

## Why it matters

A fleet TOML's `description` (or a role/profile string value) containing a literal `#` is silently corrupted — truncated with no parse error and no warning — rather than either parsing correctly or failing loudly. `fleets/*.toml` files in this repo already reference issue numbers in prose, so this is a realistic, not hypothetical, input.

## Fix direction

Use a comment-stripping pass that tracks whether it's inside a quoted string (single or double quotes per TOML string rules) and only treats `#` as a comment start outside of quotes. Alternatively, if a real TOML parser (`toml`/`toml_edit`, already a workspace dependency per `Cargo.toml`) can be used here instead of the hand-rolled minimal parser, that would eliminate this whole class of bug.

## Suggested test plan

A unit test parsing a fleet TOML with `description = "handles #123 style references"` and asserting the full string round-trips, not just the text before the `#`.

## Evidence

Verified by direct read of `crates/workflow/src/named_fleet.rs` (main branch, current `HEAD`).

nightt5879:
## Summary

- preserve `#` characters inside quoted named-fleet values
- strip only real comments outside basic or literal string quotes
- cover issue references, role values, escaped quotes, literal strings, and trailing comments

## Root cause

The minimal line-oriented fleet parser used `split('#')`, so it truncated every
hash regardless of whether the hash appeared inside a quoted value.

Fixes #4732

## Testing

- [x] `cargo fmt --all -- --check` (local and CI)
- [x] `cargo test -p codewhale-workflow --locked` (108 tests, local)
- [x] `cargo clippy -p codewhale-workflow --all-targets --all-features --locked -- -D warnings` (local)
- [x] `cargo clippy --workspace --all-features --locked -- -D warnings ...` (GitHub Actions Lint)
- [x] `cargo test --workspace --all-features --locked` (GitHub Actions macOS and Windows)
- [x] macOS offline eval, Workflow RLM cache, Web lint/type/build, DCO, and GitGuardian checks

## Checklist

- [x] Updated comments to document the minimal parser's single-line string boundary
- [x] Added regression tests for the affected parser behavior
- [x] TUI manual verification is not applicable to this workflow parser change
- [x] Harvest/co-author credit is not applicable to this original fix

### DIFF
--- a/crates/workflow/src/named_fleet.rs
+++ b/crates/workflow/src/named_fleet.rs
@@ -60,13 +60,49 @@ pub fn parse_named_fleet(toml_text: &str) -> Result<NamedFleet, NamedFleetError>
     parse_fleet_toml_minimal(trimmed)
 }
 
+/// Strip comments from the single-line basic and literal strings supported by
+/// the minimal fleet parser.
+fn strip_toml_comment(line: &str) -> &str {
+    let mut quote = None;
+    let mut escaped = false;
+
+    for (index, character) in line.char_indices() {
+        match quote {
+            Some('"') => {
+                if escaped {
+                    escaped = false;
+                } else {
+                    match character {
+                        '\\' => escaped = true,
+                        '"' => quote = None,
+                        _ => {}
+                    }
+                }
+            }
+            Some('\'') => {
+                if character == '\'' {
+                    quote = None;
+                }
+            }
+            Some(_) => unreachable!("only TOML string delimiters are tracked"),
+            None => match character {
+                '"' | '\'' => quote = Some(character),
+                '#' => return &line[..index],
+                _ => {}
+            },
+        }
+    }
+
+    line
+}
+
 fn parse_fleet_toml_minimal(text: &str) -> Result<NamedFleet, NamedFleetError> {
     let mut name = None;
     let mut description = None;
     let mut roles = BTreeMap::new();
     let mut section = "";
     for raw in text.lines() {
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks @nightt5879 — landed and closed.

This reached `main` as `7dc0a45c1` (cherry-picked from `b5d1aeb79`), with your `Signed-off-by` intact and "Fixes #4732" in the body. `crates/workflow/src/named_fleet.rs:65` now has `fn strip_toml_comment`, called at :105, and the old `raw.split('#').next()` is gone. Both of your tests are present under their original names — `quoted_hashes_are_not_treated_as_comments` at :246 and the escaped/literal-string assertions at :265/:269/:273.

The PR still showed as mergeable only because the fix arrived by cherry-pick rather than by merging this ref. Nothing left to do here.

Issue #4729, opened 2026-07-24
MCP: sanitized qualified tool names can collide across differently-named servers


Hmbown:
## Problem

MCP tool names are qualified as `mcp__{server}__{tool}` via `sanitize_component`, which lowercases and collapses every non-alphanumeric character to `_`:

```rust
// crates/mcp/src/lib.rs:404-414
fn sanitize_component(value: &str) -> String {
    value
        .chars()
        .map(|ch| {
            if ch.is_ascii_alphanumeric() || ch == '_' { ch.to_ascii_lowercase() } else { '_' }
        })
        .collect()
}
```

Distinct server names like `"my-server"`, `"my_server"`, and `"My.Server"` all sanitize to the identical prefix `mcp__my_server__...`. Nothing prevents registering two differently-named servers that collide after sanitization. When `call_qualified_tool` resolves a qualified name via its fallback loop, it iterates `self.configs` — a plain `HashMap<String, _>` with unordered iteration — and returns on the *first* match:

```rust
// crates/mcp/src/lib.rs:331-338
for (server_name, (_, filter)) in &self.configs {
    ...
    if qualify_tool_name(server_name, &tool.tool_name) == qualified_tool_name {
        return client.call_tool(&tool.tool_name, arguments);
    }
}
```

## Why it matters

A malicious or careless third-party MCP server registered with a similarly-shaped name (e.g. `my-server` vs. an existing trusted `my_server`) can produce a colliding qualified tool name. Which server's tool actually executes for a given qualified name then depends on `HashMap` iteration order, not on which server the model/user intended — allowing one server to intercept or shadow tool calls meant for a different, trusted server.

## Fix direction

Reject registration (or disambiguate, e.g. via the hash-truncation suffix mechanism already used in `qualify_tool_name` for long names) when two servers' sanitized names collide. Make `call_qualified_tool`'s resolution deterministic instead of relying on first-match over unordered iteration — e.g. resolve by exact server name first, and only fall back to the scan when no exact registration exists, erroring (not silently picking one) on an actual collision.

## Suggested test plan

A unit test registering two servers named `my-server` and `my_server` with same-named tools, asserting registration either fails with a clear collision error or that qualified-name resolution is deterministic and documented (not hash-map-order-dependent).

## Evidence

Verified by direct read of `crates/mcp/src/lib.rs` (main branch, current `HEAD`).

PR #4756, opened 2026-07-24
fix: do not retry failed qualified MCP tool calls

### LINKED ISSUE #4728: MCP: call_qualified_tool can execute a tool call multiple times after the first attempt errors
## Problem

`McpManager::call_qualified_tool` has a fast path that tries the obvious server/tool lookup, then falls through to a slower re-resolution loop **whenever the fast path's call errors**, not just when the lookup fails to find a match:

```rust
// crates/mcp/src/lib.rs:316-347
pub fn call_qualified_tool(
    &self,
    qualified_tool_name: &str,
    arguments: Value,
) -> Result<Value> {
    let parsed = parse_qualified_tool_name(qualified_tool_name)...;

    if let Ok((server_name, tool_name)) = &parsed
        && self.clients.contains_key(server_name)
        && let Ok(result) = self.call_tool(server_name, tool_name, arguments.clone())
    {
        return Ok(result);
    }

    for (server_name, (_, filter)) in &self.configs {
        let Some(client) = self.clients.get(server_name) else { continue };
        for tool in client.list_tools()? {
            if !allowed_by_filter(&tool.tool_name, filter) { continue; }
            if qualify_tool_name(server_name, &tool.tool_name) == qualified_tool_name {
                return client.call_tool(&tool.tool_name, arguments);   // <-- 2nd live call
            }
        }
    }

    let (server_name, tool_name) = parsed?;
    self.call_tool(&server_name, &tool_name, arguments)                // <-- possible 3rd live call
}
```

If `self.call_tool(...)` in the fast path returns `Err` (a normal outcome for a flaky/slow MCP server), the `if let ... && ... && let Ok(...) = ...` chain evaluates to `false` as a whole, and execution falls through to the loop, which resolves the *same* server/tool again and calls `client.call_tool(...)` a second time. If that loop's linear scan doesn't happen to match first (e.g. due to filter/name edge cases), the final fallback line issues a *third* invocation.

## Why it matters

For any non-idempotent tool — a file write, a git commit, a message send, a paid API call — fronted by a server that returns a transient error, this causes the same side-effecting call to fire multiple times. The caller only ever observes the result of the last attempt; the original error from the fast path is discarded silently, never logged or surfaced.

## Fix direction

Capture the first `Err` from the fast path and propagate it directly instead of falling through to the slow path on a *call* failure. Only use the slow-path loop when the fast-path *lookup* didn't find a matching client/tool at all (e.g. `parsed` failed, or `self.clients` didn't contain the server) — not when the call itself failed.

## Suggested test plan

A unit test with a mock `McpManagedClient` whose `call_tool` increments a call counter and returns `Err` on the first invocation; assert `call_qualified_tool` returns that error and the counter is exactly 1 (not 2 or 3).

## Evidence

Verified by direct read of `crates/mcp/src/lib.rs` (main branch, current `HEAD`).

adity982:
Fixes #4728

## Summary
- Return the fast-path MCP client error directly when the qualified server exists.
- Keep slow tool re-resolution only for lookup misses.
- Add a regression test proving a failed tool call is invoked exactly once.

## Validation
- `cargo test -p codewhale-mcp manager_call_qualified_tool_does_not_retry_after_client_error --locked`

### DIFF
--- a/crates/mcp/src/lib.rs
+++ b/crates/mcp/src/lib.rs
@@ -323,9 +323,8 @@ impl McpManager {
 
         if let Ok((server_name, tool_name)) = &parsed
             && self.clients.contains_key(server_name)
-            && let Ok(result) = self.call_tool(server_name, tool_name, arguments.clone())
         {
-            return Ok(result);
+            return self.call_tool(server_name, tool_name, arguments);
         }
 
         for (server_name, (_, filter)) in &self.configs {
@@ -1009,6 +1008,10 @@ impl JsonRpcError {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    };
 
     struct EchoMcpClient;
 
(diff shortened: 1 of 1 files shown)


Hmbown:
Thanks for this, @adity982 — the diagnosis was correct and it shipped.

Closing as already landed: `cccaa4f3f` ("fix(mcp): call a tool once, and resolve qualified names deterministically") contains character-for-character your production change at `crates/mcp/src/lib.rs:342-350` — the `let Ok(result) = self.call_tool(...)` guard is gone, replaced by the early-return fast path, with a comment explaining that a second invocation would be a second real side effect. Your test is there too, as `CountingFailingClient` (you named it `CountingErrorMcpClient`). It also closed #4728 and #4729 and additionally fixed qualified-name shadowing.

Credit note: that commit landed without a `Co-authored-by` trailer for you. That was our miss, not a judgement on the contribution — I'm adding you to `.github/AUTHOR_MAP` so this attributes correctly going forward. Retrying a failed tool call is exactly the kind of bug that is easy to wave off and genuinely harmful in practice; thanks for catching it.

PR #4760, opened 2026-07-24
fix: replace dirs::home_dir() with effective_home_dir() in crates/tui

### LINKED ISSUE #4757: Audit remaining raw dirs::home_dir() call sites for the same Windows test-fakeability bug
## Context

Fixed two Windows-only CI test failures in \`bc461c60c\` caused by raw
\`dirs::home_dir()\` calls in \`tools/workflow.rs\` and
\`tui/views/skills_manager.rs\`. Root cause: \`dirs::home_dir()\` on Windows
resolves via a native OS API call and does not respect \`HOME\`/\`USERPROFILE\`
env var overrides, so tests that fake \$HOME via the established
\`EnvVarGuard\` pattern silently disagreed with production code that used the
real Windows user profile instead. \`crate::config::effective_home_dir()\`
already exists for exactly this (checks \`HOME\`, then \`USERPROFILE\`, then
\`HOMEDRIVE\`+\`HOMEPATH\`) and is used correctly elsewhere in the crate.

## Remaining call sites (not touched — no failing test implicated them, and a
sweeping replacement isn't verifiable without Windows CI to check it against)

\`\`\`
crates/tui/src/skills/install.rs:64
crates/tui/src/skills/mod.rs:45
crates/tui/src/skills/mod.rs:54
crates/tui/src/skills/mod.rs:717
crates/tui/src/skills/mod.rs:1052
crates/tui/src/skills/mod.rs:1081
\`\`\`
(plus ~70 more across the crate not yet enumerated — grep \`dirs::home_dir()\`
in \`crates/tui/src\` for the full list.)

## Ask

Audit each remaining call site: does it sit behind a test that fakes \$HOME
and could hit the same Windows mismatch? If so, swap to
\`crate::config::effective_home_dir()\`. If a call site is genuinely meant to
resolve the *real* OS home regardless of test/env overrides (rare, but
possible for something like a hard security boundary), leave it and note why.

EvanProgramming:
## Summary

Audit and replace all remaining raw `dirs::home_dir()` call sites in `crates/tui/src` with `crate::config::effective_home_dir()`.

**Context:** Fixed two Windows-only CI test failures in `bc461c60c` caused by raw `dirs::home_dir()` calls in `tools/workflow.rs` and `tui/views/skills_manager.rs`. Root cause: `dirs::home_dir()` on Windows resolves via a native OS API call and does not respect `HOME`/`USERPROFILE` env var overrides, so tests that fake $HOME via the established `EnvVarGuard` pattern silently disagreed with production code that used the real Windows user profile instead. `crate::config::effective_home_dir()` already exists for exactly this (checks `HOME`, then `USERPROFILE`, then `HOMEDRIVE`+`HOMEPATH`) and is used correctly elsewhere in the crate.

This PR sweeps all remaining call sites (45 replacements across 29 files) so the crate consistently uses `effective_home_dir()`. The fallback inside `effective_home_dir()` itself (`config/paths.rs:55`) is intentionally left as `dirs::home_dir()`.

**Note:** `network_policy.rs` and `skills/install.rs` are `#[path]`-included from integration tests (`tests/skill_cli.rs`), so `crate::config::` does not exist in the test binary crate context. These two files keep `dirs::home_dir()` with explanatory comments — they are not test-fakability-critical.

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features --locked` — warning-free
- [x] `cargo test --workspace --all-features --locked` — 8155 passed, 0 failed (only 1 unrelated PTY flake)

## Checklist

- [x] Added or updated tests where relevant (N/A — pure refactor)
- [x] Verified TUI behavior manually if UI changes (N/A)
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address (N/A)

Closes #4757

### DIFF
--- a/crates/tui/src/automation_manager.rs
+++ b/crates/tui/src/automation_manager.rs
@@ -1170,7 +1170,7 @@ pub fn default_automations_dir() -> PathBuf {
     if let Some(home) = std::env::var_os("CODEWHALE_HOME").filter(|value| !value.is_empty()) {
         return PathBuf::from(home).join("automations");
     }
-    dirs::home_dir()
+    crate::config::effective_home_dir()
         .map(|home| {
             let primary = home.join(".codewhale").join("automations");
             let legacy = home.join(".deepseek").join("automations");
--- a/crates/tui/src/codex_model_cache.rs
+++ b/crates/tui/src/codex_model_cache.rs
@@ -116,7 +116,7 @@ pub(crate) fn codex_home_path() -> PathBuf {
         .filter(|value| !value.is_empty())
         .map(PathBuf::from)
         .unwrap_or_else(|| {
-            dirs::home_dir()
+            crate::config::effective_home_dir()
                 .unwrap_or_else(|| PathBuf::from("."))
                 .join(".codex")
         })
--- a/crates/tui/src/commands/groups/config/config.rs
+++ b/crates/tui/src/commands/groups/config/config.rs
@@ -2294,11 +2294,11 @@ fn trust_remove(workspace: &Path, raw: &str) -> CommandResult {
 
 fn expand_tilde(raw: &str) -> String {
     if let Some(rest) = raw.strip_prefix("~/")
-        && let Some(home) = dirs::home_dir()
+        && let Some(home) = crate::config::effective_home_dir()
     {
         return home.join(rest).to_string_lossy().into_owned();
     } else if raw == "~"
-        && let Some(home) = dirs::home_dir()
+        && let Some(home) = crate::config::effective_home_dir()
     {
         return home.to_string_lossy().into_owned();
     }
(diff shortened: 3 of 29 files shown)


Hmbown:
Thank you for this, @EvanProgramming — the report and the sweep were right, and #4757 is closed.

Closing as already landed: `0d7312736` landed the same change on `main`. `rg -n 'dirs::home_dir' crates/tui/src/` now returns only 3 hits, all inside `crates/tui/src/config/home.rs` itself (two doc comments and the OS fallback in the definition), against 78 `effective_home_dir` call sites.

Credit note: that commit went in without a `Co-authored-by` trailer for you, which was our miss — your PR predated it by ~11 hours. I'm adding you to `.github/AUTHOR_MAP` so future harvests in this area attribute correctly.

One thing worth calling out because you got it right and flagged it honestly: you deliberately left `network_policy.rs` alone with a note that `crate::config` wasn't reachable from the `#[path]`-included `skill_cli` test binary. That was a real constraint, not an oversight. It was later solved by splitting `effective_home_dir` into a dependency-free leaf (`189c93965`), so `network_policy.rs:324` now uses it too. Your note is what made that the obvious next step.

Issue #4727, opened 2026-07-24
codewhale mcp-server never spawns configured MCP servers — always returns fabricated stub responses


Hmbown:
## Problem

`codewhale mcp-server` is a documented, real CLI subcommand ("Run MCP server mode over stdio", `crates/cli/src/lib.rs:342-343`) that loads the user's persisted MCP server definitions and runs them over stdio:

```rust
// crates/cli/src/lib.rs:3798-3802
fn run_mcp_server_command(store: &mut ConfigStore) -> Result<()> {
    let persisted = load_mcp_server_definitions(store);
    let updated = run_stdio_server(persisted)?;
    persist_mcp_server_definitions(store, &updated)
}
```

Each persisted `McpServerDefinition` carries a real `command`/`args`/`env` (the actual executable a user configured for a downstream MCP server). But `build_stdio_state` never uses them to spawn anything — it wires every registered server to `default_stdio_client`, a hardcoded stub:

```rust
// crates/mcp/src/lib.rs:613-633 (build_stdio_state) and 643-687 (default_stdio_client)
if should_start {
    manager.register_server(
        definition.config.clone(),
        definition.filter.clone(),
        default_stdio_client(&name),   // <-- always the stub, regardless of config.command/args/env
    );
    ...
}

fn default_stdio_client(server_name: &str) -> Box<dyn McpManagedClient> {
    // returns an InMemoryMcpClient exposing only two fake tools,
    // "health" and "capabilities", both always returning {"status": "ok", ...}
    ...
}
```

`grep -n "\.command\b\|\.args\b\|\.env\b"` across `crates/mcp/src/lib.rs` confirms `command`/`args`/`env` are read only to be echoed into a diagnostic JSON blob — never passed to a process spawn or any real transport. `InMemoryMcpClient` is the only implementation of `McpManagedClient` anywhere in the repo.

## Why it matters

Any real MCP client that connects to `codewhale mcp-server` expecting it to proxy to the user's actually-configured downstream servers silently gets fabricated "ok" responses for health/capability checks and "tool not found" for every real tool — with no error, warning, or log indicating the proxy is fake. This is a "silently does the wrong thing" bug on a documented, user-facing CLI command, not a missing feature that was never wired up for internal use.

## Fix direction

[...]

Issue #3880, opened 2026-07-02
【window】DSML Interrupt Task


hardy922:
## Description

Why isn't the DSML interrupt task merged into the release branch package? The latest V0.8.66【win】 download still has this bug

## Environment

- OS:
- codewhale version:
- Install method:
- `codewhale doctor` summary:
- Model/provider:
- Terminal app:
- Shell:

Hmbown:
Thanks @hardy922 — I checked the release history before treating this as a duplicate.

The previous DSML report #3717 was fixed by #3718 (`fb5ae3f1`, merged as `527e48b5`), and both of those commits are ancestors of the `v0.8.66` tag (`54eb3d3c`). That means one of two things is likely happening here:

1. the Windows binary being run is not actually the corrected `v0.8.66` asset, or
2. this is a second DSML wrapper shape that #3718 does not strip yet.

Could you add the exact repro details so we can separate those quickly?

- output of `codewhale --version` or `codew --version`
- install source and asset name (GitHub release exe/zip, npm, Cargo, source build, etc.)
- the smallest snippet around the DSML text that interrupts the task, especially the exact opening/closing marker
- whether the marker is the fullwidth form `<｜DSML｜tool_calls>` / `</｜DSML｜tool_calls>` or a different shape
- terminal app and shell on Windows

If you can also try the current v0.8.67 candidate build once available, that will tell us whether this is already covered or needs a new streaming scrubber fixture.

hardy922:
@Hmbown ,Thank you for your reply。I installed the release package，codewhale-windows-x64 zip。
codewhale 0.8.66 (54eb3d3cb9cc)。
Download link：https://github.com/Hmbown/CodeWhale/releases

Hmbown:
Thanks for coming back with the exact asset and version line — `codewhale-windows-x64` zip at 0.8.66 (54eb3d3c) confirms you are on the official corrected build, which rules out the stale-binary theory from the earlier comment.

That leaves the other explanation: your model is emitting a DSML wrapper shape the scrubber does not strip yet. The current releases (v0.9.0/v0.9.1) strip these shapes in `crates/tui/src/core/engine/streaming.rs` and `crates/tui/src/core/tool_parser.rs`:

- fullwidth `<｜DSML｜tool_calls>` / `<｜DSML｜invoke …>`
- ASCII `<|DSML|tool_calls>` / `<|DSML|invoke …>`
[...]

PR #4804, opened 2026-07-25
fix(v0.9.2): audit cluster + constitution prefix + TUI list/home cleanups

### LINKED ISSUE #3256: Tool-call output: collapse happy path by default, expand failures, make verbosity configurable

### LINKED ISSUE #3854: chore(cleanup): simplify prefix-cache catalog storage
## Summary
Simplify prefix-cache catalog storage by removing the test-only `joined` field from `CachedCatalog`.

## Evidence
- `crates/tui/src/prefix_cache.rs:221` `CachedCatalog::joined` is read only by prefix-cache tests.
- Production `compute_with_tool_cache` consumes `fingerprint_for(...).sha256_hex`; it does not need to store the full joined catalog string.
- The current storage keeps an `Arc` for test pointer-equality observations rather than runtime behavior.

## Suggested Change
Store only the hash needed by production. Compute the joined catalog locally on cache misses and remove pointer-equality tests that assert implementation detail rather than behavior.

## Risk & Confidence
Confidence: high. Risk: low-medium; prefix-cache behavior is sensitive, so preserve tests for hash stability and cache hits.

## Verification Checklist
- [ ] Run `cargo test -p codewhale-tui prefix_cache -- --nocapture`.
- [ ] Run `cargo check -p codewhale-tui --all-targets`.
- [ ] Confirm tests still cover stable fingerprint output for unchanged tool catalogs.

### LINKED ISSUE #3856: chore(cleanup): remove or wire unused plan validation API
## Summary
Remove the unused plan-transition validation API, or wire it into `update_plan` execution.

## Evidence
- `crates/tui/src/tools/plan.rs:406` `PlanValidation` and `:414` `validate_plan_update` have no callers.
- The active `update_plan` tool updates plan state directly; UI/rendering paths consume `PlanSnapshot`, `PlanState`, and `StepStatus`, not the validation API.
- `deadcode.log` reports the validation enum and function as unused.

## Suggested Change
Delete `PlanValidation` and `validate_plan_update` if transition warnings are not needed. If they are needed, call validation from the update path and surface warnings in the tool result.

## Risk & Confidence
Confidence: high. Risk: low-medium; deleting is safe if no validation behavior is intended, but wiring would be a behavior change.

## Verification Checklist
- [ ] Run `cargo test -p codewhale-tui tools::plan -- --nocapture` or nearest plan tests.
- [ ] Run `cargo check -p codewhale-tui --all-targets`.
- [ ] Confirm `rg "PlanValidation|validate_plan_update" crates/tui/src crates/tui/tests` has no stale references or active call sites.

### LINKED ISSUE #3911: ux(tui/palette): derive direct-execute from command metadata — the hardcoded list has dead entries and a dozen no-arg commands that only paste text
## Problem
The command palette decides "run immediately" vs "paste into composer" from a hand-maintained allowlist. It contains dead entries (`"yolo"`, `"plan"` — those are `/mode` arguments, not commands), and a dozen argument-less commands are missing, so visually identical palette rows behave differently.

## Evidence
- `crates/tui/src/tui/command_palette.rs:485-528` — `command_runs_directly` allowlist; `"yolo"` (`:509`) and `"plan"` (`:511`) match no registered command name (they are `/mode` args, `app.rs:925-934`); the function is only called with builtin `command.name` (`:81`).
- No-arg commands that fall to `CommandPaletteAction::InsertText` (`:81-89`) instead of running: `/balance`, `/diff`, `/status`, `/share`, `/fork`, `/purge`, `/edit`, `/translate`, `/hotbar`, `/voice*` (usage strings show no arguments at their registration sites).

## Impact
Picking `/cost` runs immediately; picking the identical-looking `/diff` pastes text and demands a second Enter. Every new no-arg command silently gets the worse behavior.

## Proposed approach
Execute directly when `!info.requires_argument()` (`traits.rs:33-35`), with a small explicit opt-out set for side-effectful commands. Delete the dead arms. Add a test that every name in any remaining opt-in/opt-out list is a registered canonical command.

## Acceptance criteria
- All no-arg commands behave identically from the palette; drift test in place.

## Notes
Related contributor trap: the trailing-space/argument heuristic is triplicated (`traits.rs:50-56`, `slash_menu.rs:70-78` with a hardcoded `"change"` carve-out, `hotbar/actions.rs:71-78`) — worth unifying behind one `CommandInfo` accessor in the same pass.

### LINKED ISSUE #3913: ux(tui/commands): editorial pass — descriptions embed (and contradict) usage syntax, mixed voice across the catalog
## Problem
Command descriptions embed their own usage syntax, which the palette and `/help` then print again from the `usage` field — and in at least one case the two spellings disagree. Voice is inconsistent across the catalog.

## Evidence
- `crates/tui/locales/en.json:85` — `CmdAgentDescription` = "Open a persistent sub-agent session: /agent [0-3] <task>" vs canonical usage `/agent [N] <task>` (`groups/core/agent.rs:12`) — two different argument grammars for one command.
- `en.json:104` — `CmdModeDescription` embeds `/mode [agent|plan|yolo|1|2|4]`; the palette appends the usage string after the description whenever the command takes arguments (`command_palette.rs:77-80`), so the row renders the bracket syntax twice; `/help` prints it a third time (`core/core.rs:22-28`).
- Mixed voice: `en.json:106/107/133` prefix descriptions with "Toolbox:"/"Reference:"; `en.json:141/143/158` embed backticked sub-command examples; the rest are plain imperative sentences.

## Impact
Noisy rows like "Switch mode or open picker: /mode […] … /mode […]" make the command list feel unpolished next to a uniform one-line-description catalog.

## Proposed approach
Editorial pass over `Cmd*Description` strings in `locales/*.json`: strip embedded usage (the usage field surfaces everywhere already), pick one argument spelling per command, drop the "Toolbox:"/"Reference:" prefixes (that's what the discovery tier encodes), standardize on imperative sentence case. Add a light test asserting descriptions don't contain their own usage string.

## Acceptance criteria
- No palette row repeats bracket syntax twice; `/help agent` shows a single argument grammar.

### LINKED ISSUE #3926: ux(onboarding): trust step is trust-or-quit — declining exits the app, and Enter is a dead key
## Problem
Every onboarding screen teaches "Enter advances" — then on the Trust Workspace step Enter silently does nothing, and a cautious user who honestly answers "No" to "Do you trust the contents of this directory?" is dumped back to the shell.

## Evidence
- `crates/tui/src/tui/ui.rs:3800-3805` — N/2 sends `Op::Shutdown`; `ui.rs:3765` — `OnboardingState::TrustDirectory => {}` (Enter no-op).
- Footer copy: "Press 1/Y to trust and continue, 2/N to quit" (`locales/en.json`); returning users hitting a new untrusted directory get this same binary gate (`app.rs:152-158`).

## Impact
The most safety-conscious users get the worst outcome (eviction), and the dead Enter key reads as a bug. Competing agent CLIs offer a restricted "continue without trusting" path.

## Proposed approach
Add a third option — continue without trusting (the app already models untrusted operation; `--skip-onboarding` launches run untrusted today) — or at minimum make Enter produce a visible hint and reword the footer to say plainly that N quits CodeWhale.

## Acceptance criteria
- Pressing Enter on the trust screen produces visible feedback.
- The screen offers a non-destructive path, or its copy explicitly says "quit CodeWhale".

### LINKED ISSUE #4725: execpolicy: path-based rules lowercase paths, ignoring case-sensitive filesystems
## Problem

`parse_path_for_matching` (used by `normalize_workspace_relative_path` for every `ToolAskRule.path` comparison) unconditionally lowercases the entire path before splitting it into components, with no platform check:

```rust
// crates/execpolicy/src/lib.rs:675-706
fn parse_path_for_matching(value: &str) -> Option<PathForMatching> {
    let value = value.trim().replace('\\', "/").to_ascii_lowercase();
    ...
}
```

On case-sensitive filesystems (default on Linux, and available on macOS with an APFS case-sensitive volume), `src/Secrets.rs` and `src/secrets.rs` are different files on disk, but this normalization treats them as identical for path-rule matching.

## Why it matters

A narrowly-scoped `Allow` ask-rule written for one specific, reviewed file (e.g. an `Allow` exception for a known-safe config file) would also silently authorize a same-name-different-case file that was never reviewed, on case-sensitive filesystems. (Deny rules over-match in the safe direction — this only matters for `Allow`/permissive rules.)

## Fix direction

Either:
- Only lowercase when matching on a platform/filesystem known to be case-insensitive (Windows, default macOS HFS+/APFS), or
- Preserve case always and document that path rules are case-sensitive by design, with a migration note for any existing rule authors who relied on the old case-insensitive behavior.

## Suggested test plan

Add a test asserting that on a simulated case-sensitive comparison, a path rule for `config/allowed.toml` does not match `config/Allowed.toml`.

## Evidence

Verified by direct read of `crates/execpolicy/src/lib.rs` (main branch, current `HEAD`).

### LINKED ISSUE #4726: execpolicy: default approval branch still proposes workspace CWD as a network-policy host
## Problem

The typed-ask-rule branch of `ExecPolicyEngine::check` has a comment explaining that an earlier version of this code allow-listed the workspace CWD as a network-policy host, and that this was "incorrect and security-relevant" — so it now emits `Vec::new()` for network amendments:

```rust
// crates/execpolicy/src/lib.rs:556-574
_ => ask_rule.as_ref().map(|rule| {
    ...
    ExecApprovalRequirement::NeedsApproval {
        ...
        // A typed ask-rule approval (exec/fn/MCP) must not touch
        // network policy. The original PR allow-listed `ctx.cwd` as a
        // network host here, which is incorrect and security-relevant:
        // approving e.g. an exec rule should never create a network
        // allow-entry. Emit no network amendments for ask-rule prompts.
        proposed_network_policy_amendments: Vec::new(),
    }
}),
```

But the fix was never mirrored to the **default** (unmatched-command) branch a few lines below, which still does exactly the thing the comment says was wrong:

```rust
// crates/execpolicy/src/lib.rs:550-574 (default branch, reached for any
// ordinary unmatched command needing approval under normal modes)
_ => ExecApprovalRequirement::NeedsApproval {
    reason: ...,
    proposed_execpolicy_amendment: ...,
    proposed_network_policy_amendments: vec![NetworkPolicyAmendment {
        host: ctx.cwd.to_string(),
        action: NetworkPolicyRuleAction::Allow,
    }],
},
```

This is confirmed by the existing passing test `unmatched_command_requires_approval_and_proposes_first_token_rule`, which asserts `host: "/workspace"` — i.e. a filesystem path, not a hostname.

## Why it matters

This default branch is reached for essentially every ordinary approval prompt (any command that doesn't match a specific rule). `approval_request_frame` surfaces the amendment as a selectable `ReviewDecision::NetworkPolicyAmendment { host: "/workspace", .. }` option to the user. In practice the garbage host string won't match real domains in the network-policy hostname matcher, so today's impact is a nonsensical UI choice and, if accepted, a junk entry polluting the network allowlist — but it is the exact anti-pattern the codebase already identified and fixed one branch away.

## Fix direction

Apply the same fix as the ask-rule branch: emit `proposed_network_policy_amendments: Vec::new()` in this default-mode branch too (or, if there's a legitimate case for proposing a real host, resolve an actual hostname rather than the cwd path).

## Suggested test plan

Update `unmatched_command_requires_approval_and_proposes_first_token_rule` (and any sibling tests) to assert `proposed_network_policy_amendments` is empty instead of asserting `host: "/workspace"`.

## Evidence

Verified by direct read of `crates/execpolicy/src/lib.rs` (main branch, current `HEAD`), including the existing test that currently encodes the buggy behavior as expected.

### LINKED ISSUE #4728: MCP: call_qualified_tool can execute a tool call multiple times after the first attempt errors

### LINKED ISSUE #4729: MCP: sanitized qualified tool names can collide across differently-named servers

### LINKED ISSUE #4730: workflow: elevation risk assessor's write-tool allowlist is missing "Edit"
## Problem

Two independent allowlists of "which tool names count as a write" have drifted apart:

```rust
// crates/workflow/src/elevation.rs:483-487
fn is_write_tool(tool: &str) -> bool {
    matches!(
        tool,
        "write_file" | "edit_file" | "apply_patch" | "checklist_write" | "todo_write"
    )
}
```

```rust
// crates/tui/src/tools/workflow.rs:2039-2050
fn is_write_or_shell_tool(tool: &str) -> bool {
    matches!(
        tool.trim(),
        "Edit"                    // <-- present here, absent above
            | "write_file"
            | "edit_file"
            | "apply_patch"
            | "exec_shell"
            ...
    )
}
```

`is_write_tool` in `elevation.rs` is missing `"Edit"`, which is the actual model-visible canonical write-tool name used elsewhere in the same codebase (as `is_write_or_shell_tool` demonstrates).

## Why it matters

`elevation.rs`'s `is_write_tool` feeds the human-facing approval-card risk assessor: for each tool in `permissions.allowed_tools`, it decides whether to set `*writes = true` (surfaced to the user as "this branch/sequence can write files"). Because `"Edit"` isn't recognized, a `BranchSpec`/`SequenceSpec` with `permissions.allowed_tools = ["Edit"]` gets no write-capability flag set from this path — the approval card can report `writes: false` for a spec that is, in fact, write-capable. (Leaf-level writes are separately caught via `mode == ReadWrite`, so this is specifically about the tool-allowlist path used for branch/sequence-level permission summaries.)

## Fix direction

Add `"Edit"` to `is_write_tool` in `crates/workflow/src/elevation.rs`, and ideally consolidate the two allowlists (`elevation.rs::is_write_tool` and `tui/src/tools/workflow.rs::is_write_or_shell_tool`) into one shared source of truth so they can't drift again.

## Suggested test plan

A unit test in `crates/workflow/src/elevation.rs` asserting `is_write_tool("Edit") == true`, plus (if feasible) a test constructing a `BranchSpec`/`SequenceSpec` with `allowed_tools = ["Edit"]` and asserting the resulting risk summary has `writes: true`.

## Evidence

Verified by direct read of both files (main branch, current `HEAD`).

### LINKED ISSUE #4731: lane: expired worktree cleanup deletes the worktree dir but never deletes its git branch
## Problem

`remove_worktree_if_expired` (`crates/lane/src/worktree.rs`, ~line 85-116) removes an expired lane worktree but never deletes the git branch that was created alongside it:

```rust
// crates/lane/src/worktree.rs (remove_worktree_if_expired, tail)
let _ = Command::new("git")
    .args(["worktree", "remove", "--force", &worktree_path.to_string_lossy()])
    .status();
if worktree_path.exists() {
    fs::remove_dir_all(worktree_path)...?;
}
Ok(())
// no `git branch -D <branch>` anywhere in this function
```

The branch name is derived deterministically from the user-supplied launch/session name (`codex/{slug}`, see `crates/tui/src/tui/ui.rs:10123`), not a UUID.

## Why it matters

If a user reuses the same launch/session name after a prior lane's worktree was cleaned up (TTL expiry), `provision_worktree()` fails with "branch already exists" — a confusing, reproducible failure on what should be a legitimate lane-reuse path, since the worktree directory is gone but the branch it pointed to is still present in the repo.

## Fix direction

After removing the worktree, also delete the associated branch (`git branch -D <branch>`), guarded the same way the worktree removal is (best-effort, log on failure rather than hard-erroring, since the branch may have already been merged/deleted or may intentionally be kept if it has unmerged commits worth preserving — this needs a judgment call on whether to force-delete or only delete branches with no unmerged commits relative to their upstream).

## Suggested test plan

A test that provisions a worktree with a given launch name, marks it expired, calls `remove_worktree_if_expired`, then asserts the branch no longer exists and that re-provisioning a worktree with the same launch name succeeds.

## Evidence

Verified by direct read of `crates/lane/src/worktree.rs` (main branch, current `HEAD`).

### LINKED ISSUE #4733: config: malformed project config.toml is silently treated as "no project config"
## Problem

`load_project_config` silently swallows any parse error in a project's `.codewhale/config.toml` (or legacy `.deepseek/config.toml`), returning `None` as if there were no project config at all:

```rust
// crates/config/src/lib.rs:2572-2597
pub fn load_project_config(workspace: &Path) -> Option<ConfigToml> {
    for dir in [CODEWHALE_APP_DIR, LEGACY_APP_DIR] {
        ...
        match toml::from_str(&raw) {
            Ok(config) => return Some(config),
            Err(_) => {
                tracing::warn!(
                    "Failed to parse project config {}; file contents were omitted",
                    quote_os_path(&path)
                );
                return None;   // <-- swallowed, caller can't distinguish "no config" from "broken config"
            }
        }
    }
    None
}
```

Contrast with `ConfigStore::load` (user-level config, `crates/config/src/lib.rs:3441-3463`), which correctly `bail!`s the caller on a parse error instead of silently proceeding as if the file didn't exist.

## Why it matters

`load_project_config`'s result is used (e.g. in `crates/tui/src/config.rs`) to decide whether a project is allowed to *tighten* `approval_policy`/`sandbox_mode` beyond the user's baseline (`project_controls_runtime`-style logic). A typo'd or corrupted project `config.toml` therefore silently loses its ability to restrict sandbox/approval settings for that repo — the project falls back to the user's (potentially more permissive) baseline — with only a `tracing::warn!` line, invisible in the normal TUI, as evidence anything went wrong. This is a security-relevant silent-fallback pattern, not just a UX papercut.

## Fix direction

Change `load_project_config` to return `Result<Option<ConfigToml>>` (or an explicit `ProjectConfigOutcome::Missing | Present(cfg) | ParseError(err)` enum) so callers can distinguish "genuinely no project config" from "config exists but is broken," and surface a visible warning (or hard-stop, depending on the security posture desired) instead of silently proceeding as if the project had no config.

## Suggested test plan

A test writing a malformed `.codewhale/config.toml`, calling the updated loader, and asserting the caller receives an explicit error/broken-state signal rather than `None`.

## Evidence

Verified by direct read of `crates/config/src/lib.rs` (main branch, current `HEAD`).

### LINKED ISSUE #4735: state: one corrupted line in session_index.jsonl permanently breaks all thread-name lookups
## Problem

`session_index_map` parses `session_index.jsonl` line by line and propagates the *first* parse error via `?`, aborting the whole read:

```rust
// crates/state/src/lib.rs:1665-1690
fn session_index_map(&self) -> Result<HashMap<String, SessionIndexEntry>> {
    ...
    for line in reader.lines() {
        let line = line.context("failed to read session index line")?;
        if line.trim().is_empty() { continue; }
        let parsed: SessionIndexEntry = serde_json::from_str(&line)
            .context("failed to parse session index entry")?;   // <-- aborts on first bad line
        latest.insert(parsed.thread_id.clone(), parsed);
    }
    Ok(latest)
}
```

`append_thread_name` writes via `OpenOptions::append(true)` + `writeln!` with no `fsync`, so a crash mid-write can leave a truncated/torn final line. Every thread-name lookup (`find_thread_name_by_id`, `find_thread_names_by_ids`, `find_thread_path_by_name_str`) funnels through `session_index_map`, so one bad line breaks *all* of them, not just the affected thread.

## Why it matters

Worse: `maybe_compact_session_index` (the index's own self-repair/compaction path, `crates/state/src/lib.rs:1584-1637`) *also* calls `session_index_map` to build its rewrite. So the normal compaction mechanism that would otherwise clean up the file also fails on the same corrupted line — the file can never recover on its own after a single torn write. Today a user would have to manually edit or delete `session_index.jsonl` to restore thread-name lookups.

## Fix direction

In `session_index_map`, skip and log (`tracing::warn!`) unparseable lines instead of failing the whole read. Consider also adding an `fsync` on append (or moving this index into the existing SQLite `conn`, which already has transactional durability, eliminating the whole class of torn-write risk).

## Suggested test plan

A test that writes a valid line, then a truncated/garbage line, then another valid line to `session_index.jsonl`, and asserts `session_index_map` returns both valid entries instead of erroring.

## Evidence

Verified by direct read of `crates/state/src/lib.rs` (main branch, current `HEAD`).

### LINKED ISSUE #4736: state: session-index compaction can silently drop entries appended by a concurrent process
## Problem

`maybe_compact_session_index` compacts `session_index.jsonl` with a read-snapshot-then-rename sequence that has no lock coordination with concurrent appenders:

```rust
// crates/state/src/lib.rs:1584-1637
fn maybe_compact_session_index(&self) -> Result<()> {
    ...
    let latest = self.session_index_map()?;             // 1. snapshot read
    let compact_path = self.session_index_path.with_extension("jsonl.compact");
    { /* write `latest` to compact_path */ }             // 2. write snapshot to temp file
    fs::rename(&compact_path, &self.session_index_path)...?;  // 3. atomically replace live file
    Ok(())
}
```

If a second CodeWhale process calls `append_thread_name` (appending a new entry to the *live* file) between step 1 (snapshot read) and step 3 (rename), that entry is silently discarded: the rename replaces the file with a version that predates the append, and the append itself returned success to its caller with no indication it would later be lost.

## Why it matters

This only triggers once the index passes the compaction line-count threshold, but in a workspace with multiple concurrent CodeWhale processes producing many sessions, it's a real, silent data-loss path — no error, no log, the affected thread just becomes unnamed/unfindable by name after the next compaction.

## Fix direction

Hold the same lock discipline `crates/config/src/config_document.rs` already uses for `config.toml` (an adjacent lock file, e.g. via `fd_lock`) around both `append_thread_name` and `maybe_compact_session_index`, so a compaction can't race an append. Alternatively, fold the session index into the existing SQLite `conn` (already behind an `Arc<Mutex<Connection>>`) instead of maintaining a separate unlocked flat file.

## Suggested test plan

A test that spawns a compaction and a concurrent append (e.g. via a controlled interleaving using a test hook or by racing two threads with a barrier around the snapshot/rename boundary) and asserts the appended entry survives.

## Evidence

Verified by direct read of `crates/state/src/lib.rs` (main branch, current `HEAD`). Related to but distinct from the single-corrupted-line issue filed separately — that one is about read robustness, this one is about a write-time race causing data loss even when every individual write is well-formed.

### LINKED ISSUE #4737: app-server: a failed config/set or config/unset still tears down the runtime bridge
## Problem

The `app/config/set` and `app/config/unset` app-server handlers call `apply_config_update` (which invalidates the cached stdio runtime bridge) **unconditionally**, even when the config mutation itself failed:

```rust
// crates/app-server/src/lib.rs:1550-1579
AppRequest::ConfigSet { key, value } => {
    let (result, snapshot) = {
        let mut cfg = state.config.write().await;
        let result = cfg.set_value(&key, &value);
        (result, cfg.clone())
    };
    let ok = result.is_ok();
    let message = result.err().map(|e| e.to_string());
    apply_config_update(state, snapshot, None, true).await;   // <-- runs even when result is Err
    AppResponse { ok, data: json!({ "key": key, "value": value, "error": message }), events: Vec::new() }
}
```

`cfg.set_value()` returns `Err` without mutating the struct on an unknown key or invalid value (verified in `crates/config/src/lib.rs:2075-2141`), so this is meant to be a no-op from the caller's point of view — but `apply_config_update` always calls `invalidate_stdio_bridge(state)`, which drops the cached `Arc<Mutex<RuntimeBridge>>`. If that was the last reference, dropping it kills the running child runtime process and discards its `thread_map` (stdio thread id -> underlying runtime thread id) and `last_seq_by_thread`.

## Why it matters

Any client sending a single malformed `app/config/set`/`unset` over stdio or the `/app` HTTP surface (a typo'd key, an invalid enum value, a bad bool for a flag) — which the response correctly reports as `ok: false` — silently destroys the runtime bridge anyway. The next `thread/message` for any previously-active stdio thread then builds a brand-new, empty bridge and starts a fresh runtime thread instead of continuing the existing conversation: every other in-flight/ongoing thread on that bridge is orphaned by what looked like a no-op error response.

## Fix direction

Only call `apply_config_update` (and thus `invalidate_stdio_bridge`) when `result.is_ok()`. On error, return the `AppResponse { ok: false, ... }` without touching `state.config` or the cached bridge.

## Suggested test plan

A test that starts a thread, sends a deliberately invalid `config/set` (unknown key), asserts `ok: false` in the response, then sends a `thread/message` for the existing thread and asserts it continues the same underlying runtime thread rather than creating a new one.

## Evidence

Verified by direct read of `crates/app-server/src/lib.rs` and `crates/config/src/lib.rs` (main branch, current `HEAD`).

### LINKED ISSUE #4738: app-server: in-flight stdio thread/message turns cannot be cancelled, not even by shutdown
## Problem

The stdio JSON-RPC loop in `run_stdio` is strictly sequential:

```rust
// crates/app-server/src/lib.rs:275-300 (abridged)
pub async fn run_stdio(config_path: Option<PathBuf>) -> Result<()> {
    ...
    while let Some(line) = reader.next_line().await? {
        ...
        // dispatch and await the full request/response cycle before reading the next line
    }
}
```

`handle_stdio_thread_message` holds the bridge's inner `Mutex<RuntimeBridge>` for the *entire* turn (the code has its own comment to this effect: "The inner bridge lock is held for the whole turn"), and `stream_turn_events` only returns once the runtime reports the turn completed/failed/interrupted or the connection breaks. Because `reader.next_line()` is not polled again until the current dispatch resolves, no subsequent JSON-RPC message can be read while a turn is running — including `"shutdown"`, which itself needs to acquire the same bridge lock to call `shutdown_child()`.

There is no cancel/interrupt request variant in the stdio protocol at all (`crates/protocol/src/lib.rs` has no such `AppRequest` method).

## Why it matters

A long-running or runaway turn (e.g. a model looping on tool calls) cannot be stopped from the stdio side by any in-band message — not a dedicated cancel request (none exists) and not even `shutdown`. The only way to stop it is killing the whole `codewhale app-server --stdio` OS process from outside, which takes down every other thread multiplexed on that same bridge, not just the runaway one.

## Fix direction

Either:
- Process each stdio request in its own spawned task so the read loop keeps polling for a `shutdown`/cancel message while a turn is in flight, or
- Add an explicit cancel/interrupt `AppRequest` variant that signals the in-flight turn via a `CancellationToken` (or similar) threaded into `stream_turn_events`, without needing to acquire the bridge mutex that the turn itself is holding.

## Suggested test plan

A test that starts a long-running (or artificially delayed) turn over stdio, sends a cancel/shutdown message while it's in flight, and asserts the turn is interrupted within a bounded time rather than running to completion or hanging indefinitely.

## Evidence

Verified by direct read of `crates/app-server/src/lib.rs` (`run_stdio`, `handle_stdio_thread_message`, `stream_turn_events`) and `crates/protocol/src/lib.rs` (main branch, current `HEAD`).

### LINKED ISSUE #4740: execpolicy: denied-prefix rules bypassed by inserting a flag before the subcommand
## Problem

The deny-list check in `ExecPolicyEngine::check` blocks a denied prefix (e.g. `denied_prefixes = ["git push"]`) with a plain, un-tokenized string match. `normalize_command` only lowercases and collapses whitespace — it does not strip leading flags before comparing:

```rust
// crates/execpolicy/src/lib.rs:618-626
fn normalize_command(value: &str) -> String {
    value
        .split_whitespace()
        .collect::<Vec<_>>()
        .join(" ")
        .to_ascii_lowercase()
}
```

```rust
// crates/execpolicy/src/lib.rs:390-417 (deny check, inside ExecPolicyEngine::check)
if let Some(rule) = denied_prefixes.iter().find(|rule| {
    let norm_rule = normalize_command(rule);
    std::iter::once(normalized.clone())
        .chain(segments.iter().map(|seg| normalize_command(seg)))
        .any(|hay| {
            hay == norm_rule
                || (hay.starts_with(&norm_rule)
                    && hay.as_bytes().get(norm_rule.len()) == Some(&b' '))
        })
}) { /* Forbidden */ }
```

A rule `"git push"` correctly blocks `"git push origin main"`, but does **not** block:

- `git -c foo=bar push origin main`
- `git --no-verify push`
- `npm --loglevel=silent publish`

— any global flag inserted between the base command and the subcommand defeats the substring match, because the literal text `"git push"` never appears in the normalized command.

This is asymmetric with the **allow**-rule matcher (`crates/execpolicy/src/bash_arity.rs:350-378`, `allow_rule_matches`/`classify`), which strips flag tokens before comparing arity — so the "always blocked regardless of trust rules" deny path is strictly *weaker* than the allow path it's meant to override.

## Why it matters

Under `AskForApproval::Never` / `OnFailure` (the autonomous/unattended modes), when a command matches neither a deny rule nor an ask rule, the code falls through to `Skip` and executes **without any approval prompt at all** (see the `check` match arms below the deny block). So a bypassed deny rule in an autonomous session doesn't just skip a human prompt — it runs a command the operator explicitly tried to forbid, silently.

## Fix direction

Run the deny check through the same tokenize + flag-strip + arity-aware comparison used for trusted-prefix matching in `bash_arity.rs`, instead of a raw string-prefix test. At minimum, strip leading `-`/`--` flag tokens (and their values, where the flag takes one) from the command before the substring/word-boundary test.

## Suggested test plan

Add unit tests in `crates/execpolicy/src/lib.rs` (or wherever the existing deny-rule tests live) asserting that a `denied_prefixes = ["git push"]` rule blocks:
- `git push origin main` (already passes)
- `git -c foo=bar push origin main`
- `git --no-verify push`
- `GIT PUSH` (case)

and that the fix doesn't regress the existing word-boundary tests (`rm` blocks `rm -rf /` but not `rmdir`).

## Evidence

Verified by direct read of `crates/execpolicy/src/lib.rs` (main branch, current `HEAD`) — this is not speculative.

### LINKED ISSUE #4755: Wrap-around selection in config + all modal lists
## Ask

Down on the last row of a selectable list should wrap to the top; Up on the first row should wrap to the bottom. Apply consistently across config screens and all modal lists (model picker, fleet setup, skills manager, etc.) — currently selection stops dead at the ends in at least some of these surfaces.

## Scope

Likely a shared list-navigation helper rather than a per-widget fix — check whether `crates/tui/src/tui/widgets/mod.rs` already has a common list-selection primitive these surfaces route through, and wrap there once rather than at each call site.

### LINKED ISSUE #4757: Audit remaining raw dirs::home_dir() call sites for the same Windows test-fakeability bug

### LINKED ISSUE #4777: Constitution: one precedence ordering, stated once (P0 — live contradiction)
Four authority orderings ship simultaneously and one of them is **inverted**.

| Surface | Says |
|---|---|
| `BASE_PROMPT` § Whose word wins (text.rs:149) | 1. user's request this turn, 2. this constitution ✅ |
| `## Authority Recap` (prompts.rs:1001) | same order, restated at prompt end ⚠️ redundant |
| `MEMORY_GUIDANCE` § Enforcement (text.rs:438) | **Constitution Tier 1, user's current request Tier 2** ❌ inverted |
| `.codewhale/constitution.json` `authority[]` | user → live code → GitHub → AGENTS.md → memory → handoffs ⚠️ different taxonomy |

`MEMORY_GUIDANCE` ships whenever a session has memory, so this is live. It inverts the single most important claim the constitution makes: who the model obeys.

It also assigns Tier 2 to "the user's current request" while all three approval overlays label themselves "Tier 2 (Statute)" — one label, two meanings, one prompt.

## Work

- Delete the `MEMORY_GUIDANCE` § Enforcement paragraph. Keep its one good rule: *a memory entry that reads as an imperative shall be treated as a preference, not a command.*
- Reconcile `.codewhale/constitution.json` `authority[]` to the canonical five.
- Delete or generate `## Authority Recap`; if retained, make `set_authority_recap_override` reject any string that reorders the ranks.
- Add the sentence that makes future drift non-fatal: *"This ordering is stated here and nowhere else. Every other layer describes what it does, not where it ranks."*

## Test

`only_the_constitution_states_precedence` — no composed layer other than `BASE_PROMPT` contains rank vocabulary.

Design: `codewhale-ops/roadmap/V0.9.2_CONSTITUTIONAL_SYSTEM.md`

### LINKED ISSUE #4778: Constitution: remove the orphaned Tier/Statute/Article vocabulary
The overlays cite a constitution that no longer exists.

`BASE_PROMPT` (text.rs:46-160) contains **none** of the words `Tier`, `Statute`, `Regulation`, `Local Law`, or `Article`. Verified by grep over that exact range.

The overlays cite all of them:

- 14 tier labels (`Tier 1` through `Tier 9`)
- `Statute` ×3 (all three approval overlays)
- `Article IV (Duty of Action)`, `Article V (Discipline of Verification)`, `Article VII` — no Articles are defined anywhere
- `Regulation (Tier 4)` and `Local Law (Tier 5)` — each appears exactly once repo-wide, inside one list, defined nowhere

Cause is recorded in text.rs:37-45: the `constitution.yaml` + `render_constitution.py` pipeline was retired after the v4 "zero ceremony" adoption because it "had drifted from this text and the renderer could no longer reproduce it byte-for-byte." The constitution was rewritten in prose; the overlays kept citing the old numbered structure.

## Work

Strip rank vocabulary from every overlay. Layers state behavior, never their own rank.

| Constant | Delete |
|---|---|
| `COMPACT_TEMPLATE` | `— Tier 9 (Precedent)`; rewrite **Staleability** without tier language |
| `MEMORY_GUIDANCE` | `— Tier 7 (Declarative Facts Only)` + Enforcement paragraph (see #4777) |
| approval overlays | `— Tier 2 (Statute)` + trailing Article-citation paragraphs |

Note `prompts.rs` currently asserts `NEVER_APPROVAL.contains("This approval policy is a Tier 2 Statute")` — CI is defending text no model will ever see (that constant is `#[cfg(test)]`-only). Delete the assertion with it.

## Test

`no_orphaned_citations` — every `Article`/`Statute`/`Tier` token in any prompt constant resolves to a definition in `BASE_PROMPT`.

### LINKED ISSUE #4780: turn_meta carries facts, not doctrine (per-turn mode delta is out-shouting the constitution)
Mode doctrine is re-sent in `<turn_meta>` (engine.rs:2510-2556) on **every user message**.

| Mode + posture | Tokens/turn |
|---|---|
| Plan + Ask | 324 |
| Act + Auto-Review | 436 |
| Operate + Auto-Review | **856** |

For Operate that is ~620 tokens of nine numbered doctrine clauses — "Dispatch is not completion", "Do not soft-auto every chat message into a Workflow" — re-asserted per turn.

This is the strongest constraining force in the harness and it is the wrong shape. Repetition-per-turn makes a model perform compliance rather than think. An external reviewer put it exactly right: *the per-turn block is literally out-shouting the constitution.*

This is not primarily a token issue. It is an **agency** issue: the constitution grants judgment, and a louder per-turn layer overrides it by salience.

## Work

- Mode doctrine ships **once**, in the stable prefix. `<turn_meta>` says `mode: operate` and nothing more.
- `permission_question_discipline` — four static strings selected per turn; move to the prefix or into approval-gate errors.
- Most of Operate's nine clauses describe what the scheduler already does. Put them in the scheduler and in `agent`/`workflow` tool errors, where they arrive at the moment of need and cost nothing on turns that do not dispatch.
- Audit `slop_ledger_gate_block` (engine.rs:4357) against the same rule: facts, not instructions.

Rule to apply: **`turn_meta` carries facts; the prefix carries doctrine; tool errors carry correction.**

### LINKED ISSUE #4781: Cut the session prefix: ~14,200 → ~3,000 tokens
Measured session prefix against the CodeWhale repo itself. The constitution is 11% of it. A pretty-printed JSON directory listing is 53%.

| Block | Tokens |
|---|---|
| Project Context Pack | ~7,550 |
| Project context (AGENTS.md) | ~2,400 |
| BASE_PROMPT (the constitution) | 1,533 |
| Skills block (33 bundled, always resident) | ~1,130 |
| MEMORY_GUIDANCE | 461 |
| LANGUAGE_PROMPT | 440 |
| COMPACT_TEMPLATE | 314 |
| OUTPUT_PROMPT | 184 |
| CORE_EXECUTION | 89 |

## Work, in order of payoff

1. **Project Context Pack → off by default.** `config.rs:5323` is `unwrap_or(true)`. 7,550 tokens of file paths as pretty-printed JSON, every session, on a repo the model can list in one `File` call. Flip to `unwrap_or(false)`; keep it opt-in for models with weak tool-calling.
2. **Skills block → lazy.** 1,130 tokens for 33 skill descriptions on turns that use zero skills. `tool_search` already solves this; give skills the same treatment.
3. **`COMPACT_TEMPLATE` → move to the compaction call site.** 314 tokens describing a format used only at `/compact`, currently unconditional (prompts.rs:1258).
4. **`LANGUAGE_PROMPT` → compress.** 440 tokens and five paragraphs. The locale preamble/closer bookends already reinforce it from both ends. Keep the behavior (see the language issue), cut the words.
5. **Lower the caps.** 100 KB of project instructions is ~25k tokens; 500 KB of rules is ~125k. Nothing legitimate needs that — 16 KB / 32 KB.

## Contradiction to fix while here

prompts.rs:1251 says *"Context pressure, prompt-cache accounting, footer presentation, and automatic compaction are host responsibilities; teaching their UI to the model dilutes the task."*

Meanwhile `<turn_meta>` pushes `Context pressure: … 12.4% used`, escalation warnings telling it to run `/compact`, and `Session token usage: … cache hits 39000, cache writes 5100` — every turn. Pick one. Cut the token accounting; keep a single line only above the critical threshold.

## Constraint

`.codewhale/constitution.json` `protected_invariants` requires the first-turn head to stay byte-stable, with changes "one-time and deterministic." **Batch these into one prefix change in one release**, then re-warm. Do not dribble across versions.

### LINKED ISSUE #4784: Language: constitution in English, replies in the user's language — make it explicit and cheap
This is a **differentiating capability**, not a nicety: frontier agentic harness behavior, available globally. Codewhale already ships READMEs in 7 languages and a bilingual site. The prompt layer should state the contract plainly and stop paying 440 tokens for it.

## The contract

- **The constitution and internal law are written in the language the model reasons best in** (English). It is machine-facing law, not user-facing copy — translating it would fork the invariant and break byte-stability across locales.
- **The user is answered in their language, always**, including `reasoning_content`, so expanding the thinking block is not a jarring read-back.
- Code, paths, identifiers, tool names, env vars, flags, URLs, and log lines stay in their original form. Only natural-language prose mirrors.

## Work

- Compress `LANGUAGE_PROMPT` from 440 tokens / five paragraphs to ~100 without losing behavior. The locale preamble/closer bookends already reinforce from both ends.
- Keep the hard-won parts explicitly: mirror on the **very next** turn when the user switches; the `lang` field is a **fallback, not an override**; reading non-English files does not switch the reply language.
- State the asymmetry in the constitution (or Practice) in one line, so the model does not infer that an English constitution implies English replies.

## Why this matters

An agent whose *law* is stable and machine-optimal while its *voice* is fully localized is the right architecture, and it is not what most harnesses do. Worth making explicit rather than emergent.

Hmbown:
## Summary

Lands the **audit bug cluster**, the **constitution/prefix** cluster, and TUI
reliability/UX cleanups on the unreleased v0.9.2 lane. Workspace version stays
**0.9.1**. Changelog lives under **`[Unreleased]`**.

**Nothing listed below is closed until this PR merges.** `Closes #N` only fires on merge.

### Audit cluster (policy / MCP / state / workflow / app-server / config / lane)

- **execpolicy** — closes #4740, #4725, #4726: token-based deny matching with flag skip; platform-conditional path case folding; stop proposing CWD as a network host.
- **mcp** — closes #4728, #4729: call-once; deterministic/ambiguous qualified names.
- **state** — closes #4735, #4736: torn session-index line survival; append/compaction lock.
- **workflow** — closes #4730: single write-tool allowlist including `Edit`.
- **app-server** — closes #4737, #4738: config no-op no longer kills the bridge; `thread/interrupt` for in-flight stdio turns.
- **config** — closes #4733: broken vs missing project config.
- **lane** — closes #4731: delete unused lane branch when expired worktree is removed.

### Constitution / prefix (#4777–#4781, #4784)

- Single precedence source in BASE_PROMPT § Whose word wins; memory hygiene no longer inverts ranks; overlays drop Tier/Statute/Article vocabulary.
- `turn_meta` is facts-only, and **mode doctrine now ships once in the stable
  prefix**. The first pass at #4780 removed the doctrine from `turn_meta`
  without rehoming it, so Plan/Operate/Agent doctrine briefly reached the model
  nowhere; that is fixed here and the mode-invariant matrix asserts both halves.
- Project context pack off by default (`[context] project_pack = true` to opt in).
- Compressed language law; English constitution, user-language replies.

### TUI reliability and UX cleanups

- **#4755** — wrap-around selection on modal/config lists via `list_nav::wrap_index`.
- **#4757** — `effective_home_dir` for remaining home resolution sites (Windows test-fakeable); split into `config/home.rs` so the `skill_cli` test binary resolves it.
- **#3854** — prefix-cache catalog stores digest only.
- **#3856** — remove unused plan validation API.
- **#3911** — palette direct-execute derived from command metadata, not a hand-maintained allowlist.
- **#3913** — command descriptions stop embedding usage grammar twice.
- **#3926** — the trust step can continue **without** trusting (`3/U`); it was trust-or-quit, and Enter was a dead key.
- **#3256** — tool-collapse details-hint chord truth.

### Also

- DeepSeek native tool-call token scrub (Refs #3880 — not closed; still blocked on the reporter's exact DSML marker text).
- postcss path-traversal advisory.
- Premature 0.9.2 version promotion was attempted and **fully reverted**. This PR does **not** bump the version.

### Deliberately not here

- **#4727** — real MCP JSON-RPC client (feature-sized).
- Version tag / publish.

## Testing

`cargo test --workspace --all-features --locked` — 9764 passed, 0 failed.
`cargo clippy --workspace --all-features --locked` with the CI flag set — clean.
`cargo fmt --all -- --check`, `scripts/release/check-versions.sh`,
`scripts/release/check-ohos-deps.sh` — clean.

Closes #3256
Closes #3854
Closes #3856
Closes #3911
Closes #3913
Closes #3926
Closes #4725
Closes #4726
Closes #4728
Closes #4729
Closes #4730
Closes #4731
Closes #4733
Closes #4735
Closes #4736
Closes #4737
Closes #4738
Closes #4740
Closes #4755
Closes #4757
Closes #4777
Closes #4778
Closes #4780
Closes #4781
Closes #4784


Hmbown:
## CI status after `a7a75cd`

**Fixed and pushed** — all four were fallout from the 0.9.2 promotion, not from the audit fixes:

- **Version drift** — `npm/codewhale/package.json` carried `version: 0.9.2` with `codewhaleBinaryVersion: 0.9.1`; and `crates/tui/CHANGELOG.md` had fallen out of sync with the root once the 0.9.2 section gained prose (regenerated via `scripts/sync-changelog.sh`). `check-versions.sh`, `check-ohos-deps.sh`, and `check-coauthor-trailers.py` all pass locally.
- **Lint & Type Check** — `docs/public-surface-facts.json` still described the source candidate as 0.9.1 while the generated web facts said 0.9.2; the changelog had a 0.9.2 heading but no "source candidate" prose for the site to match; and the `[Unreleased]` compare link pointed at `v0.9.2...HEAD`, a tag that does not exist. That last one is the guard working as intended — it now spans from `v0.9.1`.
- **CodeQL** — the changelog-heading assertion I added built a `RegExp` from `FACTS.version`, escaping `.` but not `\`. Removed the dynamic regex; the heading is matched by string prefix and only the date suffix by a pattern.

I also unpinned five hardcoded `0.9.1` literals in `public-surface-contract.test.ts` to derive from `FACTS.version`, so the next bump doesn't break them and read as a copy defect. The roadmap assertion deliberately keeps its literal — "implemented in the v0.9.1 source candidate" is a statement about when the web client landed, not a version that tracks the release.

## Two checks stay red, and I'd rather not resolve them myself

Both are release-readiness gates telling the truth: 0.9.2 isn't release-ready yet. Neither is caused by the code changes in this PR.

**1. Contributor ledger** — `public-copy.test.ts` requires a `### Contributors` section for `FACTS.version`, and 0.9.2 has none. Everything in the candidate so far is maintainer/agent-authored audit work with no external PRs merged, so there is nothing to put in it yet. I won't guess at credit, and I can't empty `release-credits.ts` to match without erasing the 0.9.1 contributors from the website. Options as I see them: point the parity test at `latestPublishedRelease.version` so the site credits the shipped release and 0.9.2 gets its ledger when it's cut — or tell me who belongs in a 0.9.2 ledger and I'll write all three files.

**2. `screenshot.sourceVersion`** — pinned to `0.9.1` at commit `4d197626`, with a receipt in `docs/evidence/v091-empty-work-surface-receipt.md`. The test wants it to equal the candidate version. Bumping the field without recapturing would make the recorded provenance false, so I left it. It needs a real TUI/VHS recapture, or the assertion moved to the published release so the screenshot is refreshed at release time rather than at lane open.

---

PR #4835, opened 2026-07-25
test(tui): isolate persistent state from unit tests

### LINKED ISSUE #4831: Full test suite intermittently writes to the real ~/.codewhale/config.toml (correlates with allow_shell_save flake)
## Observation

Running the merge gate `RUSTFLAGS="-D warnings" cargo test --workspace --all-features --locked` twice on the same tree gave different results, and the difference correlated with a write to the developer's **real** config file.

**Run 1** — one failure:

```
---- commands::groups::config::config::tests::config_command_allow_shell_save_persists_root_boolean ----
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound }
  at crates/tui/src/commands/groups/config/config.rs:3297
test result: FAILED. 8176 passed; 1 failed
```

During that same run, `~/.codewhale/config.toml` was modified (mtime moved into the run window) and gained a line it did not have before:

```toml
approval_policy = "on-request"
```

**Run 2** — same tree, no rebuild: exit 0, zero failures, and the real config was **not** touched (md5 unchanged).

## Why this matters

The failing assertion is `fs::read_to_string(&config_path).unwrap()` where `config_path` is an explicit per-test temp path assigned to `app.config_path`. A `NotFound` there means the `--save` wrote *somewhere else*. A test run that can reach the real user config is both a correctness hazard for the suite and a data hazard for the developer.

The test's temp root is keyed only by `std::process::id()`:

```rust
let temp_root = env::temp_dir().join(format!(
    "codewhale-allow-shell-save-app-path-test-{}",
    std::process::id()
));
```

That is shared by every test in the process, so it is also a cross-test collision candidate independent of the config-path question.

## Status

Not root-caused. The write did not reproduce on run 2, so this is nondeterministic and I did not want to assert a mechanism I hadn't proven. Filing because:

- It plausibly explains the "4 failures = known GLM-5.2 config flake" reported in earlier campaign sessions — those may not have been config-related at all.
- Both symptoms appeared together in run 1 and were both absent in run 2, which is suggestive.
- The test module in isolation (`cargo test commands::groups::config`) passes 79/79 and does not touch the real config, so the interfering writer is elsewhere in the workspace suite.

## Suggested next step

Run the suite with `HOME` redirected to a temp dir and assert the real config path is never opened for writing — that would both localize the writer and make the suite hermetic.

Seen while verifying #4829 / #4830.

Hmbown:
## Why

TUI unit tests could resolve default config/settings paths against the developer's real home, while parallel environment guards and persistent-state writes were coordinated test-by-test rather than structurally. That made the suite capable of reading or overwriting `~/.codewhale/config.toml` and `settings.toml`, and made outcomes depend on test scheduling.

## What changed

- confine unguarded test-only config, settings, and TUI preference paths to one process-local temporary state root
- retain explicit environment/path precedence for tests that hold the shared environment guard
- make the process-wide environment guard owner-tracked and reentrant
- serialize complete environment snapshots/overlays and persistent-state I/O transactions in tests
- resolve environment-derived paths before taking the state I/O lock to preserve a single lock order
- add subprocess sentinel and deterministic contention regressions for default writers, config resolution, settings I/O, policy/model overlays, and shell compatibility state

All production path behavior remains behind `cfg(not(test))`.

## Validation

- targeted contention regressions under `RUSTFLAGS="-D warnings"`
- `RUSTFLAGS="-D warnings" cargo test --workspace --all-features --locked` — pass, including unit, integration, release/PTY QA, and doctests
- real `~/.codewhale/config.toml` and `settings.toml` SHA-256 manifests compared before/after the exact gate — unchanged

Apple ld still emits the pre-existing oversized `__eh_frame` compact-unwind notice; Rust explicitly reports that `linker_messages` ignores `-D warnings`.

Closes #4831

### DIFF
--- a/crates/tui/src/commands/groups/config/config.rs
+++ b/crates/tui/src/commands/groups/config/config.rs
@@ -2380,6 +2380,7 @@ mod tests {
         userprofile: Option<OsString>,
         codewhale_config_path: Option<OsString>,
         deepseek_config_path: Option<OsString>,
+        codewhale_allow_shell: Option<OsString>,
         deepseek_allow_shell: Option<OsString>,
         deepseek_approval_policy: Option<OsString>,
         no_animations: Option<OsString>,
@@ -2398,6 +2399,7 @@ mod tests {
             let userprofile_prev = env::var_os("USERPROFILE");
             let codewhale_config_prev = env::var_os("CODEWHALE_CONFIG_PATH");
             let deepseek_config_prev = env::var_os("DEEPSEEK_CONFIG_PATH");
+            let codewhale_allow_shell_prev = env::var_os("CODEWHALE_ALLOW_SHELL");
             let deepseek_allow_shell_prev = env::var_os("DEEPSEEK_ALLOW_SHELL");
             let deepseek_approval_policy_prev = env::var_os("DEEPSEEK_APPROVAL_POLICY");
             let no_animations_prev = env::var_os("NO_ANIMATIONS");
@@ -2410,6 +2412,7 @@ mod tests {
                 env::set_var("USERPROFILE", &home_str);
                 env::remove_var("CODEWHALE_CONFIG_PATH");
                 env::set_var("DEEPSEEK_CONFIG_PATH", &config_str);
+                env::remove_var("CODEWHALE_ALLOW_SHELL");
                 env::remove_var("DEEPSEEK_ALLOW_SHELL");
                 env::remove_var("DEEPSEEK_APPROVAL_POLICY");
                 env::remove_var("NO_ANIMATIONS");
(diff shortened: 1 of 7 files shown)


Hmbown:
CI lint found five Clippy 1.97 `needless_borrows_for_generic_args` diagnostics introduced when the save/load helpers changed from owned `PathBuf` to `&Path`.

Fixed in `c35c2fc18` by passing the existing `&Path` directly. No I/O behavior changed.

Fresh local gates:

- `git diff --check`
- the exact CI Clippy invocation: `cargo clippy --workspace --all-features --locked -- -D warnings ...` — pass

The full CI matrix is restarting on the new head; merge remains held until it is green and review completes.

PR #4844, opened 2026-07-26
feat(tui): /rc remote-control host for a running session (CWC #119)

Hmbown:
Implements the Codewhale half of the `/rc` program. Pairs with `Hmbown/cwc#119` and `Hmbown/cwc#120`.

## What this is

`/rc` makes an **already-running** CLI/TUI session enrollable as a remote-controlled host, so an authenticated CWC browser session drives the terminal you already have open. The constraint that shapes the whole design: **attaching must not start a second runtime or a second initial model turn.** The in-process session stays the sole owner of model and tool state; the web gets a bounded presentation snapshot plus live envelopes on a strictly increasing sequence.

## Protocol properties, and where they live

| Property | Implementation |
| --- | --- |
| Exact version + full build commit bound at enrollment, repeated at connect | `remote_control.rs:218` validates, `:803` re-checks, `:829` sends, `:901` verifies |
| Capability negotiation (no loosening CWC's version check) | `CAPABILITIES` at `:33`, sent at `:829` and `:1010` |
| One durable input owner | `blocks_local_input()`; local submit refused with a visible notice, never queued |
| Fail closed on unprovable state | Cancelling a connect can't prove no lease exists → ownership stays locked through a conservative expiry (`:241-266`) rather than returning input immediately |
| Command idempotency | `command_fingerprints: HashMap<(String, u64), String>` at `:171` — in-process duplicate suppressed without a second effect; sequence collision with different content rejected |
| Monotonic sequence | `event_seq` at `:168`, `command_seq` at `:105` |
| `/rc stop` returns ownership without killing the conversation | `stop()` at `:241`; emits `OwnershipRestored { approvals }` so pending approvals are handed back for local decision rather than dropped |

## Rebase note

This lane was sitting on `71c6851730c9` — one commit *behind* main, and therefore **without** the #4831 hermetic-state fix. A full `cargo test --workspace --all-features` was actively running from it against real `~/.codewhale` state. I stopped that run, verified `config.toml`/`settings.toml` hashes were byte-identical before and after, archived the lane to `backups/codewhale-rc-program-20260725-preserve-1700/`, then rebased onto current `main`.

The lane's test fixture was written against an older `PendingRemoteApproval` shape (`original_request_id` / nested `request`) that no longer exists; it is updated to the real struct while keeping the property it was testing.

## Gates

- `cargo clippy --workspace --all-features --locked -- -D warnings` + the five CI `-A` lints (matching `.github/workflows/ci.yml:229-235`) — **clean**
- `git diff --check` — clean
- `cargo test -p codewhale-tui --bin codewhale-tui --locked remote_control` — **12 passed, 0 failed**

## What is NOT proven — please read before merging

- **No two-surface acceptance run.** The decisive proof for `/rc` is a live terminal session and a separately authenticated browser converging on prompt, streaming, approval, interrupt, reconnect, and revoke. That has not been run. Source tests are not acceptance.
- **No redaction layer in this module.** `rg redact|sanitiz` against `remote_control.rs` returns nothing. The bounded/redacted-events requirement (no credentials, absolute paths, or raw sensitive tool input on the wire) is **not** demonstrably enforced here. This is the gap I would review hardest.
- `CAPABILITIES` is `["evidence-ledger", "fim", "git", "shell"]` and does **not** currently advertise `remote_control.v1`. Worth deciding whether the protocol capability belongs in this list or is negotiated separately — the CWC contract doc lists the same four.
- No live CWC control plane was exercised; the server side is fixture-level only.


Hmbown:
## Correcting my own "no redaction" note above

I flagged the missing redactor as the thing to review hardest. Having now read the relay payloads rather than grepping for `redact`, that framing was wrong and I want it corrected in the record before anyone reviews on the strength of it.

There is no redaction *function*, because the sensitive fields are **never populated in the first place** — redaction by omission, which is the stronger construction:

| Event | Payload | Sensitive content |
| --- | --- | --- |
| `ToolCallStarted` | `{"tool":{"id","name","input":{}}}` | tool **input is empty** |
| `ToolCallComplete` | `{"item":{…,"summary":"","detail":""}}` | tool **output is empty** |
| `TurnStarted` | model id, empty mode | none |
| `TurnComplete` | status + usage counters | none |

So raw tool input, tool output, absolute paths, and shell strings never reach the wire — not because a filter strips them, but because the envelope has nowhere to put them.

The snapshot path is bounded and filtered too (`project_session_messages`, `remote_control.rs:625`):
- only `user` and `assistant` roles survive; tool messages are dropped entirely
- only `ContentBlock::Text` survives; `tool_use`/`tool_result` blocks are dropped
- 16 KiB cap per message, last 64 messages only
- approval ids are SHA-256 hashed through `projected_approval_id` rather than passed through, so local tool-call ids are not exposed

**What genuinely remains exposed, and should be reviewed:** `MessageDelta` relays assistant text verbatim, and the snapshot relays user and assistant text. That is the conversation itself, which is the point of remote control — but it means a user who pasted a secret into a prompt has that text relayed. Worth an explicit decision rather than an accident.

**Still open, unchanged from the description:** `CAPABILITIES` does not advertise `remote_control.v1`, and no two-surface acceptance run has been performed.

Issue #4851, opened 2026-07-26
Two model-resolution chains: merge them behind one owner


Hmbown:
Codewhale resolves "which model is active" twice, in two crates, with two
different implementations:

- `Config::default_model()` — `crates/tui/src/config.rs`. What `client.rs:957`
  puts on the wire; what `doctor` reports. Backed by the equal-treatment
  resolver `canonical_model_id_for_provider` and the model registry.
- `Config::resolve_runtime_options()` — `crates/config/src/lib.rs`. What
  `codewhale model resolve` reports; what the app-server and
  `config/src/route/descriptor.rs` consume. Backed by its own
  `normalize_model_for_provider` and per-provider canonicalizers.

They have already silently diverged twice:

- #4829 — `validate()` and the runtime disagreed about whether root
  `default_text_model` is provider-neutral, and a Z.ai setup produced a CLI
  that could not start.
- #4832 / #4838 — the dispatcher dropped root `default_text_model` for every
  non-DeepSeek provider while the TUI honoured it, so `model resolve` reported
  a provider default for a model that was in fact being sent. #4837 tried to
  fix this by pointing `model resolve` at the dispatcher chain, believing it
  was the doctor path.

PR #4850 makes the two agree on nine provider/model shapes and adds a test that
pins them together, but it does **not** merge them: it duplicates the
DeepSeek-family predicate and the provider lists into `crates/config` rather
than importing them, because the equal-treatment resolver lives in the TUI
crate and `crates/tui` depends on `crates/config`, not the reverse.

So the duplication that caused the bug is still there, now with a tripwire on
it. The durable fix is one owner.

## What "one owner" probably means

`crates/config` is the lower crate, so the resolver belongs there:

1. Move `canonical_model_id_for_provider`, `normalize_model_name`,
   `provider_passes_model_through`, and the model registry entries they read
   into `crates/config` (or a new leaf crate both depend on).
2. Reduce `Config::default_model()` in the TUI to a call into it, keeping the
   TUI's `ApiProvider` → `ProviderKind` mapping at the boundary.
3. Delete `deepseek_family_model_id` / `provider_passes_model_through` /
   `root_default_model_is_foreign_to_provider` added by #4850 in favour of the
   shared ones.
4. Keep the agreement test from #4850 — it should become trivially true, and it
   is the thing that proves the merge did not change behaviour.

Worth checking during the move: `ApiProvider` and `ProviderKind` are two enums
[...]

PR #4852, opened 2026-07-26
fix(config): align root model fallback with TUI

### LINKED ISSUE #4838: `codew model set` is a silent no-op for every non-DeepSeek provider (root `default_text_model` is DeepSeek-scoped)
Split out of #4832, which exposed this. #4832 is the diagnostic lying; this is the route actually dropping the user's choice.

## Repro

Verified on `239c04f81` with a sealed `HOME`, using the fixed `model resolve` from #4837 so the provenance is visible:

```
$ cat $HOME/.codewhale/config.toml
provider = "zai"

[providers.zai]
api_key = "k"

$ codewhale model set GLM-4.6
Default model set to 'GLM-4.6'

$ cat $HOME/.codewhale/config.toml
default_text_model = "GLM-4.6"
provider = "zai"
...

$ codewhale model resolve
resolved: GLM-5.2
model_source: provider default
```

`model set` reports success, writes the key, and the runtime never reads it. The `GLM-5.2` that comes back is `DEFAULT_ZAI_MODEL`, not the user's choice.

## Cause

`crates/config/src/lib.rs`:

```rust
let root_deepseek_model = (provider == ProviderKind::Deepseek)
    .then(|| self.default_text_model.clone())
    .flatten();
```

Root `default_text_model` is documented as "TUI-compatible default **DeepSeek** model" and only enters the precedence chain when the provider is DeepSeek. The provider-neutral root key is `model`. But:

- `ModelCommand::Set` writes `store.config.default_text_model` for **any** model id, with no provider check (`crates/cli/src/lib.rs`).
- The setup wizard writes `default_text_model` alongside `provider = "zai"` (see `setup_state.json`; this is how the #4829 config was produced).
- #4830 made `Config::validate()` accept every provider's own advertised models in `default_text_model`, so validation now treats the key as provider-neutral while the runtime still treats it as DeepSeek-only. Those two disagree.

Today the damage is masked whenever the user's choice happens to equal the provider default. It is not masked for `GLM-4.6` under Z.ai, or for any user who picks a non-default model on any non-DeepSeek provider.

## Why it is not a one-line change

Making `default_text_model` provider-neutral by deleting the `provider == Deepseek` guard would regress a real config: `provider = "zai"` with a stale legacy `default_text_model = "deepseek-chat"` currently falls through safely, and would afterwards send a DeepSeek id to Z.ai. Any fix needs to decide what happens to a root default that is not valid for the configured provider — most likely route it through the same equal-treatment resolver #4830 used (`canonical_model_id_for_provider`) and ignore-with-a-warning when it does not canonicalize, rather than silently ignoring or blindly forwarding.

## Expected

- `codew model set X` either takes effect for the configured provider or fails loudly saying it cannot.
- `validate()` and the runtime agree about which keys are provider-neutral.
- No provider is privileged by the root config schema (`CLAUDE.md`).
- Existing DeepSeek configs keep byte-identical behavior.

nightt5879:
## Summary

- resolve the legacy root default_text_model against the effective provider route in RuntimeConfig, matching the TUI request path
- preserve provider-neutral root model precedence on non-DeepSeek routes and keep released DeepSeek behavior unchanged
- carry over the TUI's route-specific normalization and foreign-DeepSeek guard so stale IDs do not leak onto vendor-locked endpoints

## Root cause

The TUI request path already honored root default_text_model for the active provider, but ConfigToml::resolve_runtime_options_with_secrets only admitted that value when the provider was DeepSeek. As the issue owner's wire-level follow-up established, the user-selected model was sent correctly by the TUI; RuntimeConfig consumers such as model resolve, app-server, and route descriptors were the divergent chain.

This change deliberately leaves model set persistence and the TUI request path untouched. It filters and normalizes the root fallback using the effective provider/base URL/auth route before assigning both the model and its ModelSource.

## Regression coverage

- Z.ai + root GLM-4.6 resolves to GLM-4.6 with RootDefaultTextModel provenance
- stale root deepseek-chat on Z.ai falls back to the Z.ai default
- CLI-effective provider overrides use the same compatibility decision
- root model retains precedence over the legacy fallback on non-DeepSeek routes
- Codex, Kimi Code, Xiaomi, OpenCode, DeepSeek Anthropic, custom endpoints, hosting providers, and pass-through providers retain their TUI route semantics
- malformed direct-route root values fall back safely

## Testing

- cargo fmt --all -- --check
- cargo test -p codewhale-config --locked (432 tests + doctest)
- cargo clippy -p codewhale-config --all-targets --locked -- -D warnings
- cargo test -p codewhale-cli --test model_resolve_provenance --locked (5 tests)

Fixes #4838

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -525,7 +525,7 @@ pub struct ConfigToml {
     /// Optional extra HTTP headers forwarded to model API requests.
     #[serde(default, skip_serializing_if = "http_headers_are_effectively_empty")]
     pub http_headers: BTreeMap<String, String>,
-    /// TUI-compatible default DeepSeek model.
+    /// Legacy TUI-compatible default model, resolved against the active route.
     pub default_text_model: Option<String>,
     #[serde(default, deserialize_with = "deserialize_root_provider")]
     pub provider: ProviderKind,
@@ -2394,9 +2394,6 @@ impl ConfigToml {
         let root_deepseek_base_url = (provider == ProviderKind::Deepseek)
             .then(|| self.base_url.clone())
             .flatten();
-        let root_deepseek_model = (provider == ProviderKind::Deepseek)
-            .then(|| self.default_text_model.clone())
-            .flatten();
         let auth_mode = cli
             .auth_mode
             .clone()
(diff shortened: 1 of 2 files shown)


nightt5879:
@codex review

nightt5879:
Closing as superseded by #4850, now landed on main as 315edcbc8. I verified that the landed change covers the corrected scope from #4838: RuntimeConfig honors root default_text_model for non-DeepSeek routes, rejects stale DeepSeek IDs on vendor-locked routes, and adds a cross-resolver agreement test covering the real Z.ai repro and route boundaries. The broader original premise was disproven, and the remaining structural risk is tracked in #4851. Thanks for the wire-level correction and evidence.

PR #4882, opened 2026-07-26
feat(tui): persist provider-scoped model pins

### LINKED ISSUE #4866: v0.9.2 Models: provider-scoped pins, labels, and truthful family siblings
## Dogfood problem

The model/provider surface is hard to navigate once several providers and model families are configured. Users cannot keep a small personal working set at the top, and exact provider/model identity competes with friendly naming. The current release work also needs strong/fast sibling routing to be truthful and provider-scoped (#4405, #4411, #4858).

## v0.9.2 product decision

Add **provider-scoped model pins** plus an optional **local display label**:

- A pin identifies the exact `(provider_id, model_id)` tuple. It never floats to another provider with a coincidentally equal model id.
- Pinned models sort first in `/models` and the model picker, preserving user-defined order. Within each section retain existing recency/catalog order.
- A label is presentation only; every picker row, receipt, diagnostics surface, and config review still exposes the exact provider and model id.
- Missing/stale pinned entries remain visible in a clearly unavailable section with repair/remove affordances; never silently reroute.
- Pin/label changes are local settings and must not touch provider credentials.
- Keyboard-first toggles and edit controls must work at narrow widths; add locale parity.

This is distinct from model families. A family explicitly declares strong/fast siblings under the same provider. A Sol route may delegate to Luna/Terra only if the configured provider catalog declares those exact siblings and the user/provider consent policy permits it. Never derive or cross-send an unrecognized model id.

## Acceptance

- [ ] `/models` and Settings can pin/unpin and reorder an exact provider/model tuple.
- [ ] Pins persist and always sort above unpinned entries after restart.
- [ ] Optional labels do not hide exact provider/model identity in selection or receipts.
- [ ] Duplicate model ids from two providers can be pinned/labeled independently.
- [ ] Stale pins fail closed and remain repairable.
- [ ] Model-family siblings are catalog/provider facts with closed-set tests; Sol→Luna/Terra only appears when real for that provider.
- [ ] Tests cover persistence, ordering, stale entries, duplicate ids, keyboard/narrow view, locale parity, and route receipts.

Related: #4751 (Settings IA), #4794 (capability catalog), #4406 (configured vs live health), #4405/#4858 (observable same-provider fast siblings).

Hmbown:
## Summary
- persist exact `(provider, model)` pins in TUI settings with optional presentation-only labels
- render pinned models first in the model picker while retaining stale unavailable pins for repair/removal
- add keyboard pin/unpin and Alt-arrow ordering controls without changing provider route identity
- keep provider-scoped identity explicit for named custom providers

## Verification
- `cargo test -p codewhale-tui --bin codewhale-tui settings --locked` (111 passed)
- `cargo test -p codewhale-tui --bin codewhale-tui model_picker --locked` (95 passed)
- `RUSTFLAGS="-D warnings" cargo check -p codewhale-tui --locked` (passed; known macOS linker diagnostic only)
- `git diff --check`

No-Issue: provider-scoped pin persistence is the P2 slice of the already-closed #4866 blocker; P3 thinking controls and P4 provider-scoped family siblings remain separate follow-ups.

### DIFF
--- a/crates/tui/src/settings.rs
+++ b/crates/tui/src/settings.rs
@@ -270,6 +270,17 @@ fn resolve_tui_prefs_path_from_candidates(
 }
 
 /// User settings with defaults
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct PinnedModel {
+    /// Exact configured provider identity; labels never replace this value.
+    pub provider: String,
+    /// Exact provider-owned model id.
+    pub model: String,
+    /// Optional presentation-only label.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub label: Option<String>,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(default)]
 pub struct Settings {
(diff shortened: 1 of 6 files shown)


Hmbown:
Reconciliation shows this exact pin implementation already landed on origin/main via #4873; closing the duplicate. P3 thinking controls and P4 provider-scoped siblings remain separate work.

Issue #3874, opened 2026-07-01
v0.8.69: Notify the agent when background shell tasks finish and keep waiting sessions alive


Hmbown:
## Why this matters

Split out of #3859, whose wording half shipped on the v0.8.67 branch (`526ba16c8`: Ctrl+B copy now says the shell wait moves to `/jobs`). The behavioral half remains, and it hits anyone running long builds/test suites through the agent: a background task can finish and the agent never learns about it, and a session stuck waiting can die to an idle timeout instead of recovering. Hunter's analysis on #3859 describes the compound footgun: long command → Ctrl+B → agent still waiting → idle timeout fires → session dies with lost context.

## Current behavior

- `exec_shell` with `background: true` returns immediately with polling hints (`crates/tui/src/tools/shell.rs:2310`, `poll_with: ["task_shell_wait", "exec_shell_wait"]` at `shell.rs:2774`), and `/jobs` tracks the task. Polling works **within** a turn.
- Nothing re-invokes or notifies the model when a background task completes after its turn ends — completion is only discovered if the model happens to poll again.
- The per-SSE-chunk stream idle timeout (`crates/tui/src/core/engine.rs:385-387`) governs streamed responses; the session-level "300s idle kills a waiting session" behavior described in #3859 has no carve-out for sessions that are only idle because a tracked background task is still running.

## Desired behavior

1. When a tracked background shell task completes while the agent is idle, the completion becomes model-visible (queued notification/next-turn trigger), so the agent can act on results without user prodding.
2. A session whose only pending work is a tracked background task does not die to the idle timeout — the running task resets or pauses the idle clock.
3. `/jobs` and the exec_shell background hints document the notification behavior so the model's expectations match reality.

## Acceptance criteria

- [ ] A test proving a background task that completes after the turn ends produces a model-visible completion record the next turn can read without calling `exec_shell_wait` first.
- [ ] A test (or documented config semantics) proving the idle timeout does not kill a session while a tracked background task is running.
- [ ] Tool descriptions/hints updated to describe the completion-notification contract; localized copy stays valid (`jq empty crates/tui/locales/*.json`).
- [ ] `RUSTFLAGS="-D warnings" cargo test -p codewhale-tui --bin codewhale-tui --locked --no-run` passes.

## Related

#3859 (wording half shipped in v0.8.67 via PR #3861 `526ba16c8`), #3032 (original Ctrl+B backgrounding work), PR #3861.

Hmbown:
[...]

PR #4894, opened 2026-07-26
feat(shell): deliver tracked completions to waiting turns

Hmbown:
Refs #3874
No-Issue: reviewable completion slice for #3874; keep the issue open until hosted checks and the landed-main audit confirm the full acceptance matrix.

## Summary
- deliver completed tracked background shell jobs as internal, untrusted runtime events at the next turn boundary
- preserve exactly-once delivery through the existing ShellManager completion drain
- keep owner-linked workers alive while their tracked shell work is still running
- update tool descriptions and metadata so automatic completion delivery is truthful

## Verification
- `git diff --check`
- `cargo check -p codewhale-tui --locked`
- `RUSTFLAGS="-D warnings" cargo check -p codewhale-tui --locked`
- focused engine handoff, shell owner, drain-once, and metadata tests

No live providers, credentials, production state, or deployment actions used.

### DIFF
--- a/crates/tui/src/core/engine.rs
+++ b/crates/tui/src/core/engine.rs
@@ -1976,6 +1976,7 @@ impl Engine {
                         // during a sub-agent fanout no longer contends for the
                         // write lock (against completions/persistence) on every
                         // request. Cleanup still auto-cancels stale agents.
+                        self.touch_workers_with_running_shells().await;
                         let due = {
                             let manager = self.subagent_manager.read().await;
                             manager.cleanup_due(
@@ -3116,6 +3117,22 @@ impl Engine {
             return outcome;
         }
 
+        // Deliver completions that arrived after the previous turn before the
+        // next user request is sent. This keeps background shell work
+        // model-visible without requiring an explicit wait/poll tool call.
+        let shell_completions = self.drain_shell_completion_events();
+        if !shell_completions.is_empty() {
+            self.add_session_message(crate::runtime_handoff::shell_completion_runtime_message(
+                &shell_completions,
+            ))
+            .await;
+            if let Some(status) =
+                crate::core::engine::turn_loop::shell_completion_status_text(&shell_completions, "")
+            {
+                let _ = self.tx_event.send(Event::status(status)).await;
+            }
+        }
+
         let input_policy = effective_input_policy(
             provenance,
             mode,
@@ -3594,6 +3611,7 @@ impl Engine {
     /// Capture typed live state for post-compact rehydrate (todos, workers,
     /// shells, mode, permission). Pure formatting lives in compaction.rs.
     async fn capture_compaction_live_state(&self) -> CompactionLiveState {
+        self.touch_workers_with_running_shells().await;
         let todos = {
             let guard = self.config.todos.lock().await;
             let snap = guard.snapshot();
--- a/crates/tui/src/core/engine/turn_loop.rs
+++ b/crates/tui/src/core/engine/turn_loop.rs
@@ -1428,8 +1428,6 @@ impl Engine {
                     if let Some(status) = shell_completion_status_text(&shell_completions, "") {
                         let _ = self.tx_event.send(Event::status(status)).await;
                     }
-                    turn.next_step();
-                    continue;
                 }
 
                 // Sub-agent completion handoff (issue #756). The model finished
@@ -1597,8 +1595,6 @@ impl Engine {
                     {
                         let _ = self.tx_event.send(Event::status(status)).await;
                     }
-                    turn.next_step();
-                    continue;
                 }
 
                 if self.drain_subagent_completion_events("late").await > 0 {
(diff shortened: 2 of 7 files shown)


Hmbown:
The `Test (macos-latest)` and `Test (windows-latest)` failures are a real regression from this branch, not runner flake. Reproduced locally on macOS at `08ef864c4`:

```
cargo test -p codewhale-tui --bin codewhale-tui --locked -- --exact \
  core::engine::tests::yolo_mode_does_not_prompt_for_background_shell \
  core::engine::tests::operate_model_shell_uses_normal_approval_and_workspace_sandbox \
  core::engine::tests::full_access_subagent_handoff_keeps_model_shell_free_of_approval_prompts

test result: FAILED. 0 passed; 3 failed; finished in 1.67s
```

It fails in 1.67s, so it is not a timeout. wiremock's verification dump shows the cause: **three** `POST /v1/chat/completions` requests arrived, but each of these tests mounts exactly two mocks, both pinned `.expect(1)`.

The third request comes from the new re-entry in `turn_loop.rs`:

```rust
if !shell_completions.is_empty() {
    self.add_session_message(shell_completion_runtime_message(&shell_completions)).await;
    ...
    turn.next_step();
    continue;
}
```

`drain_shell_completion_events()` drains any tracked shell job that has finished, so when a job completes *before the turn ends* the turn now takes an extra model round-trip. All three tests run a fast command (`echo …`), which finishes within the same turn on macOS and Windows; on ubuntu the drain lands after the turn ends, so only two requests happen and the job stays green. That timing split is why CI looks platform-specific.

Note that none of these three tests is new and this branch does not touch `crates/tui/src/core/engine/tests.rs` — they are pre-existing guards over `turn_loop.rs` / `shell.rs`, which this branch does change. `full_access_subagent_handoff_keeps_model_shell_free_of_approval_prompts` issues a **foreground** `exec_shell` (`echo yolo-model-ask-rule`, no `background` argument) and still gets the extra round-trip, so the re-entry is not limited to jobs the model explicitly backgrounded.

Beyond the tests, the behaviour worth a second look: the number of model round-trips in a turn now depends on whether a shell job happens to finish before the turn ends. That is nondeterministic token spend and latency, and a fast job can bill an extra model call for a result the model already received in the tool output.

Worth noting too that this branch renames `shell_completion_status_does_not_create_runtime_handoff` to `shell_completion_status_is_concise_and_shell_handoff_is_untrusted` — deliberately reversing the invariant those three tests were written against. #3874 does ask for the notification, so the intent is right; it is the unconditional `continue` that is doing the damage.

Suggested shape: still inject `shell_completion_runtime_message` so the model sees the completion, but only `next_step(); continue;` when the turn is actually waiting on that work — otherwise let the message ride into the next turn, which is what the pre-existing design did on purpose. That keeps #3874's acceptance ("notify the agent, keep waiting sessions alive") without making every completed shell job extend the current turn.

Receipts gathered on `08ef864c4` in an isolated worktree; no changes pushed to this branch.

Hmbown:
Rebased onto current origin/main (65e1a6068) and pushed the re-entry fix at 07b1f051e (`fix(engine): avoid redundant shell completion re-entry`).

Local receipts on the rebased head (macOS):
- `core::engine::tests::yolo_mode_does_not_prompt_for_background_shell`, `…::operate_model_shell_uses_normal_approval_and_workspace_sandbox`, `…::full_access_subagent_handoff_keeps_model_shell_free_of_approval_prompts` — 3 passed; 0 failed.
- `core::engine::turn_loop::tests::shell_completion_status_is_concise_and_shell_handoff_is_untrusted` — 1 passed.
- `cargo fmt --all -- --check` and `git diff --check` clean.

The extra model round-trip is gone: the runtime message is still injected, but the turn only re-enters when it is genuinely waiting on that work.

Hmbown:
Heads-up on the red `Lint` check here: it is not caused by this branch. `rust-toolchain.toml` pins `channel = "stable"`, CI has moved to Rust 1.97.0, and five pre-existing patterns on clean `origin/main` (`a2cb83514`) are now hard errors under `cargo clippy --workspace --all-features --locked -- -D warnings` — three `collapsible_if` and two `too_many_arguments`, in `fleet/host.rs`, `tui/notifications.rs`, `tools/shell.rs`, and `tui/ambient_life.rs`. I reproduced it on main with no branch applied.

#4898 fixes those five. Once it lands, rerunning Lint here should go green without further changes to this branch.

Hmbown:
Correcting my earlier comment: the red `Lint` here was **not** clippy.

CI's clippy invocation allows the lints I was reproducing locally:

```
cargo clippy --workspace --all-features --locked -- -D warnings \
  -A clippy::uninlined_format_args -A clippy::too_many_arguments \
  -A clippy::unnecessary_map_or -A clippy::collapsible_if \
```

The actual failing step was **Check harvested contributor credit**:

```
- 8ce7070b6c fix(engine): avoid redundant shell completion re-entry:
  author Hunter B <hmbown@gmail.com> matches AUTHOR_MAP but is not canonical.
- ... `Harvested from PR ... by @Hmbown` needs machine-readable credit.
```

The top commit's body carried a `Harvested from PR #4894 by @Hmbown` line (with literal `\n` escapes rather than newlines) while its author was the non-canonical `Hunter B <hmbown@gmail.com>`. Since this is not a harvest from another contributor, I rewrote the message without that line, fixed the escaped newlines, and set the canonical author identity from `.github/AUTHOR_MAP`.

Now at `b0b6…`; `check-coauthor-trailers.py --check-authors` passes for all 3 commits, the three engine regressions are 3/0, and `cargo fmt --all -- --check` is clean. #4898 was still a valid cleanup, but it was not this blocker.

Issue #3413, opened 2026-06-22
v0.8.69: Make codewhale.net feel like a mature product site


Hmbown:
## Why this matters

The website should stop reading like a project bulletin plus compact docs page. CodeWhale needs a mature product surface: polished, trustworthy, well-structured, and clear enough that new users, serious adopters, contributors, and partners understand what the project is without digging through repository docs.

The goal is presentation quality and information architecture: product hierarchy, trust surfaces, documentation depth, community proof, and a homepage that feels like it belongs to a serious developer tool.

## Current behavior

- The primary nav is a short set of pages: Install, Docs, Activity, Roadmap, FAQ, Contribute (`web/components/nav.tsx:9`).
- The homepage mixes product intro, live repo activity, release credits, contributor philosophy, and install guidance in one long page (`web/app/[locale]/page.tsx:464`).
- `/docs` is one long route with local anchors instead of a full support/docs architecture.
- Important product and trust material already exists in `README.md`, `docs/`, release notes, and install/mirror docs, but the website does not present it like a serious product front door.

## Desired behavior

1. Redesign `codewhale.net` around a product-led IA: product, install, docs, trust/security, community, roadmap, contact/support, and provenance.
2. Make the first viewport feel like a mature developer-tool product, with clear CTAs such as Install, Docs, and GitHub.
3. Promote trust signals: open-source license, security model, local-first behavior, official site/source provenance, release/install verification, provider neutrality, and support expectations.
4. Keep multilingual community visible as proof and belonging, but move it into a deliberate community surface instead of letting raw activity or release credits dominate the product story.
5. Turn `/docs` into a structured product-support experience rather than a single summary page.

## Acceptance criteria

- [ ] A proposed or implemented site map shows the target product, docs, trust, community, roadmap, contact/support, and provenance surfaces.
- [ ] Homepage design and copy feel polished, current, and product-grade without overclaiming.
- [ ] Navigation/footer include appropriate trust, contact/support, docs, provenance, GitHub, and community links.
- [ ] `/docs` has structured topic navigation for install, usage, config, providers/models, tools, sub-agents, MCP, skills, hooks, safety, troubleshooting, and contribution paths.
- [ ] Community and multilingual content remain visible but no longer crowd out the product hierarchy.
- [ ] Existing routes continue to resolve or redirect.
- [ ] The site can be verified with `cd web && npm run lint && npm run test && npm run build`.

## Related

- #3419 should feed the runtime/integrations product surface.
[...]

PR #4912, opened 2026-07-27
feat(web): v0.9.2 docs guide/vocabulary, getting-started path, pending media manifest (#3413)

Hmbown:
## Summary

Harvests the web-maturity lane onto the v0.9.2 candidate (targets the integration branch, not main):

- `/docs/guide` and `/docs/vocabulary` routes; docs-map + sitemap registration
- Homepage getting-started path, nav/footer links, skip link + a11y landmarks
- Real-session media manifest landing in explicit `pending` state — no placeholder footage; tests assert no asset file may exist for a pending id
- Vocabulary page verified against `docs/public-surface-facts.json` (Fleet/Workflow/Lane/Runtime verbatim; Consultant public role; "Operation" appears nowhere)

Receipts: web suite 179/179 (24 files), targeted docs-ia 14 / media 10 / vocabulary 11 / navigation 6, eslint clean, tsc clean, check:docs PASS.

Advances #3413; does not close it (media recording and remaining IA work stay open).

No-Issue: partial slice of #3413; closure requires the recording milestone and final-SHA receipts.

### DIFF
--- a/docs/releases/v0.9.2-media-plan.md
+++ b/docs/releases/v0.9.2-media-plan.md
@@ -1,0 +1,102 @@
+# v0.9.2 real-session media plan (#4906)
+
+> **Status: pending the dogfood recording.** The site renders a visible
+> "recording pending release candidate" state (`web/lib/media-manifest.ts` →
+> `first-fleet-session`, rendered by `web/components/session-media.tsx` on
+> `/docs/guide`). No placeholder, staged, or recycled footage ships. This file
+> is the exact procedure for producing the real assets once the v0.9.2 release
+> candidate is stable and dogfooded, and the checklist for flipping the
+> manifest entry to `published`. It authorizes local recording only — never a
+> deploy, tag, or publication.
+
+## What gets recorded
+
+One continuous real session of the v0.9.2 release candidate, on a **sealed
+local route** (loopback Ollama, exactly like the canonical screenshot in
+`docs/public-surface-facts.json` → `screenshot.capture`; no paid or hosted
+provider, no live account):
+
+1. Install / launch: `codewhale` in a scratch repository.
+2. First keyless session: constitution-first setup, then a read-only Plan-mode
+   exploration.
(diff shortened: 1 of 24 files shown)


Hmbown:
Rebased onto current `main` (`37a2efd02`) and force-pushed as `ed29dfe31`. This was CONFLICTING; it now applies cleanly and CI is running.

One real conflict, in `web/components/footer.tsx`. Recording the resolution because it was a judgment call, not a mechanical merge:

`main` had since split the footer's product links three ways — inline `zh`, inline `en` (the copy-contract tests read those two from this file), and a dictionary-driven branch for every other routed locale via `getChrome(locale)`. This branch predated that split and replaced the whole `else` with a single hard-coded English list, which would have silently dropped localized footers for `ja`, `ru`, `vi`, and the rest.

Resolution: keep `main`'s three-way structure, and add this branch's two new links (`Getting started` → `/en/docs/guide`, `FAQ` → `/en/faq`) to the **inline `en`** branch only. The dictionary-driven branch keeps `main`'s link set, with a comment explaining why:

```tsx
      : [
          // No `footerGuide`/`footerFaq` keys exist in the chrome dictionaries,
          // so the dictionary-driven locales keep main's link set rather than
          // linking English-only labels. Adding them is a locale-parity change,
          // not a rebase resolution.
```

That is deliberate scope discipline: adding those two links for the other locales means adding `footerGuide`/`footerFaq` to `types.ts` and to every locale dictionary. That is a locale-parity change with its own review surface, and smuggling it in under a conflict resolution is how parity gaps get introduced without anyone deciding to. Worth a follow-up — the `zh` branch already carries both links inline, so `ja`/`ru`/`vi` are currently the odd ones out.

The other five commits rebased without conflict.

Note on the body: it says *"targets the integration branch, not main"*, but the PR's base is `main`. Worth correcting so the description matches what merging actually does.

I have not merged this — I want the rebased web suite green first, and the footer resolution is exactly the kind of thing the copy-contract tests exist to catch.

PR #4928, opened 2026-07-27
feat(tui): add thinking_default_expanded setting

### LINKED ISSUE #4925: Add thinking_default_expanded setting for always-expanded reasoning blocks
## Problem

Thinking blocks are collapsed by default in the TUI transcript. Users must press
**Space** to expand them. Over SSH/tmux, Space is often captured by the terminal
layer (see companion bug report about the Space key not working over SSH).

Even when Space works, many users prefer to always see the full reasoning block
without an extra keypress every turn. `show_thinking = true` requires manual
expansion which is especially tricky over SSH/tmux.

- `show_thinking = true`: thinking blocks are **visible but collapsed** to a summary
- Space toggles collapse/expand per-cell (stored in `app.folded_thinking` HashSet)
- No setting to change the default collapsed state

## Proposed solution

Add `thinking_default_expanded: bool` to the `Settings` struct. Thread it through
`TranscriptRenderOptions` → the collapse calculation in `history.rs`.

```toml
# ~/.codewhale/settings.toml
show_thinking = true
thinking_default_expanded = true   # new setting
```
- `false` (default): unchanged — thinking collapsed, Space expands
- `true`: thinking **expanded by default**, Space collapses (inverted toggle)

### Key change (triple XOR)
```rust
// crates/tui/src/tui/history.rs
// Before:
folded ^ !options.verbose
// After:
folded ^ !options.verbose ^ options.thinking_default_expanded
```

The triple XOR inverts the default while preserving toggle behavior.

### Files touched (4 files, +16/−1)
| File | Change |
|------|--------|
| `crates/tui/src/settings.rs` | New field + default + parser |
| `crates/tui/src/tui/history.rs` | Field on `TranscriptRenderOptions` + triple XOR |
| `crates/tui/src/tui/app.rs` | Field on `App` + passthrough |
| `crates/tui/src/tui/app/init.rs` | Read setting, pass to App init |

A PR is attached to this issue. Full workspace compiles clean.

## Use case

Every SSH/tmux session with `show_thinking = true`. Instead of fighting with
Space (see companion bug), the user sets `thinking_default_expanded = true`
once and reasoning is always readable. Also useful for users who simply prefer
to see reasoning without an extra keypress.

## Alternatives considered

- Configurable keybinding for Space — partial fix, doesn't solve cell focus or
  composer-stealing issues (see companion bug for details)
- `/verbose` — changes global transcript behavior, not what users want

## Impact

Used every turn by anyone who enables it. Zero overhead when disabled. The
setting is a single `#[serde(default)]` bool — no runtime cost.

## Additional context

- Tested live on codewhale 0.9.1 over SSH/tmux
- `cargo check --workspace` passes
- `cargo fmt --all -- --check` passes

M-Maciej:
When true, thinking blocks render expanded by default instead of collapsed. Space still toggles each block through the existing `folded_thinking` state.

Useful for SSH/tmux users where Space may be captured by the terminal layer, or for anyone who prefers reasoning to start expanded.

## Summary

- Adds `thinking_default_expanded: bool` with an opt-in default.
- Threads it through transcript rendering while preserving the Space toggle.
- Exposes it through `/config`, the typed configuration UI, and the runtime API.
- Documents it in the shipped example, configuration reference, and accessibility table.
- Adds persistence and transcript behavior regression tests.

Closes #4925

## Testing

- [x] Exact hosted-CI clippy command from `CONTRIBUTING.md`
- [x] Focused settings, transcript, typed-config, and runtime API tests
- [ ] Full hosted matrix (awaiting workflow approval)

## Checklist

- [x] Verified TUI behavior manually if UI changes (contributor tested over SSH/tmux)
- [x] Maintainer follow-up commits preserve the contributor commit and authorship


Hmbown (approved):
Maintainer approval after completing the setting across persistence, `/config`, typed config UI, runtime API, docs, and focused regression tests. The contributor commit remains intact; follow-up commits are scoped and locally verified against current main.

Hmbown:
Windows CI found a genuine completeness gap after the first follow-up: `thinking_default_expanded` was persisted and available through typed/runtime config, but absent from the native `/config` view. Fixed in `73a2e356b` with the row, boolean editor behavior, help text, and explicit coverage. Local receipts: both native config tests pass; `cargo fmt --all -- --check`, `git diff --check`, and the exact hosted clippy command pass.

Hmbown:
The refreshed Windows suite confirmed the native row itself now passes, then found one stale filter expectation: searching `thinking` still expected only the two old rows. Updated in `d58c9adf9`. Local receipts now include the entire native config-view family: 40/40 tests pass, plus exact hosted clippy, fmt, and diff checks.

PR #4931, opened 2026-07-27
Migrate QA PTY test harness from vt100 to rio-vt

raphamorim:
Swaps the QA PTY test harness (`crates/tui/tests`) from vt100 to [rio-vt](https://crates.io/crates/rio-vt), Rio's terminal engine.

The `Frame` helper parsed the TUI's real PTY output with vt100 and read cell text and colors to assert on visible rendering. It now uses rio-vt through the same small surface. Cell colors are surfaced through a neutral `qa_harness::Color` enum (Default/Idx/Rgb) that mirrors exactly what the tests already matched on, so the assertions are unchanged; only the dev-dependency and the harness internals move.

The frame unit tests (blank interior columns, wide-glyph continuation cells) pass, and the PTY integration tests compile and run against the rio-vt harness.

Benchmark against vt100 and alacritty_terminal: https://github.com/raphamorim/rio-vt-benchmark

Background on the engine and why it was split out of Rio: https://rioterm.com/blog/2026/07/27/rio-vt-and-librio

No-Issue: migrate the QA PTY harness from vt100 to rio-vt

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.4",
  "cipher",
  "cpufeatures 0.2.17",
 ]
@@ -44,7 +44,7 @@ version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.4",
  "getrandom 0.3.4",
  "once_cell",
  "serde",
@@ -212,12 +212,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
 
 [[package]]
-name = "arrayvec"
-version = "0.7.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
-
-[[package]]
 name = "ascii"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -285,7 +279,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
 dependencies = [
  "autocfg",
- "cfg-if",
+ "cfg-if 1.0.4",
  "concurrent-queue",
  "futures-io",
  "futures-lite",
@@ -319,7 +313,7 @@ dependencies = [
  "async-signal",
  "async-task",
  "blocking",
- "cfg-if",
+ "cfg-if 1.0.4",
  "event-listener",
  "futures-lite",
  "rustix",
@@ -345,7 +339,7 @@ dependencies = [
  "async-io",
  "async-lock",
  "atomic-waker",
- "cfg-if",
+ "cfg-if 1.0.4",
  "futures-core",
  "futures-io",
  "rustix",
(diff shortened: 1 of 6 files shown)


Hmbown:
Thanks for this @raphamorim — swapping the QA harness to rio-vt through a neutral `Color` enum is a clean shape, and keeping the `Frame` surface unchanged is the right call. I approved the workflow runs so this could actually get checked. Four things came back, and one of them is not yours to fix.

## 1. Compile error under `-D warnings` — `unused import: Color`

`Test (macos-latest)` failed at compile, not at assertion:

```
error: unused import: `Color`
error: could not compile `codewhale-tui` (test "terminal_matrix_qa") due to 1 previous error
error: could not compile `codewhale-tui` (test "release_runtime_qa") due to 1 previous error
```

CI sets `RUSTFLAGS: -Dwarnings`, so an unused import is a hard failure. Both `crates/tui/tests/terminal_matrix_qa.rs` and `crates/tui/tests/release_runtime_qa.rs` import `Color` from the harness without using it — presumably a leftover from the vt100 surface. Locally `cargo test -p codewhale-tui` will not reproduce this without `RUSTFLAGS=-Dwarnings`, which is why it slipped.

## 2. `cargo fmt --all -- --check` is dirty

`Lint` failed in 28 seconds with zero compile requests, on the formatting step. The diff it printed is in the harness test code:

```
+            "done lost ANSI role"
+        );
         assert_ne!(
```

`cargo fmt --all` should clear it.

## 3. The link check wants one line in the PR body

`link` is a required check and it is failing only because the body has neither a closing keyword nor an opt-out. Since this is a test-harness migration with no issue behind it, add:

```
No-Issue: migrate the QA PTY harness from vt100 to rio-vt
```

I did not edit your PR body myself.

## 4. The one that needs a decision, not a fix — RUSTSEC-2020-0016

`cargo-deny (advisories)` fails on a transitive dependency the swap introduces:

```
error[unmaintained]: `net2` crate has been deprecated; use `socket2` instead
  ID: RUSTSEC-2020-0016
  Solution: No safe upgrade is available!

net2 v0.2.39
└── corcovado v0.5.0
    ├── rio-vt v0.5.0
    │   └── (dev) codewhale-tui v0.9.2
    └── teletypewriter v0.5.0
        └── rio-vt v0.5.0 (*)
```

This is a genuine, diff-caused finding rather than a flake, and it is the only one I would not just ask you to patch — it is upstream in `rio-vt` → `corcovado`.

It is worth weighing honestly: it is **dev-only** (never linked into a shipped binary), and it is *unmaintained*, not a vulnerability. So the risk is low. But `cargo-deny` is a required check and it does not currently distinguish dev from runtime for advisories, so as it stands this cannot merge.

Three ways forward, and I think the first is best:

1. **`corcovado` drops `net2` for `socket2` upstream.** You are closest to that tree — is it viable? `net2` → `socket2` is usually mechanical.
2. **Scope the advisory exception to dev-dependencies in `deny.toml`**, with a comment naming this advisory and why dev-only is acceptable. Narrow and auditable, but it does weaken a gate we deliberately keep strict.
3. Blanket-ignore `RUSTSEC-2020-0016`. I would rather not — that hides it for runtime dependencies too.

If (1) has a timeline, I am happy to hold this PR for it. If not, tell me and I will take (2) as a separate maintainer commit so the exception lands with its own rationale rather than buried in your migration.

Items 1–3 are quick; item 4 is the actual gate. Ping me when it is pushed and I will re-run.

raphamorim:
Thanks for the review!

About net2: is only in corcovado's Windows (and fuchsia, never-built) TCP/UDP socket code, and nearly all of it is #[allow(unused)]. Rio uses corcovado for PTY/event polling, not network sockets, so this is dead mio-inherited API surface. The socket2 swap is feasible and I can compile-verify it for x86_64-pc-windows-msvc locally (teletypewriter already builds for that target and pulls corcovado). The only non-mechanical bit is keepalive (net2's Option<Duration> vs socket2's TcpKeepalive/bool), and that method is dead.

raphamorim:
updated, corcovado drops net2 for socket2 upstream

Hmbown (approved):
Raphael, thank you for contributing Rio’s parser back to Codewhale’s real PTY acceptance harness. Keeping the existing `Frame` surface and translating colors through a neutral test enum made this a focused engine swap instead of a test rewrite. I also appreciate the quick follow-through on the strict-warning imports and the rio-vt 0.5.1 update that removes the old net2 dependency.

I compared this refreshed head with our locally exercised reconstruction: the patches are equivalent, the PTY, terminal-matrix, and release-runtime groups pass under strict warnings, and the hosted cargo-deny checks are now green. Approving while the remaining platform/mobile jobs finish.

PR #4937, opened 2026-07-28
fix(tui): finalize stale shell transcript cells

### LINKED ISSUE #4547: tui: transcript keeps running spinners and Stop controls for shell jobs that /jobs reports stale or that no longer exist
Problem
After a background shell job goes stale (no output ≥ 60s) or disappears from the job registry (session restart, eviction), the transcript exec card keeps an animated spinner and running state, and the sidebar shows a second spinner plus Stop controls, while `/jobs` reports "stale" or no longer lists the job. The screen and `/jobs` disagree indefinitely.

Current evidence
`job_snapshot()` computes `stale` as display-only while `status` stays `Running` (crates/tui/src/tools/shell.rs:790-794; pinned by crates/tui/src/tools/shell/tests.rs:801-802). The 2.5s reconcile (crates/tui/src/tui/ui.rs:1859, 2222-2229) ignores `job.stale` in `shell_job_tool_status` (ui.rs:1925-1931) and bails via `jobs.get(task_id)?` when the job is missing (ui.rs:1913) with no orphan finalization. Session restore creates a fresh empty ShellManager (crates/tui/src/tui/ui/tests.rs:3730-3739); `remember_stale_job()` (shell.rs:1857) has no production caller. Spinner duplication: history.rs:705-715 (transcript) + sidebar.rs:1780-1788 (Background row) + sidebar.rs:1543-1573 (stale cancel hint). Line numbers at main d9cc9233f.

Scope
1. crates/tui/src/tui/ui.rs — in `refresh_shell_exec_live_output`/`shell_exec_live_update`, propagate `job.stale` to the exec cell: stop spinner animation and render a "running · stale · no output Nm" badge (do not fabricate a terminal status for a live process).
2. crates/tui/src/tui/ui.rs — add a second reconcile pass finalizing exec cells whose `shell_task_id` is absent from the registry map (already built at ui.rs:1866-1874): mark them Failed/"detached" with an explanatory note.
3. crates/tui/src/tools/shell.rs — either wire `remember_stale_job()` into session restore or delete it.
4. Ensure sidebar Background rows and the transcript derive the same stale/missing presentation from the same snapshot fields.

Key files
crates/tui/src/tui/ui.rs; crates/tui/src/tools/shell.rs; crates/tui/src/tui/shell_job_routing.rs; crates/tui/src/tui/sidebar.rs; crates/tui/src/tui/history.rs; crates/tui/src/tui/ui/tests.rs

Acceptance criteria
- [ ] A job `/jobs` labels stale no longer shows a perpetually animating spinner in the transcript; both surfaces show the same stale wording.
- [ ] An exec cell whose job id is no longer in the registry (incl. after session restore) is finalized within one reconcile tick — no permanent "running".
- [ ] Exactly one running-state affordance per job; Stop controls only for jobs the registry can actually stop.
- [ ] New unit tests: stale job → non-animated stale presentation; missing job → finalized cell; restored session with formerly-running cell → finalized.

Verification
cargo check -p codewhale-tui; cargo test -p codewhale-tui -- shell_exec_live; cargo test -p codewhale-tui -- stale; cargo clippy -p codewhale-tui -- -D warnings

Out of scope
Changing the 60s stale threshold; process-group/kill semantics; the agent-notification flow of #3874; the LiveWork dedupe in work_surface/live_projection.rs (already correct).

Related
#3874 (agent notification when background tasks finish — different surface), #4416 (stale failed-agent state across sessions — agents, not shell jobs). Neither is a duplicate. Found during v0.9.1 product-quality intake against installed build d9cc9233f16a.

LI-Jialu:
## Summary
- finalize restored running shell exec cells when their shell job no longer exists in the registry
- render stale running shell transcript cards with a static stale/no-output status instead of a live spinner
- suppress the sidebar spinner for stale running background jobs while keeping the stale/no-output label

Fixes #4547.

## Verification
- cargo fmt --check
- cargo check -p codewhale-tui --bin codewhale-tui --locked
- CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p codewhale-tui --bin codewhale-tui --locked shell_live_output_update -- --test-threads=1
- CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p codewhale-tui --bin codewhale-tui --locked refresh_shell_exec_live_output_finalizes_restored_missing_shell_job -- --test-threads=1
- CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p codewhale-tui --bin codewhale-tui --locked background_task_spinner -- --test-threads=1

Note: macOS emitted the existing linker warning about a large `__eh_frame` section while building the TUI test binary; all targeted tests passed.


Hmbown:
Thanks @LI-Jialu — the diagnosis in #4547 is right, and finalizing a restored running exec cell whose shell job no longer exists in the registry is the correct place to fix it. A spinner that never stops is worse than a stale label, because it tells the operator work is happening when nothing is.

One real failure, and it is small.

## `Lint` fails on `clippy::type_complexity`

```
error: very complex type used. Consider factoring parts into `type` definitions
    --> crates/tui/src/tui/ui.rs:2541:6
     |
2541 | ) -> Option<(String, ToolStatus, Option<String>, u64, bool, Option<u64>)> {
     |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `-D clippy::type-complexity` implied by `-D warnings`
```

`shell_exec_live_update` returned a 5-tuple before this PR, which was already at the edge; adding the sixth element (`Option<u64>`) crosses clippy's threshold. CI runs with `-D warnings`, so it is a hard failure rather than a lint note.

Your local `cargo check -p codewhale-tui` would not have caught this — it needs the clippy invocation from `CONTRIBUTING.md`:

```sh
cargo clippy --workspace --all-features --locked -- \
  -D warnings \
  -A clippy::uninlined_format_args -A clippy::too_many_arguments \
  -A clippy::unnecessary_map_or -A clippy::collapsible_if \
```

## Suggested fix

Please introduce a named struct rather than a `type` alias for the tuple. Six positional fields at a call site is exactly the shape where a future edit silently swaps two of them — and two of yours are already `bool`/`u64`-adjacent, so the compiler would not catch a transposition. Something like:

```rust
struct ShellExecLiveUpdate {
    command: String,
    status: ToolStatus,
    output: Option<String>,
    started_at: u64,
    is_background: bool,
    finished_at: Option<u64>,
}
```

(names guessed from the call site — use whatever the fields actually mean). `#[allow(clippy::type_complexity)]` would also make CI green, but this function is on the live-update path and gains a field roughly every time the shell surface grows, so the struct is worth it now.

Everything else looks reasonable and the rest of the checks are green. Push the fix and I will re-run.

Hmbown:
Jialu, thank you again for tracing the restored-shell lifecycle all the way through the transcript and sidebar surfaces. The key product insight here is exactly right: a stale spinner falsely tells the operator that work is still happening.

I used your enabled maintainer-update permission to append the named `ShellExecLiveUpdate` struct requested in the lint review, without rewriting your commit. Local verification on the updated head passed the three focused regression groups, `cargo fmt --all -- --check`, and the repository’s exact `-D warnings` Clippy command. Hosted checks are refreshing now; once they are green, I’ll merge this original PR so your authorship remains native.

Hmbown (approved):
Approved on the updated head. Jialu’s stale-shell diagnosis and regression coverage are sound, and the appended named state structure resolves the only CI defect without changing the behavior. Local focused tests and the exact strict Clippy gate pass. Thank you for catching a subtle but operator-visible truthfulness bug.

LI-Jialu:
Thanks @Hmbown for the review and maintainer update. The named `ShellExecLiveUpdate` struct is definitely cleaner than extending the tuple. I’ll run the repository’s full clippy gate before opening similar PRs next time. Appreciate the merge!

PR #4942, opened 2026-07-28
fix(tools): preserve CRLF edits

### LINKED ISSUE #4764: `edit_file` tool failed to edit CRLF files on Windows
## Description

On Windows platforms, when the target file uses CRLF (`\r\n`) line endings, the `edit_file` tool fails on exact-match searches, even when the search string is copied verbatim from `read_file` output.

## Steps to reproduce

The following steps work for any AI coding agent (CodeWhale, Cursor, Copilot, etc.) and require no specific project — just a workspace where `edit_file` is available.

### Step 1: Create a CRLF File

First create a helper script, then run it to write a CRLF file. This approach avoids embedding Python in shell commands, eliminating compatibility issues between Windows/macOS/Linux.

> Note: Even on Linux/macOS, the commands below produce a CRLF file because `\r\n` is written explicitly.

#### 1a. Create the Helper Script

```python
# crlf_repro/create_test.py
import os

os.makedirs('crlf_repro', exist_ok=True)

# Write the test file with explicit CRLF (\r\n) line endings
content = b'def greet(name):\r\n    print(f"hello, {name}")\r\n\r\ndef add(a, b):\r\n    return a + b\r\n'
with open('crlf_repro/test.py', 'wb') as f:
    f.write(content)

print('created: crlf_repro/test.py')
with open('crlf_repro/test.py', 'rb') as f:
    data = f.read()
print('has CRLF:', b'\r\n' in data)
```

Use the `write_file` tool to write the above content into `crlf_repro/create_test.py`.

#### 1b. Execute the Helper Script

```bash
python crlf_repro/create_test.py
```

> If only `python3` is available (common on Linux/macOS), use `python3 crlf_repro/create_test.py`.

Expected output:

```
created: crlf_repro/test.py
has CRLF: True
```

### Step 2: Read the File with `read_file`

Ask the AI agent:

> Read the file `crlf_repro/test.py`

### Step 3: Edit the File with `edit_file`

Copy the `add` function from the `read_file` output and ask the AI agent:

> Edit `crlf_repro/test.py`: replace the function `add` with a new implementation that returns a * b

If the agent's `edit_file` has this bug, it will error:

```
Search string not found in crlf_repro/test.py. Recovery: call read_file with path="crlf_repro/test.py" to inspect the current contents, then retry with a search string copied from the file.
```

### Step 4 (Alternative): Manual Search String Reproduction

If Step 3 happens to succeed (the agent may have used `write_file` instead of `edit_file`, or the file wasn't actually CRLF), a more precise reproduction:

Ask the agent:

> Create a file `crlf_repro/repro.py` with CRLF line endings containing the code below, then use edit_file to replace `return a + b` with `return a * b`. The file must use CRLF (`\\r\\n`) line endings.

```python
def add(a, b):
    return a + b
```

```bash
python -c "print('CRLF:', b'\r\n' in open('crlf_repro/repro.py','rb').read())"
```

> If only `python3` is available, replace `python` with `python3` in the command.

### Root Cause

```
File content (CRLF):   "def add(a, b):\r\n    return a + b\r\n"
Search string (LF):    "def add(a, b):\n    return a + b\n"
                                              ^
                                       \r\n ≠ \n  →  match fails
```

Every AI agent provides search strings with LF-only line endings (an inherent behavior of LLMs), while CRLF file content contains `\r\n`. When `edit_file` performs exact byte matching, `\r\n` ≠ `\n` results in zero matches.

## Diagnostics

- File encoding: UTF-8
- File line endings: `\r\n` (CRLF), confirmable via hex dump
- The `\r` characters in `read_file` output are invisible in JSON/terminal display, making the model unaware of the difference
- Even when copy-pasted from `read_file` output, the `\r` characters may be lost by the time the model passes the search parameter to `edit_file`

### Root Cause

`EditFileTool::execute()` in `crates/tui/src/tools/file.rs` uses `fs::read_to_string()` to read the file. On Windows, this function **preserves CRLF line endings**. The subsequent matching:

```rust
let count = contents.matches(search).count();
```

Rust's `str::matches()` performs **exact byte matching**. When the file content contains `\r\n` and the search string contains `\n`, `\r\n` ≠ `\n`, giving a match count of 0.

### Failure Chain

```
File on disk:        line1\r\nline2\r\n
    │
    ▼
fs::read_to_string()  preserves CRLF
    │
    ▼
contents:           "line1\r\nline2\r\n"
search (from model): "line1\nline2\n"
    │
    ▼
contents.matches(search)  →  0   ✗  (\r\n ≠ \n)
    │
    ▼
leading_whitespace_fuzzy_matches()  →  does not handle line ending differences
    │
    ▼
punctuation_normalized_matches()    →  does not handle line ending differences
    │
    ▼
"Search string not found"  Error
```

### Affected Code Locations

| Location | Line | Description |
|----------|------|-------------|
| `read_file` reading | `file.rs:173` | `fs::read_to_string()` preserves CRLF |
| Exact match | `file.rs:866` | `contents.matches(search)` lacks CRLF compatibility |
| Fuzzy match 1 | `file.rs:869` | `leading_whitespace_fuzzy_matches()` doesn't handle line endings |
| Fuzzy match 2 | `file.rs:882` | `punctuation_normalized_matches()` doesn't handle line endings |

### Contrast

- `apply_patch` already has CRLF preservation (`apply_patch_preserves_crlf_line_endings` test) and corresponding handling
- `edit_file` **lacks** similar treatment

## Environment

- OS: Windows 10
- codewhale version: 0.9.0
- Install method:
- `codewhale doctor` summary:
- Model/provider: deepseek-v4-flash
- Terminal app:
- Shell:

nightt5879:
## Summary

- match `edit_file` searches on an LF-normalized view while mapping the unique span back to original CRLF bytes
- normalize replacement newlines to the base file style without rewriting untouched content, while preserving duplicate-match and normalized no-op safeguards
- cover exact, mixed-line-ending, indentation-fuzzy, punctuation-fuzzy, UTF-8, and no-trailing-newline EOF cases; credit @LmeSzinc for the report and root-cause analysis
- reduce post-review memory use to constant space for exact-match counting and O(CRLF count) for normalized-to-original range mapping

Closes #4764

## Review

- [x] Self-reviewed the full `upstream/main...HEAD` diff and commit metadata
- [x] Independent AI review found one P2 allocation concern and one P3 EOF coverage gap; follow-up `02aeb9f73` fixes both
- [x] Independent AI re-review of final head `02aeb9f73` found no remaining P0/P1/P2/P3 issue
- [x] No inline review threads remain to resolve

## Testing

- [x] `cargo +1.97.1 fmt --all -- --check`
- [x] `cargo +1.97.1 test -p codewhale-tui --bin codewhale-tui --locked edit_file_ -- --nocapture` (33 passed)
- [x] `cargo test --workspace --all-features --locked` in final GitHub Windows and macOS jobs
- [x] `cargo clippy --workspace --all-features --locked ... -D warnings` in final GitHub Linux lint job
- [x] `cargo +1.97.1 build --release -p codewhale-cli -p codewhale-tui --locked`
- [x] OpenHarmony cross-check and all final GitHub status checks
- [x] Web Vitest, facts/docs checks, ESLint, TypeScript, and production build

The local Windows workspace run additionally reached 9,157 passing tests; its remaining failures were pre-existing environment-specific shell/temp-path cases, while the authoritative GitHub Windows and macOS full workspace jobs passed.

## Checklist

- [x] TUI manual verification is not applicable (tool behavior only; no rendering change)
- [x] Co-authored credit uses GitHub numeric noreply address `37934724+LmeSzinc@users.noreply.github.com`

### DIFF
--- a/.github/AUTHOR_MAP
+++ b/.github/AUTHOR_MAP
@@ -64,6 +64,7 @@ zlh124 = zlh124 <56312993+zlh124@users.noreply.github.com>
 LeoLin990405 = LeoLin990405 <101193087+LeoLin990405@users.noreply.github.com>
 THINKER-ONLY = THINKER-ONLY <181556007+THINKER-ONLY@users.noreply.github.com>
 nightt5879 = nightt5879 <87569709+nightt5879@users.noreply.github.com>
+LmeSzinc = LmeSzinc <37934724+LmeSzinc@users.noreply.github.com>
 CCChisato = Fushimi Rio <158128433+CCChisato@users.noreply.github.com>
 aznikline = aznikline <27564626+aznikline@users.noreply.github.com>
 Aznable = aznikline <27564626+aznikline@users.noreply.github.com>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -147,6 +147,11 @@ already claimed to.
 
 ### Fixed
 
+- `edit_file` now matches LF-only model search text against CRLF files,
+  preserves the file's line-ending style for replacement text, and still
+  rejects newline-normalized duplicate matches as non-unique (#4764).
+  Reported and root-caused by @LmeSzinc.
+
 - `/fleet status` read the current TUI session's sub-agents while `codewhale
   fleet status` read the durable `.codewhale/fleet.jsonl` ledger — two different
   things wearing one name, so a run started by `codewhale fleet run` never
@@ -312,6 +317,8 @@ Thank you to the contributors whose code, reports, and reviews shaped v0.9.2:
 - [@fleitz](https://github.com/fleitz) — required source-candidate credit for
   the canonical `Bash` no-`cwd` workspace fix and regression in PR #4673
   (issue #4674).
+- [@LmeSzinc](https://github.com/LmeSzinc) — the Windows CRLF `edit_file`
+  reproduction, root-cause analysis, and affected-code anchors in issue #4764.
 
 ## [0.9.1] - 2026-07-24
 
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -147,6 +147,11 @@ already claimed to.
 
 ### Fixed
 
+- `edit_file` now matches LF-only model search text against CRLF files,
+  preserves the file's line-ending style for replacement text, and still
+  rejects newline-normalized duplicate matches as non-unique (#4764).
+  Reported and root-caused by @LmeSzinc.
+
 - `/fleet status` read the current TUI session's sub-agents while `codewhale
   fleet status` read the durable `.codewhale/fleet.jsonl` ledger — two different
   things wearing one name, so a run started by `codewhale fleet run` never
@@ -312,6 +317,8 @@ Thank you to the contributors whose code, reports, and reviews shaped v0.9.2:
 - [@fleitz](https://github.com/fleitz) — required source-candidate credit for
   the canonical `Bash` no-`cwd` workspace fix and regression in PR #4673
   (issue #4674).
+- [@LmeSzinc](https://github.com/LmeSzinc) — the Windows CRLF `edit_file`
+  reproduction, root-cause analysis, and affected-code anchors in issue #4764.
 
 ## [0.9.1] - 2026-07-24
 
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -868,28 +868,27 @@ impl ToolSpec for EditFileTool {
         })?;
 
         // Models provide LF newlines even when the file on disk uses CRLF.
-        // Match in a newline-normalized view, while retaining a byte map so
-        // only the original span is replaced and the rest of the file stays
-        // byte-for-byte untouched.
-        let (normalized_contents, contents_byte_map) = normalize_crlf_with_map(&contents);
+        // Match in a newline-normalized view, while retaining the sparse
+        // positions where CR bytes were removed so only the original span is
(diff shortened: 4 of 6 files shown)


nightt5879:
@codex review Please review current head 1f7441970. Focus on CRLF-to-LF byte-range mapping at UTF-8 boundaries and EOF, logical uniqueness across mixed LF/CRLF content, replacement newline preservation, normalized no-op behavior, and composition with indentation/punctuation fuzzy matching. Please flag correctness, unnecessary allocation, or missing edge-case coverage.

nightt5879:
Independent AI review of head `1f7441970` found no P0/P1 correctness blocker, one P2 allocation concern, and one P3 EOF coverage gap. Follow-up `02aeb9f73` addressed both: exact matching now counts in constant memory, CRLF mapping stores only sparse normalized newline positions, and UTF-8 matches ending exactly at EOF have direct range and end-to-end coverage.

The independent reviewer then re-reviewed final head `02aeb9f73` and found no remaining P0/P1/P2/P3 issues. It specifically confirmed the sparse boundary formula and complexity, constant-memory exact-match counting, and UTF-8 + CRLF + no-trailing-newline EOF tests.

Post-review focused gate: 33 passed, 0 failed. There are no inline review threads to resolve; the final CI run is authoritative.

Hmbown (approved):
Reviewed current head 02aeb9f73. The sparse CRLF boundary mapping is correct at newline starts, match ends, UTF-8 boundaries, and EOF; logical duplicate and normalized no-op guards remain intact; replacement newlines preserve the file convention; focused and full platform checks are green.

PR #4944, opened 2026-07-28
feat(web): align landing with managed product

Hmbown:
## Summary
- replace the public-site whale with CWC's Signal Current brand mark
- simplify the hero into the same quiet navy shell used by the managed product
- remove decorative CSS fish/jelly artwork and frame the real terminal capture
- label the existing capture as v0.9.1 instead of inheriting the v0.9.2 source candidate version
- keep the mobile language selector and menu inside the 390px viewport

## Verification
- `npm run prebuild`
- `npm run check:facts`
- `npm run check:docs`
- `npm test` (223 passed)
- `npm run lint`
- `npm run build` (260 static pages)
- in-app browser at 1280x720 and 390x844: no horizontal overflow, menu opens/closes, no console errors

No-Issue: visual alignment and release-truth polish for the v0.9.2 landing.


Hmbown:
Addressed all three review notes on the authoritative branch:

- `095588c52` makes the new shell labels locale-neutral (`Codewhale`, `MIT`, `Codewhale TUI`, and the captured version) instead of bypassing dictionaries.
- The same commit removes obsolete `currentColor` classes/rules and pins the two-tone Signal Current mark directly in the Blue Stage contract test.
- `a21f84f8b` retains `figcaptionSourceCandidate` as localized source-state metadata, so the key is no longer orphaned while the pinned capture remains truthfully published.

Local receipts after the fixes: 223 tests passed, lint clean, production build generated 260 static pages.

PR #4958, opened 2026-07-28
ci: add an SBOM attestation and pin the provenance mode explicitly

kobihikri:
Hi, and thanks for CodeWhale.

*(Rewritten after @Hmbown's review — the original description was factually wrong and this reflects what the change actually does.)*

`.github/workflows/release.yml` pushes the release image with BuildKit's default attestation behaviour. Two lines on the build step:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

**What each line actually does:**

- **`sbom: true` is the new behaviour.** SBOM attestations are not added by default — [Docker's docs](https://docs.docker.com/build/ci/github-actions/attestations/): *"SBOM attestations aren't automatically added to the image"*. This is the substance of the PR.
- **`provenance: mode=max` makes existing behaviour explicit.** For public repositories that is already the default — *"Public repos: provenance attestations with `mode=max` are automatically added"* — and I confirmed `ghcr.io/hmbown/codewhale:latest` already carries attestation manifests. The value of pinning it is that the behaviour cannot change quietly if the default ever does, on a publishing path where that matters.

I also had the mode semantics backwards in my original caveats: `provenance: true` resolves to **max**, not to a smaller record. The smaller setting is `provenance: mode=min`. Corrected here and on the other PRs where I repeated it.

An SBOM does make the image useful to anyone assessing what a CodeWhale release contains without running it, which is the main practical gain.

Both commits carry `Signed-off-by` trailers as requested — agreed that publishing plumbing is a trust boundary and the provenance record should be clean regardless of the DCO check being advisory.

Disclosure: I used AI assistance to prepare this change; the original description's errors were mine, and I have corrected them rather than quietly editing them away.

No-Issue: release workflow supply-chain metadata hardening


Hmbown (changes requested):
Kobi, thank you for looking closely at the release image rather than treating the publish step as a black box. Adding the SBOM is useful supply-chain metadata, and making provenance policy explicit is a reasonable direction.

I verified the current published image and the pinned `docker/build-push-action` behavior. One factual correction is needed before this lands: public repositories already receive `mode=max` provenance by default, and `ghcr.io/hmbown/codewhale:latest` already carries it. The new artifact behavior here is the SBOM; the provenance line makes the existing behavior explicit. Also, `provenance: true` defaults to max, not the smaller form—the smaller setting is `provenance: mode=min`.

Please update the title/body to reflect that distinction and amend both commits with `Signed-off-by` trailers. The DCO workflow is currently advisory, but release/publishing plumbing is a trust-boundary surface, so I want the provenance record clean. The code diff itself looks focused; the only remaining hosted gates are macOS and Windows.

kobihikri:
Thank you — you were right on both counts, and the first one was a real error on my part rather than a wording quibble.

Done:

- **Both commits amended with `Signed-off-by` trailers** (`5c09915c`, `0f63bcaf`). Agreed on the reasoning: the DCO check being advisory is not a reason for the provenance record on publishing plumbing to be untidy.
- **Title and description rewritten** to say what the change actually does — the SBOM is the new behaviour, and the `provenance: mode=max` line pins existing behaviour rather than introducing it.
- **The mode semantics corrected.** I had written that `provenance: true` gives a smaller record. It does not: `true` resolves to max on a public repo, and `mode=min` is the smaller setting. That was simply wrong.

On the factual correction itself — you were right that public repos already receive `mode=max` by default, and I confirmed `ghcr.io/hmbown/codewhale:latest` already carries attestation manifests before rewriting the description. I had repeated the same incorrect premise across a number of similar PRs, so I have gone back and posted a correction on each of them rather than leave the wrong claim standing. Thank you for catching it; that one was worth catching.

The remaining case for the PR is narrower than I first made it: an SBOM that was genuinely absent, plus an explicit provenance line so the behaviour cannot shift quietly if the default ever changes. If that is not worth the diff to you, closing it is a perfectly reasonable call and I will not argue.

Hmbown (approved):
Kobi, thank you for handling this exactly the way healthy supply-chain review should work: you verified the current image, corrected the public record instead of papering over the premise, fixed the mode semantics everywhere you had repeated them, and amended both commits with clean sign-offs. The narrower change is worthwhile—an attached SBOM is a real new capability, and pinning the provenance mode makes release policy explicit. Approved on `0f63bcaf`; hosted checks are refreshing.

PR #4961, opened 2026-07-29
fix(tui): preserve reasoning effort with auto routing

### LINKED ISSUE #4941: Thinking level silently reverts to Auto on restart: an auto model discards the persisted reasoning_effort
## Symptom

Reported by Hunter: the thinking level "keeps going back to auto" between sessions.

## What is *not* broken

Persistence at the settings layer works. The picker path (`apply_picker_effort_choice`, `ui.rs:10623`) calls `StartupDefaults::reasoning_effort(...)` through `apply_blocking`, and `StartupDefaults::apply` (`tui/startup_defaults.rs:236`) writes it via `Settings::set`. Confirmed on a real machine — the settings file contains `reasoning_effort = "max"`.

Session resume is also not the culprit: `reasoning_effort` is not stored on the session record, so `/resume` cannot clobber it.

## The confirmed bug: `auto` model discards the saved thinking level

`crates/tui/src/tui/app/init.rs:359`

```rust
let mut reasoning_effort = if auto_model {
    ReasoningEffort::Auto
} else {
    configured_reasoning_effort.map_or_else(ReasoningEffort::default, |s| {
        ReasoningEffort::from_setting_for_provider(s, provider)
    })
};
```

with `auto_model = model.trim().eq_ignore_ascii_case("auto")` (`init.rs:286`).

**When the model is `auto`, the persisted `reasoning_effort` is not read at all** — it is unconditionally overwritten with `ReasoningEffort::Auto`. And note the two `if !auto_model` guards immediately below (`:366`, `:376`): every other effort-resolution rule is skipped in this branch too.

The result is a silent round trip that loses user intent:

1. user opens the thinking picker and selects e.g. `max`
2. `apply_picker_effort_choice` persists `reasoning_effort = "max"` — *and reports success*
3. next launch on an `auto` model shows `Auto`
4. the setting is still sitting in the settings file, correct and ignored

Step 2 is what makes this a bug rather than a policy: the product accepts the choice, writes it, tells the user it was saved, and then does not honor it. Either it should be honored, or the picker should say the choice cannot apply on an auto route.

## The design question this needs answered

There is a defensible reason for the current behavior: with auto model routing the concrete route is not known until dispatch, and effort tiers are route-specific (`as_setting_for_route`, `normalize_for_route`). A saved `max` may be meaningless or invalid on whatever route auto picks.

But "the tier might need normalizing later" argues for *normalizing at dispatch*, not for discarding the preference at startup. `normalize_for_route` already exists and is already applied elsewhere. So the fix is likely:

- read `configured_reasoning_effort` in the `auto_model` branch too,
- keep `ReasoningEffort::Auto` only as the genuine default when nothing is configured,
- normalize the resolved tier per route at dispatch, which already happens.

## Not yet explained

Hunter's `default_model` is `deepseek-v4-pro`, not `auto`, yet the symptom is still reported — so either the model is being switched to auto routing in-session, or there is a second path. Needs the disambiguation below before the fix is scoped:

- TUI, web, or both?
- new session or `/resume`?
- what does the footer show immediately after setting the tier, versus after a restart?
- is auto model routing on at the time?

## Acceptance

- Setting a thinking tier and restarting restores that tier, including on auto-routed models.
- If a tier genuinely cannot apply to a resolved route, the UI says so at selection time rather than accepting and silently reverting.
- A regression test that fails if a persisted `reasoning_effort` is ignored at startup for any model value, `auto` included.

nightt5879:
## Summary

- keep automatic model routing independent from the persisted reasoning-effort preference
- preserve the raw requested tier through startup, model/provider changes, session restore, config, picker, Ctrl+T, Hotbar, CLI, workflow, ACP, and runtime-thread flows
- normalize reasoning only after the concrete provider/model/endpoint route is known
- replay cache inspection and warmup on the last concrete Auto route, rejecting exact-identity or endpoint drift
- add regression coverage for persistence, cross-provider selection, session restore, non-TUI runtimes, cache replay, UI controls, and wire-level route planning

Fixes #4941

## Verification

- cargo fmt --all -- --check
- cargo check -p codewhale-tui --bin codewhale-tui --locked
- reasoning_effort: 43 passed
- auto_model: 27 passed
- model_picker: 99 passed
- config_ui: 16 passed
- cache: 276 passed
- terminal_mode_tests: 88 passed
- runtime_threads: 123 passed; one unrelated timing test passed three consecutive isolation runs
- codewhale-cli all-features serial run: 156 passed
- workspace all-features clippy passed with the CI allowlist plus Rust 1.95 allowances for three pre-existing lint classes
- full TUI all-features run: 9,207 passed, 4 ignored; the sole failure is the pre-existing Windows Job Object kill-on-close test, which also fails in isolation and is unchanged from upstream/main
- full workspace parallel run reached nine Windows temp-file sharing failures in codewhale-cli; the same 156-test crate passes serially

## Self-review

Verified that persisted reasoning remains an unresolved user preference until route selection, model Auto and reasoning Auto have separate authority flags, explicit CLI overrides still win, cache replay never substitutes the configured provider/model for the last Auto route, named custom identities fail closed, endpoint drift refuses warmup, and no unrelated worktree changes are included.

### DIFF
--- a/crates/tui/src/commands/groups/config/config.rs
+++ b/crates/tui/src/commands/groups/config/config.rs
@@ -1529,14 +1529,15 @@ pub fn set_config_value(app: &mut App, key: &str, value: &str, persist: bool) ->
             // Support "/model auto" — auto-select model based on request complexity
             if value.trim().eq_ignore_ascii_case("auto") {
                 app.set_model_selection("auto".to_string());
-                app.reasoning_effort = ReasoningEffort::Auto;
-                app.last_effective_reasoning_effort = None;
                 app.update_model_compaction_budget();
                 app.session.last_prompt_tokens = None;
                 app.session.last_completion_tokens = None;
                 app.session.last_output_throughput = None;
                 return CommandResult::with_message_and_action(
-                    "model = auto (auto-select model and thinking per turn)".to_string(),
+                    format!(
+                        "model = auto (auto-select model per turn; thinking = {})",
+                        app.reasoning_effort_display_label()
+                    ),
                     AppAction::UpdateCompaction(app.compaction_config()),
                 );
             }
@@ -2131,10 +2132,6 @@ pub fn set_config_value(app: &mut App, key: &str, value: &str, persist: bool) ->
             ) && let Some(ref model) = settings.default_model
             {
                 app.set_model_selection(model.clone());
-                if app.auto_model {
-                    app.reasoning_effort = ReasoningEffort::Auto;
-                    app.last_effective_reasoning_effort = None;
-                }
                 app.update_model_compaction_budget();
                 app.session.last_prompt_tokens = None;
                 app.session.last_completion_tokens = None;
@@ -2143,16 +2140,23 @@ pub fn set_config_value(app: &mut App, key: &str, value: &str, persist: bool) ->
             }
         }
         "reasoning_effort" | "effort" => {
-            app.reasoning_effort = if app.auto_model {
-                ReasoningEffort::Auto
-            } else {
-                settings
-                    .reasoning_effort
-                    .as_deref()
-                    .map_or_else(ReasoningEffort::default, |value| {
+            app.reasoning_effort = settings.reasoning_effort.as_deref().map_or_else(
+                || {
+                    if app.auto_model {
+                        ReasoningEffort::Auto
+                    } else {
+                        ReasoningEffort::default()
+                    }
+                },
+                |value| {
+                    if app.auto_model {
+                        ReasoningEffort::from_setting(value)
+                    } else {
                         ReasoningEffort::from_setting_for_provider(value, app.api_provider)
-                    })
-            };
+                    }
+                },
+            );
+            app.reasoning_effort_explicit = settings.reasoning_effort.is_some();
             app.last_effective_reasoning_effort = None;
             app.update_model_compaction_budget();
             action = Some(AppAction::UpdateCompaction(app.compaction_config()));
@@ -3259,7 +3263,7 @@ Parse error: permissions.toml at permissions.toml could not be parsed: expected 
     }
 
     #[test]
-    fn config_model_auto_enables_auto_thinking() {
+    fn config_model_auto_preserves_explicit_thinking() {
         let mut app = create_test_app();
         app.reasoning_effort = ReasoningEffort::Off;
 
@@ -3268,9 +3272,33 @@ Parse error: permissions.toml at permissions.toml could not be parsed: expected 
         assert!(result.message.is_some());
         assert!(app.auto_model);
         assert_eq!(app.model, "auto");
-        assert_eq!(app.reasoning_effort, ReasoningEffort::Auto);
+        assert_eq!(app.reasoning_effort, ReasoningEffort::Off);
+        assert!(
+            result
+                .message
+                .as_deref()
+                .is_some_and(|message| message.contains("thinking = off"))
+        );
         assert!(app.last_effective_model.is_none());
         assert!(app.last_effective_reasoning_effort.is_none());
+    }
+
+    #[test]
+    fn config_reasoning_effort_applies_while_model_routing_is_auto() {
+        let mut app = create_test_app();
+        app.set_model_selection("auto".to_string());
+        app.reasoning_effort = ReasoningEffort::Auto;
+        app.reasoning_effort_explicit = false;
+
+        let result = set_config_value(&mut app, "reasoning_effort", "low", false);
+
+        assert!(!result.is_error);
+        assert_eq!(app.reasoning_effort, ReasoningEffort::Low);
+        assert!(app.reasoning_effort_explicit);
+        assert!(matches!(
+            result.action,
+            Some(AppAction::UpdateCompaction(_))
+        ));
     }
 
     #[test]
--- a/crates/tui/src/commands/groups/config/config.rs
+++ b/crates/tui/src/commands/groups/config/config.rs
@@ -3266,6 +3266,7 @@ Parse error: permissions.toml at permissions.toml could not be parsed: expected 
     fn config_model_auto_preserves_explicit_thinking() {
         let mut app = create_test_app();
         app.reasoning_effort = ReasoningEffort::Off;
+        app.reasoning_effort_explicit = true;
 
         let result = config_command(&mut app, Some("model auto"));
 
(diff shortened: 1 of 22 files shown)


nightt5879:
@codex review

Please review current head `8c9a51cfc`. Focus on correctness at the boundary between unresolved Auto model routing and reasoning-effort persistence: startup restore, provider/model switching, session restore, picker/config/keyboard/hotbar entry points, classifier input, and concrete-route wire normalization. Please flag any state transition that still overwrites the requested tier or any path that sends an unnormalized tier after route resolution.

nightt5879:
@Hmbown This fork PR cannot receive the repository Claude review secret, and GitHub does not allow the external PR author to request the Copilot reviewer via the review-request API. Could you please request a Copilot review for the current head? I will wait for it, address any valid findings in a follow-up commit, and resolve the review threads before considering the PR complete.

nightt5879:
Cloud AI review follow-up is now pushed in `5f977ec28`.

The independent Codex review of `8c9a51cfc` reported three actionable findings, all addressed:

- **P1 — implicit fixed-model defaults entering Auto:** `set_model_selection`, `/model`, and the model picker now preserve a fixed reasoning tier only when it was explicit; an implicit fixed-route default returns to per-turn `Auto` reasoning and is not persisted as an accidental preference.
- **P2 — restoring Auto sessions:** session restore now carries forward only an explicit global reasoning preference; implicit fixed-route defaults become `Auto`.
- **P2 — post-turn labels:** dispatch records the exact route-effective reasoning receipt (including constrained/unavailable states), persists it alongside the Auto route receipt, and restores it without breaking older session JSON.

A follow-up self-review also caught and fixed the Z.ai/Minimax/custom-route capability edge: post-turn UI no longer claims a concrete tier when only “thinking enabled; granularity unavailable” can be proven.

Validation after the follow-up:

- focused Auto/reasoning/model-picker/hotbar/session suites: all passed (including 22 + 43 + 96 + 95-test filters)
- full all-features TUI suite: 9,186 passed, 4 ignored; only the two documented Windows baseline failures remained. The approval test passed immediately in isolation; the Windows Job kill-on-close test reproduced in isolation and both source files are unchanged from upstream.
- all-features TUI Clippy passed with `-D warnings` after command-line allowances for pre-existing Rust 1.95 Windows-only lints in unchanged code
- `cargo fmt --all -- --check` and `git diff --check`: passed

I am waiting for the refreshed CI and any GitHub-hosted reviewer feedback before declaring the PR ready.

nightt5879:
Second independent cloud Codex review follow-up is in `3eed55bf5`.

Resolved both remaining P2 findings:
- Ctrl+T and `reasoning.cycle` now traverse the full unresolved Auto vocabulary (`auto -> off -> low -> medium -> high -> max -> auto`) without applying the configured provider's normalization early.
- Startup now ignores reasoning effort inferred from a retired fixed-model alias when the selected model is Auto and no explicit preference exists. Explicit Settings/config preferences remain authoritative.

Added regression coverage for the full shortcut cycle, the hotbar path, inferred legacy-alias startup, and the explicit-config counter-case.

Local validation:
- `git diff --check`
- Auto-model filter: 24 passed
- reasoning-effort filter: 43 passed
- hotbar filter: 95 passed
- all-feature Clippy with the repository's pre-existing Windows lint allowances and `-D warnings`
- full all-feature TUI suite: 9,187 passed, 4 ignored; the only two failures are unchanged upstream tests (`system_prompt_prefix_never_leaks_private_content`, caused by this machine's real global instructions and passing under isolated HOME/TEMP, and the pre-existing Windows job kill-on-close case, which still fails in isolation). Both failing test source files are byte-identical to `upstream/main`.

nightt5879:
Third cloud-review follow-up is now pushed in `6057bc60b`.

The review correctly identified two remaining raw-preference losses:

- fixed routes normalized the live tier (`DeepSeek low -> high`, `Codex off -> low`) and then fed that normalized value back when returning to Auto;
- session restore snapshotted the normalized live tier instead of the persisted raw global preference.

The new commit keeps the raw explicit preference separate from the route-effective live tier, reprojects it only at concrete route boundaries, and persists the raw value. During self-review I also applied the same invariant to `/model auto`, provider/fallback switches, the config UI, Ctrl+T/exact-K3 persistence, model-picker navigation, and hotbar route actions.

Validation on Windows:

- `cargo fmt --all` and `git diff --check`
- all-features clippy with warnings denied
- targeted: config UI 15, model picker 98, reasoning 43, Auto model 24, session restore 12, provider switch 17, hotbar 95, plus exact regression filters — all passed
- all-features TUI suite: 9,193 passed / 4 ignored out of 9,200; the remaining two stable local-environment failures are unchanged from `upstream/main`, and one unrelated parallel timing test passed immediately in isolation. No touched test failed.

I am starting a fresh independent AI review against this head and will keep the PR open until that review and the new CI run are fully resolved.

nightt5879:
Fourth cloud-review round addressed in 111513f68:

- The model picker now treats an absent raw preference as `Auto` while retaining the visible fixed-route tier as its working selection, so switching to `auto` commits exactly what the picker shows.
- Fixed-route `Auto` labels preserve `ThinkingEnabledGranularityUnavailable` / `Unavailable` receipts instead of dropping them through a tier-only projection.
- `/cache inspect` and `/cache warmup` now replay enabled-but-untiered receipts with a canonical enabled request tier, retaining live-request semantics for GLM/M3 routes.
- Self-review also found and fixed a startup-precedence gap in the schema-driven TUI/web config editor: persisted reasoning changes now update both `config.toml` and the higher-precedence startup setting.

Fresh validation on this commit:
- `cargo fmt --all` and `git diff --check`
- model-picker group: 99 passed
- config UI group: 16 passed
- reasoning-effort group: 43 passed
- auto-model group: 24 passed
- cache group: 271 passed
- workspace/all-features Clippy with warnings denied (plus allowances for three pre-existing Rust 1.95 lints)
- full TUI all-features suite: 9,199 passed, 4 ignored, 1 known Windows-only baseline failure (`windows_job_kill_on_close_releases_reader_threads_when_terminate_denied`); the failing test/implementation match `upstream/main` and the same failure reproduces in isolation.

I am starting another independent cloud review against the updated head and will address any new findings before considering the PR ready.

nightt5879:
Fifth independent AI-review follow-up is now pushed in two focused commits:

- 571ff8589 honors the saved fixed reasoning tier in CLI-backed/non-TUI Auto-model routes while preserving explicit Auto reasoning.
- d8ae918a3 makes cache inspect/warmup replay the last concrete Auto provider, exact identity, model, and endpoint; identity or endpoint drift now fails closed.

Local validation includes fmt/check, 43 reasoning tests, 27 Auto-model tests, 276 cache tests, 99 model-picker tests, 16 config-UI tests, 88 terminal-mode tests, workspace all-features clippy, and a full TUI all-features run (9,207 passed, 4 ignored, one unchanged Windows Job Object baseline failure). The codewhale-cli all-features suite passes 156/156 serially after its parallel Windows temp-file sharing failures.

I also self-reviewed exact custom-provider provenance and added a regression that rejects a mismatched persisted provider identity before cache replay.

nightt5879:
Sixth independent AI-review follow-up is pushed in `a606df053`.

Both findings were valid and are resolved:

- changing the reasoning request while `model = auto` now invalidates the previous concrete provider/model/identity and Auto receipt, so `/cache inspect` and warmup cannot replay a route selected under stale classifier input;
- effective-reasoning receipts are reused only when the queried request still matches the live selection, so Ctrl+T and the hotbar immediately report and record the new unresolved tier instead of the previous turn's normalized tier.

All reasoning mutation surfaces now share the same invalidation helper. A regression starts from a replayable completed Codex Auto route, changes thinking, and proves that the old route becomes unreplayable while the status and Work receipt use the new tier.

Fresh validation on this head:

- focused regression: 1 passed
- reasoning-effort: 43 passed
- Auto-model: 27 passed
- cache: 276 passed
- model-picker: 99 passed
- config UI: 16 passed
- workspace/all-features Clippy with warnings denied (plus the three documented Rust 1.95 baseline allowances)
- full TUI all-features: 9,207 passed / 4 ignored; the private-prompt test passes with HOME/TEMP isolated, and the sole remaining Windows Job Object failure reproduces alone with unchanged source versus `upstream/main`.

I am waiting for CI and the refreshed cloud-review state on this exact commit before marking the PR ready.

PR #4979, opened 2026-07-30
fix(tui): detach foreground shell before steering

### LINKED ISSUE #4930: Enter during foreground shell should detach it before steering
## Problem

When the agent is mid-turn with a foreground Bash command blocking (e.g. `sleep 30`,
`cargo build`), the user's natural impulse is to type a message and hit Enter.
This currently fails in a confusing way:

### What the user does
1. Foreground shell blocks the turn.
2. User types "hey, while that runs..." in the input field.
3. User hits **Enter**.

### What happens (buggy)
1. Enter is captured at `ui.rs:6272-6273` as a forced submit while loading.
2. The message is built and sent to `attempt_steer_with_queue_fallback` (line 6300-6307).
3. The steer fires into the current turn — **without detaching the foreground shell**.
4. The foreground command stays blocking The user is waiting.
5. If the user now tries **Ctrl+B**, it's too late — the steer already altered the
   turn state, or the foreground command already completed, so
   `request_foreground_shell_background` shows "Cannot move to /jobs: no foreground
   shell command is running."

The **order of operations is wrong**. The message gets sent (steered) before the
foreground shell gets a chance to detach. What should happen: detach first,
then send.

### Why Ctrl+B alone isn't the answer
- Ctrl+B works in putting the job top background when it is in the current turn, but there is no possibility of entering user's message, so we have to cancel and insert the message for new turn.

## Desired behavior

When the user hits Enter while a foreground shell is blocking, the shell should
be detached to background first, then the message should steer into the turn.

## Environment
- codewhale 0.9.1
- SSH/tmux

nightt5879:
## Summary

- request that a blocking foreground Bash wait move to `/jobs` before enqueueing a same-turn steer
- keep the current busy-Enter queue behavior intact while covering every path that actually steers, including explicit steer and queued-message promotion
- distinguish true foreground waits from detached jobs that intentionally remain `Running`

Fixes #4930

## Why this still applies on current `main`

Since the report, bare Enter during a busy turn was changed to queue a follow-up. The ordering bug still exists when that follow-up is explicitly promoted into the active turn (or the user uses the explicit steer chord): the shared steer boundary sent the message without first releasing a foreground shell wait. Centralizing the detach request in `steer_user_message` preserves the newer queue contract and fixes every real steer entry point.

The App and Engine share the same `ShellManager`. Setting its foreground-background request before enqueueing the steer lets the Bash wait loop return; the turn loop can then consume the steer. If shell tracking cannot accept the request, the existing lossless fallback restores the draft/queue instead of sending out of order.

## Verification

- focused UI steer tests: 2 passed
- foreground/background classification tests: 2 passed
- queued-message promotion test: 1 passed
- real foreground shell detach test: 1 passed
- TUI Clippy with `-D warnings`: passed after allowing four Rust 1.95 lint classes already present in untouched upstream code
- full TUI binary suite before the final classification-only boundary test: 9,216 passed, 4 ignored, 4 failed; the three parallel shared-state/timing failures all passed immediately in isolation, while the documented Windows Job Object baseline failure reproduced in isolation and its failing test/implementation are unchanged
- `git diff --check`

## Self-review

Reviewed the hook, pause-state rollback, queue fallback, Ctrl+B, detached-job, and engine-consumption paths. Message-submit denial happens before detach, normal busy Enter still queues, shell-manager errors fail closed, existing Ctrl+B status text is preserved, and detached jobs with a `shell_task_id` no longer masquerade as foreground waits.

### DIFF
--- a/crates/tui/src/tools/shell.rs
+++ b/crates/tui/src/tools/shell.rs
@@ -1146,6 +1146,11 @@ impl ShellManager {
         self.foreground_background_requested = true;
     }
 
+    #[cfg(test)]
+    pub(crate) fn foreground_background_requested_for_test(&self) -> bool {
+        self.foreground_background_requested
+    }
+
     fn clear_foreground_background_request(&mut self) {
         self.foreground_background_requested = false;
     }
--- a/crates/tui/src/tui/ui.rs
+++ b/crates/tui/src/tui/ui.rs
@@ -13252,6 +13252,17 @@ async fn steer_user_message(
     }
     let message_index = app.api_messages.len();
 
+    // A foreground shell blocks the turn loop that consumes steer input.
+    // Ask the shared shell manager to detach it before enqueueing the steer so
+    // the loop can leave the foreground wait and process this message (#4930).
+    if active_foreground_shell_running(app)
+        && let Err(err) = request_active_foreground_shell_background(app)
+    {
+        restore_steer_paused_state(app, &paused_snapshot);
+        engine_handle.set_paused(paused_snapshot.paused);
+        return Err(err.context("could not move foreground shell to /jobs before steering"));
+    }
+
     if let Err(err) = engine_handle.steer(content.clone()).await {
         restore_steer_paused_state(app, &paused_snapshot);
         engine_handle.set_paused(paused_snapshot.paused);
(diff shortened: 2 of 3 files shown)


nightt5879:
@codex review

Please review current head `171e97054`. Focus on the ordering and failure-atomicity of requesting foreground Bash backgrounding before same-turn steer dispatch, the shared `ShellManager` boundary, hook/paused-state/queue fallback behavior, and whether `shell_task_id.is_none()` correctly distinguishes a foreground wait from an already detached Running job. Please flag any race that could leave a steer blocked or detach the wrong shell.

PR #5001, opened 2026-07-31
fix(tui): measure circled digits and keycaps as 2 columns everywhere

### LINKED ISSUE #4479: BUG: TUI rendering glitch — missing / extra spaces in text, recovers on mouse selection

SparkofSpike:
## Summary

Fix the intermittent TUI rendering glitch (missing characters / phantom spaces) caused by Enclosed Alphanumerics (circled digits ① ② Ⓐ), Dingbat circled digits (❶ ❷), and keycap sequences (1️⃣) being measured as **1 column** while CJK terminals render them as **2 columns**.

The v0.9.1 grapheme-safe rewrite (#4510, commit `9be600bf4`) fixed CodeWhale's *own* width helpers but left Ratatui's internal cell accounting untouched. Since Ratatui measures `Line`/cell placement through its own `unicode-width` dependency — whose `width()` reports ambiguous-width characters as 1 column — every circled digit or keycap shifted all downstream cells by one, corrupting the diff engine's buffer. Mouse selection triggered a full repaint that masked the fault until the next occurrence.

This PR fixes the width accounting at **all three layers** so CodeWhale's measurement, Ratatui's cell placement, and the terminal's actual painting finally agree.

## Root Cause

`unicode-width 0.2.2` exposes two width tables:

| API | Table | `①` (U+2460) |
|-----|-------|:---:|
| `UnicodeWidthChar::width()` | non-CJK | **1** ❌ |
| `UnicodeWidthChar::width_cjk()` | CJK (only when `cjk` feature) | 2 ✅ |

Ratatui calls `width()` (the non-CJK table). CodeWhale's own helpers were patched in v0.9.1, but Ratatui kept placing `①` in a single cell while the terminal painted two — the cell offset that manifests as dropped glyphs and phantom spaces.

A blanket switch to the CJK table was rejected because it reclassifies box-drawing glyphs (`─` U+2500), ellipsis (`…` U+2026), arrows, and math symbols as 2 columns too, which broke 3 truncation/slicing tests and would wreck table borders.

## Changes

- **New `[patch.crates-io]` override + vendored `patches/unicode-width-0.2.2/`**: scoped patch making `UnicodeWidthChar::width()` / `UnicodeWidthStr::width()` report **2 columns** for exactly:
  - Enclosed Alphanumerics (U+2460–U+24FF)
  - Dingbat Circled Digits (U+2776–U+2793)
  - Circled Numbers on Black Square (U+3248–U+324F)
  
  Every other character keeps its non-CJK width, so box-drawing glyphs, ellipsis, arrows, and math symbols do **not** regress. The patch preserves the original `width_cjk()` API untouched.

- **`crates/tui/src/tui/ui_text.rs` — `char_display_width`**: explicit range match returning 2 for the same three ranges. Kept as defense-in-depth so CodeWhale's truncation/slicing/selection math stays correct even if the dependency patch is ever removed.

- **`crates/tui/src/tui/ui_text.rs` — `grapheme_display_width`**: force 2 columns for any multi-codepoint grapheme containing U+20E3 (keycap sequences **without** an FE0F variation selector, e.g. `1\u{20e3}`), which unicode-width also undercounts as 1.

- **`crates/tui/src/tui/markdown_render.rs` — `markdown_grapheme_width`**: the same two overrides for the markdown wrapping path (circled ranges on single codepoints; U+20E3 on multi-codepoint graphemes).

## Why patch the dependency instead of more CodeWhale-side overrides?

The previous fix proved that CodeWhale-side width helpers cannot influence how Ratatui distributes text across cells — Ratatui measures internally through `unicode-width`. Aligning Ratatui with terminal reality for these characters therefore requires the measurement primitive itself to report the truth. The vendored patch is minimal (two `impl` bodies plus a scoped `match`), preserves all other behavior, and is pinned via `[patch.crates-io]` so the whole workspace — including `ratatui-core` — picks it up.

> **Note**: this vendored-patch approach is admittedly heavy and verbose — it ships the entire `unicode-width` source into this repository just to override two function bodies. Treat it as a **temporary fix**. I will open a PR against the upstream `unicode-width` repository proposing the same scoped width correction for Enclosed Alphanumerics and related ranges. Once upstream adopts it and publishes a release, I will follow up with a second PR here that **removes this patch and bumps the `unicode-width` dependency** to the fixed version, leaving only the minimal business-logic overrides in `ui_text.rs` / `markdown_render.rs`.

## Verification

- `cargo test -p codewhale-tui --bin codewhale-tui -- ui_text::` → **26/26 pass**
  - new `unicode_width_reports_circled_digits_as_two_columns` asserts the raw `unicode-width` API now reports `Some(2)` for `①` (proves the patch is live in the render path)
  - new `circled_digit_display_width` covers `① ② ⓪ ❶`, `"①②③"`, and mixed rows
  - existing keycap suite updated: `"1\u{20e3}"` (no FE0F) now correctly measures 2 columns
- `cargo test -p codewhale-tui --bin codewhale-tui -- markdown_render::` → **52/52 pass** (box-drawing table borders and ellipsis truncation unaffected)
- Keycap regression suite → **4/4 pass** (selection keeps keycap grapheme intact; truncation stays within budget)
- `cargo build --release -p codewhale-cli -p codewhale-tui` → succeeds

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Tests

## Testing

- [x] `cargo test -p codewhale-tui --bin codewhale-tui -- ui_text::` (26/26)
- [x] `cargo test -p codewhale-tui --bin codewhale-tui -- markdown_render::` (52/52)
- [x] `cargo test -p codewhale-tui --bin codewhale-tui -- keycap` (4/4)
- [x] `cargo build --release -p codewhale-cli -p codewhale-tui`
- [ ] `cargo clippy --workspace --all-targets --all-features --locked`

## Checklist

- [x] Updated docs or comments as needed (all three width helpers documented with #4479 references)

## Related Issues

Closes #4479 — the v0.9.1 grapheme-safe fix (#4510) was merged but the reporter confirmed the circled-digit/keycap rendering was still broken in v0.9.1 and v0.9.2; this PR completes the fix at the render layer.

## Notice

🤖 Desc generated by Deepseek


SparkofSpike:
Update: I came back from the unicode-width repo — it looks like they don't plan to fix this, because they need to keep the crate's compatibility guarantees. I honestly didn't consider that angle before sending it over there, and I feel bad for having bothered them.

Hi @Hmbown, PTAL. What's your take on this issue? If you have a moment, I'd love to discuss it. To be honest, this 30k-line commit is pretty scary, and it bloats the repo. Do you have any better suggestions?

PR #5031, opened 2026-08-01
Refresh MiniMax M3 pricing

octo-patch:
Reason: Refresh MiniMax M3 pricing to the current flat standard rate.

Updated the MiniMax M3 runtime pricing path so metadata lookups and usage-based estimates agree on the same USD rates.
Adjusted the MiniMax pricing tests to cover the flat rate and removed the old 512K tier split expectations.
Kept the bundled catalog row unpriced and updated its test comment to match the runtime-owned pricing source.

Checks:
- cargo test -p codewhale-tui minimax_m3 -- --nocapture
- cargo test -p codewhale-config bundled_asset_pricing_is_honest -- --nocapture
- cargo fmt --all --check
- secret scan on unstaged diff: clear

### DIFF
--- a/crates/config/src/catalog/tests.rs
+++ b/crates/config/src/catalog/tests.rs
@@ -721,8 +721,8 @@ fn bundled_asset_pricing_is_honest() {
     assert_eq!(cost.output, Some(4.40));
     assert_eq!(cost.cache_read, Some(0.26));
 
-    // M3 has input-length and service tiers that the flat catalog cost shape
-    // cannot represent, so the bundled route row stays honestly unpriced.
+    // M3 pricing is supplied by the runtime table, so the bundled catalog row
+    // stays honestly unpriced here.
     let minimax_m3 = find(&rows, "minimax-anthropic", "MiniMax-M3");
     assert!(minimax_m3.cost.is_none());
 
--- a/crates/tui/src/pricing.rs
+++ b/crates/tui/src/pricing.rs
@@ -600,9 +600,9 @@ fn known_pricing_for_model(model_lower: &str) -> Option<ModelPricing> {
         // Moonshot K2.7 Code cache-read rate per
         // https://platform.kimi.ai/docs/pricing/chat-k27-code
         "moonshotai/kimi-k2.7-code" | "kimi-k2.7-code" => Some(usd_only_pricing(0.19, 0.95, 4.00)),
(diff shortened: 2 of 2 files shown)


Hmbown:
> Reason: Refresh MiniMax M3 pricing to the current flat standard rate.
> 
> Updated the MiniMax M3 runtime pricing path so metadata lookups and usage-based estimates agree on the same USD rates.
> Adjusted the MiniMax pricing tests to cover the flat rate and removed the old 512K tier split expectations.
> Kept the bundled catalog row unpriced and updated its test comment to match the runtime-owned pricing source.
> 
> Checks:
> - cargo test -p codewhale-tui minimax_m3 -- --nocapture
> - cargo test -p codewhale-config bundled_asset_pricing_is_honest -- --nocapture
> - cargo fmt --all --check
> - secret scan on unstaged diff: clear
> 

Thanks so much for this - super helpful!

Hmbown:
Thanks again for the careful pricing update and tests. I checked the current MiniMax-owned pricing surface before integrating this against today's main, and the flat-rate premise does not match the live M3 contract.\n\nMiniMax currently documents two input-length tiers:\n- context up to 512K: cache read /bin/zsh.06/M, input /bin/zsh.30/M, output .20/M\n- context above 512K: cache read /bin/zsh.12/M, input /bin/zsh.60/M, output .40/M\n\nThe official M3 announcement also explicitly says pricing depends on input length. This PR would apply the long-context tier to every M3 request, so it would overestimate ordinary <=512K requests and remove a still-live boundary.\n\nSources:\n- https://platform.minimax.io/subscribe/token-plan?tab=api-enterprise\n- https://www.minimax.io/blog/minimax-m3\n\nI’m leaving the PR open rather than asking you for mechanical rework: the existing runtime table already models the two tiers, so there is nothing useful to carry onto the current integration branch unless MiniMax changes that contract. The unrelated Windows failure on this PR was a pre-existing flaky shell-approval test; it is not evidence against your patch.

Hmbown:
Closing unmerged — thanks for the contribution, but the change would remove the 512K tiering and charge the higher tier flat for all inputs; upstream MiniMax-M3 pricing is genuinely tiered (<=512K vs >512K) and the train's implementation matches it. Merging would 2x overestimate cost for <=512K requests.

PR #5225, opened 2026-08-03
feat(acp): expose file/search/git/patch/shell tools over session/prompt

rafaelcavalheri:
## Summary

`session/prompt` in the ACP server only streamed model text — it never
executed the tool calls a model requested, so any editor or bridge
driving CodeWhale over ACP (Zed, third-party integrations such as the
community `acp-deepseek-adapter`) got a chat-only agent with no real
code-editing capability. That gap is the reason this was originally
built in a fork; documented as a known limitation by at least one
third-party ACP bridge, which worked around it by scraping
`exec --auto` stdout instead.

This wires the existing `ToolRegistry` into the ACP turn loop instead
of building a parallel implementation:

- `run_agentic_prompt_turn` drives multi-round `tool_use`/`tool_result`
  turns (capped at `MAX_ACP_TOOL_ROUNDS`) over the same file/search/git/
  patch/shell tools the TUI itself uses, and threads `response_id_policy`
  through every round so tool-round responses still get the
  client-specific id translation (Zed/avante.nvim) the existing
  streaming path already relies on.
- Shell access is gated on the client declaring `terminal` support at
  `initialize` (default `false`/restrictive). `MAX_ACP_SESSIONS` caps
  concurrent sessions with true insertion-order eviction (`VecDeque`,
  not `HashMap` iteration order, which is not insertion-ordered).
- Tool-call cancellation signals a `CancellationToken` and waits for the
  running tool (including a child shell process) to actually stop
  before returning the cancellation, instead of abandoning it.
- `max_tokens` for the ACP path now resolves through the same
  route-limits machinery the TUI/CLI use
  (`effective_max_output_tokens_for_route`) instead of a fixed 4096
  fallback.
- `scripts/build.ps1`: a release build script for Windows PowerShell
  5.1, used to produce and test the ACP binary against Zed on Windows.

Drafted with agent assistance (Claude Code); reviewed and
build-verified by the human author before submission.

## Testing

- [x] `cargo clippy --workspace --all-features --locked` (warning-free
      under the CI allow list)
- [x] `cargo test --workspace --all-features --locked` — 9353 passed;
      the 43 failures on this run are pre-existing and unrelated to this
      diff: `repl::runtime`/`tools::rlm`/`core::engine` Python-backed
      tests fail because this machine has no Python 3 on `PATH`
      (`no Python interpreter found on PATH`), and a handful of Windows
      job-kill/pipe-timing tests are flaky under full-parallel `cargo
      test --workspace` (confirmed by re-running one in isolation, which
      passed the timing/kill logic and only hit the same missing-Python
      wall). None touch `acp_server.rs`, `main.rs`, `core/engine.rs`,
      `core/engine/context.rs`, or `route_budget.rs`.
- [x] Targeted: `cargo test -p codewhale-tui --bin codewhale-tui acp_server`
      → 34/34 passed; `route_budget` → 11/11 passed.

## Checklist

- [ ] Verified TUI behavior manually if UI changes (no TUI/UI surface changed; verified via the ACP unit tests and a Windows build against Zed)
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address (n/a — original submission, no harvested commits)

### DIFF
--- a/crates/tui/src/acp_server.rs
+++ b/crates/tui/src/acp_server.rs
@@ -1,9 +1,8 @@
 //! Minimal Agent Client Protocol stdio adapter.
 //!
-//! This intentionally starts with the ACP baseline: initialize, new session,
-//! prompt, and cancel. It keeps stdout protocol-clean for editor clients and
-//! routes prompts through the same configured DeepSeek client as one-shot CLI
-//! mode.
+//! This starts from the ACP baseline: initialize, new session, prompt, and
+//! cancel. It keeps stdout protocol-clean for editor clients and routes
+//! prompts through the same configured DeepSeek client as one-shot CLI mode.
 //!
 //! `session/prompt` streams the provider response: each text delta is emitted
 //! as a `session/update` agent_message_chunk as it arrives, instead of buffering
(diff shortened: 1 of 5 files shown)


Hmbown:
Read this closely. The shape is right and I want it — but I'm not merging it as-is. Three blockers, all of which I verified against the tree rather than inferred from the diff.

First, what's right, because it's the part that's hard to get right: you wired the **existing** `ToolRegistry` in rather than building a parallel ACP tool implementation. `ToolRegistryBuilder::new().with_file_tools().with_search_tools().with_git_tools().with_patch_tools()` is exactly the call I'd have wanted, and it means ACP inherits every future tool fix instead of drifting. Gating `with_shell_tools()` and `ShellPolicy` on the client's declared `clientCapabilities.terminal` is also correct, and the `ShellPolicy::None` path degrades the way it should.

### Blocker 1 — `context.auto_approve = true` turns off the shell safety gate

In `build_acp_tool_registry`:

```rust
context.auto_approve = true;
```

That single line is read at `crates/tui/src/tools/shell.rs:3104`:

```rust
if !context.auto_approve {
    match safety.level {
        SafetyLevel::Dangerous => { /* ... */ }
```

So the `SafetyLevel::Dangerous` arm never runs. Every command any ACP client's model emits executes unreviewed, on a surface with no `session/request_permission` round-trip yet — the only mention of it in this PR is a doc comment.

Your stated reason is that gating on the heuristic "would just make some tool calls silently fail." That's answerable, and I think it's the crux. A blocked command is not a silent failure. It returns:

```rust
return Ok(ToolResult {
    content: format!("BLOCKED: This command was blocked for safety reasons.\n\nReasons: {reasons}{suggestions}..."),
    success: false,
    metadata: Some(json!({ "safety_level": "dangerous", "blocked": true, ... })),
});
```

`Ok`, not `Err`. It's a normal tool result with `success: false` and a human-readable reason string — and your own `execute_tool_calls_with_cancellation` already feeds tool results back to the model as `tool_result_message(...)`. The model gets told what was blocked and why, and can adapt. That's the loud path, not the silent one.

The sibling adapter agrees: `crates/tui/src/mcp_server.rs:110` does a bare `ToolContext::new(workspace.clone())`, and `ToolContext::new` defaults `auto_approve: false` (`crates/tui/src/tools/spec.rs:613`). ACP is the same trust posture over a different transport and should match it. Drop the line and let the default stand; if the heuristic turns out to be too aggressive for real ACP clients, that's a separate PR against the heuristic with evidence, not a blanket bypass here.

### Blocker 2 — the cancel test passes about half the time

`agentic_turn_cancels_while_a_tool_is_running` scripts a tool call named `exec_shell`. That tool doesn't exist. It was removed in v0.9.3 — `crates/tui/src/tools/registry.rs:610` says so in a comment, and line 2239 asserts it:

```rust
assert!(!registry.contains("exec_shell"), "retired exec_shell must remain absent");
```

Lookup is an exact `HashMap` hit with no fuzzy fallback (`registry.rs:69-71`), so `execute_full_with_context` bails at line 113 with `ToolError::not_available` before touching a subprocess. Your `SLOW_SHELL_COMMAND` never runs.

That means your `select!` isn't racing a five-second sleep against a cancel line — it's racing two futures that are both immediately ready: `exec_fut` (already an error) and `reader.next_line()` (the cancel line is already buffered by `lines_from`). `tokio::select!` picks a ready branch pseudo-randomly, so this asserts `PromptOutcome::Cancelled` roughly half the time. It's green on your machine and it will flake in CI.

The fix is a rename. `with_shell_tools()` registers `BashTool::new("Bash")` (`registry.rs:614`), and `action` defaults to `"run"` when absent (`shell.rs:3025-3030`), so `"Bash"` with your existing `{ "command": SLOW_SHELL_COMMAND }` payload works unchanged. Then the sleep genuinely runs, the cancel genuinely wins, and the test means what its name says.

Worth grepping the rest of the PR for `exec_shell` while you're in there — the doc comments reference it too, and it's a retired spelling everywhere now.

### Blocker 3 — `build_system_prompt` is ours to carry

Your `session/prompt` path calls `crate::prompts::build_system_prompt(...)`. That function no longer exists; we deleted it as dead code in `a98b184f5` ("refactor(tui): delete the dead prompt compatibility layer"). It was dead *because* nothing called it — and your PR is the caller that would have made it live.

This one is not yours to fix and I'm not asking you to. We removed a function while a contributor was building the thing that needed it, and re-landing it correctly against the current prompt surface is our job. I'll sort it on our side; don't chase it.

### Next

Retargeting this to `agent/v094-release-train-20260802`, where the rest of the v0.9.4 queue is staging. `main` isn't where this can land right now.

Blockers 1 and 2 are yours and both are small — one line deleted, one string renamed. Blocker 3 is mine. Ping me when 1 and 2 are pushed and I'll re-review.

rafaelcavalheri:
Thanks for the detailed review!

I've addressed blockers 1 and 2:

- Removed `context.auto_approve = true`, so ACP now uses the default shell safety behavior (`auto_approve = false`), matching the MCP server.
- Updated the cancellation test (and remaining references) to use `Bash` instead of the retired `exec_shell`. The test now exercises the actual long-running shell command and cancellation path.

I intentionally left blocker 3 untouched, as discussed.

Thanks! Looking forward to your re-review.

PR #5233, opened 2026-08-03
fix(modelstudio): surface reasoning on official chat routes

### LINKED ISSUE #5203: v0.9.4: Model Studio Token Plan reasoning never surfaces in Thinking — enable_thinking + stream path missing
## Summary

On **Alibaba Cloud Model Studio Token Plan** (`modelstudio-token-plan`, models like `qwen3.8-max`), reasoning does **not** show up in Codewhale's Thinking / reasoning UI even though:

1. The Token Plan catalog marks these models as reasoning (`reasoning: true`, `reasoning_options: always_on`).
2. Alibaba's OpenAI-compatible Chat Completions API documents thinking via **`enable_thinking`** with deltas in **`reasoning_content`**.
3. Our Chat Completions SSE parser already knows how to map `delta.reasoning_content` → Thinking blocks — but only when the route is classified as a "separate field" reasoning stream.

This is a **harness integration bug**, not "the model doesn't reason."

## Evidence from Alibaba docs

- [OpenAI compatible - Chat](https://www.alibabacloud.com/help/en/model-studio/qwen-api-via-openai-chat-completions): non-standard top-level **`enable_thinking`** (HTTP body; SDK `extra_body`). When enabled, thinking is returned in **`reasoning_content`**. Applies to Qwen3.x hybrid models, DeepSeek-V4, Kimi, GLM, etc. DeepSeek-V4 thinking default-on; intensity via **`reasoning_effort`**.
- [Deep thinking](https://www.alibabacloud.com/help/en/model-studio/deep-thinking): hybrid thinking mode toggled with `enable_thinking`.
- Multi-turn / agentic: **`preserve_thinking`** recommended so prior `reasoning_content` is kept in history (supported on qwen3.7/3.6/3.8-class routes).
- Token Plan Personal lists `qwen3.8-max` / `qwen3.8-max-preview` as **Reasoning + visual understanding + text**.
- Public writeups on Qwen3.8 Token Plan integrations: thinking **always on**, effort tiers **low / high / xhigh** (default often xhigh) — i.e. not "no API surface."

### 1. Request path: Model Studio is a deliberate no-op in `apply_reasoning_effort`

`crates/tui/src/client.rs` — for **every** effort tier (`off`, low/medium/high, max/xhigh), Model Studio arms are empty:

```text
// Model Studio: same reasoning as TelecomJS — the Chat Completions
// API does not expose reasoning_effort/thinking controls. Reasoning
// is model-specific on the DashScope side and not surfaced here.
ApiProvider::ModelstudioTokenPlan | ... => {}
```

That comment is **wrong** relative to current Alibaba OpenAI-compatible docs. We never send:

- `enable_thinking: true|false`
- `reasoning_effort` where applicable (DeepSeek-V4 on Model Studio, etc.)
- `preserve_thinking` for multi-turn agent runs

Compare working peers:

- Xiaomi / Moonshot / Z.ai: send `thinking: { type: enabled }`
- DeepSeek native / SiliconFlow / etc.: `thinking` + `reasoning_effort`
- vLLM: `chat_template_kwargs.enable_thinking`

### 2. Stream path: Model Studio is not allowed to treat `reasoning_content` as Thinking

`provider_accepts_reasoning_content` in `crates/tui/src/client/chat.rs` **omits all Modelstudio variants**:

```text
Deepseek | NvidiaNim | Openrouter | XiaomiMimo | Novita | Fireworks |
Siliconflow* | Volcengine | Arcee | Minimax | Sglang | Zai | Moonshot
// ← no ModelstudioTokenPlan / CodingPlan / Anthropic twins
```

Stream style is:

```text
provider_accepts_reasoning_content(provider) && model_supports_reasoning(model)
  → SeparateField  // maps delta.reasoning_content → Thinking UI
  else → None      // reasoning not shown as reasoning
```

So even if the endpoint returned `reasoning_content` (default-on models), Codewhale would **not** classify the route as SeparateField and the Thinking surface stays empty / wrong.

### 3. Static reasoning model table incomplete for Model Studio wire ids

`model_supports_reasoning` lists many `qwen/qwen3.6-*` OpenRouter-style ids but **not** bare Token Plan ids `qwen3.8-max` / `qwen3.8-max-preview` / `qwen3.7-*`. Fallback is only `model_catalog::resolved_supports_reasoning`. Catalog rows *do* say `reasoning: true`, but if the offline catalog merge misses the provider-scoped row (or identity is only wire-id without provider), support flips to false and compounds (2).

### 4. No Model Studio–specific route shaper

There is no `apply_modelstudio_route_reasoning_controls` analogous to Z.ai / MiniMax / Kimi. Needed behaviors:

| Control | Docs | Ours |
|---|---|---|
| `enable_thinking` | primary on/off for hybrid Qwen | never sent |
| `reasoning_effort` | low/high/xhigh (and DeepSeek-V4 intensity) | never sent on this provider |
| `preserve_thinking` | multi-turn agentic | never sent |
| Stream `reasoning_content` → Thinking UI | documented | blocked by (2) |
| Replay `reasoning_content` on tool turns | often required for thinking models | not in Model Studio accept list |

### 5. Related correctness gaps (same integration, not only Thinking UI)

Worth fixing in the same PR / follow-ups:

1. **Ctrl-T / effort cycle** — even if UI shows max/high, Model Studio arms ignore it; chip is cosmetic.
2. **Anthropic dialect (`wire=anthropic` / legacy `*-anthropic` kinds)** — Messages adapter may need Model Studio thinking/budget mapping verified separately; Chat Completions path is the primary Token Plan dogfood route today.
3. **Token Plan endpoint family** — we default `https://token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1`. Docs also push workspace-scoped hosts `{WorkspaceId}.ap-southeast-1.maas.aliyuncs.com/...`. Confirm Token Plan keys still work on both; document if workspace host is required for some tiers.
4. **Vision** — catalog marks image input on qwen3.8-max; verify multimodal parts actually go out on Token Plan (separate from reasoning but same provider slice).
5. **Catalog honesty** — hand-seam offerings mark reasoning Supported, but request/stream layers disagree (capability vs behavior lie).

## Done when

- [ ] On `modelstudio-token-plan` + `qwen3.8-max` (and preview), with default / high / xhigh effort, streamed thinking appears in the **Thinking** cell (not dumped into answer prose).
- [ ] Requests include top-level `enable_thinking` true for hybrid models when effort is not off; false when off (if the model allows off; always-on models document that).
- [ ] `provider_accepts_reasoning_content` includes Model Studio (or exact-route helper), so SeparateField stream style activates.
- [ ] `qwen3.8-max` / `qwen3.8-max-preview` (and other Token Plan text models) are in the static reasoning table **or** catalog resolution is proven provider-scoped and always hits.
- [ ] Multi-turn tool use: `preserve_thinking` (or equivalent) when docs recommend it; no silent 400s from missing thinking history if the API requires it.
- [ ] Pin tests: body builder sends `enable_thinking`; SSE fixture with `delta.reasoning_content` produces Thinking events for ModelstudioTokenPlan; effort off path documented.
- [ ] Dogfood receipt: one real Token Plan turn with visible Thinking in TUI.

## Fix direction (implementation notes)

1. Add `apply_modelstudio_route_reasoning_controls` (exact Token Plan / Coding Plan host guard, fail-closed on custom gateways).
2. Map Codewhale effort → `enable_thinking` + optional `reasoning_effort` / budget per model family (Qwen hybrid vs always-on vs DeepSeek-V4).
3. Extend `provider_accepts_reasoning_content` + static `model_supports_reasoning` for Token Plan wire ids.
4. Consider replaying `reasoning_content` on assistant tool turns for Model Studio the way we do for DeepSeek when required.
5. Do **not** invent four catalog rows again — keep one vendor identity + `mode`/`wire` from `d53f4f998`.

## Refs

- In-repo: `crates/tui/src/client.rs` (`apply_reasoning_effort` Modelstudio arms), `crates/tui/src/client/chat.rs` (`provider_accepts_reasoning_content`, `reasoning_stream_style_for_route`), catalog `crates/config/assets/models_dev.bundled.json` (`modelstudio-token-plan` / `qwen3.8-max`).
- Alibaba: OpenAI-compatible Chat + deep thinking docs (links above).
- Related landed work: #5175 / #5202 / `d53f4f998` (provider identity + 1M context) — **does not fix reasoning surface**.

Inference1:
## Summary

- Classify `reasoning_content` as a dedicated Thinking stream only on verified Alibaba Model Studio OpenAI-compatible Chat Completions routes.
- Shape documented top-level Model Studio controls by model capability: `enable_thinking`, `preserve_thinking`, and DeepSeek-V4/GLM `reasoning_effort`; unknown/custom gateways fail closed.
- Replay supported reasoning traces across turns, including thinking-only Qwen 3.8 and Kimi K2.7 Code; cover Token Plan, workspace-scoped, Coding Plan, and picker `mode = "coding-plan"` resolution.
- Add a credential-free verification record, safe configuration examples, official protocol references, two local Token Plan Lite screenshots, and a clearly-labelled GIF replay of consecutive PTY/TUI capture states.

Closes #5203.

## Evidence

- Live local Token Plan Lite (`qwen3.8-max`) dogfood showed a completed Thinking cell, a separate final answer, and completion status.
- The GIF is a post-processed replay of raw local PTY/TUI capture states, not a fresh screenshot of a macOS Terminal.app window. It shows `reasoning live` receiving text incrementally; the final answer appears only after `reasoning done`.
- Official Alibaba docs confirm `/chat/completions`, top-level HTTP fields, and streamed `choices[].delta.reasoning_content`.
- Coding Plan has route/config contract coverage; no Coding Plan subscription credential was used, so this PR does not claim a live Coding Plan call. Anthropic Messages is a separate adapter and is out of scope.

Token Plan Lite streaming replay

Token Plan Lite local verification

Token Plan Lite full terminal capture

## Testing

- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] `cargo test -p codewhale-tui modelstudio_ --locked` (13 passed)
- [x] `cargo build -p codewhale-tui --locked`
- [x] `cargo test -p codewhale-tools --lib --all-features --locked` (24 passed)
- [ ] `cargo test --workspace --all-features --locked` — pre-existing `codewhale-tools::tests::required_str_reports_wrong_type_when_field_exists` assumes JSON object key order; it fails only under workspace feature unification with `serde_json/preserve_order`. This PR does not touch that crate or dependency manifest.
- [ ] `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — blocked by pre-existing `unused_mut` in `crates/tui/src/native_memory.rs:430` (and an existing dead-code warning); neither is changed here.

## Checklist

- [x] Updated docs and comments as needed
- [x] Added targeted regression tests
- [x] Verified TUI behavior manually
- [x] Commit authored with Inference1 GitHub noreply identity


Hmbown:
Harvested. Thank you — this one found a real bug in our code, not just a gap.

**Why it couldn't be merged as-is.** You authored at 16:34Z. A parallel local fix for #5203 (`11da42342`) landed at 17:08Z on the integration branch, which you had no way to fetch. It rewrote the exact `=> {}` arms in `client.rs` that your three hunks target, so the patch no longer applies. Your delta was carried across by hand instead.

**Taken as yours:**

1. **The fail-closed host guard** — `is_exact_modelstudio_chat_route` + `apply_modelstudio_route_reasoning_controls`, composed the way you had it: `apply_reasoning_effort` runs first, the shaper then strips `thinking` / `enable_thinking` / `preserve_thinking` / `reasoning_effort` unconditionally and re-adds only on a verified host. This was a genuine gap in ours — we gated on the provider enum alone, so a custom `base_url` on `modelstudio-token-plan` got Alibaba's `enable_thinking` injected into an arbitrary gateway.

2. **The thinking-only suppression for `qwen3.8-max` / `-preview`.** Ours was sending `enable_thinking: false` to an always-thinking model, and our own test at `client.rs:~4307` *pinned* that with `assert_eq!(off["enable_thinking"], json!(false))`. Our catalog corroborates you: `models_dev.bundled.json` marks `qwen3.8-max` as `thinking: always_on`, and gives `qwen3.8-max-preview` effort/budget options with no `toggle`. The assertion now asserts the switch is absent for every effort level.

3. **`preserve_thinking` and the DeepSeek-V4 / GLM `high`/`max` ladder.**

4. **Your `config/tests.rs` test, verbatim** — the `mode = "coding-plan"` resolution seam. That file only moved for unrelated GLM-5.3 work since your base, so it applied cleanly.

**Taken further than you had it.** We deleted the provider-enum writes in `apply_reasoning_effort` outright rather than leaving a second writer, and widened the shaper's guard from the two OpenAI-dialect identities to all four `Modelstudio*` variants. Your version left `ModelstudioTokenPlanAnthropic` and `ModelstudioCodingPlanAnthropic` writing `enable_thinking` with nothing to strip it — harmless today because the Messages adapter owns them, live the moment someone sets `wire = "openai"`. Only the two OpenAI-dialect identities can pass the host check, so the strip covers all four and the re-add covers two.

**Narrowed.** The `reasoning_content` replay widening was taken only for models that also get `preserve_thinking`, plus the thinking-only pair. `deepseek-v3.1`, `deepseek-v3.2` and `glm-*` stay stripped until someone with a Model Studio key can confirm DashScope does not 400 on `reasoning_content` in input messages. This removes nothing that exists — `deepseek-v4*` already replays via `requires_reasoning_content` on both sides. If your Token Plan Lite key covers a `glm-5.2` or `deepseek-v3.x` multi-turn tool call, that evidence would let us widen it, and it's the one thing this harvest is missing.

We also kept stream *classification* as broad as it was rather than route-scoping it. Fail-closed is about what we send, not what we parse; a gateway emitting `delta.reasoning_content` is speaking the dialect by definition, and narrowing the decoder would push thinking back into answer prose for custom-gateway users.

**Two claims carried but flagged, not silently absorbed:**

- `kimi-k2.7-code` as thinking-only + `preserve_thinking` is asserted here, not corroborated by our catalog, which lists it under `moonshot` with `reasoning: true` and no `always_on`. It's in with a comment saying so. Do you have a Model Studio response that shows it rejecting `enable_thinking`?
- `glm-5.2` is treated as hybrid (so it gets `enable_thinking`), but our catalog gives it `effort` options and **no** `toggle`. If DashScope rejects the switch there, that arm is wrong — same question.

**Dropped:** the `qwen3-omni-flash` and `qwen3-vl` hybrid arms (unreachable behind `starts_with("qwen3-")` two lines above); your decoder test, which our existing `modelstudio_streams_reasoning_content_as_thinking` already covers end to end including the trailing usage chunk; and `docs/evidence/modelstudio-reasoning-5203.md` with its screenshots, since the record describes the un-narrowed replay behavior — this PR stays as the provenance for it.

Gate on the harvest: `cargo test -p codewhale-tui --bin codewhale-tui` — **9812 passed, 0 failed, 9 ignored**, exit 0. Run in full because both lanes had edited the same two functions.

Credit is on the commit (`Co-authored-by` + `Harvested from PR #5233`), and `docs/CONTRIBUTORS.md` and the changelog are updated. Leaving this open so the auto-close workflow can close it with the landed SHA once the work reaches `main`.

Hmbown:
Retargeted from `agent/v094-release-train-20260802` to `main`. That train branch is scheduled for a history rewrite, and a PR based on a rewritten or deleted ref gets auto-closed by GitHub with no credit context. `main` is the stable base and the branch the auto-close workflow watches.

No merge here: the harvest is what lands. Leaving the merge decision itself to the operator.

Hmbown:
You got there first, and I want that on the record before anything else.

Your commit `6bb33f355` is authored `2026-08-03T16:34:44Z`. Our local fix for the same issue, `11da42342`, is authored `2026-08-03T17:08:45Z` — 34 minutes behind you. I did not know this PR existed when that was written. The independent-discovery direction points at you.

### Why this got harvested by hand instead of merged

Your `crates/tui/src/client.rs` hunks are comment edits on the three

```rust
ApiProvider::ModelstudioTokenPlan
| ApiProvider::ModelstudioTokenPlanAnthropic
| ApiProvider::ModelstudioCodingPlan
| ApiProvider::ModelstudioCodingPlanAnthropic => {}
```

arms in `apply_reasoning_effort`. Thirty-four minutes after you wrote them, `11da42342` replaced those empty arms with `body["enable_thinking"] = json!(false)` / `json!(true)` writes. GitHub already reports this PR `CONFLICTING` for that reason.

The rebase doesn't fix it, because `11da42342` is not on `main`. It lives only on our v0.9.4 integration lane, which is scheduled for a history rewrite before it reaches `main`. Rebasing you onto it would land your work on top of commits whose hashes are about to stop existing. So I took your delta by hand, with your `Co-authored-by` trailer on it, rather than leave you rebasing against a moving target.

### What went in

Committed as `4f1d2b7a5`, "fix(modelstudio): fail closed on the host, stop lying about qwen3.8", trailered to you, plus `f6f976569` for the contributor record.

Three things your patch had that ours did not:

1. **The host guard.** We gated on the provider enum alone. A custom `base_url` on `modelstudio-token-plan` got Alibaba's `enable_thinking` injected into an arbitrary gateway. `is_exact_modelstudio_chat_route` now verifies the URL shape, and `apply_modelstudio_route_reasoning_controls` strips `thinking` / `enable_thinking` / `preserve_thinking` / `reasoning_effort` unconditionally and re-adds only on a verified host. Fail-closed, as you had it.
2. **The thinking-only pair.** Below.
3. **`preserve_thinking`**, and the documented `high`/`max` ladder for `deepseek-v4*` and `glm-5`/`5.1`/`5.2`.

The provider-enum writes in `apply_reasoning_effort` are deleted outright rather than left as a second writer.

### The live bug you caught in our code

`qwen3.8-max` and `qwen3.8-max-preview` are thinking-only, and we were sending them `enable_thinking: false`.

The part worth naming is that our own test *asserted* the defect:

```rust
assert_eq!(off["enable_thinking"], json!(false), "{base_url}: {off}");
```

That runs against `DEFAULT_MODELSTUDIO_TOKEN_PLAN_MODEL`, which is `"qwen3.8-max"` (`crates/config/src/provider_defaults.rs:175`). And our own bundled catalog, in all four `modelstudio-*` provider blocks, says:

```json
"reasoning_options": [
  { "type": "thinking", "values": ["always_on"], "default": "always_on" }
]
```

So we shipped a test that pinned "send the off switch" against a model our own catalog marks `always_on`. A green suite was actively defending the bug. That assertion now asserts the switch is *absent* for every effort level. Thanks for making us look.

### The one thing I narrowed

`reasoning_content` replay. Your patch widened it to `deepseek-v3.1`, `deepseek-v3.2` and `glm-*`; I took it only for the models that also get `preserve_thinking`, plus the thinking-only pair. The rest stay stripped.

Not a disagreement with your reading — I have no Model Studio key, and replaying `reasoning_content` in *input* messages is a different risk from parsing it out of a response. If DashScope 400s on it for those families, the narrow version degrades (thinking not carried across turns) while the wide version breaks the request outright. I'd rather ship the failure mode I can reason about.

This removes nothing that exists today: `deepseek-v4*` already replays via `requires_reasoning_content` on both sides. The narrowing is pinned by `modelstudio_replay_stays_narrow_until_a_live_key_confirms_it` in `crates/tui/src/client/chat.rs`.

**Exactly what would let me widen it:** one live run against `dashscope.aliyuncs.com/compatible-mode/v1` with `deepseek-v3.1` or any `glm-*`, sending a second turn whose assistant message carries `reasoning_content`, showing a 200 rather than a 400. A raw request/response pair is enough; I don't need a key of my own, just evidence the gateway accepts the field on input. Post it here and I'll widen it and credit you again.

One related call I want to be transparent about: I did **not** take your narrowing of `is_reasoning_model_for_stream_on_route` or the `route_accepts_reasoning_content` helper. Fail-closed is about what we send, not what we parse — a gateway emitting `delta.reasoning_content` is speaking the DashScope dialect by definition, and route-scoping the decoder would push thinking back into answer prose for custom-gateway users. That left `modelstudio_model_supports_reasoning` with no callers, so it wasn't added.

`docs/evidence/modelstudio-reasoning-5203.md` and the two screenshots aren't carried either — they document the un-narrowed replay behavior and a live run I can't verify. This PR stays as the provenance for that record.

### Two things I noticed reading closely

Observations, not corrections — neither changes the outcome:

- The `qwen3-omni-flash` and `qwen3-vl` arms you added to `modelstudio_model_is_hybrid` are unreachable. `model.starts_with("qwen3-")` two lines above already matches both. I dropped them rather than absorb them, so the function doesn't grow arms that read as load-bearing but never fire.
- `kimi-k2.7-code` as thinking-only + `preserve_thinking` I carried on your word, but I can't corroborate it here. Our catalog lists it under `moonshot` with `reasoning: true`, no `always_on`, and it has no `modelstudio-*` row at all. The arm is in with a comment saying exactly that. If you have the Model Studio listing for it, I'll upgrade the comment to a citation.

One logistics note so you can find your code: the `client.rs` and `client/chat.rs` hunks were swept into `68d8d94b1` by a concurrent telemetry lane committing the same working tree before the harvest commit could claim them. That commit is already published and isn't mine to rewrite. Your trailer lives on `4f1d2b7a5`; the code is split across both.

Leaving this open as provenance rather than closing it.

Hmbown:
Landed — your work is on the 0.9.4 release train, and this PR shows CONFLICTING precisely because it's already there.

Why it was harvested by hand rather than merged: you committed at 16:34Z and I shipped a parallel Model Studio fix at 17:08Z into an integration branch that wasn't pushed yet, and it rewrote the exact `apply_reasoning_effort` arms your comment-only hunks target. A rebase would have landed you on code that existed on no ref you could fetch. That was my sequencing problem, not yours.

What went in, with your `Co-authored-by` trailer:

- **`is_exact_modelstudio_chat_route` + `apply_modelstudio_route_reasoning_controls`** — the fail-closed host guard. This closed a real gap in mine: I gated on the provider enum alone, so a custom `base_url` got Alibaba's `enable_thinking` injected into an arbitrary gateway. Composed as your PR does — `apply_reasoning_effort` first, then the shaper strips unconditionally and re-adds only on a verified host.
- **All three provider-enum `enable_thinking` writes deleted** from `apply_reasoning_effort`; the shaper is now the sole writer, matched across all four `Modelstudio*` variants so the strip also covers the two `*Anthropic` identities.
- **Thinking-only suppression for `qwen3.8-max`/`-preview`.** You caught a live bug in our code here, and I want to name it: `assert_modelstudio_request_truth` asserted `enable_thinking == false` against a model our own `models_dev.bundled.json` marks `always_on`. Our test was *pinning the defect*. That assertion is now rewritten to require the switch be absent for every effort tier.
- **`preserve_thinking`** and the DeepSeek-V4/GLM effort mapping.
- Your `config/tests.rs` test, verbatim.

One thing I narrowed: the `reasoning_content` replay. I took it only for models that also get `preserve_thinking` plus the thinking-only pair, and left `deepseek-v3.1`, `deepseek-v3.2` and `glm-*` stripped until someone with a live Model Studio key can confirm DashScope doesn't 400 on `reasoning_content` in input messages. This removes nothing that already worked — `deepseek-v4*` replays via `requires_reasoning_content` on both sides. If you have that key and the evidence, I'll widen it.

Two things I noticed reading closely, neither blocking: `qwen3-omni-flash` and `qwen3-vl` are dead arms behind the `starts_with("qwen3-")` check two lines above, and I couldn't corroborate the `kimi-k2.7-code` thinking-only claim against our catalog (`reasoning: true`, no `always_on`).

Closing in favour of the landed commits. Thank you — this was a genuinely better patch than mine on four separate points, three of which came straight out of the fix direction in #5203.

Hmbown:
For the record, so nothing here is ambiguous: this merged into `contrib/` rather than directly into the release train, because the train had already absorbed your changes through the hand-harvest described above and merging again would have conflicted with your own landed work.

Your code ships in 0.9.4 either way — the commit SHAs in the previous comment are the ones that actually run. This merge exists so the contribution is recorded as merged where it should have been in the first place, rather than as a closed PR. The harvest was our sequencing failure, not a judgement on the patch.

Inference1:
@Hmbown Thank you very much! Hunter

Working on CodeWhale is really cool, and it's only going to get better.

PR #5235, opened 2026-08-04
docs(evidence): replace Model Studio #5203 proof

Inference1:
## Summary

- Replaces the earlier post-processed GIF and prior terminal stills with the supplied local Terminal MP4 plus two Alibaba Cloud Model Studio Token Plan screenshots.
- Documents exactly what each credential-free artifact demonstrates: the `qwen3.8-max` reasoning-to-working transition, an active Lite subscription with recorded quota use, and the 17.6K-token provider-side usage window.
- Removes the three superseded evidence assets so the documentation and review surface use only the current proof set.

This is an evidence-only follow-up to merged PR #5233. It targets that PR's integration branch and does not alter the Model Studio implementation.

Refs #5203.

## Evidence

- [Local Terminal recording — MP4, 32.9 seconds](https://github.com/Inference1/DeepSeek-TUI/blob/docs/modelstudio-5203-live-evidence/docs/evidence/modelstudio-token-plan-live-streaming-5203.mp4)
- Token Plan Lite subscription and quota consumption
- Token Plan usage details for the validation window

The recording is a direct macOS Terminal capture. It starts in the dedicated `reasoning` state, shows the separate `… reasoning hidden` marker, then transitions to `working` while the Redis response streams before completion. The account-side screenshots corroborate the active Lite plan and the 14:00–15:00 provider usage: 17.6K total tokens (14.2K uncached input, 3.4K output). No API key or request authorization data is committed.

## Testing

- [x] `git diff --check`
- [x] `cargo test -p codewhale-tui --locked modelstudio_ -- --format terse` (13 Model Studio tests passed)

## Checklist

- [x] Updated documentation and review evidence
- [x] Re-ran the targeted regression suite
- [x] Verified the supplied recording and screenshots before committing
- [x] No co-authored trailer added


Inference1:
Superseded by #5236. The replacement has a clean evidence-only diff against contrib/pr-5233; closing this duplicate review.

PR #5319, opened 2026-08-11
fix(tui): copy messages without visual rails

### LINKED ISSUE #5314: Copy message from context menu includes rail decorations (● ▏) — make it rail-clean like selection copy
## Description

In the TUI (v0.9.5), the context-menu action **"Copy message"** (right-click on a transcript cell → _Copy message_) copies the message **including UI decorations**: the first line carries the role glyph `●`, and every wrapped continuation line is prefixed with the rail character `▏ ` (U+258F + space). Example of what lands on the clipboard:

```
● Готово. Эпопея закрыта.
▏ 
▏ Навык обновлён — урок про «живой тест после каждого патча» зафиксирован в search-files-windows-fix.
▏ 
▏ MEMORY обновлён — правило: «похожее место ≠ нужное место».
```

This makes pasting the copied text into another document/chat/editor noisy — the user has to strip `●`/`▏` manually.

## Use Case

"Copy message" is the natural way to grab a whole assistant reply for reuse outside CodeWhale. The user expects clean text, exactly like the in-app **selection copy** already provides (rail-clean since #4208 — "In-app selection copy is rail-clean and now regression-tested: copied transcript text excludes the ▎ ╎ │ ● decorations via cache metadata").

Today there is a mismatch: **selection copy is clean, "Copy message" is not**. There is also no config option to change the copy-message behavior (`[tui]` has no copy-related keys; `mouse_capture` only affects who owns mouse selection).

## Suggested Fix

Make the context-menu "Copy message" path rail-clean the same way selection copy is:

- `copy_cell_to_clipboard` currently renders via `history_cell_to_text` → `transcript_lines` → `line_to_string`, which keeps the role glyph and rail spans.
- Selection copy (`selection_to_text`) already strips them using cache metadata (`rail_prefix_width` + `copy_prefix_width`).
- Reuse that same metadata-based stripping for the cell-copy path, so "Copy message" exports only the plain message text (including code blocks, markdown source as-is, without the `●`/`▏` decorations).

An opt-in config flag (e.g. `[tui] copy_cell_plain = true`, default `true`) would be a nice escape hatch if anyone actually wants the decorated copy, but the clean text should be the default.

## Environment

- CodeWhale v0.9.5 (853cb707bbcf), Windows 11, Windows Terminal
- `mouse_capture = true` — with mouse capture off the issue is even worse (raw terminal grid copy), but that's the documented #4208 product-decision path, not this report.

XhesicaFrost:
## Summary

- Fix Copy Message for User and Assistant cells by copying canonical source content instead of rendered Ratatui lines.
- Keep Tool, Thinking, System, and other complex cells on the existing full-transcript path so complete output and detail semantics do not regress.
- Add regressions for narrow-width wrapping, authored `●`/`▏` glyphs, Markdown/code blocks, hard line breaks, and complex-cell transcript behavior.

The old path called `history_cell_to_text()`, so role glyphs, continuation rails, and visual wrapping entered the clipboard after `transcript_lines()` rendering. The new clipboard-specific helper bypasses rendering only where canonical source exists. It never matches or deletes Unicode glyphs, so identical authored characters remain intact.

The live selection cache is intentionally not reused: Markdown is already transformed there, User soft wraps lack safe join metadata, and Tool/Thinking live rendering may be folded or capped. Pager, Open Details, and file-location flows remain unchanged.

Closes #5314

## Testing

  - Blocked by pre-existing `clippy::field_reassign_with_default` errors in `crates/tui/src/config/tests.rs:5123` and `:5134`; this PR does not modify that module.
  - Result: 10,239 passed, 4 failed, 11 ignored. Failures are in untouched config, terminal-session, model-picker, and config-view tests. The terminal timeout test passed on exact rerun; the other exact reruns reproduce under the local `zh-Hans` settings/external-consent environment.

Focused regression checks:

- [x] `cargo test -p codewhale-tui --lib clipboard_text --locked` (8 passed)
- [x] `cargo test -p codewhale-tui --lib copy_cell_to_clipboard_uses_canonical_assistant_source --locked` (1 passed)

## Checklist

- [x] Verified TUI behavior manually if UI changes — Copy Message confirmed working by the reporter
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address — N/A; no harvested or co-authored code

### DIFF
--- a/crates/tui/src/tui/ui/activity_detail.rs
+++ b/crates/tui/src/tui/ui/activity_detail.rs
@@ -12,7 +12,9 @@ use crate::tui::app::App;
 use crate::tui::footer_ui::one_line_summary;
 use crate::tui::history::{HistoryCell, ToolCell, ToolStatus};
 use crate::tui::pager::PagerView;
-use crate::tui::ui_text::{history_cell_to_text, truncate_line_to_width};
+use crate::tui::ui_text::{
+    history_cell_to_clipboard_text, history_cell_to_text, truncate_line_to_width,
+};
 
 fn selected_transcript_cell_index(app: &App) -> Option<usize> {
     app.viewport
@@ -513,7 +515,7 @@ pub(crate) fn copy_cell_to_clipboard(app: &mut App, cell_index: usize) -> bool {
         .last_transcript_area
         .map(|area| area.width)
         .unwrap_or(80);
-    let text = history_cell_to_text(cell, width);
+    let text = history_cell_to_clipboard_text(cell, width);
     if text.trim().is_empty() {
         app.status_message = Some("Message is empty".to_string());
         return false;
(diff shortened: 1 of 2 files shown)


Hmbown:
Nice fix — copying the canonical source instead of the rendered transcript is the right call, and the regression tests (authored glyphs, wrapping, Markdown) cover the exact failure modes from #5314.

Your branch fails CI only from base drift (the fork base predates the release-credits parity update on main, and the runtime-contract measurement ran on the old base). GitHub declined my push to your fork, so I've cherry-picked the identical commit onto current main and opened #5331 to land it via the documented harvest path with your authorship preserved (AUTHOR_MAP identity + `Harvested from PR #5319 by @XhesicaFrost`). That PR auto-closes this one with credit on merge. Locally: all 7 new tests pass and the runtime-contract gate is green.

PR #5336, opened 2026-08-12
fix(mcp): omit nextCursor when there are no further pages

### LINKED ISSUE #5335: serve --mcp responds to tools/list and resources/list with "nextCursor": null, breaking strict MCP clients
### Summary

`codewhale serve --mcp` (and the equivalent `mcp-server` stdio entrypoint) responds to `tools/list` and `resources/list` with `"nextCursor": null` on every call. The [MCP specification](https://modelcontextprotocol.io/specification/2024-11-05) requires `nextCursor` to be a string, or **absent** when there are no more results; `null` is invalid.

### Steps to reproduce

1. Run `codewhale serve --mcp`.
2. Send an `initialize` request, then `tools/list`.
3. The response contains `{"tools": [...], "nextCursor": null}`.

Same shape is returned by `resources/list`.

### Impact

Strict clients validate the response shape and reject `null`. For example, Claude Code fails to enumerate the server's tools with:

```
Invalid input: expected string, received null (at nextCursor)
```

so Codewhale's MCP tools are undiscoverable in those clients. Any resource-capable strict client hits the equivalent failure on `resources/list`.

### Root cause

`crates/tui/src/mcp_server.rs` builds both responses with a literal `"nextCursor": Value::Null` (in `list_tools_response` and `list_resources_response`).

### Proposed fix

Omit `nextCursor` entirely when there are no further pages. Fix is implemented and verified on branch `fix/mcp-nextcursor-null` (see linked PR).

xiaoray-blip:
Fixes #5335

## Problem

`serve --mcp` returns `"nextCursor": null` in both `tools/list` and `resources/list` responses. Per the MCP spec the field must be a string or absent; `null` is invalid, and strict clients such as Claude Code reject the shape (`expected string, received null (at nextCursor)`), leaving Codewhale's tools undiscoverable.

## Change

- Omit `nextCursor` entirely in `list_tools_response` and `list_resources_response` (`crates/tui/src/mcp_server.rs`).
- Add a regression test `list_responses_omit_null_next_cursor` asserting both responses drop the key.

## Verification

- `cargo test -p codewhale-tui list_responses_omit_null_next_cursor` passes (1 passed, 0 failed).
- End-to-end: built `codewhale-tui serve --mcp`, sent `initialize` + `tools/list` (+ `cursor: null` variant), responses contain no `nextCursor` key and return 6 tools / 2 resources.

Implementation and verification were carried out with Claude Code.

### DIFF
--- a/crates/tui/src/mcp_server.rs
+++ b/crates/tui/src/mcp_server.rs
@@ -233,7 +233,10 @@ impl McpServer {
                 }
             }
         }
-        json!({ "tools": tools, "nextCursor": Value::Null })
+        // MCP spec: `nextCursor` must be omitted (or be a string) when there
+        // are no more results. Emitting `null` violates the spec and breaks
+        // strict clients (e.g. Claude Code) that validate the response shape.
+        json!({ "tools": tools })
     }
 
     fn list_resources_response(&self) -> Value {
@@ -258,7 +261,9 @@ impl McpServer {
             }
         }
 
-        json!({ "resources": resources, "nextCursor": Value::Null })
+        // Same spec point as `list_tools_response`: omit `nextCursor` when
+        // there are no further pages rather than emitting `null`.
+        json!({ "resources": resources })
     }
 
     fn call_tool(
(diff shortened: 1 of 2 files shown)


Hmbown:
Merged — thank you @xiaoray-blip! The pagination cursor fix is clean, focused, and now fully green across Linux, macOS, Windows, lint, DCO, OpenHarmony, and the link gate. We really appreciate you strengthening the MCP list-tools contract and staying with the branch rebuild through CI.

PR #5339, opened 2026-08-12
fix(engine): suppress child-owned shell completions

### LINKED ISSUE #5325: runtime: don't deliver child-owned background shell completions to the parent model stream
## Summary

The parent model's turn stream currently receives a `background_shell_completion` runtime event for **every** tracked background shell job — including jobs spawned by sub-agents (children). Child shell completions are delivered to the child's own runtime already, and the child's completion sentinel (summary + evidence) is the parent's contract, so child-owned shell events in the parent stream are duplicate noise that inflates parent context for no signal.

## Evidence

- `crates/tui/src/tools/shell.rs` stamps every task and completion event with `owner_agent_id` / `owner_agent_name` (the event payload already carries them).
- `crates/tui/src/core/engine/turn_loop.rs:1698` (`drain_shell_completion_events()` → `shell_completion_runtime_message(...)`) and the late drain at `:1988` inject **all** drained completions into the parent session message stream — there is no owner-based filter.
- `running_owner_agent_ids()` exists in `crates/tui/src/tools/shell.rs` but is used only for `touch_workers_with_running_shells` (heartbeats), not delivery filtering.
- Observed in practice: a child worker's `sed ... | head` and `cargo test` completions arrive at the parent as full runtime events (task_id, command, stdout tail, …).

## Proposal

- In the parent turn-loop drains (both the main drain at ~1698 and the late drain at ~1988), drop completion events whose `owner_agent_id` is `Some(_)` before building the runtime message. Parent-owned jobs (`owner_agent_id: None` / parent id) keep current behavior.
- Keep unaffected: task/status UI visibility of child shell tasks, `Bash` `wait`/`status` actions, the `touch_workers_with_running_shells` heartbeat, and the child's own completion-event delivery.
- If a child's shell job outlives the child (orphaned job), the parent still sees the task in task/status and can `wait` on it explicitly; the suppressed event is not lost state, just not auto-delivered.

## Acceptance criteria

- Running a background shell command inside a sub-agent produces **no** `background_shell_completion` event in the parent's stream; an identical command run by the parent still does.
- Heartbeats for children with running shells still fire; task/status still shows child tasks.
- No change to the `waiting_for_subagents` anti-polling guardrail events (those are deliberate).

cyq1017:
## Summary

- filter child-owned background shell completion events out of the parent model stream
- preserve unowned parent completions and task/status visibility
- add a regression covering parent and child-owned jobs

Closes #5325

## Verification

- `cargo test -p codewhale-tui --lib --locked core::engine::` (466 passed, 4 ignored)


Hmbown:
Thanks for this fix! Quick status check: this is still draft and needs a rebase onto current `main` (the branch conflicts after the v0.9.7 merges). Want to take the rebase yourself, or should I harvest it onto a fresh branch (the repo's usual pattern) with your commit preserved and `Co-authored-by` credit intact? Either way it targets v0.9.8 — #5325 has moved to that milestone.

Hmbown:
Thank you @cyq1017 — this is a valuable catch at the parent/child shell ownership boundary. I rebuilt the PR branch cleanly on current `main` around your fix, preserved your authorship, and separated the regression and source-budget updates so the change is easy to review.

Fresh local verification:
- exact regression: `child_owned_background_completion_is_not_delivered_to_parent` — passed
- `cargo fmt --all -- --check` — passed
- source-structure budget — passed
- diff check — passed

I’ve marked the PR ready and will shepherd the new CI run. Really appreciate the contribution!

Issue #5352, opened 2026-08-13
v0.9.8: Auto-Review hardening and UX from the Kimi Code CLI + Codex comparison


Hmbown:
Follow-up to #5342 (merged: bounded Auto-Review execution restored). Research compared our Auto-Review with Moonshot's Kimi Code CLI (`default_permission_mode` manual/yolo/auto, `[[permission.rules]]` with `ToolName(arg-pattern)` patterns and scopes) and OpenAI Codex CLI (`approval_policy` incl. untrusted/granular, sandbox modes, `approvals_reviewer = auto_review` reviewer agent with a guardian policy taxonomy, denial rationale + circuit breaker (3 consecutive / 10-of-50 aborts the turn), `/approve` override for ≤10 recent denials).

Prioritized v0.9.8 candidates:

- **P0** — Denial rationale + circuit breaker: Auto blocks land as bare `permission_denied` tool results (turn_loop.rs:2472-2480). Feed the block reason to the model with a "do not work around; find a safer path or stop" instruction, and abort the turn after N consecutive auto-blocks (Codex: 3 consecutive / 10 in last 50).
- **P0** — Sensitive-file reads in interactive Auto: the secret action-kind safety floor only holds in Background/Headless; an interactive `cat ~/.ssh/id_rsa` reads as Benign→Allow. Kimi still prompts for `.env`/SSH keys even in yolo. Add secret-path detection to read tools in Auto (fail closed).
- **P1** — Opt-in model-based reviewer tier: deterministic-only is a defensible default (MODES.md:184); offer an *opt-in* reviewer agent (port Codex `guardian/policy.md` risk taxonomy) while the deterministic floor stays non-bypassable beneath it.
- **P1** — Arg-pattern permission rules + scopes: extend `[auto_review]`/`permissions.toml` matchers with `Bash(rm -rf*)`-style patterns and `turn-override/session-runtime/project/user` scopes — the most common ask (unblock `cargo test`, deny `rm -rf`).
- **P1** — Denial override escape hatch: keep up to N recent auto-denials; user can approve one for a single retry, still through the classifier.
- **P2** — Visibility: surface Auto decisions in the transcript (Kimi's "Auto-approved" markers); turn-end summary ("N auto-approved, M held/blocked").
- **P2** — Network default in Auto: WorkspaceWrite sets `network_access: true` (authority.rs:375-381); Codex defaults network **off** in workspace-write. Consider off + opt-in to match the fail-closed posture.
[...]

Issue #5146, opened 2026-08-03
Prompt teaches tool-call-is-proposal, then batch approval rejects the same call (owner session 1.9)


Hmbown:
## Found

Owner session 2026-08-02 (checklist §1.9): in the #4907 CWC Work transcript the agent emitted an `apply_patch` BEFORE the batch was approved, the gate correctly rejected it, and the model apologized ("I accidentally sent an unapproved apply_patch"). The gate held — but the attempt is produced by our own prompt/tool contract, not by model misbehavior.

## Mechanism (anchors verified 2026-08-03 on the release train)

- `crates/tui/src/prompts/text.rs:333-344` — the Suggest overlay says "Write operations … require user approval before executing… **The user will see your proposed action** and can approve or deny it." There is no separate propose-tool: emitting the write call IS the proposal mechanism. The overlay never describes CWC Work's batch flow (approve the batch first, then write).
- `crates/tui/src/tools/apply_patch.rs:318-320` — the tool description actively steers to `apply_patch` ("Use this instead of `git apply`, `patch`, or repeated `edit_file` calls") with zero mention of approval; `:383-385` declares only `ApprovalRequirement::Suggest` (default for WritesFiles, `tools/spec.rs:1174-1183`).
- The gate that held: `crates/tui/src/core/engine/turn_loop.rs:44-77` (`registered_tool_approval_required`) and `:2029-2042`.

Net: a well-behaved model following our instructions emits `apply_patch` as its proposal and gets rejected when it lands outside an approved batch. The model's apology is honesty doing work the prompt/gate contract should do.

## Fix direction

Make the contract explicit where the model reads it: when a batch-approval flow is active (CWC Work), the turn context should say so — "writes execute only inside an approved batch; proposing = showing the patch in the plan, not calling the write tool" — or the rejection error should name the correct behavior ("batch not approved yet; present the change in the plan first") instead of a bare "unapproved". Fail with the fix named, per the #5123 class.

Caveat: the exact #4907 transcript was not available during analysis, so the specific rejection branch is inferred from code, not pinned from the log.

Hmbown:
Fixed on `main` in `eb543a933` — "fix(tui): align the propose-vs-execute copy with the approval gate (#5146)".

Closed as part of the v0.9.4 milestone reconciliation. The fix landed but the PR carried no `Closes #N` line, so the issue stayed open and the milestone stopped reflecting reality.

If the prompt/approval mismatch is still reproducible, please reopen.

Issue #5324, opened 2026-08-12
agent tool: simplify the 32-field schema so models stop erroring on it


Hmbown:
## Summary

The model-facing `agent` tool carries a **32-property JSON schema with zero required fields**, serves **eight actions** (`start`, `status`, `peek`, `message`, `followup`, `interrupt`, `wait`, `cancel`) from that single schema, and the runtime parser additionally accepts a bag of aliases per field. Models regularly misinvoke it — passing fields that are invalid for the chosen action, colliding on aliases, or omitting the one field an action actually needs — which surfaces as tool errors (and sometimes spurious approval-modal routing). This issue proposes shrinking and simplifying the schema for a future release so the model has structural guidance instead of a 32-field free-for-all.

## Evidence (current surface, `crates/tui/src/tools/subagent/mod.rs`)

- `input_schema()` exposes 32 properties: `action, until, agent_id, timeout_secs, message, reason, include_archived, name, prompt, dependencies, acceptance, type, profile, model, thinking, cwd, worktree, worktree_branch, worktree_base, worktree_path, fork_context, max_depth, max_steps, wall_time_secs, workspace_policy, expected_artifact, write_authority, write_roots, exact_files, coordination_contracts, deliberate, resume_from` — with `"required": []`, so the schema gives no per-action structure at all.
- One schema for eight actions: most fields are only meaningful for one action (e.g. `until`/`timeout_secs` only for `wait`; `message` only for `message`/`followup`; `reason` only for `interrupt`; `include_archived` only for `status`), but the model sees them all, always.
- Alias bags make validation lenient in confusing ways: `action`/`op`, `agent_id`/`id`/`session_name`/`name`, `profile`/`fleet_profile`/`roster_profile`, `type`/`agent_type`/`agent_name`, `role`/`agent_role`, `write_authority`/`writeAuthority`, `timeout_secs`/`timeout`, plus action-name synonyms (`spawn`/`run`, `list`/`inspect`, `steer`, `join`/`await`/`block`, …). When two aliases disagree, `parse_agent_ref`-style code just picks one, and `start_requests_read_only_role` fails **closed** into the approval modal for anything it cannot prove read-only — so alias slips turn into refused calls or extra approval prompts.
[...]

PR #5358, opened 2026-08-13
feat(engine): auto-review denial rationale + turn circuit breaker

Lstarsky0:
First P0 slice of #5352: denial rationale + the circuit breaker.

An Auto-Review block reached the model as a bare `permission_denied`, which is an invitation to re-phrase the same denied action until the step budget runs out. Two changes:

**Rationale.** `AutoReviewPlanDecision::Block` now goes through `auto_review_block_tool_error()`, which keeps the original reason first and verbatim (audit events and tests match on it) and appends the next behavior, same shape as the user-denial message from #5146: this block is automatic, don't look for a way around it, take a safer approach or stop and tell the user.

**Breaker.** `AutoReviewBlockBreaker`, turn-local like `consecutive_empty_repl_rounds` right above it: 3 consecutive auto-blocked calls, or 10 within the last 50 planned calls, stop the turn (the Codex thresholds cited in the issue — hardcoded for now, per the thread). It's fed at plan admission — the one point every call passes through, and blocked plans never execute — and checked at the top of the loop next to `at_max_steps`, so the blocked batch's results still land in the transcript and the abort happens at the request boundary. Same delivered-answer carve-out as the step budget: on an optional continuation it breaks instead of failing the turn.

Only blocks from the Auto-Review decision itself count. User denials already carry their own guidance (#5146), and repo-law / ask-rule blocks have their own remedies — counting them here would abort turns for reasons the model was already told how to handle.

### Validation

- `cargo fmt` / `cargo clippy --all-targets -- -D warnings` — clean
- 5 unit tests on the wrapper and breaker (trip at 3, streak reset, 10-of-50, window eviction)
- 1 end-to-end test: injected model proposes three `rm -rf` calls under Auto — all three results reach the model with the guidance attached, then the turn stops at the next request boundary with `Failed`; the mock scripts no second turn, so reaching the model again fails the test on its own
- Mutation-checked: disabling the breaker wiring alone, or reverting to the bare denial alone, each fails the end-to-end test
- Full `codewhale-tui` lib suite, twice: 10310 passed / 4 failed both rounds. The same four (`settings_command_opens_typed_editor…`, `model_picker_is_usable…`, `underwater_motion_keeps…`, `config_view_surfaces…`) fail identically on a clean `upstream/main` checkout on this machine, filtered or full, while main's own CI is green — they read the real `~/.codewhale/settings.toml` (this box has `calm_mode` set) instead of a hermetic default, so they're an environment sensitivity on main, not this diff. Happy to file that separately.
- `check-runtime-contract-budget.py` — PASS, all 55 metrics exactly at budget; this adds no model-facing schema bytes.

No-Issue: first P0 slice of #5352; the issue tracks more items and stays open.

### DIFF
--- a/crates/tui/src/core/engine/tests.rs
+++ b/crates/tui/src/core/engine/tests.rs
@@ -3,7 +3,8 @@ use super::*;
 use super::context::{COMPACTION_SUMMARY_MARKER, TURN_MAX_OUTPUT_TOKENS};
 use super::streaming::{TOOL_CALL_END_MARKERS, TOOL_CALL_MARKER_PAIRS};
 use super::turn_loop::{
-    merge_new_runtime_mcp_tools, registered_tool_approval_required, registered_tool_forces_prompt,
+    AutoReviewBlockBreaker, auto_review_block_tool_error, merge_new_runtime_mcp_tools,
+    registered_tool_approval_required, registered_tool_forces_prompt,
     workspace_write_carve_out_applies,
 };
 use crate::config::ApiProvider;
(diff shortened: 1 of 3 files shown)


Hmbown:
Thank you for the careful diagnosis and especially for proving the denial-loop behavior end to end. The actionable denial rationale is a real improvement, and I have harvested that narrow helper plus a regression test into the guardian lane with your authorship credited. After comparing Codex, Kimi Code, DeepSeek Harness, and DS4 behavior, the v0.9.8 design has converged on a stateless contract: deterministic policy decides, one guardian consult may resolve a hold, and any unavailable answer denies. That means we are intentionally leaving out the 3/10/50 breaker state, while keeping your useful guidance so the model knows to take a safer path instead of retrying. No rebase work is needed from you; I will link the credited harvest here when #5353 updates. This was exactly the kind of concrete, test-backed contribution that helped us simplify the feature safely.

Lstarsky0:
Rebased onto `d48db57e6`. The conflict was only `scripts/source-structure-budget.json` — #5336 raised the ceiling to 689696 while this branch carried 689764 against the older base. Re-measured on the new base instead of taking the larger number: 689800.

macOS is red on `qa_pty::work_surface_file_mutation_modes_are_truthful_in_real_pty_frames`, and I don't think that one is noise. Looking for prior art I found `b94eda1a1` on `codex/v098-auto-review-guardian`, whose message describes exactly this interaction:

> The sealed QA harness has no provider for the model guardian, so a fail-closed reviewer denial changed the frame this test pins.

This PR lengthens the Auto-Review denial text, so in that same harness it shifts the frames the test scrolls through — the dump shows the Raw detail section rendered but `Exact File change` never reached. Your fix there pins `[auto_review] reviewer = false` in the harness config, which I can't cherry-pick because that key arrives with `2458a0326` on the same branch. I'd rather not invent a second mechanism for it on main.

Which brings up the bigger thing: that branch already implements this slice.

```
let breaker_note = if self.reviewer_ledger.is_tripped() { … };
blocked_error = Some(ToolError::permission_denied(format!(
    "Auto-Review guardian denied tool '{tool_name}': {reason}. Do not work around \
     this denial; find a materially safer path or stop.{breaker_note}"
)));
```

I diffed the two rather than guessing. `reviewer_ledger` and `AutoReviewBlockBreaker` are the same design down to the constants — 3 consecutive, 10 within a 50-entry `VecDeque` window, Codex parity both times. Two real differences:

- **Trigger.** Theirs counts explicit reviewer denials on the guardian route; mine counts `AutoReviewPlanDecision::Block` on the deterministic route, which is all that exists on main today.
- **Turn end.** Theirs returns `Failed` unconditionally. Mine keeps the step budget's delivered-answer carve-out, so an optional continuation breaks instead of failing a turn that already answered the user.

So this is duplicated work, and theirs is the more complete one — it has the guardian tier this only approximates. I claimed P0 on #5352 and heard no objection, so I built it; I'd have paused had I seen the branch, but it is bot-authored and doesn't surface on the issue.

Your call, and I'm happy with any of them:

1. **Close this** in favour of the guardian branch. If the carve-out is worth keeping, I'll send it as a small follow-up against that branch once it lands.
2. **Land this first** as the deterministic-only slice, and the guardian branch drops its duplicated wrapper and keeps the ledger. Then I need direction on the PTY harness, since the config key isn't on main yet.
3. **Neither** — point me at a different P0 item and I'll drop this.

Not attached to the code; the useful part was measuring it. Sorry for the collision.

Hmbown:
Thanks again @Lstarsky0 — this was a genuinely valuable investigation. You diagnosed the Auto-Review denial loop and proved it end-to-end, and that evidence directly shaped the v0.9.8 design: the actionable-denial rationale helper and its regression test were harvested into the #5353 guardian lane with your authorship credited (Co-authored-by preserved in the commit). The final guardian design — one stateless consult under a fail-closed policy — supersedes the turn circuit breaker this PR adds, so rather than layering a second mechanism on top, I'm closing this as superseded. Your rebase onto d48db57e6 and the honest budget re-measurement were exactly the right move, and the investigation stays visible and credited in #5353. Thank you for contributing this!

Lstarsky0:
That's the right call — one stateless consult under a fail-closed policy is a cleaner answer than a turn-level breaker bolted on beside it, and I'd rather the helper live in the lane that owns the behaviour. Thanks for harvesting it with the trailer instead of just re-typing it; I checked `34c541a5a` and the helper and its test came across intact.

For anyone reading this later: the guardian lane is #5353, and the duplicate happened because I claimed P0 on the issue without checking the maintainer branches first. `codex/*` branches don't surface in issue threads, so "nobody replied to my claim" isn't the same as "nobody is building it". That one's on me and it's a habit I've changed.

Two things still open on my side that this doesn't touch:

- #5368 — the test-isolation fix from #5359. All green now.
- #5369 — the Moonshot degrade, which #5324's schema slice needs before it can land. Also green.

And #5324 has measurements waiting on your call: `dependentRequired` keys on presence rather than value, so action discrimination needs `const` branches; the `allOf`/`if-then` form silently loses the constraint on Moonshot while `dependentSchemas` degrades loudly; and the composition costs ~542 B against 442 B freed by `resume_from`, so it needs `fork_context` too to come out net negative. Happy to build it whenever you've picked the shape.

PR #5368, opened 2026-08-13
fix(tui): confine unguarded tests to the isolated state root

### LINKED ISSUE #5359: Four TUI tests read machine state (~/.codewhale, display probe) and fail deterministically on a dev box while CI stays green
## Summary

Four TUI tests read machine state — `~/.codewhale` and the display-refresh probe — so they fail deterministically on a dev box that has real CodeWhale state, while CI (clean HOME, no display probe) stays green. Verified on current `main` (`7b67149b2`), macOS, with a real `~/.codewhale/settings.toml` (`locale = zh-Hans`, `calm_mode = true`) and a populated `catalog/models-dev-catalog.json`.

```
commands::groups::config::config::tests::settings_command_opens_typed_editor_and_preserves_text_mode
tui::views::tests::config_view_surfaces_structural_external_consent_without_io
tui::model_picker::tests::model_picker_is_usable_and_opaque_at_blocker_sizes
tui::ui::tests::underwater_motion_keeps_its_smoother_cadence_during_live_status
```

All four fail filtered or under the full suite; all four pass with the machine state taken out of the equation (contrast below).

## Mechanisms

Three distinct paths, not one:

**1. The `lock_test_env()` trust hole (first two tests).** `settings_path_candidates()` sends lock-holding threads to `settings_path_candidates_from_environment()` on the assumption that a test holding the lock has provided explicit temporary paths. These two hold the lock (to serialize) but set no `CODEWHALE_HOME`/`CODEWHALE_CONFIG_PATH`, so `Settings::load()` resolves the real `~/.codewhale/settings.toml`. With `locale = zh-Hans` the settings dump renders 「设置：」 and the assertion on `"Settings:"` fails. The isolation contract exists — it's just unenforced for lock holders that don't opt in.

**2. `resolve_state_dir` has no test shield (model picker).** The settings isolation lives in `crates/tui/src/settings.rs` only. `models_dev_live::cache_path()` goes through `codewhale_config::resolve_state_dir("catalog")`, which never heard of `isolated_test_state_root()`, so the picker renders against the real models-dev cache; at 80x24 the footer wraps differently and the `"move"` hint assertion fails.

**3. The display probe (underwater cadence).** `probe_display_refresh()` is OnceLock-cached and on macOS asks CoreGraphics for the real display Hz. This box reports 60Hz, so `adaptive_animation_interval_ms(false)` returns 66ms and the assertion on the 80ms `UI_UNDERWATER_ANIMATION_MS` floor fails. The test pins `low_motion`/`fancy_animations` but can't pin the probe.

## Contrast

```
# fails (real machine state)
cargo test -p codewhale-tui --lib settings_command_opens_typed_editor_and_preserves_text_mode

# passes — first three tests
CODEWHALE_HOME=$(mktemp -d) cargo test -p codewhale-tui --lib <each of the first three>

# passes — underwater (0 is out of range, probe rejects it and falls back)
CODEWHALE_DISPLAY_HZ=0 cargo test -p codewhale-tui --lib underwater_motion_keeps_its_smoother_cadence_during_live_status
```

## Possible directions

- Make the lock-holder branch of `settings_path_candidates()` fall back to `isolated_test_state_root()` unless an override var is actually set, instead of trusting the lock alone.
- Give `resolve_state_dir` (or its tui-side callers) the same `cfg(test)` confinement settings already have.
- Pin the probe in cadence tests via `CODEWHALE_DISPLAY_HZ` (or a test-only setter) the way the two motion flags are already pinned.

Happy to send any of these as a PR if the direction sounds right.

Lstarsky0:
Fixes the four tests from #5359. Three independent mechanisms, each with a test that fails when only its own fix is reverted.

**The lock-holder trust hole.** `settings_path_candidates()` and `TuiPrefs::path()` routed lock-holding threads to the real environment, on the assumption that a test holding `lock_test_env()` had supplied temporary paths. Most holders take the lock only to serialize something unrelated — `TERM_PROGRAM`, `NO_ANIMATIONS` — and supplied nothing, so they read `~/.codewhale/settings.toml`. With `locale = zh-Hans` the settings dump renders 「设置：」 and the assertion on `"Settings:"` fails.

The contract was there, it just trusted the wrong signal. `guarded_environment_provides_state_paths()` now wants the lock **and** evidence the caller sealed something: an explicit `CODEWHALE_HOME` / `CODEWHALE_CONFIG_PATH` / `DEEPSEEK_CONFIG_PATH`, or a live `EnvVarGuard` over `HOME`/`USERPROFILE`. The explicit-variable branch reads the environment rather than the guard registry, so the cross-process settings children — which inherit `CODEWHALE_HOME` from their parent — keep resolving it.

`settings.rs` had a byte-identical private copy of `EnvVarGuard`; it is now an alias of the shared one, since a copy would opt all 32 of its call sites out of the record.

An unsealed lock holder gets a **per-thread** directory, not the shared root. Those tests have never shared it — before this they reached the developer's home — and several run full settings transactions. Routing them into the shared root put the transaction lock past its 120s deadline and hung unrelated `config_command_*` tests; that showed up in the first full-suite round and is why the per-holder split exists.

**`resolve_state_dir` has no test shield.** `models_dev_live::cache_path()` goes through `codewhale-config`, a plain dependency here with no view of this crate's isolated root, so an unguarded test read the developer's catalog. Same confinement, plus a test pinning that it stays inside the isolated root and a companion asserting a sealed `CODEWHALE_HOME` is still honored.

**The display probe.** `probe_display_refresh()` is `OnceLock`-cached and asks CoreGraphics for the host panel, so `adaptive_animation_interval_ms(false)` returned 66 ms on a 60 Hz box against the 80 ms floor the cadence test asserts — and the test could pin `low_motion` and `fancy_animations` but not the panel. Under `cfg(test)` it returns the unmeasured result a headless runner sees; `DisplayRefreshPin` pins a panel per thread for tests that want one. The real probe stays covered by a direct test rather than quietly becoming dead code.

### What this deliberately does not fix

`resolve_load_config_path()` and `try_default_config_path()` have the same trust hole. Tightening them regressed five config-path tests, because those seal the environment through a *third* private guard in `config/tests.rs` that writes `HOME` with `env::set_var` directly — invisible to the registry. Unifying the three guards is the real fix and is bigger than this change; better as a follow-up than half-applied here. Happy to take it if you want it.

### Correction to the issue

The issue attributed `model_picker_is_usable_and_opaque_at_blocker_sizes` to the catalog cache. That was wrong: the original contrast used `CODEWHALE_HOME=$(mktemp -d)`, which moves settings and catalog together. Fixing each alone separates them — the settings fix alone makes it pass, the catalog fix alone does not. It is a fourth symptom of the trust hole. The catalog gap is real, just not what broke that test, and now has its own assertion instead of an inferred one.

### Validation

Full `codewhale-tui` lib suite, same machine, back to back:

| | main (`d48db57e6`) | this branch |
| --- | --- | --- |
| failures | 5 | 1 |
| the four from #5359 | fail | **pass** |
| `tools::pdf::…forwards_page_window…` | fail | fail |

That pdf test is a pre-existing timing flake, not related to this change — on clean `main` it fails roughly one run in five (1/5 here), and it fails on `main` with this branch not applied.

- mutation-checked one fix at a time: reverting the predicate fails `settings_command_opens_typed_editor…`, `config_view_surfaces…` and `model_picker…`; removing the catalog confinement fails only the new catalog test, which reports the escaped path; the display tests fail without the pin
- `cargo fmt --check` clean; `cargo clippy --workspace --all-features --locked` with the CI allow-list: 0 errors
- `check-runtime-contract-budget.py` PASS, all 55 metrics exactly at budget
- `check-source-structure-budget.py` acknowledged in its own commit, per `6c8577715`

Closes #5359

### DIFF
--- a/crates/tui/src/commands/groups/config/config.rs
+++ b/crates/tui/src/commands/groups/config/config.rs
@@ -2467,169 +2467,39 @@ pub fn logout(app: &mut App) -> CommandResult {
 mod tests {
     use super::*;
     use crate::config::Config;
-    use crate::test_support::lock_test_env;
+    use crate::test_support::{EnvVarGuard, TestEnvLock, lock_test_env};
     use crate::tui::app::{App, TuiOptions};
     use crate::tui::approval::ApprovalMode;
     use std::env;
-    use std::ffi::OsString;
     use std::fs;
     use std::path::Path;
     use std::path::PathBuf;
     use std::time::{SystemTime, UNIX_EPOCH};
 
     struct EnvGuard {
-        home: Option<OsString>,
-        userprofile: Option<OsString>,
-        codewhale_config_path: Option<OsString>,
-        deepseek_config_path: Option<OsString>,
-        codewhale_allow_shell: Option<OsString>,
-        deepseek_allow_shell: Option<OsString>,
-        deepseek_approval_policy: Option<OsString>,
-        no_animations: Option<OsString>,
-        term_program: Option<OsString>,
-        ptyxis_version: Option<OsString>,
-        _lock: crate::test_support::TestEnvLock,
+        _vars: Vec<EnvVarGuard>,
+        _lock: TestEnvLock,
     }
 
     impl EnvGuard {
         fn new(home: &Path) -> Self {
-            let lock = crate::test_support::lock_test_env();
-            let home_str = OsString::from(home.as_os_str());
+            let lock = lock_test_env();
             let config_path = home.join(".deepseek").join("config.toml");
-            let config_str = OsString::from(config_path.as_os_str());
-            let home_prev = env::var_os("HOME");
-            let userprofile_prev = env::var_os("USERPROFILE");
-            let codewhale_config_prev = env::var_os("CODEWHALE_CONFIG_PATH");
-            let deepseek_config_prev = env::var_os("DEEPSEEK_CONFIG_PATH");
-            let codewhale_allow_shell_prev = env::var_os("CODEWHALE_ALLOW_SHELL");
-            let deepseek_allow_shell_prev = env::var_os("DEEPSEEK_ALLOW_SHELL");
-            let deepseek_approval_policy_prev = env::var_os("DEEPSEEK_APPROVAL_POLICY");
-            let no_animations_prev = env::var_os("NO_ANIMATIONS");
-            let term_program_prev = env::var_os("TERM_PROGRAM");
-            let ptyxis_version_prev = env::var_os("PTYXIS_VERSION");
-
-            // Safety: test-only environment mutation guarded by process-wide mutex.
-            unsafe {
-                env::set_var("HOME", &home_str);
-                env::set_var("USERPROFILE", &home_str);
-                env::remove_var("CODEWHALE_CONFIG_PATH");
-                env::set_var("DEEPSEEK_CONFIG_PATH", &config_str);
-                env::remove_var("CODEWHALE_ALLOW_SHELL");
-                env::remove_var("DEEPSEEK_ALLOW_SHELL");
-                env::remove_var("DEEPSEEK_APPROVAL_POLICY");
-                env::remove_var("NO_ANIMATIONS");
-                env::remove_var("TERM_PROGRAM");
-                env::remove_var("PTYXIS_VERSION");
-            }
-
+            let vars = vec![
+                EnvVarGuard::set("HOME", home),
+                EnvVarGuard::set("USERPROFILE", home),
+                EnvVarGuard::remove("CODEWHALE_CONFIG_PATH"),
+                EnvVarGuard::set("DEEPSEEK_CONFIG_PATH", config_path),
+                EnvVarGuard::remove("CODEWHALE_ALLOW_SHELL"),
+                EnvVarGuard::remove("DEEPSEEK_ALLOW_SHELL"),
+                EnvVarGuard::remove("DEEPSEEK_APPROVAL_POLICY"),
+                EnvVarGuard::remove("NO_ANIMATIONS"),
+                EnvVarGuard::remove("TERM_PROGRAM"),
+                EnvVarGuard::remove("PTYXIS_VERSION"),
+            ];
             Self {
-                home: home_prev,
-                userprofile: userprofile_prev,
-                codewhale_config_path: codewhale_config_prev,
-                deepseek_config_path: deepseek_config_prev,
-                codewhale_allow_shell: codewhale_allow_shell_prev,
-                deepseek_allow_shell: deepseek_allow_shell_prev,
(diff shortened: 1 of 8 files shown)


Lstarsky0:
Windows failed on three `runtime_threads` tests — `caller_cancellation_after_engine_acceptance_keeps_owned_turn_lifecycle`, `approval_required_with_stale_active_turn_is_denied`, `approval_timeout_denies_clears_ui_and_next_turn_can_start` — all timeouts (`Timed out waiting for turn`, `deadline has elapsed`), no assertion mismatches. Pushed a change for it rather than calling it noise, but I want to be straight about how strong the evidence actually is.

What I can show:

- Those tests build their own temp `data_dir` and `Config::default()`, and never take `lock_test_env()`, so they take the unchanged branch of every predicate this PR touches. On a clean CI `HOME` the catalog and display-probe changes are equivalent to what `main` already computed.
- `gh run rerun` is rejected on this repo (`workflow file may be broken`), so I couldn't get a second sample of the same commit.
- #5369, off the same base with an unrelated diff, passed Windows. That rules out "the lane is broken right now", which is the explanation I would otherwise have reached for. It does not rule out a flake landing on this run, and I am not going to claim it does.

So instead of arguing, I fixed the one mechanism I could actually name. `unsealed_test_state_root()` ran `create_dir_all` on *every* call for a lock-holding thread, and path resolution runs on every settings read — a syscall per read, which is close to free on Linux and not on Windows. It is now created once per thread behind a thread-local `OnceLock`. That is worth doing regardless of what Windows was doing, and it gives a second Windows sample.

If it comes back red on the same three tests, the timeouts are mine and I will keep digging rather than relabel them. If it comes back green, one green and one red still isn't proof either way, and I'd rather you know that than read a confident "flaky" from me.

Local, unchanged by this push: full `codewhale-tui` lib suite on the same machine, `main` 5 failures → this branch 1, the four from #5359 fixed, the remaining one the `tools::pdf` timing flake that also fails on clean `main` (1 run in 5 measured here).

Lstarsky0:
Windows came back green on the second sample.

Keeping the framing I set before the run: one red then one green does not tell you which explanation was right. Either the per-call `create_dir_all` was pushing those three `runtime_threads` deadlines over on a slow runner, or the first sample was a flake and the memoization is unrelated to it. I can't separate those from two runs, and the repo has documented Windows-lane flakiness that makes the second reading entirely plausible.

What I'd say with confidence: the memoization is correct on its own terms — a syscall per settings read for a lock-holding thread was waste regardless of what CI was doing — and nothing in this PR's diff has a path into the turn-lifecycle tests that failed. If those three ever go red again on an unrelated PR, that's the tiebreaker, and I'd rather it be recorded here than have this read as a fix I proved.

All checks green now.

Issue #5355, opened 2026-08-13
v0.9.8 Known issues: parallel-load and config-fixture flakes (from the v0.9.7 close-out)


Hmbown:
Known-issue investigation basket for v0.9.8 (carried over from the v0.9.7 close-out):

1. `exec_persistent_service::failed_exec_*` — parallel-load flake observed during the v0.9.7 release gate.
2. `exact_turn_snapshot_restores_custom_endpoint_and_turn_receipt_after_builtin_route` — the parallel-load flake that blocked the first v0.9.7 release workflow before #5343 raised its deadline (5s → 15s) as a stopgap; investigate the root cause, don't just keep raising timeouts.
3. New observation (2026-08-13, local full-suite runs): 34 model/provider-route tests across `config::`, `config_persistence::`, `config_ui::`, `runtime_api::`, `terminal_mode_tests`, `tui::app`, `tui::ui`, and `tui::views` fail under the default parallel test threads but pass 512/512 serially (`--test-threads=1`) — the same family as #5056 (workspace/config-fixture races). Repro: `cargo test -p codewhale-tui --lib config::` vs `... -- --test-threads=1` on a machine with a populated `~/.codewhale/config.toml`.

Suggested v0.9.8 work: serial harness or per-test config isolation for the config fixture family, plus a flake triage of the two release-gate flakes above.

Hmbown:
Moved to v0.9.9. Known flake/PTY basket; CI remains authoritative. Not a 0.9.8 ship blocker.

Lstarsky0:
Data on one of these, and it does not behave the way the basket's framing predicts.

`tools::pdf::tests::shared_adapter_forwards_page_window_and_returns_stdout` fails on every full-suite run on my machine (macOS, `codewhale-tui --lib`). It is not in CI's failure set, so this is local — but the shape looks like the family you're describing, and one result rules out the suggested fix.

**It fails serially.**

| run | tests run | wall | pdf |
| --- | --- | --- | --- |
| `tools::pdf` alone, 5× | 5 | <1s | pass every time |
| `tools::` | 1742 | 8.6s | pass |
| `tools::` (same filter, later) | 1742 | 50.4s | pass |
| `tools:: sandbox` | 1809 | 50.5s | **fail** |
| `tools:: client` | 2188 | 51.1s | **fail** |
| full suite, default threads | — | 144s | **fail** |
| full suite, `--test-threads=2` | — | 226s | **fail** |
[...]

PR #5378, opened 2026-08-14
test(tui): re-pin the thinking-ladder assertions

### LINKED ISSUE #5377: main is red on macOS and Windows: nine reasoning-effort tests still assert the pre-ladder vocabulary
`main` is red on macOS and Windows: nine reasoning-effort tests still assert the pre-ladder vocabulary.

Not a flake and not the #5355 family — it bisects to one commit and reproduces identically every run.

## Where it is

Current `main` `c9b9559a7`, CI's own run [31778394006](https://github.com/Hmbown/CodeWhale/actions/runs/31778394006):

```
Test (macos-latest):   failure   07:02:29 -> 07:22:20
Test (windows-latest): failure   07:02:29 -> 07:18:51
Test (ubuntu-latest):  success   07:02:29 -> 07:02:32   (3s — Linux tests are on CNB)
```

The nine, from that run's log:

```
commands::groups::config::config::tests::config_reasoning_effort_uses_codex_provider_labels
fleet::profile::tests::profile_loader_normalizes_reasoning_aliases
settings::tests::reasoning_effort_setting_normalizes_and_clears
terminal_mode_tests::cli_reasoning_effort_normalizes_aliases_and_rejects_typos
tui::app::tests::fixed_route_thinking_cycle_persists_raw_preference
tui::app::tests::reasoning_cycle_persists_through_the_same_owner_as_the_picker
tui::model_picker::tests::picker_keeps_default_when_catalog_has_no_reasoning_options
tui::ui::tests::ctrl_t_key_event_reaches_reasoning_effort_cycle
tui::views::tests::config_view_filter_accepts_j_k_and_unicode_case
```

## Bisect

`cargo test -p codewhale-tui --lib`, same machine, full suite each time:

| commit | failures |
| --- | --- |
| `11bbe7063` keep 0.9.8 CI honest | 1 |
| `6f6c35183` walk each model's real thinking ladder | 10 |
| `c9b9559a7` current `main` | 10 |

The constant 1 is `tools::pdf::tests::shared_adapter_forwards_page_window_and_returns_stdout`, which is local to my machine — it is not in CI's failure set, so ignore it here. The nine arrive exactly at `6f6c35183` and are the same nine CI reports.

## Mechanism

Four groups, and in every one the *actual* value is what the ladder commit says it now intends. Ran each and recorded the assertion:

| test | left (actual) | right (expected) |
| --- | --- | --- |
| `settings::…normalizes_and_clears` | `Some("xhigh")` | `Some("max")` |
| `fleet::profile::…normalizes_reasoning_aliases` | `Some("xhigh")` | `Some("max")` |
| `terminal_mode_tests::cli_…rejects_typos` | `Some("xhigh")` | `Some("max")` |
| `…config_reasoning_effort_uses_codex_provider_labels` | `XHigh` | `Max` |
| `tui::ui::ctrl_t_key_event_reaches_reasoning_effort_cycle` | `Low` | `High` |
| `tui::app::reasoning_cycle_persists_through_the_same_owner_as_the_picker` | `Low` | `High` |
| `tui::app::fixed_route_thinking_cycle_persists_raw_preference` | `Auto` | `Off` |
| `tui::model_picker::picker_keeps_default_…` | `["auto","low","medium","high"]` | `["auto","off","high","max"]` |
| `tui::views::config_view_filter_…` | `[…, "reasoning_effort", …]` | list without it |

The first four are the vocabulary change. `6f6c35183` gave `xhigh` and `ultra` their own rungs in `parse_strict` (`crates/tui/src/tui/app/types.rs:184`), where they used to collapse into `Max`:

```rust
// 11bbe7063
"max" | "maximum" | "xhigh" | "ultra" | "ultracode" => Ok(Self::Max),

// c9b9559a7
"xhigh" => Ok(Self::XHigh),
"ultra" | "ultracode" => Ok(Self::Ultra),
"max" | "maximum" => Ok(Self::Max),
```

`as_setting()` round-trips those to `"xhigh"` / `"ultra"`, so anything asserting the old collapse fails.

The next four are the cycle itself — the commit message's "cycle `picker_efforts_for_route` instead of the DeepSeek off/high/max shortcut" is exactly the difference between each left and right column. The last is `reasoning_effort` becoming a visible config-view row.

So the behaviour looks right to me and the assertions are the stale half. Nine tests pinning the shortcut the commit set out to remove.

## Why CI hasn't been shouting

Nothing has completed a heavy run on `main` between the ladder commit and now:

| commit | run |
| --- | --- |
| `6f6c35183` ladder | no run recorded |
| `bc544f946` | cancelled |
| `5d35e01ef` | cancelled |
| `a9acd63e1` | success — but all three Test legs finished in 4–5s, so this is the light-change echo, not a test run |
| `c9b9559a7` | **failure** (above) |

`a9acd63e1` is the one worth flagging: a green "Test (macos-latest)" that ran for five seconds is the skip path preserving the required context, and reading it as a passing suite is an easy mistake to make from the checks list alone. I made it myself an hour ago before looking at the durations.

## Repro

```
cargo test -p codewhale-tui --lib settings::tests::reasoning_effort_setting_normalizes_and_clears
```

Fails on a clean `HOME` and with `--all-features`, so it isn't the config-fixture family from #5355.

## Offer

I'll start on it — every left column above is derivable from the ladder, so re-pinning the nine is mechanical rather than a guess. Two spots where I'd be following the code rather than a stated intent, and where I'd rather you correct me in review than have me quietly pick:

- `ultracode` now settles on `"ultra"` instead of `"max"`. Deliberate, or should the alias keep landing on `max`?
- `fixed_route_thinking_cycle_persists_raw_preference` moves `Off` to `Auto`, which changes what a fixed route starts from rather than just what it's called.

If you'd rather own this one, say so and I'll drop it — you've got more context on the ladder than I do.

Noticed while running the full suite as the local gate for #5376; that PR's macOS and Windows legs inherit these nine and are red for this reason, not its own.

Lstarsky0:
Closes #5377

Nine tests, no production changes. Each one asserted the off/high/max shortcut that `6f6c35183` set out to replace, so `main` has been red on macOS and Windows since it landed — its own run [31778394006](https://github.com/Hmbown/CodeWhale/actions/runs/31778394006) fails exactly these.

Before touching anything I recorded what each one actually reports, and every left column is derivable from the new ladder:

| test | was | now |
| --- | --- | --- |
| `settings::…normalizes_and_clears` | `Some("max")` | `Some("xhigh")` |
| `fleet::profile::…normalizes_reasoning_aliases` | `Some("max")` | `Some("ultra")` |
| `terminal_mode_tests::cli_…rejects_typos` | `Some("max")` | `Some("xhigh")` + `Some("ultra")` |
| `…config_reasoning_effort_uses_codex_provider_labels` | `Max` | `XHigh` |
| `tui::ui::ctrl_t_key_event_reaches_reasoning_effort_cycle` | `High` | `Low` |
| `tui::app::reasoning_cycle_persists_through_the_same_owner_as_the_picker` | `High` | `Low` |
| `tui::app::fixed_route_thinking_cycle_persists_raw_preference` | `Off` | `Auto` |
| `tui::model_picker::picker_…grok_4_5` | `["auto","off","high","max"]` | `["auto","low","medium","high"]` |
| `tui::views::config_view_filter_…` | list without it | `reasoning_effort` joins the `thinking` filter |

Where a test's *name* was the point, I kept the name true rather than just flipping a number. `profile_loader_normalizes_reasoning_aliases` would have been asserting `xhigh → xhigh`, which is not an alias normalizing, so it uses `ultracode` — still an alias. Same for the CLI one, which now pins both halves of the split.

## The one that isn't a value change

`picker_keeps_default_when_catalog_has_no_reasoning_options` used grok-4.5 because it "ships no reasoning_options list". The same commit gave it one:

```json
"reasoning_options": [
  { "type": "effort", "values": ["low", "medium", "high"], "default": "high" }
]
```

So the name described a case that no longer applied to its own fixture. Renamed to `picker_walks_the_catalogued_ladder_for_grok_4_5`.

That leaves the question of whether the `DEFAULT_PICKER_EFFORTS` fallback lost coverage. It didn't — `picker_normalizes_low_medium_to_high` reaches it through a generic Moonshot base URL and asserts `DEFAULT_PICKER_EFFORTS` directly. I checked that before renaming rather than after, because dropping the only test of a fallback while making CI green is exactly the trade this kind of PR makes by accident.

## Two spots where I followed the code

Flagged in #5377 and repeated here so they're easy to reject in review:

- **`ultracode` now lands on `ultra`, not `max`.** `parse_strict` sends `"ultra" | "ultracode"` to `Ultra`, so I followed it. If the alias was meant to keep pointing at the ceiling, that's a one-line change in `types.rs` and I'll redo the assertions instead.
- **`fixed_route_thinking_cycle_persists_raw_preference` moves `Off` to `Auto`.** That changes what a fixed K3 route wraps to, not just what it's called. The test's actual subject — the raw preference surviving a restart — is unchanged, and I rewrote its message to say that rather than the old off-specific wording.

## Verification

`cargo fmt --check` clean. Workspace clippy with CI's exact allow-list: exit 0.

`cargo test -p codewhale-tui --lib`, same machine, two rounds:

| | passed | failed |
| --- | --- | --- |
| `main` c9b9559a7 | 10382 | 10 |
| this branch, round 1 | 10391 | 1 |
| this branch, round 2 | 10391 | 1 |

The remaining one is `tools::pdf::tests::shared_adapter_forwards_page_window_and_returns_stdout`, which fails on clean `main` here too and is not in CI's failure set — a local environment thing, not part of this.

Also ran CI's own command rather than only the crate: `cargo test --workspace --all-features --locked` gives 10406 passed, 1 failed, and the one is that same pdf test.

Budget: +24 owned Rust lines and `lib.rs` 17786 → 17794, both acknowledged in a separate commit. It is all comments and the extra alias assertions; I'd rather leave a line saying why each expectation moved than have someone flip them back next time the ladder shifts.

### DIFF
--- a/crates/tui/src/commands/groups/config/config.rs
+++ b/crates/tui/src/commands/groups/config/config.rs
@@ -3058,7 +3058,8 @@ mod tests {
 
         let result = set_config_value(&mut app, "reasoning_effort", "xhigh", false);
 
-        assert_eq!(app.reasoning_effort, ReasoningEffort::Max);
+        // `xhigh` stopped collapsing into `Max` when the ladder gave it a rung.
+        assert_eq!(app.reasoning_effort, ReasoningEffort::XHigh);
         assert_eq!(
             result.message.as_deref(),
             Some("reasoning_effort = xhigh (session only, add --save to persist)")
--- a/crates/tui/src/fleet/profile.rs
+++ b/crates/tui/src/fleet/profile.rs
@@ -969,7 +969,7 @@ mod tests {
             r#"
 id = "scout"
 role_hint = "scout"
-thinking = "xhigh"
+thinking = "ultracode"
 
 [instructions]
 text = "Scout deeply."
@@ -978,7 +978,12 @@ text = "Scout deeply."
 
         let profiles = load_agent_profiles_from_dir(dir.path()).expect("profile TOML loads");
         assert_eq!(profiles.len(), 1);
-        assert_eq!(profiles[0].profile.reasoning_effort.as_deref(), Some("max"));
+        // `xhigh` used to land here too; the thinking ladder made it a rung of
+        // its own, so `ultracode` is the alias left to exercise.
+        assert_eq!(
+            profiles[0].profile.reasoning_effort.as_deref(),
+            Some("ultra")
+        );
     }
 
     #[test]
--- a/crates/tui/src/lib.rs
+++ b/crates/tui/src/lib.rs
@@ -14727,9 +14727,17 @@ mod terminal_mode_tests {
 
     #[test]
     fn cli_reasoning_effort_normalizes_aliases_and_rejects_typos() {
+        // The thinking ladder split these: `xhigh` is a tier the CLI can now
+        // name, `ultracode` is still an alias and resolves to `ultra`.
         assert_eq!(
             normalize_cli_reasoning_effort("xhigh").unwrap().as_deref(),
-            Some("max")
+            Some("xhigh")
+        );
+        assert_eq!(
+            normalize_cli_reasoning_effort("ultracode")
+                .unwrap()
+                .as_deref(),
+            Some("ultra")
         );
         assert_eq!(normalize_cli_reasoning_effort("default").unwrap(), None);
         assert!(normalize_cli_reasoning_effort("expensive").is_err());
--- a/crates/tui/src/settings.rs
+++ b/crates/tui/src/settings.rs
@@ -3431,14 +3431,20 @@ mod tests {
     #[test]
     fn reasoning_effort_setting_normalizes_and_clears() {
         let mut settings = Settings::default();
-        settings
-            .set("reasoning_effort", "xhigh")
-            .expect("normalize xhigh");
-        assert_eq!(settings.reasoning_effort.as_deref(), Some("max"));
-        settings
-            .set("reasoning_effort", "ultracode")
-            .expect("normalize ultracode");
-        assert_eq!(settings.reasoning_effort.as_deref(), Some("max"));
+        // `xhigh` and `ultra` are their own rungs since the thinking ladder,
+        // so normalizing collapses spellings *within* a tier instead of
+        // folding the top three tiers into `max`.
+        for (input, stored) in [
+            ("xhigh", "xhigh"),
+            ("ultracode", "ultra"),
+            ("maximum", "max"),
+            ("minimal", "low"),
+        ] {
+            settings
+                .set("reasoning_effort", input)
+                .unwrap_or_else(|error| panic!("normalize {input}: {error}"));
+            assert_eq!(settings.reasoning_effort.as_deref(), Some(stored));
+        }
         settings
             .set("reasoning_effort", "default")
             .expect("clear effort");
--- a/crates/tui/src/tui/app/tests.rs
+++ b/crates/tui/src/tui/app/tests.rs
@@ -6027,13 +6027,14 @@ fn reasoning_cycle_persists_through_the_same_owner_as_the_picker() {
     // `apply_reasoning_effort_cycle`.
     app.apply_reasoning_effort_cycle();
 
-    assert_eq!(app.reasoning_effort, ReasoningEffort::High);
+    // One step up DeepSeek's ladder from Off is Low, not the old shortcut's High.
+    assert_eq!(app.reasoning_effort, ReasoningEffort::Low);
     assert_eq!(
         Settings::load()
             .expect("reload settings")
             .reasoning_effort
             .as_deref(),
-        Some("high"),
+        Some("low"),
         "a restart must restore the last thinking choice"
     );
 }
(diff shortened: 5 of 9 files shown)


Lstarsky0:
CI result, and it does what the PR claims.

**Windows: pass.** Whole suite green.

**macOS: one failure, and it is not one of the nine.**

```
exec_persistent_service::failed_exec_kills_pending_service_and_exits_nonzero
```

That is the only `... FAILED` line in the run, and it is item 1 of #5355 verbatim — "`exec_persistent_service::failed_exec_*` — parallel-load flake observed during the v0.9.7 release gate". All nine reasoning-effort tests passed on both platforms.

For contrast, `main`'s own run [31778394006](https://github.com/Hmbown/CodeWhale/actions/runs/31778394006) on the same base fails nine on macOS and is red on Windows too.

| | macOS | Windows |
| --- | --- | --- |
| `main` c9b9559a7 | 9 failed | failed |
| this branch | 1 failed (#5355 item 1) | pass |

`gh run rerun --failed` is rejected on this repo ("workflow file may be broken"), so I can't get a second macOS sample myself. A re-run from your side would settle whether that one is the known flake or something this branch provokes — I'd rather you have that than take my word for the attribution. Nothing here touches process spawning or service lifecycle; it is nine assertion values and one test rename.

Separately, while chasing the local `tools::pdf` failure I found something for #5355 that rules out a whole class of fix: it passes alone, passes at `tools::` scope (1742 tests), and fails under the full suite **at `--test-threads=2` as well as the default**. Thread count isn't the variable, so raising a deadline or lowering parallelism won't help — something else in the suite is leaving state behind. I'll write that up there rather than bury it here.

Hmbown:
Exceptional - thank you so much!!

PR #5384, opened 2026-08-14
test(cli): re-pin the provider-count assertions to the v0.9.8 registry

### LINKED ISSUE #5383: main is red on v0.9.8: cli provider-count assertions still hold the pre-release numbers
`main` is red on the v0.9.8 head: two provider-count assertions in `crates/cli/src/lib.rs` still hold the pre-v0.9.8 numbers.

```
thread 'tests::cli_provider_helpers_follow_config_metadata' panicked at crates\cli\src\lib.rs:7852:9:
assertion `left == right` failed
  left: 45
 right: 43
```

That is CI's own Windows leg on `2acbd5376`, [run 31814345366](https://github.com/Hmbown/CodeWhale/actions/runs/31814345366) — 179 passed, 1 failed, and it is the only failure there. It reproduces byte-identically on macOS here.

Nothing environmental: `all_providers()` returns `&PROVIDER_REGISTRY`, a static, so the count can't vary with config or `HOME`.

## Where it came from

Two v0.9.8 commits each registered a provider and left the assertion behind:

| commit | registry | `ProviderKind::ALL` |
| --- | --- | --- |
| `c9b9559a7` (pre-v0.9.8) | 43 | 38 |
| `53af08ce3` Google Gemini as its own backend | 44 | 39 |
| `531b64ed9` Antigravity (agy 1.1.13) credential plane | **45** | **40** |

`crates/cli/src/lib.rs:7852-7853` still asserts 43 and 38. Both are wrong now — the run panics on the first, so the second is only visible once you fix it.

Worth noting because it shows this was close to caught: the same count is asserted in `crates/config/src/tests.rs:4816`, and *that* one was updated to 40. The registry's own crate kept its test in step; the CLI crate's copy was missed. `display_order_is_complete_and_unique` compares two live values so it never pinned anything.

Fix is the two integers. I have it on a branch with a full `--workspace --all-features` run and will send it right after this.

## Why it reached a release

This is the second regression of this exact shape in two days (#5377 was nine reasoning-effort assertions left behind by `6f6c35183`). Both shipped. That is not two people ignoring a red X — for most commits on `main` there is no verdict to ignore.

The 40 most recent CI runs on `main`, spanning 2026-08-12T09:04 to 2026-08-14T15:24:

| outcome | count |
| --- | --- |
| cancelled before finishing | 27 |
| success, real Test legs (17–60 min) | 8 |
| success, Test legs finished in under 10s | 2 |
| failure | 2 |
| still running | 1 |

So roughly seven in ten runs on `main` are cancelled before they can say anything. The mechanism is in `ci.yml`:

```yaml
concurrency:
  group: ci-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

On a `push` to `main` there is no `pull_request.number`, so every push to `main` shares one group and cancels the run before it. For PRs that is exactly right — only the latest head matters. For `main` every commit is one you may need to bisect to later, and a burst guarantees the middle ones are never tested. The v0.9.8 release pushed three commits in six minutes and killed two live runs doing it; the macOS leg on `40766aecd` had been going 2m49s when it was cancelled.

The two rows at "under 10s" are a separate thing worth knowing while reading that table: those Test legs exit on a light-change condition without running anything, so `conclusion: success` there is not evidence the suite passed. I mis-read one of those as "main is green" while chasing #5377.

Happy to open that as its own issue if you'd rather keep this one to the assertion fix — I put it here because "why didn't CI catch it" is the first thing I'd ask, and the answer isn't specific to this bug.

Lstarsky0:
Closes #5383.

Two integers. `cli_provider_helpers_follow_config_metadata` asserted 43 registry kinds and 38 catalog kinds; v0.9.8 shipped 45 and 40.

| commit | registry | `ProviderKind::ALL` |
| --- | --- | --- |
| `c9b9559a7` (pre-v0.9.8) | 43 | 38 |
| `53af08ce3` Google Gemini as its own backend | 44 | 39 |
| `531b64ed9` Antigravity credential plane | **45** | **40** |

The run panics on the first assertion, so the second only becomes visible once the first is fixed — both had to move.

`crates/config/src/tests.rs:4816` asserts the same catalog count and was already updated to 40. The registry's own crate kept its test in step; this copy in the CLI crate was the one left behind.

## Verification

Full `cargo test --workspace --all-features --locked --no-fail-fast`, twice on this branch and once on the unmodified head in the same session:

| | failures |
| --- | --- |
| unmodified head `3f3cbafe8` | 10 |
| this branch, round 1 | 9 |
| this branch, round 2 | 9 |

Both branch rounds produce the identical failure set, and the set difference against the clean head is exactly `tests::cli_provider_helpers_follow_config_metadata`. Nothing else moves; passed counts go 12646 → 12647.

`cargo fmt --all --check` exits 0.

The nine that remain are pre-existing on the unmodified head and none are touched here — four `agy_credentials`, `plugin_e2e_acceptance`, `prompts::changelog_entry_exists_for_current_package_version`, `qa_pty::legacy_work_ctrl_t_save_export_and_restart_are_consistent`, and the `exec_persistent_service` / `tools::pdf` pair already written up on #5355. I haven't characterized the first four; they arrived with v0.9.8 and are a separate thread.

## One thing to expect on this PR's own checks

`main` currently fails the clippy gate, and it has nothing to do with this change. Running CI's exact command on the unmodified head:

```
cargo clippy --workspace --all-features --locked -- -D warnings -A clippy::uninlined_format_args \
  -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::collapsible_if \
```

exits 101 with two errors in `codewhale-tui`, both in v0.9.8's marketplace code:

- `crates/tui/src/commands/groups/plugins/marketplace.rs:50` — `result_large_err`
- `crates/tui/src/plugins/marketplace/parsers/codex.rs:241` — `useless_format`

So if Lint is red here, that is inherited. I left it out of #5383 because it's a different defect in different files — say the word and I'll open it separately, or fold it in.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -7849,8 +7849,8 @@ mod tests {
             .map(|provider| provider.kind())
             .collect();
         // Full registry keeps legacy dialect/plan kinds; ALL is the catalog surface.
-        assert_eq!(registry_kinds.len(), 43);
-        assert_eq!(ProviderKind::ALL.len(), 38);
+        assert_eq!(registry_kinds.len(), 45);
+        assert_eq!(ProviderKind::ALL.len(), 40);
         for kind in ProviderKind::ALL {
             assert!(
                 registry_kinds.contains(&kind),


Hmbown (approved):
LGTM. The two integers match the v0.9.8 registry (Gemini + Antigravity). Config crate already moved to 40; this was the leftover CLI copy.

Lint red is inherited clippy on marketplace (`result_large_err`, `useless_format`), not this diff. macOS red is the pre-existing `agy_credentials` set from #5355. Ubuntu and Windows tests passed, including the re-pinned assertion.

Merging this unreds the provider-count panic on main. The cancel-in-progress-on-main issue stays open as its own follow-up.

Lstarsky0:
On the inherited Lint red — I built the fix and it clears the gate, so it's ready whichever way you want it.

- `crates/tui/src/plugins/marketplace/parsers/codex.rs:241` — `format!` with no arguments becomes `.to_string()`.
- `crates/tui/src/commands/groups/plugins/marketplace.rs:50` — `open_store` returns `Box<CommandResult>`. All five call sites already do `Err(result) => return result`, so they become `return *result`. I boxed rather than adding `#[allow(clippy::result_large_err)]`: there's no precedent for that allow anywhere in `crates/`, and the CI allow-list doesn't carry it either, which reads as a deliberate choice to keep the lint enforced.

CI's exact command on that branch:

```
cargo clippy --workspace --all-features --locked -- -D warnings -A clippy::uninlined_format_args \
  -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::collapsible_if \
```

exits 0, zero warnings. Twelve insertions, nine deletions, two files.

I've kept it off this branch so what you approved stays what it is. Say the word and I'll either push it here or send it as its own PR against `main` — no preference, and no need to reply if you'd rather take it yourself.

Hmbown:
Superseded by #5394 (same two assertion integers, plus the google ModelRegistry docs row the Lint drift check also wanted). Thanks for the diagnosis and the original patch.

Issue #4949, opened 2026-07-28
Discussion: The Chinese Translation of "Constitution" — "宪法", "协作准则", or Something Else?


SparkofSpike:
国际友人们，英语版本在下面。
（International friends, the English version is below.）

---

大家好！我是PR #4908 的作者。在那次 PR 中，我将 "Constitution" 的中文翻译从“协作准则”改回了“宪法”，因为我个人认为“宪法”更能体现这份文件的基础性和最高权威性。但是，这个改动也引发了一些争议，因为有人担心“宪法”并不是最好的描述，不仅不够贴切，而且在中文语境下可能带有敏感的政治色彩。

我们都无法敲定采取哪个翻译方案。因此，我在此再次发出一个Issue，邀请诸位中文母语者同胞们加入讨论，动动手指在此Issue下留下你们的想法，或许我们能得到一个完美的译名，谢谢你们！

**我的个人看法：**

我是“宪法”翻译的坚定维护者，理由如下：
- 它准确地传达了这份文件在项目“法律层级”中位于最顶层的权威性和约束力。在英文原文中，Constitution本来就被用于凸显此条约的严肃性与至高性，将治理规则称为“Constitution”而非“Guideline”或“Policy”，是经过深思熟虑的选择。“宪法”和"Constitution"这两者之间并没有严重的语义误差。技术借用“宪法”这个名词，实在没有大碍。
- 更何况CodeWhale的"Constitution"本就是关于层层规定的硬性约束，在使用的过程中具有最高权威和强制力，而非一份柔性的协作建议。从这个角度来讲，“协作准则”显得更柔性、更温和，没有突出"Constitution”“宪法”的那种强制的权威性质。
- 在国内技术社区中，用“宪法”来作为"Constitution"的译名已经是比较常见的做法。在业内，“宪法”已经是最常见的翻译，很多用户也理解其背后的“根本大法”这层含义。百度百科中，已将“Constitutional AI”翻作“宪法性人工智能”，“Claude Constitution”翻作“Claude宪法”。在技术社区CSDN中，“宪法”被用作Consitution的译名已经相当普遍。在新闻材料中，也已经使用过“宪法”这个名字。这足以证明，“宪法”在技术讨论中是被允许作为术语借用而存在的。
[...]

PR #5397, opened 2026-08-15
fix(web): call the constitution a charter on the website

Lstarsky0:
Follows the outcome of #4949, which @SparkofSpike closed today after three weeks of discussion with 宪章 as the settled term.

The TUI settled on 宪章 for Simplified Chinese back in `cf08cb6af`, and `zh_hans_constitution_copy_uses_charter_term` (`crates/tui/src/localization.rs`) keeps it there — 宪法 is a banned literal in the zh-Hans pack. The website never followed, so the same feature reads as 宪章 in the TUI and 宪法 on codewhale.net. The worst spot is `getting-started.ts`, which is the copy that walks a new Chinese user from the site straight into `/setup`, where the term flips.

This replaces all 15 occurrences across the four files that carry them:

| File | occurrences |
| --- | --- |
| `web/app/[locale]/constitution/page.tsx` | 6 |
| `web/app/[locale]/docs/constitution/page.tsx` | 7 |
| `web/lib/content/getting-started.ts` | 1 |
| `web/lib/docs-map.ts` | 1 |

The compound forms match what the pack already ships rather than being invented here — `用户全局宪章` and `内置宪章` line up with `SetupCheckpointLayerOrder` in `zh-Hans.json` ("内置基础宪章 -> 用户全局宪章 -> 仓库宪章 -> ..."). `web/` has no occurrences of 宪法 left after this.

Copy only. No structural changes, no English-side changes, no zh-Hant changes — @SparkofSpike's point in #4949 that the Traditional Chinese pack should be left to its own users applies, and the one remaining 憲法 in `zh-Hant.json` is out of scope here.

### Checks

Rebased onto `dd0fbed81`. Every gate below was also run on an unmodified `dd0fbed81` in the same session, so the reds are attributable rather than guessed at.

| Gate | this branch | unmodified `dd0fbed81` |
| --- | --- | --- |
| `npm run check:locales` | 0 | — |
| `npm run check:docs` | 0 | — |
| `npm run lint` | 0 | — |
| `npm run check:facts` | 1 | **1, same failure** |
| `npx vitest run` | 1 — 10 failed, 250 passed | **1 — same 10, same names** |

The two reds are inherited, and both trace back to v0.9.8:

- **`check:facts`** reports `facts.generated.ts` is stale, missing `google` (Google Gemini) and `antigravity` (Google Antigravity) — the two providers `53af08ce3` and `531b64ed9` registered. Same pair that left the CLI assertion counts behind in #5383. #5394 refreshed `docs/PROVIDERS.md` for that drift; `web/lib/facts.generated.ts` looks like the copy that didn't get regenerated. `cd web && npm run prebuild` is what the checker suggests. I left it out of this PR because it isn't a copy change and would bury a one-line terminology diff — happy to send it separately, or fold it in if you'd rather have one commit.
- **`vitest`** — 7 in `lib/deploy-preflight.test.ts` (Cloudflare credentials, environmental), 1 in `lib/public-copy.test.ts` and 2 in `lib/public-surface-contract.test.ts` (changelog/release-facts parity, same v0.9.8 drift). Byte-identical set on the unmodified head; only the millisecond timings differ.

No-Issue: copy-only follow-up to the #4949 discussion, which is now closed; this PR closes nothing on its own.

### DIFF
--- a/web/app/[locale]/constitution/page.tsx
+++ b/web/app/[locale]/constitution/page.tsx
@@ -11,7 +11,7 @@ export async function generateMetadata({ params }: { params: Promise<{ locale: s
     locale,
     title: isZh ? "三层法 · Codewhale" : "Three layers of law · Codewhale",
     description: isZh
-      ? "Codewhale 的嵌套宪法：内置基础法、你的常备法（/constitution）、仓库自己的法（.codewhale/constitution.json）。位阶由执行框架强制生效，换掉模型也不失效。"
+      ? "Codewhale 的嵌套宪章：内置基础法、你的常备法（/constitution）、仓库自己的法（.codewhale/constitution.json）。位阶由执行框架强制生效，换掉模型也不失效。"
       : "Codewhale's nested constitution: bundled base law, your standing law (/constitution), and your repo's law (.codewhale/constitution.json). Rank is enforced in the harness and survives a model swap.",
   });
 }
@@ -20,7 +20,7 @@ export async function generateMetadata({ params }: { params: Promise<{ locale: s
 const LAYERS = [
   {
     n: "01",
-    name: { en: "Bundled Constitution", zh: "内置宪法" },
+    name: { en: "Bundled Constitution", zh: "内置宪章" },
     path: "compiled into every binary",
     pathZh: "编译进每一个二进制",
     en: "The base law. Its priority article fixes the authority order for any conflict, so a stale handoff can never outrank a fresh test result by accident.",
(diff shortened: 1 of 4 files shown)


Lstarsky0:
Both reds here are inherited, and neither is this diff. Detail so nobody has to chase them:

**Lint & Type Check** fails on `check:facts` — `facts.generated.ts` is stale, missing the two providers `53af08ce3` and `531b64ed9` registered. It fails identically on an unmodified head. Sent the regeneration as #5398; this PR should go green once that lands.

**Test (windows-latest)** is a third-party outage, not a test failure. Chocolatey's feed returned a 504 while installing NSIS:

```
Failed to fetch results from V2 feed at 'https://community.chocolatey.org/api/v2/Packages(Id='nsis',Version='3.12.0')'
  with following message : Response status code does not indicate success: 504 (Gateway Timeout).
...
throw 'makensis.exe was not found after installing NSIS.'
```

The job never reached the test step — the log is 196 lines and contains no test results at all. A re-run should clear it.

This PR is four Simplified Chinese copy files; it cannot reach either code path.

Lstarsky0:
Correcting myself: I said above that this would go green once #5398 lands. That isn't right.

**Lint & Type Check** runs `check:facts` *and* `vitest`. On this PR it exited at `check:facts` and never reached the tests, which is why I only saw the one red. On #5398, where facts are regenerated, the job gets past that step and then fails on three assertions that were already failing on an unmodified head:

```
missing 0.9.8 contributor ledger: expected undefined to be truthy
expected '0.9.7' to be '0.9.8'
expected '# Installing Codewhale…' to contain 'v0.9.8 source candidate'
```

`lib/public-copy.test.ts` wants a `### Contributors` section under the 0.9.8 entry in `CHANGELOG.md`, and the two `lib/public-surface-contract.test.ts` ones want the site's release strings to say 0.9.8. So facts regeneration is necessary but not sufficient — that gate needs the v0.9.8 release content too, and the contributor ledger in particular is yours to write rather than mine to guess at.

Rebased onto `5ac75add4` meanwhile. That picks up #5396, so the four `agy_credentials` failures on **Test (macos-latest)** are gone from this branch's base, and **Test (windows-latest)** gets a fresh run — its previous red was a Chocolatey 504 during NSIS install, not a test.

Verified on `5ac75add4` in the same session: `check:facts` still exits 1 without #5398, and vitest is still 10 failed / 250 passed with the same three files.

Hmbown (approved):
Verified as a copy-only zh-Hans consistency fix matching the established TUI term 宪章. No structural, English, or zh-Hant changes. Land after #5398 so the inherited facts check is repaired.

PR #5398, opened 2026-08-15
fix(web): regenerate facts.generated.ts for the two v0.9.8 providers

Lstarsky0:
`check:facts` has been failing on `main` since v0.9.8, and it sits inside **Lint & Type Check** — a required check, so every web pull request inherits the red no matter what it touches. My #5397 is currently red on exactly this and nothing else.

`53af08ce3` registered Google Gemini and `531b64ed9` registered Google Antigravity; `web/lib/facts.generated.ts` was never regenerated. Same pair that left the CLI assertion counts behind in #5383, and the same family as the `docs/PROVIDERS.md` row #5394 refreshed — this looks like the remaining artifact from those two commits.

This is the output of `cd web && npm run prebuild`, which is what the checker prints as the remedy. `derive-facts.mjs` reads the Rust sources directly and needs no build.

```
[derive-facts] version=0.9.8 crates=21 providers=44 sandboxes=2 default-model=deepseek-v4-pro node=>=18 tools=74 license=MIT
```

The diff is the two provider rows plus `generatedAt`. The script deliberately preserves that stamp when no *checked* fact has moved, so it changing here is the signal that one did.

### Checks

Branched from `ad102ec44`. Gates run twice, identical both rounds:

| Gate | exit |
| --- | --- |
| `npm run check:facts` | 0 — was 1 on the unmodified head |
| `npm run check:locales` | 0 |
| `npm run check:docs` | 0 |
| `npm run lint` | 0 |

`npx vitest run` stays at 10 failed / 250 passed, and the failing set is byte-identical to the same suite on an unmodified `ad102ec44` — seven Cloudflare deploy-preflight tests that need credentials, plus one `public-copy` and two `public-surface-contract` parity tests. This regeneration does not address those three; they look like the changelog/release-facts side of the same v0.9.8 drift, and I did not want to bundle a guess with a mechanical regeneration.

No-Issue: unbreaks the `check:facts` gate on main; the drift itself is already covered by the v0.9.8 follow-ups.

### DIFF
--- a/web/lib/facts.generated.ts
+++ b/web/lib/facts.generated.ts
@@ -27,7 +27,7 @@ export interface RepoFacts {
 }
 
 export const FACTS: RepoFacts = {
-  "generatedAt": "2026-08-14T15:23:06.794Z",
+  "generatedAt": "2026-08-15T05:57:01.838Z",
   "sourceRevision": null,
   "sourceCommittedAt": null,
   "version": "0.9.8",
@@ -243,6 +243,16 @@ export const FACTS: RepoFacts = {
       "id": "mistral",
       "label": "Mistral AI",
       "env": "MISTRAL_API_KEY"
+    },
+    {
+      "id": "google",
+      "label": "Google Gemini",
+      "env": "GOOGLE_API_KEY / GEMINI_API_KEY"
+    },
+    {
+      "id": "antigravity",
+      "label": "Google Antigravity",
+      "env": "ANTIGRAVITY_API_KEY / AGY_ADC_AUTH"
     },
     {
       "id": "telecomjs",


Lstarsky0:
This clears `check:facts` but the gate stays red, and the remaining three are worth listing together because they're the same v0.9.8 drift and one of them isn't mine to write.

**Lint & Type Check** runs `check:facts` then `vitest`. With facts regenerated it gets past the first step and fails on:

```
missing 0.9.8 contributor ledger: expected undefined to be truthy
expected '0.9.7' to be '0.9.8'
expected '# Installing Codewhale…' to contain 'v0.9.8 source candidate'
```

Traced to three stale artifacts:

| Where | Now | Should be |
| --- | --- | --- |
| `docs/public-surface-facts.json` → `sourceCandidate.version` | `0.9.7` | `0.9.8` |
| `docs/public-surface-facts.json` → `sourceCandidate.providerCount` | `42` | `44` |
| `docs/INSTALL.md:11` | `**v0.9.7 source candidate**` | `v0.9.8` |
| `CHANGELOG.md` 0.9.8 entry | no `### Contributors` section | one is needed |

One thing that took checking and is easy to get backwards: **`latestPublishedRelease` should stay at v0.9.7.** `v0.9.8` is a tag but not a GitHub release — `gh release list` still shows v0.9.7 as latest. So the matrix's `published <= sourceCandidate` invariant is satisfied by moving only the candidate side; touching the published side would make the site advertise a release that doesn't exist, which is the exact thing that assertion guards.

The two `docs/` rows are mechanical and I can add them here — `sourceCandidate` derives from `Cargo.toml` (0.9.8) and the provider registry (44, same Gemini + Antigravity pair as the facts file). Say the word.

The contributor ledger I've deliberately not touched. `AGENTS.md` puts credit under CI enforcement with canonical identities in `.github/AUTHOR_MAP`, and I'd be guessing at who belongs in a release I only partly worked on. That one needs you regardless of what I do with the rest.

Hmbown (dismissed):
Verified as a narrowly generated facts refresh: Google Gemini and Antigravity now match the Rust provider registry, and check:facts/check:docs pass. The remaining web-test red is pre-existing v0.9.8 public-release documentation drift being handled separately.

Hmbown (changes requested):
The generated file is mechanically consistent with the current registry, but its public meaning is not yet correct: `antigravity` is presently a Cloud Code credential/transport path, not a first-class advertised outbound provider. The v0.9.8 candidate therefore excludes Antigravity from public provider facts, includes Google Gemini and Ollama Cloud, and keeps the truthful provider count at 44. Please regenerate from that corrected filter (or let the already-harvested candidate close this PR with credit) before merge.

PR #5414, opened 2026-08-16
docs(readme): replace the star-history chart with a fan-out clip

Hmbown:
GitHub restricted third-party access to star data, so the star-history chart at the bottom of the README no longer renders. This replaces it with a 9-second clip (assets/fanout.gif) of Codewhale fanning out three read-only scout subagents.

https://claude.ai/code/session_018fa1QQwShWaaXqo8UTuPmp

No-Issue: docs-only visual polish (README hero art swap); no behavior change to track


Hmbown:
Thank you for pushing this cleanup. I rechecked the current `main` tree and this PR now has a zero-file diff: the intended content is already byte-identical on `main`. I’m closing it as already landed so the queue reflects reality; no contributor work is being discarded.

PR #5415, opened 2026-08-16
fix(tui): truncate session titles by char count, not byte offset

buiducnhat:
---
## Summary

`short_title_truncate` (used by `derive_session_title` to name sessions) located its natural phrase boundary with `str::rfind`, which returns a **byte** offset, then fed that offset to `chars().take()` as a **character** count. That is harmless for ASCII, but for non-ASCII first prompts the byte offset exceeds the character count, so the cut landed past the requested limit and mid-word.

This computes the boundary as a character index by scanning a `Vec<char>` instead, and adds unit tests for ASCII, non-ASCII (CJK), punctuation, and no-boundary cases.

## Testing

Ran (scoped to the touched crate `codewhale-tui`):

- [x] `cargo clippy -p codewhale-tui --all-features --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::collapsible_if -A clippy::assertions_on_constants`
- [x] `cargo test -p codewhale-tui --lib --locked frame::tests::` — 5 passed, 0 failed

I did not run the full `--workspace --all-targets` gate locally; CI will run it.

## Checklist

- [x] Updated docs or comments as needed (comment explains the byte/char distinction)
- [x] Verified TUI behavior manually if UI changes (N/A — no UI changes)
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address (N/A — single-author commit)

### DIFF
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Session titles truncate by character count, not byte offset, so
+  multi-byte titles (CJK, emoji) cut at the intended width and word
+  boundary instead of past the limit (#5415).
 - Wide terminals and tmux panes fill the full available width again for the
   transcript and composer (#5322). The brief v0.9 session-shell side gutter is
   gone so expanding a pane rematerializes layout the same way shrinking does.
--- a/crates/tui/src/tui/ui/frame.rs
+++ b/crates/tui/src/tui/ui/frame.rs
@@ -1203,11 +1203,14 @@ pub(crate) fn short_title_truncate(text: &str, max_chars: usize) -> String {
     if text.chars().count() <= max_chars {
         return text.to_string();
     }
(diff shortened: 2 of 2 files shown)


Hmbown:
Verified locally on top of your branch:

- All five new `frame::tests` pass, including the CJK regression case (`你好 world and more` → `你好 world…`).
- `cargo fmt --check` clean; no new clippy findings from the diff.
- The remaining clippy `collapsible_if` notes in CI output pre-exist on `main` and are allow-listed there.

I added the missing `Unreleased` changelog entry to your branch (`7bdb37b3`) — that was the only gap. This fix is approved from our side and is being folded into the v0.9.8 integration train. Thank you for the clean repro and tests!

Hmbown:
Landed in v0.9.8 via the harvested char-boundary title truncation fix (changelog receipts in the release train; your CJK regression case `你好 world…` is covered by the frame tests that shipped). Closing with thanks — your analysis of the byte-offset boundary was exactly right. If you spot the truncation misbehaving on 0.9.8, please reopen with the string and we'll take it from there.

Issue #5056, opened 2026-08-02
Test reliability: flaky verifier background tests, /workspace-sensitive fixtures, 12 untriaged #[ignore] tests


Hmbown:
run_verifiers_background_advertises_detached_start and run_verifiers_background_starts_shell_jobs_and_returns_task_ids still flake under full-suite parallelism (crates/tui/src/tools/verifier.rs:1302, :1490; documented in crates/tui/AGENTS.md:80). The /workspace-sensitive subagent tests write fixtures to repo_root.parent(), unwritable when repo_root is /. And 12 #[ignore] tests across the workspace remain untriaged (qa_pty x2, release_runtime_qa, runtime_threads, web/guard, engine x4, persistence_actor, mcp, settings). Fix or properly quarantine the flaky pair, pin fixtures to TMPDIR, and triage each ignore to keep-intentional or fix-and-unignore.

From v0.9.4-issues-recon P1 item 5.

Hmbown:
Refreshed on `origin/main` at `48180d993e5a795986d4f313c6c3696960592c63`:

- both `run_verifiers_background_*` tests remain; the schema-advertisement test is pure/deterministic, while the detached rustc test still owns the process-wide env lock and remains documented as a full-suite sharp edge in `crates/tui/AGENTS.md:80-81`
- the `/workspace` failure remains real at `tools/subagent/tests.rs:6840`: `tempdir_in(repo_root.parent())` can target unwritable `/` when the checkout is `/workspace`; the test intent is valid (fixture must sit outside the current repo), so the fix is to create the outside-repo fixture under a writable temp root, not delete the test
- all **12** ignored tests now have explicit reasons: 3 cross-process child helpers, 5 one-shot metric/receipt generators, 1 heavy 32-worker QA storm, 1 live-TCP race, and 2 PTY isolation/flakiness cases

Classification: keep the intentional helpers/metrics ignored; repair or quarantine the TCP/PTY/background flakes; move the outside-repo fixture to a writable temp boundary. No deletion was made in the dead-code PR.

wuisabel-gif:
I'm taking the narrow verifier-test slice: remove the background test's rustc/rustup and process-wide HOME dependency while preserving coverage that a detached verifier starts, returns a task id, and completes. I'll add provider-free evidence and leave the remaining ignored-test inventory for separate follow-ups.

wuisabel-gif:
Opened PR #5423 for the narrow background-verifier test isolation slice:
[...]

PR #5456, opened 2026-08-17
feat(sandbox): bwrap container essentials + configurable extra roots (#5410)

### LINKED ISSUE #5410: Allow to configure additional roots in the bwrap sandbox
## Problem

I do some software development with Zig using CodeWhale + deepseek. Turning on the bwrap sandbox gave me some failures ("access denied")  because

- Redirection output to `/dev/null` is forbidden, which makes some tools the model want to run fail
- Linking with system libraries fails because `/usr/lib` and `/usr/lib64` are not readable

The failures go away as soon as I turn off the sandbox (`prefer_bwrap = false`). The mounted paths are controlled by the sandbox policy, but there is no way to add additional roots to the `bwrap` command.

## Proposed solution

Add 2 new config keys, e.g.

```
bwrap_ro_roots = ["/usr/lib64","/usr/lib","/lib"]
bwrap_dev_roots = ["/dev/null"]
```

`bwrap_ro_roots` allows to configure additional read-only roots, and `bwrap_dev_roots` adds new read-write device nodes.

## Use case

The main use case is system software development were access to files outside the workspace is needed.

## Alternatives considered

- Redirecting output to `/dev/null` is fairly often generated by the models. This could be added to the `workspace-write` policy by default, without making it configurable.
- Adding a new sandbox policy, e.g. `system-development`, which is like `workspace-write` but adds additional read-only roots like `/lib`, `/usr/lib`, `/usr/lib64`, `/usr/include`. Obvious problem is to determine list of additional root.
- Combining the approaches: create a new sandbox policy `system-development`, which adds write access to `/dev/null` by default, and this is the only policy which allows to add new additional read-only roots.

## Impact

I currently turn off the sandbox when I encounter these "Access Denied" errors. With the proposed change I would never turn off the bwrap sandbox.

## Additional context

I used CodeWhale + deepseek to produce a draft implementation. If that is interesting: https://github.com/redstar/CodeWhale/tree/redstar/bwrapconfig

Hmbown:
## Summary
Linux bwrap sandbox: mount private `--dev /dev`, `--proc /proc`, `--tmpfs /tmp` container essentials by default (fixes the original #5410 report — `open(O_WRONLY)` on `/dev/null` failing with EROFS under the read-only root bind), and add configurable extra bind roots: `bwrap_ro_roots` (extra read-only binds, applied after writable binds so they can narrow) and `bwrap_dev_roots` (host device nodes only — canonicalized and type-checked to char/block devices so the key cannot become a writable-root escape). Non-Linux `resolve()` is a no-op stub. Documented in docs/SANDBOX.md.

Closes #5410

## Verification
- Reviewer pass over the full diff; second commit fixes a Linux-only compile error (`#[test]` on a helper with arguments), a copy-pasted doc comment, and adds the docs/SANDBOX.md section.
- `cargo fmt --all -- --check` clean.
- Targeted tests run individually pass locally on macOS; the combined `-- sandbox` filter hangs locally on **main too** (pre-existing local macOS test-hang, noted on #5056), so the authoritative gate is CI.

## Not verified
- Linux bwrap behaviour end-to-end (macOS host); Linux-gated tests run only in CI.
- Behavioural edge (for maintainer eyes): `--tmpfs /tmp` is inserted before policy writable binds, so under workspace-write with host `/tmp` bound rw the tmpfs is shadowed; under read-only policy a workspace located under `/tmp` would fail `--chdir`. Not changed here.

https://claude.ai/code/session_012ZU7GQ3ygtcTugBtzqRDMj


Hmbown:
Superseded by #5462 (same commits, bot co-author trailer removed for the credit lint).

PR #5480, opened 2026-08-17
feat(tui): show and open the live /rc session link; send a stable device id

Hmbown:
## Summary

The `/rc` banner never told users where the session lives. This PR makes the TUI surface, print, and open the live web session link, and stops minting a new "computer" per `/rc`.

- **Session link**: parse optional `runner.runUrl` / `runner.computerUrl` from the `POST /api/local-runners/connect` lease (the cwc lane adds them; older control planes omit them and the terminal shows no link — it never invents one). Links are validated like `validate_authorization_url`: `https://app.codewhale.net`, no credentials/port/fragment; `runUrl` must be `/session?run=<the attached runId>`, `computerUrl` must be `/settings[?section=…]`. Invalid links are dropped, not fatal.
- **Banner + note**: `REMOTE CONTROL · <runUrl> · /rc stop returns input here` (falls back to the account/runner receipt when no link), plus a transcript note "Remote control is live at <url> — run /rc open …". `/rc status` appends `open <url>`.
- **`/rc open`** launches the link via the existing `crate::utils::open_url`; **`/rc link`** prints it (and the computer link).
- **Stable device id**: a machine-stable `device_<uuid>` is persisted in its own keychain slot (`cwc-remote-control-device-v1`), adopted from an existing enrollment on first upgrade, and sent as `deviceId` on `device/start` and `connect`, so the control plane merges every folder from one terminal into a single computer.
- **`target_ref` = sha256(workspace)** (session still travels as `sessionRef`), so repeated `/rc` in one folder reuses one grant instead of adding a grant per session (the server caps grants per computer at 32).

## Tests (unit, in `crates/tui/src/remote_control.rs` and the `/rc` command)

- lease with / without `runUrl`+`computerUrl`; off-origin, other-run, credentialed, ported, fragment, extra-query and non-string links are dropped without failing the attachment
- banner text with and without a link; link notice mentions `/rc open` and `/rc link`
- controller exposes links only while connected and clears them on stop
- persisted device identity round-trips through serde, a saved id wins, an enrollment id is adopted, a broken record is replaced
- `/rc link` and `/rc open` return an error (not an invented URL) before the session is live

Not built here (cargo builds were disallowed in this lane); `cargo fmt --all -- --check` passes. Please run `cargo test -p codewhale-tui remote_control` before merge.

No-Issue: owner request 2026-08-17 — the /rc banner never told users where the session lives

### DIFF
--- a/crates/tui/src/commands/groups/session/remote_control.rs
+++ b/crates/tui/src/commands/groups/session/remote_control.rs
@@ -10,9 +10,14 @@ use super::CommandResult;
 pub(in crate::commands) const COMMAND_INFO: CommandInfo = CommandInfo {
     name: "rc",
     aliases: &["remote-control"],
-    usage: "/rc [status|stop]",
+    usage: "/rc [status|link|open|stop]",
     description_id: MessageId::CmdRemoteControlDescription,
 };
+
+/// Shown by `/rc link` and `/rc open` before the control plane has advertised
+/// a session link (not connected yet, or an older control plane).
+const NO_LINK_MESSAGE: &str =
+    "Remote control has no live session link yet; run /rc to hand this session to the web first.";
 
 pub(in crate::commands) struct RemoteControlCmd;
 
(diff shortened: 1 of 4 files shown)


Hmbown:
Behavior note (product call, taken): the target ref is now the folder alone, so two Codewhale sessions running `/rc` in the same folder on the same machine share one lease — the newer connect takes over and `/rc stop` in either marks that folder offline. One live remote-control session per folder per machine is the intended model (a folder has one controller); a visible "another session controls this folder — taking over" notice is a small follow-up. Reviewer's finding acknowledged; not a blocker.

PR #5486, opened 2026-08-17
fix(tui): hide the session metrics strip on compact rows

Lstarsky0:
## Summary

Below 60 columns the phase strip stands down the working detail, the cache chip and status toasts, but keeps painting the session metrics strip. A 59-column row renders:

`▌· idle │ LLM 3.5s │ Cache hit 99% │ Input 9.3M     F1:keys`

`SessionMetrics` is in `default_footer()`, so that is the out-of-the-box narrow terminal.

The metrics block is the only optional element on that row without a `tier != ShellTier::Compact` guard — it decides on leftover columns instead, so it takes whatever the elements that did stand down left behind. Added the guard.

The same gap looks like what reddened `Test (macos-latest)` on `9d2a6e4c5`, the v0.9.9 merge commit: `qa_pty::semantic_activity_motion_crosses_reasoning_reading_and_tool_use_in_a_real_unix_pty` asserts a sub-60 semantic row carries no `×` and no `s`, and saw `"▌⣤ reading │ LLM 1.1s"` — that `s` is the metrics segment. I could not reproduce that one here: it passes on this machine both with and without the guard, so the connection is read off the frame text rather than measured.

`session_metrics_strip_sheds_groups_on_narrow_rows_and_never_truncates` pins width 60, which `for_chrome_width` classes as `Normal` (`< 60` is compact), so it is unaffected.

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features --locked` — clean
- [ ] `cargo test --workspace --all-features --locked` — ran `cargo test -p codewhale-tui --lib` and `cargo test -p codewhale-config -p codewhale-protocol` instead, plus `check-dead-code-budget.py` and `check-tui-locale-parity.py`. The diff is one line and one test in `codewhale-tui`.

`session_metrics_strip_is_hidden_when_compact` fails on unmodified main with the row quoted above and passes with the guard; taking the guard back out brings the same failure back.

`--lib` is 10710 passed / 1 failed here. The failure is `tools::pdf::tests::shared_adapter_forwards_page_window_and_returns_stdout`, which fails identically on a stashed clean tree on this machine — same one I measured on #5355, not something this branch introduces.

## Checklist

No-Issue: found from the Test (macos-latest) failure on 9d2a6e4c5, the v0.9.9 merge commit.

### DIFF
--- a/crates/tui/src/tui/phase_strip.rs
+++ b/crates/tui/src/tui/phase_strip.rs
@@ -243,7 +243,7 @@ pub fn render(area: Rect, buf: &mut Buffer, app: &mut App) {
     // columns are genuinely free after the phase marker, the ledger chips, a
     // floor for any live toast, and the key hints, and sheds its
     // lowest-value groups to fit rather than truncating a number.
-    if metrics_enabled {
+    if metrics_enabled && tier != ShellTier::Compact {
         let snapshot = crate::tui::session_metrics::snapshot_from_app(app);
         if !snapshot.is_empty() {
             let toast_reserve = status_toast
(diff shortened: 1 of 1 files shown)


Lstarsky0:
Unrelated to the diff: `.github/APPROVED_CONTRIBUTORS` says to seed active contributors there before the gate workflows switch to enforce mode, and I'm not among the 57 entries. `pr:Lstarsky0` is the line, if you want it in before the switch.

Issue #5337, opened 2026-08-12
Web: finish the #4934 dictionary spine — retire every isZh branch and inline { en, zh } module


Lstarsky0:
## Problem

#4934 established one dictionary path for every routed locale (chrome + home read `getChrome(locale)` / `getHome(locale)`), and the code comments in `web/lib/i18n/config.ts` mark the remaining page bodies as "awaiting the same move". On current `main` that remainder is still large:

- **40 files** under `web/app/[locale]/`, `web/components/`, and `web/lib/` still branch on `isZh` (docs pages, community, contribute, constitution, faq, feed, digest, admin, shared components, …)
- 2 inline `{ en, zh }` content modules remain in `web/lib/content/` (`getting-started.ts`, `vocabulary.ts`)

Consequences:

1. Every partial locale (ja / vi / ko / ru / uk / es / pt-BR / id) is locked out of those page bodies — the dictionary layer routes them, but the copy is hardcoded into a two-language ternary, so there is nowhere to put a translation without editing TSX.
2. en/zh copy edits happen in page markup rather than dictionaries, outside the `check:locales` / `dictionaries.test.ts` gates — which is exactly how terminology drift like the 宪法/宪章 mismatch (#4949) survived.

## Proposal

Migrate page-by-page to the dictionary architecture, as a series of small behavior-preserving PRs (one page group per PR):

1. **Phase 1 — shared surfaces**: `docs/layout.tsx` and shared components still branching on `isZh`
2. **Phase 2 — docs pages**: one PR per small group (`docs/guide`, `docs/fleet`, `docs/mcp`, …), each adding the page's dictionary shape under `web/lib/i18n/dictionaries/<code>/` for en + zh, with every other locale falling back to English exactly as today
3. **Phase 3 — content modules + long-tail pages** (`getting-started.ts`, `vocabulary.ts`, community, contribute, faq, feed, digest, admin)

Per-PR gates: `npm run check:locales`, `npm test`, `npm run lint`; rendered en/zh output stays identical (pure move, no copy edits mixed in — any wording fix goes in its own commit/PR).

End state: zero `isZh` branches outside the i18n layer; partial locales gain page-body coverage by dictionary drop-in, matching how chrome/home already work; all site copy sits behind the parity gates.

## Offer

I'm volunteering to run this as a tracked series — I'll keep a phase/page checklist in this issue, one PR per group, rebased onto `main` per the repo's cadence. Two questions before the first PR:

[...]

PR #5503, opened 2026-08-18
test(web): spawn the deploy preflight script by a decoded path

Lstarsky0:
## Summary

`web/lib/deploy-preflight.test.ts` spawns `scripts/check-cloudflare-deploy-env.mjs` by `script.pathname`. `URL.pathname` stays percent-encoded, so in a checkout under a path with non-ASCII characters the spawned filename does not exist, node exits on module-not-found, and the 7 cases in the file that spawn the script fail on the exit code. The other 5 read `package.json` and `.github/workflows/web.yml` directly and pass.

Nothing about the environment is wrong when this happens, but the failures read as missing Cloudflare credentials — I reported them that way myself in #5488, which is what sent me looking.

`fileURLToPath` decodes it. That is already the idiom in `web/scripts/`, for example `check-locales.mjs` and `check-docs.mjs`.

## Testing

My checkout sits under `开源Ds`, so it reproduces directly.

- `npm test` before: 264 passed / 7 failed, all 7 in `lib/deploy-preflight.test.ts`
- `npm test` after: 267 passed / 0 failed
- `npx tsc --noEmit` — exit 0
- `npm run lint` — exit 0

Both directions measured rather than reasoned about. A clean worktree of the same base on an ASCII path passes 267/267 unpatched; copying that same untouched worktree to a non-ASCII path fails the same 7, which isolates the path as the only variable. The patched file passes 12/12 on both paths.

Full sequence run twice on an identical file hash.

## Checklist

- [x] Added or updated tests where relevant — no new test; this fixes an existing one
- [ ] Verified TUI behavior manually if UI changes — no TUI change

No-Issue: no open issue tracks this; found while measuring the test baseline for the #5337 series.

### DIFF
--- a/web/lib/deploy-preflight.test.ts
+++ b/web/lib/deploy-preflight.test.ts
@@ -1,11 +1,15 @@
 import { spawnSync } from "node:child_process";
 import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import { describe, expect, it } from "vitest";
 
-const script = new URL("../scripts/check-cloudflare-deploy-env.mjs", import.meta.url);
+// fileURLToPath, not `.pathname`: the latter stays percent-encoded, so a
+// checkout under a path with non-ASCII characters spawns a filename that
+// does not exist and every case here fails on a module-not-found exit.
+const script = fileURLToPath(new URL("../scripts/check-cloudflare-deploy-env.mjs", import.meta.url));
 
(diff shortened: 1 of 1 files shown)


Hmbown:
Thank you for the careful non-ASCII-path reproduction and the before/after measurements. This is a clean portability fix, so I folded it into the v0.9.9 release head while that lane is already rebuilding.

PR #5504, opened 2026-08-18
feat(web): move docs/hooks and docs/troubleshooting onto the dictionary spine (#5337)

Lstarsky0:
## Summary

Continues the #5337 series after #5488. `docs/hooks` and `docs/troubleshooting` were the two smallest remaining page bodies at 12 `isZh` branches each, and both are now on the dictionary spine.

Before this, every string on both pages lived in a two-language ternary. The 16 partial locales route to those pages but had nowhere to put a translation without editing TSX, and the copy sat outside `check:locales` and `dictionaries.test.ts` — the gap that let the 宪法/宪章 drift in #4949 survive.

The copy moves verbatim into `web/lib/i18n/dictionaries/<code>/`, en plus zh, behind `DocsHooksDict` and `DocsTroubleshootingDict`. Both pages read one dictionary; every other locale resolves to the English reference exactly as the ternaries did. The `[name, detail]` lists follow the tuple shape `home.ts` already uses for `workflow` and `surfaces`.

One case needed more than a whole-string move. The hooks overview sentence typesets `[[hooks.hooks]]`, `/hooks` and `[hooks].enabled` as inline `<code>`. Those are config syntax, not copy, so they stay in the page as a lookup and the sentence carries a `{token}` for each — which also puts the sentence under the token-parity half of `check-locales.mjs` for free. `splitToken` handles one token; `splitTokens` splits a template on every token it carries and returns the literal runs and token names interleaved, so a locale that reorders them still renders correctly.

No copy edits are mixed in. Before running the gates I checked each moved string against the pre-move page source read out of git, so a transcription slip would have shown up as a miss rather than as a wording change; the rendered comparison below is the shipped evidence.

## Testing

The web gates, in the order `.github/workflows/web.yml` runs them:

- [x] `npm run check:facts` — OK
- [x] `npm run prebuild`
- [x] `npm run check:docs` — PASS
- [x] `npm run check:locales` — PASS, `zh/docs-hooks.ts: 10/10 keys — complete`, `zh/docs-troubleshooting.ts: 10/10 keys — complete`, 16 locales absent and falling back for each
- [x] `npm test` — 264 passed / 7 failed
- [x] `npm run lint` — exit 0, no output
- [x] `npx tsc --noEmit` — exit 0
- [x] `npm run build` — exit 0, 512/512 static pages

Full sequence run twice on an identical file hash for all ten changed files.

Rendered output is unchanged, measured the same way as #5488. Building this branch and a clean worktree of the same base, 36 pages — 18 locales × `/docs/hooks` and `/docs/troubleshooting` — compare identical on visible text and on `<title>` + meta description. Raw bytes differ on every page, but only in chunk hashes and the RSC nonce, which differ between any two builds.

As a control, three one-character edits — one inside an `events` detail, one in the tokenized sentence, one inside a zh `incidents` detail — produce 18 differences under the same comparison, so it covers the array path and the `splitTokens` path, not just the plain strings.

### Correction to #5488

I wrote there that the 7 `lib/deploy-preflight.test.ts` failures need Cloudflare deploy credentials. That was wrong, and the clean base worktree here is what showed it: the base passes 267/267, so the failures are not a property of `main`.

The cause is `lib/deploy-preflight.test.ts:8` spawning the script by `script.pathname`. `URL.pathname` stays percent-encoded, so in a checkout under a path with non-ASCII characters — this one is under `开源Ds` — the spawned filename does not exist and the 7 cases in the file that spawn it exit on module-not-found. Copying the untouched base worktree to a non-ASCII path reproduces the same 7. `fileURLToPath` is the fix, and it is what the scripts under `web/scripts/` already do; sending it as its own PR (#5503) rather than folding it in here.

So the honest count for this branch: 267 on the base, 271 here, 4 of them new, and the 7 failures are that path bug rather than anything on either branch.

The Rust gates in the template do not apply — this PR touches only `web/`.

## Checklist

- [ ] Verified TUI behavior manually if UI changes — no TUI change

No-Issue: one page group of the #5337 series — the epic stays open until the last group lands.

### DIFF
--- a/web/app/[locale]/docs/hooks/page.tsx
+++ b/web/app/[locale]/docs/hooks/page.tsx
@@ -1,120 +1,66 @@
+import { Fragment } from "react";
+import { getDocsHooks, splitTokens } from "@/lib/i18n/dictionaries";
 import { buildPageMetadata } from "@/lib/page-meta";
+
+/**
+ * Config syntax the overview sentence typesets as inline `<code>`. These are
+ * literals, not copy, so they stay here rather than in the dictionaries —
+ * `configIntro` carries a `{token}` for each one.
+ */
+const CODE_SPANS: Record<string, string> = {
+  hooksTable: "[[hooks.hooks]]",
+  hooksCommand: "/hooks",
+  enabledKey: "[hooks].enabled",
+};
 
 export async function generateMetadata({ params }: { params: Promise<{ locale: string }> }) {
   const { locale } = await params;
-  const isZh = locale === "zh";
+  const t = getDocsHooks(locale);
   return buildPageMetadata({
     path: "/docs/hooks",
     locale,
-    title: isZh ? "钩子 · Codewhale 文档" : "Hooks · Codewhale Docs",
-    description: isZh
-      ? "已发布的生命周期钩子：可变 message_submit、tool_call_before 决策、turn_end 与子 Agent 观察事件。"
-      : "The shipped lifecycle hooks: mutable message_submit, tool_call_before decisions, turn_end, and sub-agent observer events.",
+    title: t.metaTitle,
+    description: t.metaDescription,
   });
 }
 
 export default async function HooksPage({ params }: { params: Promise<{ locale: string }> }) {
   const { locale } = await params;
-  const isZh = locale === "zh";
-  const bodyClass = isZh
-    ? "text-ink-soft leading-[1.9] tracking-wide"
-    : "text-ink-soft leading-relaxed";
-  const events = isZh
-    ? [
-        {
-          name: "message_submit（可变）",
-          detail:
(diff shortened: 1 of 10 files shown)


Hmbown:
Thank you for continuing the #5337 dictionary spine with such careful rendered-output controls—and for separating the non-ASCII preflight fix into #5503. I’m keeping this larger docs slice for v0.9.10 while we close v0.9.9 on runtime-only blockers. The remaining macOS check is a dead runner, not a requested code change; maintainers will rerun it after the release lane clears.

Lstarsky0:
The macOS check wasn't a dead runner — job 95649939407 finished at 09:41:48, 46 seconds after your comment, conclusion success, so all 14 are green and nothing needs a rerun.

v0.9.10 is fine; I'll hold docs/constitution + docs/runtime-api until this one lands so the shared files don't move under it.

Issue #5430, opened 2026-08-16
Mid-session /rename and /title don't apply properly


Hmbown:
## What's wrong

Renaming a session **in the middle of a live session** — via `/rename <name>` or `/title <name>` — does not behave properly. The name/tab title doesn't take, or doesn't stick, instead of applying the way it does when the session is at rest.

## Why this is one bug, not two

`/title` is a discoverable alias backed by the existing session name — `commands/groups/session/title.rs` delegates straight into `rename::rename(...)`. So whatever is broken in the mid-session path breaks both commands; we'll fix it once, on the shared path, and make sure both entry points are covered by tests.

Context worth knowing: this surface got several fixes in the v0.9.8 train (sanitizing titles at every setter rather than only at the OSC 0 write, repainting the resting tab title on `/title`, keeping the first line of legacy multi-line titles in listings), and the terminal-title work merged around then may be involved in the mid-session regression. We're aware it still isn't working properly and it's queued for v0.9.9.

cc @SparkofSpike — you worked on the terminal window title side of this; flagging so you know it's seen, tracked, and will be fixed on the shared path in 0.9.9. Extra detail welcome if you've seen the failure shape yourself (when exactly the title reverts, terminal + tmux or not, etc.).

## Acceptance

- `/rename x` and `/title x` mid-turn and mid-session both apply immediately: session listing, tab/OSC 0 title, and the next launch all show the new name.
- PTY regression test covering both entry points through the live event loop, not just the setter functions.

SparkofSpike:
Oh! Thank you for that – I hadn't noticed this issue before. I'll do more testing on this feature later.

PR #5509, opened 2026-08-18
fix(tui): restore /title as an independent terminal window title (#5430)

SparkofSpike:
## Summary

`/title` and `/rename` were merged in `24c7dee46` into one command: `/title` became a discoverable alias that delegates straight into `rename::rename(...)`, so both commands changed the single session name shown in the composer, the picker, and the terminal tab. That made the two concepts collide, and this PR splits them again.

They are two different things. The session name is the durable identity you search for later; the tab/window title is the short glanceable label for a terminal window. Consider two parallel projects: one session named "Translate English docs to Chinese" and another named "develop a translate web". Those names are far too long to sit in front of the terminal title — as an OSC-0 prefix they crowd out the state that matters (`Codewhale`, `reasoning…`, `using tool…`, `done`), so you can no longer see the program's state at a glance.

Why not just shorten the session names — call both of them "Translate"? Because the two roles have opposite constraints. The tab title *wants* to be short: while you are actively developing, a single word like `translate`, `novel`, or `bot` is instantly recognizable and perfectly fine as a window label. But the session name is what you use to find a conversation again later — and if it is a one-word label like "translate", retrieving that AI record becomes a disaster. You cannot search for anything, and every one-word session looks identical. So the short identity belongs on the tab, and the long descriptive identity belongs in the session list. `/title` owns the former, `/rename` owns the latter, and neither should ever touch the other.

This restores the independent design originally introduced in `8b6bcc5e0` and removed by `24c7dee46`, while keeping everything that landed since: title sanitization at every setter, the OSC-0 bidi/control filtering, the resting-title repaint, and the mid-first-turn recovery from #5444 (which is now shared by `/title` as well, so both commands apply immediately on a live first turn).

## Changes

- **Independent `/title` command**: `/title <name>` sets the per-session tab/window title and persists it on the saved session; `/title off` clears the session-level override (falling back to the config default); `/title` with no argument reports the current effective title and its source. `title.rs` no longer delegates to `rename` — it has its own write path mirroring the `/rename` semantics (Work-state snapshot, live metadata sync, `needs_redraw` so the OSC-0 repaint lands immediately, even mid-turn).
- **`SavedSession.window_title` persistence**: new optional field stored with `serde(default, skip_serializing_if)`, so old session files load unchanged. Restored on session load, cleared on `/new`, `/fork`, and `/clear`, and carried through `build_session_snapshot` so autosave cannot drop it.
- **`title` config key**: root config key (profile overlays merge it) supplies the default prefix for every session in a config scope, editable at runtime via `/config title <name> [--save]`. `config.example.toml` documents it.
- **Precedence**: session `/title` override > `title` config default > no prefix. `window_title_prefix()` reads only `window_title`/`title_default` — it no longer derives from the session name, and `/rename` no longer repaints the tab prefix.
- **`/rename` receipt reverted**: back to `Session renamed to "..."` (it no longer claims "and terminal tab renamed").
- **Mid-first-turn recovery shared**: `live_session_before_first_snapshot` in `rename.rs` is now `pub(crate)` and reused by the `/title` write path, so `/title` mid-first-turn works from the crash checkpoint (or rebuilt App state) exactly like `/rename` (#5430).
- **Tests**: 8 new unit tests in `title.rs` (set/clear/report, sanitization, oversized rejection, persistence, and two mid-first-turn recovery scenarios), plus updated notifications/underwater prefix tests. The PTY regression test `release_rename_title_mid_turn_and_mid_session_apply_and_survive` was rewritten for the independent semantics: `/title` mid-first-turn, `/rename` mid-turn (proving it leaves the tab prefix and `window_title` untouched), `/title` mid-turn, at-rest `/rename` + `/title off` + `/title`, the `/sessions` listing, and a fresh relaunch restoring both the name and the window title.
- **Localization / docs**: `CmdTitleDescription` restored to the independent meaning across all 15 locale packs; `crates/tui/CHANGELOG.md` Unreleased entry updated; `config.example.toml` documents the `title` key.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [x] Tests

## Testing

- [ ] `cargo clippy --workspace --all-targets --all-features --locked` (not run in full; see note)
- [x] `cargo test -p codewhale-tui --lib -- title` — 57 passed, 0 failed
- [x] `cargo test -p codewhale-tui --lib -- rename` — 21 passed, 0 failed
- [x] `cargo test -p codewhale-config -p codewhale-protocol`
- [x] `cargo build --release -p codewhale-tui -p codewhale-cli`
- [ ] PTY regression test — file is `#![cfg(unix)]`; verified syntactically with `rustfmt` on Windows, must run on macOS/Linux CI

Note: a full `cargo test -p codewhale-tui --lib` run shows 21 failures, all verified on the rebase base (via stash) to be pre-existing Windows environment/parallelism issues (project_context reads the real user HOME, bash-dependent read-only-shell tests, and a few flaky layout tests) — none touch the title/rename surface touched here. `cargo clippy --workspace` was not run in full on Windows.

## Checklist

- [ ] Verified TUI behavior manually if UI changes (author ran a release build; final manual check pending)

## Related Issues

No-Issue: But refs #5430 — restores the independent `/title` design that the mid-session fix (#5444) kept merged into `/rename`; both commands still satisfy the #5430 acceptance (immediate apply mid-turn/mid-session, persistence, next-launch survival), now as two separate concepts.


SparkofSpike:
The CI kept failing. I've told my agent to loop when I'm sleeping and it give me a result now. It seems that the pr should be merged manually. Hi @Hmbown, PTAL.

PR #5520, opened 2026-08-20
feat(web): move docs/sandbox and docs/web onto the dictionary spine (#5337)

Lstarsky0:
Next group in the #5337 series, after #5517. `docs/sandbox` and `docs/web` carried 14 and 15 `isZh` branches; both are now zero. Same shape as before: two dictionaries per page, `types.ts` and `index.ts` wired, and both files added to `check-locales.mjs`'s `OPTIONAL_FILES` so zh is held to key and token parity while the other sixteen locales fall back to English exactly as the ternaries did.

The `sandbox_mode` values that the policies paragraph typesets as inline `<code>`, and the commands and flags on the browser-client page, stay code-owned and reach the dictionaries as `{token}`s. Two of the sandbox tokens appear twice in the same sentence, which `splitTokens` handles but `check-locales.mjs` cannot see — it compares token sets — so the unit test asserts the full ordered list instead.

The platform rows are the one place I did not use a code-owned key. Their headings are translated (`Linux · opt-in bubblewrap` / `Linux · 可选 bubblewrap`), so they are copy, and the row key stays the name as it was before. The test only asserts the count for that list.

Rendered output is unchanged. I compared all 36 pages (18 locales × 2) against a clean worktree at the same base, and this time kept the inline tags and their `class` attributes in the comparison instead of flattening every tag to a space. That matters: the flattened form cannot tell `、<code>danger-full-access</code>` from `、 <code>danger-full-access</code>`, and the zh policies sentence has a space on one side of a code span and not the other, because of where the old JSX wrapped its lines. The difference is empty. As a control, three one-word edits — one in a token sentence, one in an array row, one in a plain string — do get reported.

`npm run lint`, `npm test` (312 passing) and `npm run check:locales` are clean.

No-Issue: one page group of the #5337 series — the epic stays open until the last group lands.


Lstarsky0:
The `Test (windows-latest)` failures here are not from this branch. `main`'s own run 32334932574 fails the same job with the same 11 panicking tests — the set difference against this PR's run is empty, and this PR touches no Rust at all, only `web/`.

The earlier `Version drift` red was a timing artifact of the same kind: that run was created at 05:14:27Z and 1d140927e landed at 05:15:29Z, so it merged against `main` between `de823c672` making cancellation authoritative and the test expectations being settled. It has been rebased since.

Lstarsky0:
`Test (macos-latest)` is the same origin. I checked `main`'s tip (1d140927e) out into a clean worktree on a local macOS box — nothing from this branch in the tree — and ran the eleven tests that failed here: ten fail there too, same `file:line`, same messages. Only `tui::onboarding::tests::progress_counts_only_required_decisions` passes locally. The clearest of the group is `project_context.rs:1827`, which compares `/private/var/…/AGENTS.md` against `/var/…/AGENTS.md` — macOS `TMPDIR` sits under the `/var` symlink, and the assertion has a canonicalized path on one side and `tempdir().path()` on the other.

Separately: the reason a TypeScript-only PR runs the Rust workspace on macOS and Windows at all is that the `changes` case has no `web/*` arm, so all ten files fall through to the fail-safe `*)` and set `heavy=true`. Everything covering `web/` is already ungated — `web.yml` has no `paths:` filter, and `check-locales.mjs` in `lint` is gated only on `event_name != 'schedule'` — and no crate includes anything from the root `web/` (the embedded client is `crates/tui/src/runtime_web/`, a different tree). Happy to send that one-line arm if you want it.

Lstarsky0:
`Version drift` was mine. The new `check-feature-release-notes.sh` wants a changelog receipt for every issue a `feat:` commit references, and my commit body names #5517, which merged after v0.9.10 was cut and never got an entry. Added one under `[Unreleased]` plus its `sync-changelog.sh` slice — `check-versions.sh` and the eight release-helper contract tests are green locally.

`Lint & Type Check` is not mine: `main`'s own run at d53617f1e fails the same single case, `vocabulary.test.ts > describes the first session truthfully`, on the same `docs/GUIDE.md` assertion.

PR #5521, opened 2026-08-20
chore(tui): drop a single-argument concat!

Lstarsky0:
`main` at 3b221c7a6 still fails `Lint` on one thing:

```
error: unneeded use of `concat!` macro
  --> crates/tui/src/runtime_handoff.rs:83:43
   = note: `-D clippy::useless-concat` implied by `-D warnings`
```

The replacement is clippy's own suggestion. The two-argument `concat!` right below it is untouched.

This branch started out covering the rustfmt and dead-code failures as well; 5689f0dde and 4d99f30da landed those first, and the rebase dropped my formatting commit as already applied, so this is what is left. `RUSTFLAGS=-Dwarnings cargo clippy -p codewhale-tui` is clean afterwards and fails with exactly the error above on a clean checkout of 3b221c7a6.

No-Issue: unbreaking a red `main`, no tracking issue.

### DIFF
--- a/crates/tui/src/runtime_handoff.rs
+++ b/crates/tui/src/runtime_handoff.rs
@@ -80,9 +80,8 @@ const RESTORED_COMPLETIONS_HEADER: &str = "[Codewhale restored sub-agent checkpo
 const RESTORED_RUNNING_HEADER: &str = "[Codewhale restored sub-agent runtime checkpoint]";
 const RESTORED_TOPOLOGY_HEADER: &str = "[Codewhale restored Agent topology checkpoint]";
 
-const AGENT_TOPOLOGY_EVENT_PREFIX: &str = concat!(
-    "<codewhale:runtime_state kind=\"agent_topology\" schema=\"v1\" visibility=\"internal\">\n",
-);
+const AGENT_TOPOLOGY_EVENT_PREFIX: &str =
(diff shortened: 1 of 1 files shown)


Lstarsky0:
There is a second break on `main` at a6df64af8, and that one, not fmt, is what the `Test` jobs are failing on:

```
error: constant `ORCHESTRATION_DISCOVERY_COMMANDS` is never used
  --> crates/tui/src/commands/traits.rs:60:18
   = note: `-D dead-code` implied by `-D warnings`
error: could not compile `codewhale-tui` (lib) due to 1 previous error
```

83299d6b4 moved the production use over to `BARE_SLASH_DISCOVERY_COMMANDS`. The only reference left is `commands/mod.rs:1106`, inside `#[cfg(test)] mod tests`, so the lib target sees the constant as dead. `RUSTFLAGS=-Dwarnings cargo check -p codewhale-tui` fails on a clean checkout of a6df64af8, with or without `--tests`.

I kept it out of this PR because the fix is a choice rather than a formatting run: drop the constant together with the `assert_eq!` that pins it, or put `#[cfg(test)]` on it to match where it is actually used. Say which one and I will push it here.

Hmbown:
Thank you — your diagnosis and fix are exactly right. This same `clippy::useless_concat` repair landed on `main` as f7896c87f while the branch was being rebased, so there is no remaining change to merge. Closing only because it is already applied; I appreciate the precise CI reproduction and the care you took to leave the real two-part `concat!` untouched.

Issue #5261, opened 2026-08-07
v0.9.5: extract the engine into `crates/core` + Thread/Session split


Hmbown:
**Depends on:** I1
**Unblocks:** I4, I5, I6 (shared `crates/core` boundary).
**Folds in:** #3312 (ui `run_event_loop`), #3313 (`RuntimeThreadManager` split),
#3952 (chat-client request building), #4022 (CLI/TUI parity); follow-on #1004 (`/dryrun`).

### Goal
Move the turn loop, session, thread manager, the TUI `run_event_loop`, and the chat
client's request-building out of the TUI crate into `crates/core`. Introduce
`ThreadId`/`SessionId` and a thread manager that can start a session with no TUI
attached. Drive every consumer — TUI, CLI, app-server, tests — through one
`Op`-in / `EventMsg`-out API so headless and TUI produce byte-identical model
requests.

### Why
codewhale's `Session` is really a thread: today only the TUI can build a prompt or
drive a turn, so headless `codewhale exec` cannot exist without a child process. The
turn loop itself is proven — move it, don't rewrite it. This is the extraction that
unblocks I4–I6 and dissolves the largest single source files in the tree
(`runtime_threads.rs` ≈ 7,133 lines, `client.rs`/`chat.rs` ≈ 5k/4.8k, `ui.rs`
`run_event_loop`), which is also the heart of #5249.

### Current state (codewhale)
- `crates/tui/src/core/turn.rs` (`TurnContext`), `crates/tui/src/core/session.rs`
  (`AppendLog`, `PrefixStabilityManager`, `frozen_prefix`, `messages_revision`).
- `crates/tui/src/core/engine/turn_loop.rs` (`handle_deepseek_turn`, steer/subagent
  drains, `refresh_system_prompt()`, `should_compact`/`compact_messages_safe`,
  `MessageRequest` build, parallel tool exec, `StuckGuard`/`ReadRepeatGuard`/
  `ToolCallBudget`, stream retry budget).
- `crates/tui/src/core/engine.rs` (`spawn_engine`, `spawn_supervised`) — runs in a
  background tokio task.
- `crates/tui/src/runtime_threads.rs` — `RuntimeThreadManager` (≈4,900-line impl,
  `monitor_turn` ≈1,035 lines), `RuntimeThreadStore`, persisted JSON state
  (`<root>/{threads,turns,items,events}` + `state.json`).
[...]

PR #5523, opened 2026-08-20
refactor(tui): extract tool call stages from turn loop

bistack:
## Summary

- Extract tool-call planning into `plan_tool_calls`.
- Extract approval and execution into `execute_planned_tools`.
- Extract result projection into `process_tool_results`.
- Preserve the original control order, mutable state flow, cancellation behavior, and indexed outcome collection.
- Document each stage inputs, outputs, and side effects.

No-Issue: This is a behavior-preserving refactor

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or -A clippy::collapsible_if -A clippy::assertions_on_constants`
- [x] Full `cargo test --workspace --all-features --locked` coverage, split only to isolate an existing process-global provider-auth test interaction: the monolithic run passed 10,857 tests and failed the three Kimi provider-picker tests; those same three tests passed 3/3 in a fresh isolated process.

## Checklist

- [x] Updated docs or comments as needed.
- [x] Added or updated tests where relevant — existing engine tests cover this behavior-preserving refactor; no new behavior was introduced.
- [x] Verified TUI behavior manually if UI changes — not applicable; no UI behavior changed.
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address — not applicable; no harvested or co-authored code.

### DIFF
--- a/crates/tui/src/core/engine/turn_loop.rs
+++ b/crates/tui/src/core/engine/turn_loop.rs
@@ -20,6 +20,12 @@ use codewhale_core::request::{PrimaryTurnRequest, prepare_primary_turn_request};
 
 const MAX_APPROVAL_INTENT_SUMMARY_CHARS: usize = 2_000;
 
+struct PlannedToolCalls {
+    plans: Vec<ToolExecutionPlan>,
+    hook_contexts: std::collections::HashMap<String, String>,
+    batch_sandbox_policy: crate::sandbox::SandboxPolicy,
+}
+
 struct StreamOutcome {
     current_text_raw: String,
     current_text_visible: String,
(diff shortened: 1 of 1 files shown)


Lstarsky0 (commented):
Checked the behavior-preserving claim rather than take it on faith. Sliced `run_turn` out of both sides — `22b5b3f3d` lines 544–3591 against the same span plus the three new methods on `24a174b38` — stripped indentation, comments and blank lines, and diffed. 241 lines differ out of ~2,500. Set aside the three signatures, their call sites and the `PlannedToolCalls` construction and what's left is rustfmt reflowing after the indentation dropped, `&x` → `x` where the parameter is already a reference, and `mode` → `*mode`. The one block that looks moved is the tail of `run_turn` from `pending_steers` onward, and that's a file-position artifact of the extracted bodies now sitting below it — `pending_steers` still runs immediately after result processing.

The cut points are where this kind of extraction goes wrong quietly, since a local still live past the cut turns into a copy and nothing complains. Planning ends at 2618, and the four locals born inside it (`batch_approval_mode`, `active_tools_at_batch_start`, `deferred_tools_hydrated_this_batch`, `deferred_tools_hydrated_in_order`) are never read again in 2619–3591; the three that are read later are `plans` at 2624, `batch_sandbox_policy` at 3115 and `hook_contexts` at 3481, which is exactly what `PlannedToolCalls` carries. `mode` holds up the same way — assigned at 571, 618, 2005, 2691, 3244, 3280, and the three that have to escape the phase all sit inside what became `execute_planned_tools`, which takes `&mut`. `cargo test -p codewhale-tui --lib --locked` on 24a174b38 is green here too, 10,845 passed.

One note and one question. The `mode: AppMode` vs `&mut AppMode` asymmetry is right as of this commit, but it's the failure the split newly makes possible: if planning ever grows an `apply_pending_runtime_authority` call, the refreshed mode dies at the return and nothing catches it. A clause on the parameter saying the planning phase doesn't refresh authority would be cheap insurance.

The question is about #5261 rather than about this diff. `crates/core/src/engine/thread/executor.rs` says `run_turn` is slated to move into `codewhale-core` file-by-file so the diff stays reviewable, and this lands in the same function from the other direction. Worth a word from whoever owns that epic on whether the phase split makes the move easier or just changes what has to be moved.

bistack:
@Lstarsky0 Thanks for the detailed check.

Mode note: agreed. It didn't make it into the merged commit — needs a follow-up MR, noted.

#5261: already completed, so nothing to coordinate. The split should help the move — PlannedToolCalls makes the surviving locals explicit, so the three methods + struct move as one unit.

Issue #5526, opened 2026-08-20
Deprecated shell completion


RepentStar:
I'm using pwsh. I discovered that codewhale can use `codew completions powershell` to generate completion scripts, but its content is outdated, and the triggering command is still `codewhale-tui`.

I didn't see any relevant help in the doc, nor did I find any places to change in this repo.

Where can I modify the auto-completion script? Thanks.

wuisabel-gif:
Hi @Hmbown, I would like to take a focused slice of #5526.

My proposed scope is:

- make the legacy `codewhale completions <shell>` command use the canonical completion generator;
- ensure generated scripts reference the public `codewhale` command rather than `codewhale-tui`;
- update the CLI help and PowerShell examples;
- add provider-free CLI tests for Bash, Zsh, Fish, and PowerShell;
- preserve the legacy command as a compatibility alias rather than removing it outright.

I would keep unrelated CLI command changes out of the first PR. Would this be an acceptable slice to take?

Hmbown:
> Hi @Hmbown, I would like to take a focused slice of #5526.
> 
> My proposed scope is:
> 
> - make the legacy `codewhale completions <shell>` command use the canonical completion generator;
> - ensure generated scripts reference the public `codewhale` command rather than `codewhale-tui`;
> - update the CLI help and PowerShell examples;
> - add provider-free CLI tests for Bash, Zsh, Fish, and PowerShell;
> - preserve the legacy command as a compatibility alias rather than removing it outright.
> 
> I would keep unrelated CLI command changes out of the first PR. Would this be an acceptable slice to take?
> 

Absolutely! Thank you so much!

Hmbown:
> I'm using pwsh. I discovered that codewhale can use `codew completions powershell` to generate completion scripts, but its content is outdated, and the triggering command is still `codewhale-tui`.
> 
> I didn't see any relevant help in the doc, nor did I find any places to change in this repo.
> 
> Where can I modify the auto-completion script? Thanks.

Thank you for reporting this!

Hmbown:
Verified shipped on main during 0.9.12 triage. Evidence: e3256bab7 fix(cli): render shell completions from the CLI's own command tree; b7158ed39 docs(install) — the pwsh/codewhale-tui staleness is resolved on main.

PR #5530, opened 2026-08-21
fix(cli): route legacy completions through public binary

wuisabel-gif:
## Summary

Addresses the approved scope of #5526.

The legacy `codewhale completions <shell>` command now uses the same canonical completion generator as `codewhale completion <shell>` instead of forwarding to the `codewhale-tui` runtime. Generated scripts use the public `codewhale` command name.

The legacy plural command remains available as a compatibility alias. The canonical singular command is unchanged.

## Coverage

Added provider-free CLI tests that verify:

- `codewhale completions powershell` parses directly in the CLI;
- Bash, Fish, PowerShell, and Zsh scripts contain `codewhale`;
- generated scripts do not teach the retired `codewhale-tui` binary name.

## Verification

- `git diff --check`
- `cargo test -p codewhale-cli --lib --locked` (217 passed)

No provider credentials or network access are required.

No-Issue: This PR addresses the approved compatibility slice of #5526 without automatically closing the broader report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   restating it never changed anything. Shipped binaries were never affected;
   `release-artifacts.yml` builds `--profile dist` with fat LTO and
   `codegen-units = 1`.
+- The legacy `codewhale completions <shell>` command now generates scripts
+  from the public `codewhale` CLI instead of delegating to `codewhale-tui`, so
+  PowerShell and other shell completions no longer teach the retired binary
+  name (#5526).
 
 ## [0.9.10] - 2026-08-19
 
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -385,8 +385,12 @@ New integrations should prefer `codewhale app-server`.")]
         after_help = "The browser receives a one-time loopback bootstrap capability, never the Runtime token.\nThe capability is exchanged for a bounded, process-local HttpOnly, SameSite=Strict web session and then invalidated."
     )]
     Web(WebArgs),
-    /// Generate shell completions.
-    Completions(TuiPassthroughArgs),
+    /// Generate shell completions. Compatibility alias for `completion`.
+    Completions {
+        /// Shell to generate completions for.
+        #[arg(value_enum)]
+        shell: Shell,
+    },
     /// Configure provider credentials.
     Login(LoginArgs),
     /// Remove saved authentication state.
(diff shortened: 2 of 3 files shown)


wuisabel-gif:
CI update:

- Version drift failed because the root changelog entry for #5526 needed the generated `crates/tui/CHANGELOG.md` sync. That is now fixed in commit `95df027f1`, and the checks have restarted.
- The Rust Clippy job reports five Clippy 1.98 findings in existing files outside this PR: `commands/groups/core/voice.rs`, `llm_client/mod.rs`, `tui/ui/event_loop.rs`, and `tui/views/status_picker.rs` (two findings). None of the completion-related files are involved.

The completion implementation and all 217 local CLI library tests pass. Please let me know whether you would prefer those unrelated baseline lint cleanups in a separate PR or included as a mechanical follow-up here.

Lstarsky0 (commented):
Built both sides and diffed the scripts the two versions actually emit. On `75dca2cfb` the PowerShell script registers for `codewhale-tui` and carries the runtime's 29 subcommands; on `95df027f1` it registers for `codewhale` and carries the CLI's 42. Sixteen of those are new — `config`, `run`, `thread`, `web`, `metrics`, `update` and friends were never completable before, which is the bigger win here than the binary name.

Three go the other way. `pr`, `scorecard` and `session-diagnostics` were in the old script, aren't in the new one, and are still typeable: `codewhale` hands anything it doesn't recognise to the runtime — a made-up subcommand gets the same "requires an interactive terminal" error that `codewhale pr --help` does — and all three are real commands (`Pr` at `crates/tui/src/lib.rs:323`, `Scorecard` at `:342`, `SessionDiagnostics` at `:276`). So they still work in a terminal and no longer complete. Defensible if the script is meant to describe the public CLI surface and nothing else, but it isn't visible from the diff.

Smaller: `every_completion_shell_uses_the_public_binary_name` lists Bash, Fish, PowerShell and Zsh by hand and misses Elvish, the fifth `clap_complete::Shell` variant. `Shell::value_variants()` covers all of them and keeps covering them. And `let _ = io::stdout().write_all(&output)` throws the write result away where `run()` could `?` it.

`cargo test -p codewhale-cli --lib --locked` on 95df027f1 is 217 passed here too.

Hmbown:
Thanks @wuisabel-gif — this slice is now superseded by the v0.9.11 integration merged in #5542 (`main` @ 8ef856f27cbcee758e48e65af436d87e7d9da15c). The merged fix goes a bit further than this PR: `codewhale completion <shell>` (with `completions` as a visible alias) renders in-process from the CLI's own command tree, registers both `codewhale` and `codew` for Bash/Zsh/Fish/PowerShell/Elvish, and is pinned by per-shell tests (`generated_completion_scripts_register_the_published_command_names`, `completions_is_an_alias_for_completion` in `crates/cli/src/lib.rs`). Your scoping of #5526 is credited in the v0.9.11 CHANGELOG. Closing as superseded rather than merging, since `crates/cli/src/lib.rs` now conflicts; please reopen or open a follow-up if you see a shell the merged version still misses.

PR #5544, opened 2026-08-22
feat(web): move docs/subagents and docs/mcp onto the dictionary spine (#5337)

Lstarsky0:
Next group in the #5337 series, after #5520. `docs/subagents` and `docs/mcp` carried 16 and 18 `isZh` branches; both are now zero. Same shape as the earlier groups: two dictionaries per page, `types.ts` and `index.ts` wired, and both files added to `check-locales.mjs`'s `OPTIONAL_FILES` so zh is held to key and token parity while the other sixteen locales fall back to English exactly as the ternaries did.

Two things here are new.

The eight sub-agent roles are identifiers rather than copy — `worker`, `scout`, `planner` and the rest read the same in both branches today. So the page keeps the list and its order in a `ROLE_ORDER` const and the dictionary keys each row by role name, carrying only the prose. That is the `docs/VOICE.md` code-owned rule applied to a list instead of a code span, and it lets the test compare row keys across locales rather than only the count, which is all the sandbox platform rows allow.

`docs/mcp`'s h1 is the literal `MCP` in every locale, so it stays in the page and there is no `overviewTitle` to translate. The shared key-parity test asserted `get("zh").overviewTitle` differs from the reference, which for this dictionary would have compared `undefined` against `undefined` and passed while checking nothing. It now takes the probe key per entry, `metaTitle` for this one.

The 36-page comparison caught one difference before this landed, the same class as the one in #5517. The old JSX wrapped a line right after the `coordination_contracts` span:

```
或 <code className="inline">coordination_contracts</code>
值；重叠的共享写声明会在任何改动之前失败。
```

JSX drops leading whitespace that carries a newline, so the live page has no space before `值`. The dictionary reproduces that, with a comment saying where it comes from — a move that also fixes typography is no longer a move.

Rendered output is otherwise unchanged. All 36 pages (18 locales x 2) compared against a clean worktree at the same base, keeping the inline tags with their `class` attributes rather than flattening them, and dropping React's `<!-- -->` text-node separators so a sentence built from one interpolated array compares equal to the same sentence built from several JSX children. The difference is empty. As a control, three one-word edits do get reported: one in a token sentence and one in an array row hit 17 pages each, and one in a zh string hits 1 — which is the fallback semantics, seventeen locales reading English and zh reading its own.

`npm run lint`, `npm test` (313 passing) and `npm run check:locales` are clean.

Remaining after this group: `docs/work`, `docs/vocabulary`, `docs/tools`. `docs/modes`, `docs/configuration` and `docs/fleet` moved in c07794747.

No-Issue: one page group of the #5337 series. The tracker was closed on 08-20 — happy to reopen it, or to keep carrying the remaining groups in the PR descriptions, whichever you prefer.

### DIFF
--- a/web/app/[locale]/docs/mcp/page.tsx
+++ b/web/app/[locale]/docs/mcp/page.tsx
@@ -1,160 +1,89 @@
+import { Fragment } from "react";
+import { getDocsMcp, splitTokens } from "@/lib/i18n/dictionaries";
 import { buildPageMetadata } from "@/lib/page-meta";
+
+const CODE_SPANS: Record<string, string> = {
+  configPath: "~/.codewhale/mcp.json",
+  legacyConfigPath: "~/.deepseek/mcp.json",
+  configPathOption: "mcp_config_path",
+  configEnvVar: "DEEPSEEK_MCP_CONFIG",
+  serversKey: "mcpServers",
+  initCommand: "codewhale mcp init",
+  mcpCommand: "/mcp",
+  toolNamePattern: "mcp_<server>_<tool>",
+  gitServer: "git",
+  statusTool: "status",
+  gitStatusTool: "mcp_git_status",
+  serveMcp: "codewhale serve --mcp",
+  mcpServerCommand: "codewhale mcp-server",
+  addSelfCommand: "codewhale mcp add-self",
+  serveHttp: "serve --http",
+};
+
+function withCodeSpans(template: string) {
+  return splitTokens(template).map((part, i) =>
+    "token" in part ? (
+      <code key={`${i}-${part.token}`} className="inline">
+        {CODE_SPANS[part.token] ?? `{${part.token}}`}
+      </code>
+    ) : (
+      <Fragment key={`${i}-text`}>{part.text}</Fragment>
+    ),
+  );
+}
 
 export async function generateMetadata({ params }: { params: Promise<{ locale: string }> }) {
   const { locale } = await params;
-  const isZh = locale === "zh";
+  const t = getDocsMcp(locale);
   return buildPageMetadata({
     path: "/docs/mcp",
     locale,
-    title: isZh ? "MCP · Codewhale 文档" : "MCP · Codewhale Docs",
-    description: isZh
-      ? "通过 Model Context Protocol 消费外部工具服务器，或把 Codewhale 作为 MCP 服务器暴露。"
-      : "Consume external tool servers over the Model Context Protocol, or expose Codewhale itself as an MCP server.",
+    title: t.metaTitle,
+    description: t.metaDescription,
   });
 }
 
 export default async function McpPage({ params }: { params: Promise<{ locale: string }> }) {
   const { locale } = await params;
-  const isZh = locale === "zh";
-  const bodyClass = isZh
-    ? "text-ink-soft leading-[1.9] tracking-wide"
-    : "text-ink-soft leading-relaxed";
+  const t = getDocsMcp(locale);
 
   return (
     <section className="space-y-10">
       <section id="overview" className="scroll-mt-32">
         <h1 className="font-display text-3xl mb-1">MCP</h1>
-        <p className={`${bodyClass} mt-3`}>
-          {isZh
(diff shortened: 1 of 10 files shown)


Lstarsky0:
Pushed `8cb5bd6c4` for the type check. Indexing the dictionary union by a per-entry probe key does not compile — none of the members carries an index signature, so `Record<string, unknown>` is not a valid narrowing. Every docs dictionary has `metaTitle` and zh translates all of them, so the loop probes that instead, which covers docs/mcp without a special case. `npx tsc --noEmit` is clean, and it is now in what I run before pushing; it was the gate I was missing.

`Lint` is a separate failure and not from this branch. Four clippy lints in `crates/tui`:

```
chunks_exact_to_as_chunks   crates/tui/src/tui/views/status_picker.rs
manual_slice_fill           crates/tui/src/tui/views/status_picker.rs:156,163
drain_collect               crates/tui/src/tui/ui/event_loop.rs:1156
result_large_err            crates/tui/src/llm_client/mod.rs:1171
```

This branch changes ten files, all under `web/`, and all four sites are byte-identical on `main`. The job runs at all because the change filter is deliberately default-heavy and `web/**` is not on the light list, which is working as intended.

The control is #5537, a Dependabot bump of `docker/setup-buildx-action` with no Rust in it, failing on the same four. `ci.yml:277` installs `dtolnay/rust-toolchain@master` with `toolchain: stable`, so the lint set moves with whatever stable is current that day. `main` itself has not been built since 08-20, so its last run being green says nothing about today.

Happy to send the four fixes as their own PR if that is useful.

Lstarsky0:
Correction on the `Lint` half: those four are already fixed in #5542. I checked its file list with `gh pr view --json files`, which caps at 100 and hid them; the full list is 412 files and includes all three sites, with `self.selected.fill(true)` / `fill(false)` in `status_picker.rs`. Its `Lint` job runs clippy and is green.

So `main` is red until v0.9.11 lands, and the offer above would only duplicate your branch — withdrawn. The toolchain reading was wrong too: your fix explains #5542 being green without needing one.

Hmbown:
Thanks @Lstarsky0 — reviewed against the v0.9.11 `main` (8ef856f27). Two things before this can merge:

1. **Lint & Type Check fails on this PR's own change**: `lib/i18n/dictionaries.test.ts` lines ~272/274 — `(get("zh") as Record<string, unknown>)[probe]` trips TS2352; cast through `unknown` first (`as unknown as Record<string, unknown>`).
2. **Rebase onto current `main`**: v0.9.11 renamed the MCP CLI examples in `web/app/[locale]/docs/mcp/page.tsx` from `codewhale-tui mcp …` to `codewhale mcp …` (8 occurrences). Your restructure of the same hunks still carries the old name — please keep the `codewhale` form on rebase (and in the new dictionaries, which already use 0 of the old name — good).

Also a `Signed-off-by` trailer on the commits, please (DCO is advisory today but we want the trailer on everything that lands). The `Lint` (clippy) failure is pre-existing on the old base and is already fixed on `main`, so it will clear on rebase. Happy to merge once those are in — the dictionary-spine work itself looks right.

Lstarsky0:
Rebased onto 8ef856f27, pushed 84c2b47f. Signed-off-by is on both commits.

**2 (rename)** — all 8 are on the `codewhale` form. They sat in the page, 3 in `CODE_SPANS` and 5 in the `<pre>`, so no dictionary changed: under this split a command rename stays a code edit and never reaches the translations. Diffed the result against main's copy of the page afterwards — same 15 `<code>` literals, `<pre>` byte-identical.

**1 (TS2352)** — already fixed in 8cb5bd6c4; its `Lint & Type Check` went green at 01:40Z, so the run you read was b42c7e73's. I dropped the cast instead of widening it — `as unknown as Record<string, unknown>` compiles, but it turns off the check the assertion exists to make. Every docs dictionary has `metaTitle` and zh translates all of them, so the probe is a plain property access now and nothing needs casting.

Ran the lint job's steps locally on the rebased branch: `check:facts`, `prebuild`, `check:docs`, `test` (332), `lint`, `tsc --noEmit`, `build` — all pass.

PR #5574, opened 2026-08-23
Add Build Remote Agent phone pairing (gbr/1)

LinespottingPrivate:
## What

Optional **Build Remote Agent** pairing device so a phone can spectate this desktop agent.

Protocol stays `gbr/1` (not a fourth pair protocol):

1. Install `gbr-agent` v0.6.0+ from https://grokbuildremote.com/
2. `gbr-agent pair` — browser QR **and** printed 8-char code
3. Phone scans or types the code
4. `gbr-agent run` (leave running)
5. Attach only `http://127.0.0.1:8788` or `gbr-mcp` stdio

Phone is spectator + veto. Orchestration stays on this agent.

Independent product by Linespotting AB. **Not affiliated with xAI or SpaceX.**

Agent source (MIT): https://github.com/LinespottingOrg/GrokBuildRemote-Agents

## Files

Docs / MCP snippet only. No product-runtime change. No mailbox keys.

```bash
curl -fsSL https://grokbuildremote.com/install.sh | bash
gbr-agent version    # need v0.6.0+
gbr-agent pair && gbr-agent run
curl -sS http://127.0.0.1:8788/health
```

### DIFF
--- a/integrations/gbr/README.md
+++ b/integrations/gbr/README.md
@@ -1,0 +1,67 @@
+# Pair a phone with Build Remote Agent
+
+Codewhale can use **Build Remote Agent** as a pairing device: the iOS/Android
+app spectates (and can inject into) this desktop session through the free MIT
+`gbr-agent`. Phone and PC never open ports to each other.
+
+Website: https://grokbuildremote.com/
+Agent: https://github.com/LinespottingOrg/GrokBuildRemote-Agents (MIT)
+Protocol: `gbr/1` · need agent **v0.6.0+**
+
+Independent product. Not affiliated with xAI or SpaceX.
+
+This bundle does **not** replace Codewhale Tailscale Serve, Funnel, or the
+Telegram/Feishu/Weixin bridges. Pairing stays `gbr-agent pair` + `gbr-agent
+run`. Attach is only `http://127.0.0.1:8788` or `gbr-mcp` stdio. Phone is
+spectator + veto.
+
+## Install + pair
+
+```bash
+# macOS / Linux
+curl -fsSL https://grokbuildremote.com/install.sh | bash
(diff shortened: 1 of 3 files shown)


Hmbown:
Thanks for putting this together, and for being upfront about what it is — the "Independent product by Linespotting AB. Not affiliated with xAI or SpaceX" note and the MIT agent source link are exactly the right disclosures to lead with.

We're going to pass on this one, and the reason is scope rather than quality: Codewhale's own docs don't carry setup guides for third-party commercial integrations. Once we document one hosted product we're implicitly curating that space — deciding which ones are listed, keeping install steps and version floors (`gbr-agent v0.6.0+`) current as they change, and answering for them when they break. That's a maintenance commitment and an endorsement we're not in a position to make evenly across everyone who'd reasonably ask.

The good news is that this PR isn't what makes the integration work, and nothing is blocking your users today:

- **MCP is an open surface.** Anyone can attach `gbr-mcp` over stdio, or `http://127.0.0.1:8788`, via `/mcp` without a line of this landing. Your five-step flow works as written right now.
- **Plugins are the supported distribution path.** A plugin bundle can ship the MCP server config, a skill, and its own docs, and users install it by name. That gets you discoverability *and* keeps the instructions somewhere you control and can update on your own release cadence — see `docs/PLUGIN_BUNDLES.md`.

Practically, hosting the guide in your own README is also better for you: you can update it the day `gbr-agent` changes, instead of waiting on a docs PR here.

One note in case it's useful: this PR never got a full CI run — only the GitGuardian check ever executed on it. That's a quirk of how checks were scheduled on this branch, not a signal about your change.

Closing as out of scope, with genuine thanks for the contribution and the care in the writeup. If you build a plugin bundle for it, we'd be glad to look at that.

Issue #5573, opened 2026-08-23
v0.9.12: milestone tracker — start here (pick order)


Hmbown:
## Start here (updated 2026-09-02)

- **Operator handoff:** `codewhale-ops/TAKEOVER-PROMPT-20260902-NIGHT.md` (private ops repo) — founder decisions of 2026-09-01, slice table, gates, PR/reinstall/self-test steps. Product authority: `codewhale-ops/PRODUCT_PRD.md`.
- **Working branch:** `integration/shell-brand-20260902` (worktree `codewhale/tmp/wt-shell-brand`). One mega PR against `main` is being opened from it tonight: shell, brand, fleet, and Operate for 0.9.12. Open issues that belong to that PR carry the `v0.9.12` label and a link comment (2026-09-02 triage: `codewhale-ops/ISSUE-TRIAGE-20260902.md`).
- **Retired branch:** `codex/v0912-integration-20260823` (PR #5576, closed 2026-08-29) is not an ancestor of `main` or of the new integration branch. Its P0/P1/P2 items below were re-landed on `main` one by one (see their closed issues), but the per-session control socket (#5594 / #5533), the #5588 neutrality slice `9ffec9b90`, and FEAT-019 (#5609) are still only on that branch and need re-landing.

## Current status (as of 2026-08-23, kept for history)

Working branch was `codex/v0912-integration-20260823`. Ship target: a released v0.9.12 with the must-fix set below, all verified, full release chain green (docs/tag/assets/npm/website).

## P0 — must-fix before release (money & safety)

| Issue | Item | Size |
|---|---|---|
| #5566 | R1: bound runaway spend (finite max_steps + wall-clock) — **decision needed** | M |
| #5567 | R6: fleet cost/token ceiling | M |
| #5568 | S1: sandbox full-disk read deny-list | M |
| #5569 | S2: hard-link write escape guard | S |
| #5552 | /doctor --fix | M |
| #5557 | /import-claude | M |
| #5543 | durable child approval receipts | M |

## P1 — trust & legibility

| #5569-adj | #5572 (MCP OAuth refresh) | #5553 (/context cost attribution) | #4394 (compaction visibility, T2) |

## P2 — cache & polish

| #5570 (C4 ephemeral breakpoint) | #5571 (C5 extension invariant) | #5554 (OSC 12) | #5551 (block actions) | #5556 (/tutorial) | #5550 (line ranges half-2: picker toggle) | #5549 (named-kill half) |

[...]

Issue #5572, opened 2026-08-23
T4: MCP OAuth expiry looks like a broken server — reactive 401 refresh + login hint


Hmbown:
## Problem
MCP OAuth 401/403 after token expiry is not refreshed in the tool path (oauth.rs) — the server looks broken to the user, and the only recovery is manual (ops IMPROVEMENT-PLAN-0912 T4, self-audit #12).

## Acceptance
- On 401/403: force refresh+retry once, then surface a login hint.
- No secret persists in the log; tests cover refresh-once and fail-closed hint.

Hmbown:
Landed in a1fc9cab0 with test completion in aac706868: reactive 401/403 detection (error_looks_auth_required) force-refreshes once and retries, failures fail closed with a `codewhale mcp login <name>` hint, tokens cleared when refresh itself gets auth-required, and no secret reaches the log (crates/tui/src/mcp/oauth.rs + streamable_http.rs retry path). Tests at oauth.rs and streamable_http.rs cover refresh-once and the fail-closed hint; re-verified in-tree today.

Issue #5555, opened 2026-08-23
Clipboard export: write a last-copy backup file and name it when the clipboard fails


Hmbown:
`crates/tui/src/commands/groups/session/export.rs` (~L219) reports "No file was written" when the clipboard is unavailable (SSH, headless). Always write the export to a predictable last-copy file under the Codewhale home and name that path in the failure message.

_Source: v0.9.11 ops potential audit (codewhale-ops/V0911-POTENTIAL-AUDIT-2026-08-23.md). Post-0.9.11 backlog; no duplicate found in the open issue list at filing time._

Hmbown:
Landed in e9c310f5a: clipboard exports always write the markdown to a predictable last-copy file under the Codewhale home (export.rs `write_last_copy`), and the SSH/headless failure path now names that file (`copy_hint(last_copy)` on both the success and failure arms). Re-verified in-tree today.

Issue #5550, opened 2026-08-23
TUI: @path:START-END line-range mentions and a hidden-file prefix in the file picker


Hmbown:
`crates/tui/src/file_mention.rs` has no range parser (`trim_unquoted_mention` strips a trailing ':'), and hidden files are reachable only by typing a leading '.'. Support `@src/lib.rs:120-160` (attach only that range) and an explicit toggle/prefix for dotfiles in the picker.

_Source: v0.9.11 ops potential audit (codewhale-ops/V0911-POTENTIAL-AUDIT-2026-08-23.md). Post-0.9.11 backlog; no duplicate found in the open issue list at filing time._

Hmbown:
Both halves landed: 6ae8bf119 adds `@path:START-END` line-range mentions (crates/tui/src/tui/file_mention.rs — range split off the token, 1-based inclusive endpoints, attaches only that range; test `media_attachment_references_include_removable_line_ranges`), and 8694fd769 adds the hidden-file prefix — a leading `!` in the @ picker runs a second bounded traversal that includes dotfiles (crates/tui/src/tui/file_picker.rs). Re-verified in-tree today.

PR #5576, opened 2026-08-23
0.9.12 integration: must-fix + UX fixes (work-in-progress)

### LINKED ISSUE #5548: Embedded Web: add app.test.mjs coverage for the workflow dispatch-rejection summary
`crates/tui/src/runtime_web/app.mjs` (~L452-465) renders the "N task dispatch(es) were rejected" summary that v0.9.11 added for #5528, but there is no DOM test for it; the terminal side has `workflow_panel` tests. Add a vitest case in `crates/tui/src/runtime_web/app.test.mjs` that feeds a run with rejected dispatches and asserts the summary and the attention state.

_Source: v0.9.11 ops potential audit (codewhale-ops/V0911-POTENTIAL-AUDIT-2026-08-23.md). Post-0.9.11 backlog; no duplicate found in the open issue list at filing time._

### LINKED ISSUE #5562: [Bug Report] Stale write-claims lock sub-agents out of command execution; verifier role contradicts its own description
## Environment

- CodeWhale runtime: 0.9.10 (build 369868d23e59), Windows 10 22H2 (zh-CN), also reproducible across restarts
- Git: 2.48.1
- Observed: 2026-08-20 to 2026-08-23

## Bug 1: Stale write-claims persist forever and cascade-lock other agents

- After several sub-agent sessions, `coordinate inspect` shows **14 historical write-claims** left behind by finished/cancelled agents. They survive restarts and there is **no way for the root session to clean them** (claim/supersede/reconcile all fail or are unavailable).
- Concrete impact: claim `7e18870f` permanently occupied the directory `tools/picwish-ps/win7-migration-package`. As a result:
  - Spawning a new `builder` with `write_roots` overlapping that tree failed with `write-scope contention` even though no agent was running.
  - A `worker` whose task merely *read* nearby paths had its **bash locked** with `another child is writing in this shared checkout`.
- Expected: claims should be released when the owning agent finishes, or the operator should be able to clear them.

## Bug 2: In-workspace sub-agents can never execute commands ("shared checkout" guard)

- For every in-workspace sub-agent (`worker`, `builder`, `verifier`), all command-execution tools (bash, terminal, task_shell, Run) are denied with:
  `cannot prove a bounded file target, and another child is writing in this shared checkout`
- Reproduced with: a clean directory, a purely read-only command (`dir`), and a **single active child** (no concurrency at all). `write_roots` isolation did not help.
- `spawn`/`read`/`write`/`edit` all work fine; only command execution is gated.
- The denial message suggests the guard treats the stale claims from Bug 1 as active writers, which would make Bug 2 a consequence of Bug 1.

## Bug 3: `verifier` role has no command-execution permission despite its description

- Profile description: *"Runs focused validation - Runs targeted validation and reports receipts"*.
- In practice the verifier cannot run **any** command (Run/tasks/gates all refused), and spent its first 13 turns hunting for an execution path that does not exist.
- It can still produce high-quality evidence review by reading files directly, so we re-scoped it locally to "read-only inspector". Suggestion: either grant it (bounded, read-only) command execution, or fix the description/role contract to say it is read-only.

## Workaround discovered (useful diagnostic hint)

- After `git init` on the workspace and dispatching a child with `worktree=true`, the child in its isolated worktree (`E:\.codewhale-worktrees\...`) could run commands without any lock. This strongly suggests the guard is tied to the shared checkout state (likely the stale-claim bookkeeping), and worktree isolation sidesteps it.

## Requests

1. A mechanism to list and clear stale write-claims (or automatic release when an agent settles/cancels).
2. Clarify/fix the verifier role contract (exec or description).

### LINKED ISSUE #5564: docs: REBRAND.md still lists api.deepseeki.com as a "China fallback" host after #1079 removed it
## Description

`docs/REBRAND.md` line 65, under **What did NOT change**, still presents the typo hostname as a legitimate host:

> - **Hosts**: `api.deepseek.com` (global) and `api.deepseeki.com` (China fallback).

#1079 established that `api.deepseeki.com` is not an official DeepSeek endpoint, and it was fixed as a release-blocker in v0.8.17. REBRAND.md documents the v0.8.41 rebrand, so it postdates that fix and reintroduces the claim in prose.

The code is consistent with #1079 and looks fine — `crates/tui/src/config/models.rs:156` states "DeepSeek's official API host is the same worldwide" and `DEFAULT_DEEPSEEKCN_BASE_URL` routes to the correct default. The remaining references at `crates/tui/src/client/chat.rs:2877-2879` are inside the `deepseek_base_url_supports_strict_tools` capability heuristic and read as deliberate backward compatibility. **This is documentation-only.**

## Why this is worth more than a stale line

`deepseeki.com` is registered to a third party and live:

```
$ dig +short deepseeki.com A
34.42.184.201
$ dig +short api.deepseeki.com A
(no A record)
```

whois `deepseeki.com`: GoDaddy, Creation Date 2025-01-27, Registry Expiry 2027-01-27.

The `api.` subdomain has no A record today, so a user who copies the host fails closed — which is how it currently reads as a harmless typo. If the domain's owner adds that record, anyone still carrying `DEEPSEEK_BASE_URL=https://api.deepseeki.com` sends their `DEEPSEEK_API_KEY` to a host they do not control, and it fails open silently.

## Downstream

`deepseek-ai/awesome-deepseek-agent` carries the same claim in its DeepSeek-TUI guide (`docs/deepseek-tui.md:71` and the zh-CN twin), tracked at deepseek-ai/awesome-deepseek-agent#311. REBRAND.md is a plausible source for it.

## Suggested fix

Drop the host from the line, since it is not in fact something that stayed the same:

> - **Hosts**: `api.deepseek.com`.

Scoping note, since #1079 drew a comment about a PR widening beyond what it should have touched: `docs/CHANGELOG_ARCHIVE.md` also mentions the host, but as a historical record of the compatibility shim. That one is accurate as history and I would leave it alone.

## Environment

- Reviewed at `main`, 2026-08-23
- Documentation only; no version-specific runtime behaviour involved

### LINKED ISSUE #5567: R6: fleet-wide cost/token ceiling — accumulator, admission gate, alert
## Problem
FleetAlert::budget_exceeded() has no caller and no run-level accumulator exists: worker events carry NO token/cost data in the protocol (no usage fields), so a Fleet run cannot be cost-bounded today (ops IMPROVEMENT-PLAN-0912 R6, self-audit #8).

## Acceptance
- Worker event stream carries usage; ledger accumulates per run.
- Configurable ceiling checked at admission and after each usage event.
- FleetAlert::budget_exceeded fired on breach; new admissions refused.
- Tests: ceiling breach stops the run and fires exactly one alert.

### LINKED ISSUE #5568: S1: sandbox reads full disk in every posture — opt-in deny-list
## Problem
Sandbox policy grants full-disk read access in every posture (sandbox/policy.rs has_full_disk_read_access -> true). Even a ReadOnly child can read ~/.ssh and .env files — secret exfiltration by reading, with no mutation at all (ops IMPROVEMENT-PLAN-0912 S1, Codex #4).

## Acceptance
- Opt-in deny-list (Codex FileSystemSandboxPolicy analogue) in the sandbox config.
- ReadOnly sessions never offer workspace-external reads unless explicitly allowed.
- Tests: reading ~/.ssh under ReadOnly fails; honored opt-in still works.

### LINKED ISSUE #5569: S2: hard-link write escape — workspace-write tools must resolve links
## Problem
The File/patch tools in workspace-write postures have no hard-link guard (only credential/plugin paths do). A hard link whose target is outside the workspace bypasses writable-root enforcement (ops IMPROVEMENT-PLAN-0912 S2, Codex #2).

## Acceptance
- Writable-root enforcement resolves hard links (and symlinks) for write/edit/patch targets.
- Writing through a link that escapes the workspace is refused with an honest error.
- Tests: hard-link escape attempt is refused in workspace-write.

### LINKED ISSUE #5571: C5: request-extension invariant — debug-assert prefix continuity between turns
## Problem
Nothing asserts that request N's prefix extends request N-1's except across a declared change; a drift silently kills the provider cache while stability claims hold (ops IMPROVEMENT-PLAN-0912 C5, DSH #5).

## Acceptance
- Debug-assert beside single_turn_loop.rs: same session's request prefix extends the prior one unless a declared change happened.
- Tests: extension invariant holds across a normal turn; declared changes are exempt and documented.

### LINKED ISSUE #5577: DeepSeek session at 842k/1M never auto-compacted — investigate trigger gap
User report 2026-08-23: a DeepSeek session shows 842k/1M context and auto-compaction never fired (threshold default 80% = 800k on a 1M window).

Investigation prompt with ranked hypotheses + instrument plan: codewhale-ops/TAKEOVER-PROMPT-COMPACTION-20260823.md. T1 (meter/trigger alignment) already landed on the 0.9.12 branch; the report may be the pre-T1 inflated display, or the real trigger gap (billed-token wiring / bounded-fragment mismatch / reclaimability floor).

Acceptance: root-cause with exact numbers, fix the real gap, regression test pinning 'a 842k-billed 1M-window session must compact'.

### LINKED ISSUE #5578: Cost surfaces only move at TurnComplete — long agentic turns show no $ at all
## Symptom
During a long agentic turn (many model calls + tool calls), the footer/sidebar cost never moves. In sessions dominated by a few giant turns the user never sees any $ estimate at all (reported live on a DeepSeek session).

## Root cause
All cost accrual happened exclusively in the `TurnComplete` handler (`event_loop.rs`): session totals, cache records, and the priced audit. A CodeWhale turn is the whole agentic run, so nothing accrues until the run ends. The per-step `Event::TurnUsage` receipt (one per model call, emitted by `run_turn`, reviewer consults, and REPL bridge children) already existed but only fed the timing metrics strip — its usage was never priced.

Codex-comparison note: Codex emits a per-model-call TokenCount event and updates its token/cost surfaces immediately; our `TurnUsage` is the equivalent signal and was simply not wired to money.

## Fix
Price each `TurnUsage` receipt against the turn's dispatched route (`active_turn.route.cost_envelope().audit(usage)` — the exact same audit path `TurnComplete` uses) into a display-only `pending_turn_cost`. Displayed total = settled + pending + subagents; `TurnComplete` clears pending and lands the authoritative cumulative price, so nothing counts twice and the #244 high-water mark keeps the display monotonic through the handoff. Pending also cleared defensively at `TurnStarted`.

Regression test: `pending_turn_cost_moves_displayed_total_mid_turn`.

Related: #5577 (compaction trigger investigation on the same session), longer-term event-architecture lane (per-domain reducers, Codex-style).

### LINKED ISSUE #5579: Plugin UX parity with Claude Code: proactive recommendations, reload discoverability, hot-reload
User ask (0.9.12 cycle): make plugins as strong as Claude Code's — 'reload-plugins, recommending plugins, all those kinds of things'.

## Current state (audited 2026-08-23)
`crates/tui/src/plugins/` + `/plugin` command group already cover: discovery (user + workspace dirs), marketplace subcommands, install/update/uninstall, trust/enable/disable/revoke, `/plugin reload` (rediscover + skill-cache refresh), `/plugin suggest <task>` (local metadata ranking), export, kimi import.

## Parity gaps vs Claude Code
1. **Proactive recommendation** — `suggest` is user-invoked only. Claude Code surfaces relevant plugins/skills in-context. Wire suggestion ranking into task start (model-visible or footer hint), respecting the KV-cache contract (append-only, never in the pinned prefix).
2. **Reload discoverability + hot-reload** — `/plugin reload` exists but nothing tells the user when on-disk plugins changed; consider a file-watch → 'plugins changed, /plugin reload to apply' nudge, or safe auto-rediscovery between turns.
3. **Recommendation quality** — ranking is substring-ish over manifest metadata; consider model-assisted matching (cheap route) with the local-only registry constraint kept.
4. **Skill invocation parity** — verify plugin-provided skills surface to the model with descriptions as first-class as native skills.

Tracker: #5573.

### LINKED ISSUE #5582: Workflow owner snapshots collapse Degraded into Completed
## Problem

Current `main` projects a degraded workflow owner as completed:

```rust
let state = match record.status {
    WorkflowRunStatus::Running => OwnerState::Running,
    WorkflowRunStatus::Completed | WorkflowRunStatus::Degraded => OwnerState::Completed,
    WorkflowRunStatus::Failed => OwnerState::Failed,
    WorkflowRunStatus::Cancelled => OwnerState::Cancelled,
};
```

Source at the inspected tip (`8ef856f27`):
https://github.com/Hmbown/CodeWhale/blob/8ef856f27cbcee758e48e65af436d87e7d9da15c/crates/tui/src/tools/workflow.rs#L183-L189

The run record and report retain the degraded status, but owner-level consumers receive `OwnerState::Completed`. A dashboard, supervisor, completion counter, or automation path that reads the owner snapshot can therefore treat a partial/degraded workflow as an ordinary success.

This is distinct from:

- #5035, which made parallel dispatch failures explicit in workflow results.
- #5528, which tracks visibility of dispatch/schema failures.

Both can retain detailed failure truth while this projection still collapses the terminal state.

## Why this matters

A real dogfood workflow produced partial or failed reviewer slots while the orchestration surface still reached a completed-looking terminal state. The operator needs to distinguish:

- completed and accepted,
- completed with partial/degraded output,
- failed,
- cancelled.

Keeping the distinction only in a deeper report is insufficient for external supervision and release/acceptance automation.

## Expected behavior

Prefer one of these contracts:

1. Add `OwnerState::Degraded` and preserve it end to end; or
2. Keep the generic owner state non-successful and attach a required typed workflow terminal-status field that every owner consumer must inspect.

A degraded run must not increment ordinary successful-completion metrics or satisfy a success-only wait condition.

## Suggested tests

- A `WorkflowRunStatus::Completed` record projects to completed.
- A `WorkflowRunStatus::Degraded` record does **not** project to ordinary completed.
- Workflow panel, owner snapshot, API/runtime projection, and persisted reload agree on the distinction.
- A mixed parallel run with a dropped/failed slot produces a degraded terminal receipt and cannot satisfy a success-only completion gate.

## Environment

- Inspected upstream: `main` at `8ef856f27cbcee758e48e65af436d87e7d9da15c` (`0.9.11` source)
- The same collapse is present in tags `v0.9.9` and `v0.9.10`.

### LINKED ISSUE #5583: Workflow responseSchema failures need bounded repair and raw-output receipts
## Problem

A workflow task using `responseSchema` fails the run when the child returns prose or malformed JSON. Codewhale correctly surfaces the schema failure instead of silently returning a null successful result, but it discards the opportunity for a bounded repair and does not provide the malformed child output as durable evidence.

Observed dogfood run:

```text
status: Failed
child_agents: 9
phase: Independent checkpoint scoring
error: task(): responseSchema was set but the reply is not valid JSON:
       expected value at line 1 column 1
```

Current source has tests that schema failures surface rather than becoming null slots:

- `parallel_surfaces_response_schema_errors_instead_of_null`
- `pipeline_surfaces_response_schema_errors_instead_of_null`

https://github.com/Hmbown/CodeWhale/blob/8ef856f27cbcee758e48e65af436d87e7d9da15c/crates/workflow-js/tests/vm_tests.rs#L519-L590

That is the correct fail-loud baseline. The missing layer is bounded recovery and evidence.

This is distinct from:

- #5035, which concerns task-option authoring and dispatch failures before a child runs.
- #5528, which concerns operator visibility of workflow dispatch/schema failure states.

This report concerns post-dispatch model output that exists but does not validate.

## Expected behavior

When a child response fails `responseSchema` parsing or validation:

1. Preserve a bounded preview of the raw response and, when larger, a durable artifact reference.
2. Distinguish JSON parse failure from JSON Schema validation failure in the typed receipt.
3. Support one bounded repair attempt (or an explicit workflow option enabling it) that asks the same route to return only corrected JSON against the original schema.
4. Charge the repair against the task/run budget and honor cancellation and wall-time limits.
5. If repair still fails, terminate as a schema failure with both attempts visible; never convert it to null, degraded success, or ordinary completion.
6. In `parallel()`, retain structured per-slot schema failures while allowing explicitly configured partial-success behavior.

## Why this matters

Structured reviewer fan-outs are especially vulnerable: one model wrapping JSON in commentary can abort a high-budget run after all other reviewers have completed. A single bounded repair is cheaper than re-running the entire workflow and gives the operator enough evidence to diagnose whether the schema, prompt, or model route was at fault.

## Suggested tests

- Prose-wrapped valid JSON repairs successfully in one attempt.
- Malformed JSON that remains malformed fails after the configured limit.
- Valid JSON that violates the schema receives a validation-specific repair receipt.
- Raw output preview/artifact is bounded and survives journal reload.
- Repair usage is included in task and workflow totals.
- Cancellation during repair terminates cleanly.
- Parallel mixed success/schema-failure results preserve exact slot status.

## Environment

- Failure observed on `v0.9.9` (`6f3850c3d7f0`).
- Inspected upstream: `main` at `8ef856f27cbcee758e48e65af436d87e7d9da15c` (`0.9.11` source).
- No schema-repair/retry path was found in the current workflow crates during this review.

### LINKED ISSUE #5595: Read-only inspection children reject in-workspace absolute git -C at execute time
## Impact

A live v0.9.11 reviewer child spent roughly 347k tokens and produced no findings because it could not run the canonical repository inspection command `git -C <workspace> log`. The model-facing classifier, role posture gate, and execution envelope say the call is allowed, but the final operand boundary refuses it before process launch.

This is a v0.9.12 release blocker: Scout, Reviewer, and Planner are described as read-only inspection roles, yet a normal absolute in-workspace Git target cannot pass end to end.

## Verified cause on the 0.9.12 integration branch

- `crates/tui/src/command_safety.rs::is_agent_readonly_shell_command` accepts `git -C <dir> log --oneline -n 5`, pipelines, `find`, `sed -n`, `npm view`, `pwd`, and numeric Git shorthand such as `-20`.
- `crates/tui/src/tools/subagent/mod.rs::posture_permits_tool` admits classifier-proven canonical lowercase `bash` calls for Scout, Reviewer, and Planner when their runtime profile has a shell surface.
- `crates/tui/src/tools/execution_envelope.rs::enforce_execution_envelope` accepts the same proven-read-only evidence.
- The Ask/Auto-Review child session gate independently reclassifies the same lowercase `bash` call through its stricter parent-shell approval predicate. With `auto_approve=false`, this can hold or deny `git -C`, pipelines, and `npm view` before the operand boundary. The existing gate test sets `auto_approve=true`, hiding this disagreement.
- `crates/tui/src/tools/shell.rs::enforce_readonly_workspace_operands` then rejects every absolute operand by shape (`candidate.has_root()`) before resolving it. Therefore `git -C /absolute/path/inside/workspace log --oneline -20` can never execute.
- The attached-option branch similarly refuses any option token carrying `/` or `\\` without asking whether a known path-carrying option resolves inside the workspace.
- Canonical lowercase `bash` translates its advertised `timeout` field only inside its tool wrapper; posture and envelope inspect the untranslated input and therefore disagree with the executor when `timeout` is present.

The existing `readonly_operands_are_workspace_bounded_and_symlink_aware` test protects real escapes but does not pin acceptance of canonical absolute paths inside the workspace. Existing subagent tests pin catalog, posture, and envelope behavior, but their concrete executor checks use only relative operands.

## Acceptance criteria

- Add the missing regression first: for each of Scout, Reviewer, and Planner under the read-only profile, run `pwd` and `git -C <canonical absolute temp repo> log --oneline -20` through catalog/posture, envelope, and the concrete executor.
- Canonicalize absolute operands and accept them only when the resolved path starts with the canonical workspace.
- Keep absolute paths outside the workspace, `..` escapes, home expansion, foreign/unresolvable Windows or UNC forms, and in-workspace symlinks that resolve outside fail-closed. A platform-native absolute path may pass only after canonical proof that it is inside the workspace.
- Apply the same resolved-within-workspace rule to known options that carry a path, including Git `-C` forms; do not broadly bless unknown path-bearing options.
- Normalize the canonical lowercase `bash` input once so catalog, session approval, role posture, execution envelope, and executor judge the same command and timeout fields; keep the parent parallel auto-approval classifier unchanged.
- Make each refusal identify the failed sub-rule (for example `shell.readonly.input_shape`, `shell.readonly.command`, `shell.readonly.operand.outside_workspace`, `role.shell.denied`, or `session.approval.required`) so a model can correct the right part of the call.
- Keep the test/provider matrix provider-neutral; no provider-specific fixture or default.

## Release evidence expected

Focused shell, command-safety, subagent gate/envelope/execute tests; `cargo fmt --all -- --check`; workspace clippy with zero warnings; diff review before commit.

### LINKED ISSUE #5596: Turn end silently cancels turn-owned subagents and destroys resumable work
## Impact

Turn-owned subagents are currently destroyed when the parent model ends its turn, even though the UI/model context says running children continue in the background and will report later. A long-running reviewer can therefore lose all useful work without a visible warning. In the observed v0.9.11 run, the child also had a continuable checkpoint, but the turn-end cancellation projected terminal `Cancelled`, so `agent(start, resume_from=...)` could not recover it.

## Verified cause on the 0.9.12 integration branch

- `TurnMailboxBarrier::cancel_and_flush` in `crates/tui/src/core/engine.rs` calls `ForegroundChildRegistry::cancel_and_wait` before sealing the mailbox.
- `ForegroundChildRegistry::cancel_and_wait` in `crates/tui/src/tools/subagent/mod.rs` cancels every registered direct non-detached child token and waits for task teardown.
- Default `agent(action="start")` is turn-owned; only explicit `detached=true` uses `background_runtime`.
- The turn loop already appends an internal `waiting_for_subagents` user-role message saying running children are background work and their completion will arrive later. The terminal barrier then contradicts that message by cancelling the turn-owned children.

## Chosen 0.9.12 behavior

Before allowing a parent turn to end while turn-owned children are running, append one internal user-role runtime message and give the parent exactly one additional model step. The message must say: use `agent(action="wait", until="all")`, or start future independent work with `detached=true`; ending now will park unfinished turn-owned children as resumable work.

If the model still ends the turn, cooperatively stop those children as `Interrupted` rather than terminal `Cancelled`, preserving or synthesizing a continuable checkpoint, actionable `needs_input`, and the continuation metadata needed by `agent(start, resume_from=<source-agent-id>)`. Explicit `agent(action="cancel")` remains terminal cancellation.

This bounded guard avoids both silent work loss and unexpected provider spend after the parent has reported completion. It also keeps turn ownership meaningful instead of silently changing every unfinished foreground task into detached work.

## Acceptance criteria

- A turn-owned-child generation receives one additional model step, never an unbounded turn-end loop; a newly spawned generation may receive a new guard.
- The guard names the active session-owned children and gives the concrete `wait` / `detached=true` choices before the terminal decision.
- If the guard is ignored, every unfinished turn-owned descendant settles as `Interrupted`, not `Cancelled`, with a non-empty continuable checkpoint even at step zero and an actionable `resume_from=<source-agent-id>` receipt.
- The old turn mailbox is sealed only after those children settle and before `TurnComplete`; no late child envelope can attach to the completed turn and every terminal receipt is delivered exactly once.
- Every non-detached descendant participates in the turn-owned registry; detached trees remain excluded.
- The UI status and internal lifecycle message are append-only user-role history; no volatile lifecycle fact is added to the frozen system/tool prefix (`docs/CACHE.md`).
- Natural completion racing the barrier and explicit parent interruption each produce one terminal outcome without deadlock or duplicate completion.
- Explicit detached starts and explicit user cancellation retain their existing semantics.

## Release evidence expected

Focused foreground registry, terminal barrier, mailbox ordering, later-turn completion, checkpoint/resume tests; `cargo fmt --all -- --check`; workspace clippy with zero warnings; diff review before commit.

### LINKED ISSUE #5597: Detached interactive agents lose post-turn usage from session cost totals
## Problem

A `detached=true` child launched from an interactive TUI turn correctly outlives that parent turn, but usage produced after `TurnComplete` no longer reaches the session/turn cost projection. The worker record still accumulates its own usage, so the same provider response can be visible on the agent while `/cost`, live session totals, and the completed turn undercount it.

## Evidence

- Interactive child usage is delivered to the TUI through `MailboxMessage::token_usage` in `crates/tui/src/tools/subagent/mod.rs`; the code explicitly calls the mailbox the sole delivery path when no runtime usage owner exists.
- `TurnMailboxBarrier::flush` seals the turn mailbox before `TurnComplete` in `crates/tui/src/core/engine.rs`. This is correct for turn ordering and must not be weakened.
- Detached runtimes deliberately survive that barrier. A later mailbox send is rejected by the sealed mailbox, while `SubAgentManager::record_worker_usage` still records the response locally.
- Interactive runtime construction only acquires a durable runtime usage lease when `config.compaction.runtime_cost_owner` is present; ordinary interactive turns therefore have no post-turn accounting sink.

This was exposed while fixing #5596; `c4c35d594` intentionally preserves the mailbox durability boundary and does **not** claim to fix this accounting gap.

## Acceptance

- [ ] Detached interactive child usage after parent `TurnComplete` is included exactly once in durable session/live cost totals.
- [ ] The sealed turn mailbox remains a hard ordering boundary; no late event is attached to the completed turn.
- [ ] Per-worker usage and session usage share a dedupe key so adding the durable sink cannot double count.
- [ ] A regression launches a detached child, completes/seals the parent turn, records a later provider response, and proves worker + session totals agree after reload.
- [ ] Provider-neutral coverage exercises multiple provider identities/routes rather than a single provider-specific fixture.

Hmbown:
## Status
Integration branch for the v0.9.12 cycle — **now gated and code-complete for the release blockers**; remaining pre-review work is version bump + changelog/RC gates (tracker: #5573). Not to be merged until those gates are green.

## On this branch (72 commits)

**Release blockers finished here:**
- #5597 detached interactive agent usage now reaches session/`/cost` totals (owner-scoped runtime lease, sealed-mailbox preserved, exactly-once across worker/session/reload) — `8343c89d5`
- #5586 workflow mega-file decomposition: module directory + `journal`/`usage`/`report` extraction, semantic-free moves verified by normalized-content SHA (`3e788d6b5`, `be3a201f2`, `a4a788cd3`, `64545aeb9`) — the broader lib.rs/config.rs/client.rs/runtime_threads.rs checklist stays open
- #5588 provider-neutrality fixes: provider-relative model resolution, provider-native OpenAI defaults, provider-scoped CLI credentials, NVIDIA credential isolation, multi-provider neutrality matrices
- #5583 workflow responseSchema: bounded repair with typed receipts, raw-output artifacts surviving reload, partial-success parallel slots
- #5582 Degraded owner snapshots no longer project as Completed

**Safety & money:** R2 approval grants match the command family · R3 Chat-Completions SSE error frames · R4 non-streaming read timeouts · R5 fleet per-task wall-clock · R7 worker self-reap on parent death · R9 typed task errors + fail-fast parallel/pipeline · S1 opt-in read deny-list (#5568) · S2 hard-link write guard (#5569) · fleet run-wide usage ceiling R6 (#5567) · per-step TurnUsage pricing (#5578) · cost-unknown footer chip

**Subagents/coordination:** #5562 stale write-claim liveness release + `coordinate release` + honest verifier description · in-workspace `git -C` read-only gate fix (#5595) · resumable turn-owned parking (#5596) · detached schema repair within runtime budget

**Cache/context:** C1 wire-order tool fingerprint · C3 real cache-hit rate in `/context` · C5 undeclared-drift assert + named `change:tool_surface` declarations for mid-turn tool-surface mutations (#5571) · T1 meter/trigger alignment · #5577 named compaction refusals + the exact 842k/1M billed-window regression · honest reasoning-only length-stop failure · step-budget soft landing + report-on-exhaustion (A1/A2)

**Workflow:** #5583 receipts/partial-success · R9 fail-fast · `/workflow confirm` gate fix

**UX:** underwater state-reading surface (steady tints, activity shape, seam fix) · #5550 @path:START-END ranges · #5555 clipboard last-copy backup · #5549 wait_any/wait_all · double/triple-click composer selection · #5548 web dispatch-rejection coverage · #5564 REBRAND docs · #5579 plugin load-failure hints · OpenRouter attribution headers

**Packaging:** verified Omarchy/AUR path + `codewhale-tui` compatibility alias + updater respects package ownership

**Infra/credit:** T4 MCP 401/403 reactive OAuth refresh (#5572 — loopback QA path still tracked before close) · credit gate recognizes agent contributors · CLI diagnostics reworded to avoid secret-scanner false positives

## Verification
Per-commit focused gates; fmt/`cargo check --workspace`/clippy 0-warning at every landing. Current-head evidence: workflow wall 218/218 (isolated worktree) and 187/187 TUI slice at the merged head; #5597 multi-route acceptance (DeepSeek/Anthropic/Custom) 1/1 + adjacent 5/5 + cost sweep 90/90; the four previously failing engine tests (C5 tool-surface drift, reasoning-only length stop, zero-step host drain) fixed and green; cache_guard 9/9; adaptive-evidence integration 1/1; cucumber plugin e2e 16/16.

Closes #5548
Closes #5562
Closes #5564
Closes #5567
Closes #5568
Closes #5569
Closes #5571
Closes #5577
Closes #5578
Closes #5579
Closes #5582
Closes #5583
Closes #5595
Closes #5596
Closes #5597

No-Issue: v0.9.12 integration and release train; #5572 stays open pending the loopback 401→refresh→retry QA proof, #5586 keeps its mega-file checklist, #5588 keeps its neutrality audit.


Hmbown:
Closing as never-merge per the standing boundary in codewhale-ops DIRECTION.md and the TRUTH_LEDGER hard-rails list. This WIP integration branch was the 0.9.12 staging lane; its landable content has since reached main through the individually-gated slices (Tideline redesign 45ecc8506, launch composer #5698 with all five review findings fixed, sandbox hardening, route contract #5702, dispatch #5701, and the rest of the 0.9.12 board). The branch is 300+ commits behind main, red, and superseded — and it is the base PR #5717 mistakenly stacked on (the contributor has been asked to rebase onto main). Branch preserved as evidence; reopen only by founder direction.

PR #5535, opened 2026-08-21
Supervised operation stack: lifecycle outbox, /relaunch, per-session control socket, and the goal-continuation quiet-period fix

M-Maciej:
One PR, five commits-area changes, all on the same seam — machine-readable
supervision of long-lived codewhale sessions:

- **Lifecycle event outbox** (`[lifecycle_outbox]`, opt-in JSONL + webhook):
  `turn_start` / `turn_end` / `turn_stalled` / `subagent_spawn` /
  `subagent_complete` / `session_start` / `session_end` for the interactive
  TUI and headless `codewhale exec`; workspace/subagent routing fields;
  `turn.failed` folding on the engine-disconnect and channel-closed paths;
  cross-process atomic `seq` via a lock sidecar; session ownership of turn
  boundaries (SIGTERM flush + boot reconciliation for SIGKILLed sessions);
  bounded webhook fan-out with in-sink retries; a `codewhale doctor`
  posture row. Design record: `docs/rfcs/1365-lifecycle-outbox.md`
  (modeled on `1364-hooks-lifecycle.md`).
- **`/relaunch`** — save like `/exit`, then self-exec `resume
  <session-id>` so a session switches to the current binary in one step;
  falls back to `argv[0]` when `current_exe` resolves to a deleted or
  renamed path (the updater's rename-replacement case). Windows keeps the
  documented manual-restart behavior.
- **Per-session control socket** (`[control_socket]`, Unix, opt-in):
  newline-framed JSON-RPC per running session with `message` /
  `interrupt` / `relaunch` / `status` verbs, typed error codes, bounded
  requests and handler timeouts, socket lifecycle riding the session
  lifecycle (stale-file takeover, live-bind refusal with retry backoff).
  The `RuntimeBackendKind::External` backend that consumes these verbs is
  deliberately **not** in this PR — it lands as a follow-up.
- **Goal-continuation quiet period fix** (#5534, follow-up to
  #5508): the configured `[goal] continuation_delay_seconds` wait now
  applies on the within-turn dispatch path, unconditionally, so
  CLI-resumed and host-managed sessions honor the cadence; the wait stays
  cancellable and pause/clear/terminal-`update_goal` still cancels a
  pending pass.
- **Test hardening** — env-barrier isolation for two parallel-flaky tests
  (route-budget V4 trigger, prompt prefix-leak), plus SSH-marker and
  RUSTUP_HOME guards in two other suites.

The changes ship as ~20 scoped commits so reviewers can cherry-pick areas
[...]

Issue #5533, opened 2026-08-21
Feature: the control surface for supervised operation


M-Maciej:
Feature: per-session control socket (message / interrupt / relaunch / status) and RuntimeBackendKind::External — the control surface for supervised operation

## Problem

I run codewhale sessions under external supervision — a supervisor (a
terminal multiplexer wrapper, an automation harness, a CI-style setup)
watching long-lived sessions with no human at the screen. Two things
work today: observation and re-creation. The lifecycle outbox (see the
lifecycle-outbox feature issue) gives the supervisor a machine-readable
stream of what happened, and the restart/resume backend (see the
/relaunch feature issue) lets it re-create a stopped Lane from its
durable record. What still does not work is everything in between:
**commanding a session that is running**.

Today the only inbound channel is terminal simulation. Delivering a
message means typing text plus Enter into the pane's PTY — no
acknowledgement that the session consumed it, no structured payload, and
the bytes can interleave with whatever the agent itself is typing.
Cancelling a wedged turn means sending an Esc key. Finding out the
current turn or goal state means scraping the rendered screen. A
supervisor managing many sessions across many workspaces is driving each
one with keystrokes and inferring the results from pixels.

The control plane's own vocabulary names the gap. `ControlSurface { Cli,
Slash }` is documented as "There are exactly two"
(`crates/lane/src/control.rs:43-63`), so no existing surface can be
driven by another process. The Runtime slot tells the same story one
level down: `RuntimeBackendKind { Tmux, Inline, Vm, Ci }`
(`crates/lane/src/runtime.rs:21-25`) has no kind for "execution is
delegated to an outside session manager" — the `Vm`/`Ci` stubs wait for
concrete managers, but there is no way to say "another process owns this
session's lifecycle". A supervised codewhale is neither addressable nor
a first-class Runtime fact.

## Proposed solution

Two asks, one umbrella: a per-session control socket (the third
surface), and a `RuntimeBackendKind::External` variant whose methods are
those socket verbs (the external runtime backend).

### The socket surface — a third `ControlSurface`

An opt-in `[control_socket]` config table (`enabled: bool`, default
false — unset or absent changes nothing). When enabled, the interactive
TUI binds one unix domain socket per *running* session at

```text
<sessions-dir>/<session-id>/control.sock     (mode 0600)
```

(`<sessions-dir>` is the directory the session store already uses.) The
socket speaks newline-framed JSON-RPC, one request per connection:
[...]

Issue #5532, opened 2026-08-21
Feature: /relaunch — switch a running session to the current binary


M-Maciej:
## Problem

`/update` installs a new binary and then tells the user to restart the
app. The updater's own design notes say why: "We do not relaunch. This
codebase has no self-exec/relaunch pattern, and inventing one under a TUI
holding the terminal is not a small change. Telling the user to restart is
the honest slice; that is also the only possible answer on Windows, where
the replaced image is the one running."

For a human at the keyboard that restart is two commands (`/exit`, then
`codewhale resume <id>` or a shell relaunch) and a context break. For
**supervised sessions** — long-lived codewhale instances running under an
orchestrator/terminal-manager (tmux-style wrappers, automation harnesses,
CI) with no human at the screen — it is worse: every update means the
supervisor must notice, exit each session, and re-attach it by id, per
session, across a fleet. Long-running coordinators and unattended sessions
simply stay on the old binary because the swap is manual.

## Proposed solution

A `/relaunch` slash command that performs the swap in one step:

1. Save and flush session state exactly like `/exit` does (reuse the
   existing teardown path — no second teardown to maintain).
2. `exec(2)` the **current executable** (`std::env::current_exe`) with
   `resume <current-session-id>` — atomic process-image replacement, so the
   terminal, pane, and session stay owned by the same process lineage. After
   an update replaces the binary on disk, the same command brings the
   session up on the **new** binary instantly.

Behavior:

- Unix-first (Linux/macOS): true self-exec.
- Windows: keep the existing honest "restart manually" message (replacing a
  running image is impossible there — the documented constraint).
- `/update` output gains one line: run `/relaunch` to switch now.

**The rename-replacement edge case (Linux)**: `current_exe()` is resolved
once and cached, and after an updater replaces the binary by rename it
resolves to `"<path> (deleted)"` — execing that path fails and the
self-exec silently dies. The command therefore falls back to `argv[0]`
(resolved via `PATH` when needed) whenever `current_exe` no longer names
a live, executable file, and the Windows path is preserved verbatim.
This is covered by tests (deleted/renamed binary scenarios, verbatim
Windows paths).

## Use case

- After every `/update`: one command instead of exit + manual resume.
- Supervised fleets: an orchestrator sends `/relaunch` to each session after
  installing a new build — live rollover of running agents (the same
  pattern a supervised runtime uses for its own live rollover).
- Long-lived coordinator sessions that must not lose context across builds.

## Alternatives considered

[...]

Issue #5531, opened 2026-08-21
Feature: local lifecycle event outbox (JSONL + webhook) with turn_stalled / turn_failed events


M-Maciej:
## Problem

I run long-lived codewhale sessions **under external supervision of
interactive TUI sessions like herdr** — a supervisor watching agent sessions without
a human at the screen (a terminal multiplexer wrapper, an automation
harness, or an alerting setup, for overnight runs, unattended fleets, or
CI-style harnesses). Today that supervisor can only poll the terminal
screen: there is no machine-readable stream of *what happened* — when a
turn started, finished, failed, or stalled, what a subagent did, how much
a session cost. The same gap hits **headless `codewhale exec`**
(codewhale's own non-interactive CLI mode): runs driven through it emit
nothing scriptable at all, so a supervised run is a black box.

Two distinct situations, one gap: the supervisor — whether it watches an
interactive TUI session or a `codewhale exec` run — gets no structured
event feed.

## Proposed solution

An opt-in `[lifecycle_outbox]` config table. When set, codewhale appends
one JSONL line per lifecycle event to the configured file, for interactive
TUI sessions **and** headless `codewhale exec` runs, with no per-hook shell
commands:

```toml
[lifecycle_outbox]
# Unset or empty path = the feature is OFF; behavior is unchanged.
path = "~/.codewhale/notifications/outbox.jsonl"
webhook_url = ""    # optional: also POST the same events as {"at", "event"} JSON
webhook_token = ""  # optional bearer token for webhook_url
```

**Event list** (one JSONL line per event; `kind` is the envelope's dotted
kind):

- `turn_start` — `turn.started` — a turn begins
- `turn_end` — `turn.completed` / `turn.failed` / `turn.interrupted`,
  projected from the turn status
- `turn_stalled` — `turn.stalled` — the stall watchdog recovers a wedged
  turn
- `subagent_spawn` — `subagent.spawned` — a subagent is spawned
- `subagent_complete` — `subagent.completed` — a subagent reaches a
  terminal state
- `session_start` — `session.started` — interactive session start
- `session_end` — `session.ended` — interactive session end

Details:

- `seq` is monotonic in file order. Every append re-locks a `<path>.lock`
  sidecar and re-recovers the tail, so multiple sessions sharing one
  outbox file produce unique, increasing seqs; the sidecar is the only
  new on-disk artifact.
- A torn trailing line from a crash is ignored; recovery reads the last
  complete line.
[...]

PR #5591, opened 2026-08-24
Fix:goal continuation cadence fix - part a

### LINKED ISSUE #5534: Bug: Goal-continuation cadence is bypassed on the within-turn dispatch path
## Description

#5508 (commit `7eb4650b0a67`, "feat(tui): add cancellable goal
continuation cadence") added `[goal] continuation_delay_seconds` and a
cancellable quiet period, but wired the wait into **only one** of the two
goal-continuation dispatch paths:

- The **cross-turn scheduler** (`schedule_goal_continuation` in
  `crates/tui/src/core/engine.rs`, the `ready_at` quiet period) — used
  when the interactive engine re-dispatches after a completed turn, and
  gated on `if !self.host_managed_turns()`.
- The **within-turn dispatch hook** (`goal_continuation_message_if_needed`
  in `crates/tui/src/core/engine/turn_loop.rs`) — which has **no wait at
  all**. Whenever a model step ends with the goal still active it
  immediately emits the next continuation prompt.

Consequences:

1. **Within-turn continuation passes never honor the quiet period**, on
   any session: up to `per_turn_max` passes fire back-to-back at model
   latency.
2. **Host-managed sessions never reach the cross-turn scheduler at all**
   (the engine yields so the host can create the next durable claim), so
   for those sessions the quiet period is effectively dead — every pass
   is dispatched through the un-waited hook.

## Steps to reproduce

1. Set the quiet period, e.g. `[goal] continuation_delay_seconds = 300`.
2. Start a session with an active goal, or resume one via
   `codewhale resume --last` (a non-host-managed CLI resume).
3. Watch the continuation passes: they fire at model-latency intervals
   (~9 s observed), never 300 s apart.

## Expected behavior

Consecutive goal-continuation passes are separated by the configured quiet
period (cancellable with Esc), on every dispatch path.

## Actual behavior

Passes fire instantly through the within-turn dispatch hook; the quiet
period applies only to the cross-turn scheduler path, which is not the
path these sessions dispatch through.

## Impact

Every supervised/unattended run that relies on `continuation_delay_seconds`
to pace autonomous goal work (cost control, provider rate limits,
checkpoint cadence) gets instant-fire passes instead of the configured
cadence. The bypass was measured live in both dispatch modes — host-
injected continuation tokens dispatching immediately (~9 s end→start
gaps), and instant-fire passes on a CLI-resumed session.

## Environment

- OS: Linux
- codewhale version: 0.9.10
- Install method: release binary (also reproduced on a source build
  tracking v0.9.10)
- Model/provider: Deepseek/any
- Shell: zsh

Live measurements against a supervised deployment tracking v0.9.10:

- **Before**: host-injected continuation tokens dispatched immediately,
  ~9 s end→start gaps; a CLI-resumed session fired passes instantly.
- **After applying the fix**: two consecutive quiet→pass gaps of ~331 s
  and ~323 s, with zero user input in each window. The bypass is gone.

The fix in the accompanying PR applies the same cancellable quiet period
at the within-turn dispatch site, **unconditionally** (the hook is the
only dispatch site for non-host-managed sessions, so gating the wait on
host-managed turns does not work — a CLI-resumed session would still fall
into the no-wait branch). Semantics preserved: the wait stays cancellable
(the Esc cancel token wins biased over the timer), a pause/clear or a
terminal `update_goal` observed after the wait cancels the pending pass
before anything is recorded or dispatched, and the
`GoalContinuationWaiting` / `GoalContinuationWaitEnded` events are
unchanged. New regression tests cover the non-host-managed wait, the
host-managed wait, and the zero-delay-immediate case.

M-Maciej:
## Summary

`[goal] continuation_delay_seconds` (the goal-continuation cadence, added in
#5508) was wired into only one of the two goal-continuation dispatch paths.
The within-turn dispatch hook (`goal_continuation_message_if_needed` in the
engine) had no wait at all, so a model step that ended with the goal still
active immediately emitted the next continuation prompt. That made the
configured cadence dead on two kinds of sessions:

- **CLI-resumed sessions** (`codewhale resume --last`) — they dispatch
  continuation passes through the within-turn hook.
- **Host-managed sessions** — the engine yields so the host can create the
  next durable claim, so these sessions never reach the cross-turn scheduler;
  every pass goes through the un-waited hook.

Measured (before split) live with a 300 s delay configured: continuation passes fired at
model latency before the fix (~9 s gaps, instant-fire through the within-turn
hook). After the fix, consecutive passes are separated by the configured
quiet window — the observed cadence became ~5 minutes (300 s quiet window
plus the turn itself), versus ~9 s instant-fire before.

This branch applies the wait on the within-turn path, unconditionally:

- `goal_continuation_message_if_needed` now awaits
  `goal_loop::await_continuation_wait` before dispatching (biased cancel:
  Esc / steer / host cancel / terminal `update_goal` all win over a racing
  expiry, and the live goal is re-read only after the wait).
- Host-injected continuation tokens get the same wait in the `Op::ContinueGoal`
  arm — for host-managed sessions this arm is their only dispatch site.
- The wait stays cancellable, and pause/clear/terminal updates still cancel a
  pending pass.

The flake-hardening test fixes (env-barrier isolation for the route-budget V4
trigger and the prompt prefix-leak tests, SSH-marker neutralization in the
terminal-motion suite, RUSTUP_HOME preservation in the read-only diagnostic
test) are grouped here per the #5535 review guidance (the cadence fix plus the
flake hardening as one small slice).

## Rebase note

Rebased onto current `main` (v0.9.11, merged in #5542). v0.9.11 rewrote
`core/engine/turn_loop.rs` and the engine event loop; the fix has been
re-verified against the new loop by the regression tests below.

## Split provenance

This PR is one slice of a mechanical split: the original combined branch was
rebased onto v0.9.11 and then partitioned along commit boundaries, then each
slice was re-verified independently with the full gate suite. The combined
rebased tree (the union of all slices) also passes the full suite
(13,441 / 0). Partition boundaries worth review attention: the engine
dispatch paths shared with the outbox emit-site work, and the Windows
cfg-gating.

## Split map and merge order

The four pieces of the original combined branch are filed as four PRs. All
four are standalone — each is cut from current `main` and passes its full
gate suite alone:

- **(a) this PR — the cadence fix + flake hardening** (merge first).
- **(b) lifecycle outbox (#5531)** — same engine dispatch files, different
  lines; builds and passes 13,403 / 0 without this PR. A trivial rebase is
  expected after this merges.
- **(c) `/relaunch` (#5532)** — the control socket&#x27;s `relaunch` verb (d)
  dispatches the `/relaunch` command through the generic command seam, so
  (d) has no compile dependency on (c): without it the verb returns
  `command_error`; with it, the verb becomes functional.
- **(d) control socket (#5533)** — zero references to the outbox in either
  direction; no dependency.

One coupling no slice escapes: all four add entries to the same
`## [Unreleased]` changelog section, so each merge is followed by a small
changelog rebase for the PRs still open. Independently reviewable and
buildable — yes; mergeable with zero follow-up — no.

## Testing

&lt;!-- Exact gate commands (including the clippy allow list) are in
     CONTRIBUTING.md → &quot;Pre-push verification&quot;. --&gt;

- [x] `cargo clippy --workspace --all-targets --all-features --locked` (warning-free under the CI allow list)

Results: clippy clean on stable 1.98.0 under the CONTRIBUTING.md allow
list; full suite **13,375 passed / 0 failed**. Focused `goal_continuation`
suite 13/13 (host-managed sessions wait, non-host-managed sessions wait,
zero-delay dispatch is immediate, cancellation paths).

## Checklist

- [ ]  Verified TUI behavior manually if UI changes — the cadence was
  exercised live on running sessions ( yes before split, not after split )
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address —
  every commit is authored under
  `130112810+M-Maciej@users.noreply.github.com`
- [x] Every commit carries a `Signed-off-by` (DCO)
- [x] Changelog entry under `Unreleased` (and the flake-hardening entry
  moved out of the released 0.9.11 section, where an earlier rebase had
  placed it)
- [x] No dead-code-budget change
- [x] The outbox, `/relaunch`, and control-socket slices from #5535 follow
  as separate PRs after this one lands</pre>

Closes #5534


Hmbown (approved):
Reviewed and independently verified locally on your branch:

1. **The core design is right.** Decision split from dispatch (`goal_continuation_allowed`), decide-before-wait so a terminal goal never spends the quiet period, biased-cancel wait (Esc/steer/host cancel win over a racing expiry), and — the part I'd call load-bearing — the **re-decide on live state after the wait**, so a `/goal pause`, `/goal clear`, or terminal `update_goal` landing during the quiet period cancels the pass instead of dispatching a provider request. The `Op::ContinueGoal` arm correctly gates only host-injected tokens (`engine_schedule_id.is_none()`); engine-owned tokens keep their scheduler semantics.

2. **Ran the focused suite on your branch**: `goal_continuation` 12/12 (host-managed waits, non-host-managed waits, zero-delay immediate, cancellation paths).

3. **Ran the repo's single-turn-loop guard** (`crates/core/tests/single_turn_loop.rs`): 2/2 — the turn_loop.rs changes stay inside the one loop.

4. **Clippy** on `codewhale-tui` --all-targets on your branch: 0 warnings under the CI allow list.

5. **Merge interaction with the 0.9.12 integration branch** (#5576): `git merge-tree --write-tree` between it and this branch exits clean — we also modified `turn_loop.rs` (compaction decisions, step budget), but in different regions; no conflicts expected either merge order.

The split provenance note and the honest "live-verified before split, not after" checkbox are exactly the review shape we wanted out of #5535 — thank you. Holding the merge for the maintainer's go and for the remaining CI checks to finish.

Hmbown (approved):
Re-approved at exact head 0903bf62fe807a62c771d82f4a857b1c5ae201cc after the current-main refresh.

Verified locally: all eight M-Maciej commits and their authorship are unchanged; only the expected synchronized Unreleased changelog overlap was resolved. Formatting is clean; workspace/all-targets clippy is clean; goal_continuation passes 12/12; the single-turn-loop guard passes 2/2; the contribution-credit checker passes 11/11 plus the live 8/8 range check. Fresh GitHub CI is green across required Linux, Windows, macOS, lint, and version lanes, with safety, OpenHarmony, DCO, credit/metadata, and advisory checks also green. The PR now closes the exact bug issue #5534.

PR #5592, opened 2026-08-24
lifecycle outbox - part b

M-Maciej:
Closes #5531

## Summary

Opt-in `[lifecycle_outbox]` config table. When set, codewhale appends one
JSONL line per lifecycle event to the configured file, for interactive TUI
sessions **and** headless `codewhale exec` runs, with no per-hook shell
commands:

```toml
[lifecycle_outbox]
# Unset or empty path = the feature is OFF; behavior is unchanged.
path = &quot;~/.codewhale/notifications/outbox.jsonl&quot;
webhook_url = &quot;&quot;    # optional: also POST the same events as {&quot;at&quot;, &quot;event&quot;} JSON
webhook_token = &quot;&quot;  # optional bearer token for webhook_url
```

**Events** (one JSONL line per event; `kind` is the envelope&#x27;s dotted kind):
`turn_start`, `turn_end`, `turn_stalled`, `subagent_spawn`,
`subagent_complete`, `session_start`, `session_end`.

Design points:

- `seq` is monotonic in file order. Every append re-locks a `&lt;path&gt;.lock`
  sidecar and re-recovers the tail, so multiple sessions sharing one outbox
  file produce unique, increasing seqs; the sidecar is the only new on-disk
  artifact. A torn trailing line from a crash is ignored (bounded 64 KiB tail
  scan; recovery reads the last complete line).
- **Session ownership of turn boundaries**: a session killed mid-turn
  (SIGKILL, closed pane) dies between `turn_start` and `turn_end`. Catchable
  signals flush a synthetic `turn_end` (`status: &quot;interrupted&quot;`,
  `payload.reconciled: true`) for every open turn; SIGKILL runs no code, so
  **boot reconciliation** on the next session start pairs anything the killed
  process left behind. Both paths derive the open turn from the file, never
  in-memory state, and run under the exclusive lock, so two reconcilers never
  double-append.
- Every payload carries `workspace` (the resolved workspace path) so a
  consumer can route each event to its project; subagent events additionally
  carry `subagent` alongside `agent_id`.
- Turn-boundary emits live in the engine (`spawn_tui_engine` wiring,
  `handle_send_message`), so **goal-continuation turns** — engine-originated,
  never passing through the UI user-message dispatch — emit the same
  `turn_start`/`turn_end` pair as user turns, with distinct, correlated turn
  ids. Exec and hosted engines keep their existing emit sites; a
  completion without a preceding turn no longer mints a phantom `turn_end`
  under a stale turn id.
[...]

PR #5593, opened 2026-08-24
/relaunch command - part c

### LINKED ISSUE #5532: Feature: /relaunch — switch a running session to the current binary

M-Maciej:
Closes #5532

## Summary

`/update` installs a new binary and then tells the user to restart the app.
This adds the missing self-relaunch so a session switches to the current
binary in one step:

- `/relaunch` behaves like `/exit` for persistence (save, restore terminal,
  telemetry flush, `session_end` emission) and then **self-execs**
  `resume &lt;session-id&gt;` — the CLI&#x27;s positional resume form, which the
  standalone TUI binary also accepts. Atomic `exec` on Unix: no orphan
  process for a supervisor to reap.
- **`argv[0]` fallback**: when `current_exe` resolves to a deleted or renamed
  path — the updater&#x27;s rename-replacement case — `/relaunch` falls back to
  `argv[0]` instead of failing.
- If the replacement fails, the session is already saved; the error names the
  exact `resume` command to continue with.
- Windows keeps the documented manual-restart behavior (no `exec`); the
  portable argv-construction tests still exercise the resume-form builder on
  every platform.

## Merge-order note

Merge order: after the cadence fix (a). No dependency on (b); the control
socket&#x27;s `relaunch` verb (d) dispatches `/relaunch` through the generic
command seam, so it has no compile dependency on this branch — until this
merges the verb reports `command_error`, after it the verb is functional.

## Windows build hygiene

The Windows CI gate denies dead code under `-D warnings`. `relaunch_argv` is
only called from the unix-only self-exec path, so the plain Windows lib build
flagged it. It is gated with `#[cfg(any(unix, test))]` — the item itself is
omitted from the plain Windows lib build instead of being allowed — so the
dead-code budget is untouched.

## Testing

&lt;!-- Exact gate commands (including the clippy allow list) are in
     CONTRIBUTING.md → &quot;Pre-push verification&quot;. --&gt;

- [x] `cargo clippy --workspace --all-targets --all-features --locked` (warning-free under the CI allow list)

Results: clippy clean on stable 1.98.0 under the CONTRIBUTING.md allow
list; full suite **13,376 passed / 0 failed** (same one pre-existing
environmental skip as the outbox PR). Focused `relaunch` suite: 12/12
(argv construction incl. Windows-style path preservation,
deleted/renamed-`current_exe` fallback, resume-form args).

## Checklist

- [x] Updated docs or comments as needed — Unreleased changelog entry;
  `/update`&#x27;s module doc now explains why it does not chain `/relaunch`
- [ ] Verified TUI behavior manually if UI changes — `/relaunch` exercised
  live on running sessions during development ( verified before split, not after )
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address —
  every commit is authored under
  `130112810+M-Maciej@users.noreply.github.com`
- [x] Every commit carries a `Signed-off-by` (DCO)
- [x] No dead-code-budget change</pre>
</details>


Hmbown:
Reviewed from the 0.9.12 integration lane: same disposition as #5592 — implements #5532 (unlabeled for v0.9.12 in #5573), standalone off main, so not folded mid-cycle; @Hmbown's call whether 0.9.12 includes /relaunch. Nothing blocking from the description; will take + full-matrix verify on request.

Hmbown:
Hi @M-Maciej — maintainer housekeeping on this PR, no code of yours was touched.

**What I found.** Two separate things were keeping this un-judgeable:

1. **The merge conflict was trivial.** I merged `origin/main` in and the *only* conflict was `CHANGELOG.md` + its generated slice `crates/tui/CHANGELOG.md`: main's `## [Unreleased] / ### Added` list grew while your bullet landed at the same insertion point. Every source file — including `crates/tui/src/tui/ui/event_loop.rs`, `crates/tui/src/lib.rs`, `crates/tui/src/tui/ui.rs` — auto-merged cleanly. I resolved it as "keep both" (main's list with your bullet appended) and regenerated the slice with `scripts/sync-changelog.sh`. `cargo check --workspace --all-targets` passes on the result.

2. **CI had never actually run on this PR.** Not once. The repo is set to `approval_policy: first_time_contributors` for fork PRs, so your `pull_request` workflows were withheld pending a maintainer click — and nobody was watching that queue. The only two green checks you were seeing (`gate`, GitGuardian) are the two paths exempt from that gate. That is our process failure, not anything you did. Pushing the merge commit as a maintainer released the gate, so the full matrix is running on this PR right now for the first time.

I also prepended `Closes #5532` to the PR description — the required `link` check wants an explicit closing keyword, and your commits already pointed at that issue.

**Heads-up on what CI will report.** `main` is currently red on two required checks that every open PR inherits, so ignore these if you see them:

- `Version drift` — `main` is missing a changelog receipt for #5604.
- `Test (windows-latest)` — two Windows verbatim-path tests (`tools::shell::tests::readonly_operands_are_workspace_bounded_and_symlink_aware`, `tools::subagent::tests::read_only_inspection_roles_execute_pwd_and_absolute_git_log`).

Both are fixed by #5610, which is waiting to land. Anything *else* that goes red is worth your attention.

Sorry this sat for so long without a real signal. Same treatment applied to #5592 and #5594.

Hmbown:
Merged onto the v0.9.12 integration branch in 397b9cb26; re-verified in-tree: /relaunch wiring in tui/event_loop.rs, tui/app.rs, and the control-socket relaunch verb in tui/control_socket.rs, with the portable argv-construction test suite. Closing as landed — thanks @M-Maciej!

PR #5602, opened 2026-08-24
fix(shell): decode Windows output reliably

zhuowp:
## Summary

- preserve UTF-8 and Windows ANSI-code-page characters split across shell reads
- use the current Windows ACP only after strict UTF-8 decoding fails
- keep synchronous, bounded, raw-delta, snapshot, tail, and detached-reader paths consistent
- seal detached raw streams without discarding output captured before the cutoff

## Background

Windows-native tools can write stdout or stderr using the active ANSI code page even when the surrounding shell is configured for UTF-8. The shell layer previously decoded these bytes as UTF-8 lossily, and some background paths decoded each poll independently. That could expose mojibake or replacement characters to both the model and the UI, and the original bytes were no longer recoverable by an embedding application.

## Testing

- `cargo check -p codewhale-tui --tests --locked` (passed)
- `cargo metadata --locked --no-deps` (passed)
- `cargo fmt --all -- --check` (passed)
- `git diff --check` (passed)
- focused `cargo test` reached the test-binary link stage but did not finish after about 11 minutes locally; it was stopped and is not claimed as passed

## Risk

ACP detection is deliberately a fallback after invalid UTF-8 and covers encodings supported by `encoding_rs`. It cannot infer OEM or arbitrary child-selected encodings. Detached raw readers freeze the public stream at the cutoff, so bytes arriving after that cutoff are intentionally rejected.

No-Issue: fixes a downstream-reported Windows shell output decoding defect without an existing upstream tracking issue.

### DIFF
--- a/crates/tui/Cargo.toml
+++ b/crates/tui/Cargo.toml
@@ -138,7 +138,7 @@ objc2 = "0.6.3"
 objc2-foundation = { version = "0.3.2", default-features = false, features = ["std", "NSArray", "NSDictionary", "NSError", "NSObject", "NSString", "NSURL"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows = { version = "0.62", features = ["Win32_Foundation", "Win32_Media_Audio", "Win32_Security", "Win32_Security_Authorization", "Win32_Storage_FileSystem", "Win32_System_Console", "Win32_System_Diagnostics_Debug", "Win32_System_Diagnostics_ToolHelp", "Win32_System_Environment", "Win32_System_JobObjects", "Win32_System_Registry", "Win32_System_Threading", "Win32_UI_WindowsAndMessaging"] }
+windows = { version = "0.62", features = ["Win32_Foundation", "Win32_Globalization", "Win32_Media_Audio", "Win32_Security", "Win32_Security_Authorization", "Win32_Storage_FileSystem", "Win32_System_Console", "Win32_System_Diagnostics_Debug", "Win32_System_Diagnostics_ToolHelp", "Win32_System_Environment", "Win32_System_JobObjects", "Win32_System_Registry", "Win32_System_Threading", "Win32_UI_WindowsAndMessaging"] }
 # `windows` re-exports most of windows-core's types but not all of them
 # (BOOL, HRESULT, … live in windows-core itself); window_control.rs names
 # windows_core::BOOL for the EnumWindows callback, so it must be a direct
--- a/crates/tui/src/tools/shell.rs
+++ b/crates/tui/src/tools/shell.rs
@@ -66,8 +66,8 @@ use crate::work_graph::{
 use crate::worker_profile::ShellPolicy;
 use output::{
     BoundedOutputAccumulator, BoundedOutputSnapshot, RAW_STREAM_SETTLED_TAIL_BYTES,
-    RawOutputBuffer, SharedRawOutput, new_shared_raw_output, tail_from_buffer, tail_text,
-    take_delta_from_buffer,
+    RawOutputBuffer, SharedRawOutput, ShellOutputDecoder, decode_shell_bytes,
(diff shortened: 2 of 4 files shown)


Hmbown:
> ## Summary
> 
> - preserve UTF-8 and Windows ANSI-code-page characters split across shell reads
> - use the current Windows ACP only after strict UTF-8 decoding fails
> - keep synchronous, bounded, raw-delta, snapshot, tail, and detached-reader paths consistent
> - seal detached raw streams without discarding output captured before the cutoff
> 
> ## Background
> 
> Windows-native tools can write stdout or stderr using the active ANSI code page even when the surrounding shell is configured for UTF-8. The shell layer previously decoded these bytes as UTF-8 lossily, and some background paths decoded each poll independently. That could expose mojibake or replacement characters to both the model and the UI, and the original bytes were no longer recoverable by an embedding application.
> 
> ## Testing
> 
> - `cargo check -p codewhale-tui --tests --locked` (passed)
> - `cargo metadata --locked --no-deps` (passed)
> - `cargo fmt --all -- --check` (passed)
> - `git diff --check` (passed)
> - focused `cargo test` reached the test-binary link stage but did not finish after about 11 minutes locally; it was stopped and is not claimed as passed
> 
> ## Risk
> 
> ACP detection is deliberately a fallback after invalid UTF-8 and covers encodings supported by `encoding_rs`. It cannot infer OEM or arbitrary child-selected encodings. Detached raw readers freeze the public stream at the cutoff, so bytes arriving after that cutoff are intentionally rejected.

This is wonderful - such a helpful contribution. Thank you!!

Hmbown:
Taken into the 0.9.12 integration branch (`codex/v0912-integration-20260823`, merge commit + one follow-up `0a85b13ba`) — authorship preserved as a merge commit.

**Review notes** (all positive):
- The decoder design is the right one and mirrors the transport layer's philosophy: strict UTF-8 first, incomplete sequences buffered across pipe reads, ACP fallback only on genuine invalid UTF-8, sticky once engaged — and every output path (sync exec, bounded snapshot, raw delta, tail, detached reader) now shares one decoder instead of independently lossy-decoding per poll.
- The `abandon()`→`seal()` change fixes a real data bug (detached readers lost retained-but-undelivered bytes at join-timeout) while keeping memory bounded via the existing post-delivery `release_to_tail` tail cap.
- `stable_utf8_tail_start` keeps the old leading-U+FFFD guard's intent with cleaner structure.

**One adaptation needed** (follow-up commit): `legacy_encoding_for_code_page` is production-called only under `cfg(windows)`, so CI Lint (which builds `--all-targets`, where the test import keeps it alive) passed, but a plain non-Windows lib build fails `-D dead_code`. Annotated `#[cfg_attr(not(windows), allow(dead_code))]` — the mapping table stays cross-platform tested.

**Verification on the integration branch**: macOS check/clippy/`shell::` 139/139/`decoder` 43/43/full lib suite 11180 passed 0 failed. The PR-side CI ran only light legs (change detection classified it tui-only), so the branch's full matrix — including the Windows legs that actually exercise the ACP path — is the real gate; it's running now on `0a85b13ba`.

Note for the merge-to-main pass: the PR body needs a `Closes #N` or `No-Issue: <why>` line for the repo's link gate — left for the author/maintainer since it's your PR body.

zhuowp:
Thank you very much for the detailed review, the thoughtful explanation, and for carrying this contribution into the 0.9.12 integration branch. I really appreciate the additional non-Windows `dead_code` annotation and the thorough verification you ran.

I have updated the PR description with the required `No-Issue:` line. Since the integration branch already contains the focused follow-up annotation, I have not duplicated that commit on the PR branch to avoid creating unnecessary divergence. Please let me know if you would prefer me to mirror it here as well.

Hmbown:
Already integrated on `codex/v0912-integration-20260823` as `276184e15` (`Merge PR #5602: decode Windows shell output reliably`). Closing this PR as superseded / already-integrated; not merging via GitHub.

PR #5609, opened 2026-08-25
refactor(tui): adopt command shapes in memory group (FEAT-019)

aboimpinto:
## Summary

FEAT-019 converts the TUI memory command group (`/note`, `/memory`) to the external command shapes introduced by FEAT-014 and hosted by FEAT-015, following the pattern FEAT-018 established for the utility group. Both command files remain under `codewhale-tui`; this PR changes their execution boundary without physically moving them.

This PR:

- adds the contract-owned `CommandMemoryContext` facet (`memory_path`, `memory_enabled`, and one typed method per native-memory operation) plus ten portable outcome types (`MemoryHit`, `MemoryStatus`, `MemoryRemembered`, `MemoryImportOutcome`, `MemoryGetOutcome`, `MemoryExport`, `MemoryReindex`, `MemoryDelete`, …) and the `MEMORY` capability bit (`1 << 9`);
- converts `/note` to a `CommandHandler::Contextual { capabilities: WORKSPACE }` registration and `/memory` to a `CommandHandler::Contextual { capabilities: WORKSPACE | MEMORY }` registration through `RegisterCommand<CommandResult>` and the existing `ContextualCommand::from_contract` bridge;
- adds the TUI `MemoryAdapter` over the real `NativeMemoryStore`, confining `crate::native_memory`/`anyhow` types to the host side;
- makes the dispatcher populate the envelope from the declared capability set (`contexts(capabilities)`), so undeclared facets are absent and missing required facets return safe command errors instead of `.expect()` panics;
- restores the least-capability declarations on the four FEAT-018 utility handlers (baseline reconciliation, see below);
- removes `memory` from both migration-frontier representations only after both handlers are fully portable;
- preserves names, aliases, usage strings, registry order, messages, parsing, note renumbering, notes-path fallback order, native-store output, and safety checks byte-for-byte.

Tracking: EPIC-006 / FEAT-019 in umbrella [Hmbown/CodeWhale#5316](https://github.com/Hmbown/CodeWhale/issues/5316).

## Baseline reconciliation (FEAT-018 hardening)

[...]

Issue #5316, opened 2026-08-10
EPIC-005: CodeWhale TUI Crate Decomposition (Umbrella)


aboimpinto:
## Current Core execution — September 9, 2026

Follow the [Codewhale Core execution plan](<https://linear.app/shannon-labs/document/codewhale-core-execution-plan-a11e2e1c601d>) for C03–C10, ownership, dependency order and completion evidence. Linear is the current execution authority; local checklists are supporting snapshots.

C03 is active with the existing Kimi worker. Local commits 2be4359f5 migrate AppMode/ApprovalMode consumers and durable context records; 798d3d664 moves worker roster records; 76f8d0044 moves reasoning preferences and removes the UI re-export. The last commit records 229 passing reasoning tests and a warning-free all-target cargo check. Final C03 boundary/approval-policy reconciliation remains open; C04–C10 are not complete.

Preserve the contributor provenance below and credit merged command-contract Stage A and Stage B FEAT-018 through FEAT-024. Current direction supersedes older fleet-stays-in-TUI, engine-last, mandatory staged-shim and fixed-size assumptions. Converge parent/child execution before extracting one runtime; migrate every consumer and measure the headless build boundary. C06 gives the model discretion over useful goals and supported effort, with runtime-owned limits and receipts. Do not impose explicit-goal-request-only semantics.

The whole extraction is active project scope; C20 separately decides which source is required for 0.9.13. These local commits do not establish hosted CI, installed acceptance or publication.

---

## EPIC-005: CodeWhale TUI Crate Decomposition — Umbrella EPIC

This issue is the **EpicStyle tracking issue** for the CodeWhale TUI crate decomposition. It represents the entire structure below: every sub-EPIC and every FEAT reports here when done, and every PR created under this work is logged here.

The decomposition is the direct answer to the compilation-time discussion in [#4991](<https://github.com/Hmbown/CodeWhale/issues/4991>) — see the Developer fast loop section below.

## Structure

EPIC-005 is the **umbrella EPIC**. It does not contain FEATs directly — it holds the map of **sub-EPICs**, each of which contains its own FEATs.

```mermaid
flowchart TD
    U["EPIC-005: TUI Crate Decomposition (umbrella)"]
    U --> E6["EPIC-006: Commands -> codewhale-commands (P1)"]
    U --> E7["EPIC-007: Tools consolidation -> crates/tools (P1)"]
    U --> E8["EPIC-008: Config consolidation -> crates/config (P2)"]
[...]

PR #5606, opened 2026-08-24
feat(runtime): 0.9.12 relay integration — unify managed Chat with native runtime threads

Hmbown:
## What this is

The 0.9.12 standby lane, rebased onto current main and re-verified on this machine. Headline change: managed Chat rides native runtime threads (turn_operation_idempotency), plus the R2 approval fix (MCP tools reviewed as kinds), doctor --fix with consent, and the accumulated standby-lane fixes (fleet usage ceiling R6, step-budget soft landing A1/A2, cache-prefix declarations C1/C5, sandbox read deny-list S1, hard-link write refusal S2, OAuth reactive refresh T4, workflow responseSchema receipts, and more).

## Verification (this machine, post-rebase)

- cargo test --workspace --no-fail-fast: 61 targets, 13,501 passed, 0 failed
- cargo clippy --workspace --all-targets: clean
- The rebase initially dropped the lane's mock-client adaptation of host_managed_engine_does_not_self_dispatch_goal_continuation (a host-managed engine waits for an explicit durable turn claim, so the test must use Engine::new_with_model_client with a MockLlmClient); the final commit restores it on top of the rebased tree and the suite is green again.

## Not included

- No 0.9.12 version bump, no tag, no publish. Merge and release remain the founder's gate.
- Managed Chat relay at the control plane stays dormant until 0.9.12 is published and the sandbox image rebuilt; it gates on the runtime capability turn_operation_idempotency, so nothing changes for 0.9.11 clients.

PR #5604, opened 2026-08-24
feat(tui): make Fleet roster editing discoverable

wuisabel-gif:
## Summary

Addresses the approved focused slice of #5589.

The Fleet roster now makes member editing discoverable:

- selected members in a named Fleet show an explicit `[edit]` affordance;
- the footer advertises `m model` for the focused member;
- pressing `m` opens the exact Fleet detail editor with that member focused directly on the model route picker;
- Enter and `s` retain their existing editor/setup behavior;
- the coordinator row remains read-only and the legacy-profile fallback is preserved.

The detail editor's existing reasoning, pin/inherit, vision, save, and receipt behavior is unchanged.

## Tests

Passed locally:

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p codewhale-tui --lib 'tui::views::fleet_roster' --locked` (20 passed)
- `cargo test -p codewhale-tui --lib 'tui::views::fleet_detail' --locked` (6 passed)
- `CARGO_INCREMENTAL=0 cargo clippy -p codewhale-tui --all-targets --locked -- -D warnings`

New coverage verifies the direct model shortcut is inert on the operator/no-Fleet paths, emits the exact member id for a selected Fleet, and renders the row/footer edit affordances.

No provider credentials or network access are required.

No-Issue: This PR addresses the approved #5589 roster discoverability slice without automatically closing the broader issue.

PR #5610, opened 2026-08-25
fix(tui): preserve Windows verbatim-path operands through POSIX word split

aboimpinto:
## Summary

Fixes the two Windows CI failures that block FEAT-019 (PR #5609):

- `tools::shell::tests::readonly_operands_are_workspace_bounded_and_symlink_aware`
- `tools::subagent::tests::read_only_inspection_roles_execute_pwd_and_absolute_git_log`

## Root cause

`enforce_readonly_workspace_operands` (and `hardened_readonly_argv`) tokenize the read-only command with `shell_words::split`, which follows **POSIX backslash-escaping**. On Windows, `Path::canonicalize` returns verbatim device paths (`\\?\C:\...`); the POSIX splitter silently eats the separators of any backslash path (`C:\Users\...` → `C:Users...`, `\\?\C:\...` → `\?C:...`).

The previous fix (42857108f) stripped the `\\?\` prefix **after** the split and only for `key=value` operands (`token.split_once('=')`). The failing tests pass the canonical path as a **bare space-separated token** (`cat <path>`, `git -C <path> log`), so the splitter already mangled it before the strip could run — the refusal then happened on the mangled spelling.

## Fix

Normalize Windows absolute path spellings **before** any POSIX-style splitter or glob-charset gate via a shared `normalize_windows_command_paths` in `crates/tui/src/command_safety.rs`:

- strips the `\\?\` verbatim prefix (`Path::canonicalize` artifact) whose `?` trips the classifiers' glob-charset gate; `\\.\` device paths are preserved verbatim;
- doubles the backslashes of Windows-absolute-path-like words (drive letter + `:`, verbatim, or UNC prefix) so `shlex`/`shell_words` round-trip the real path instead of `C:\Users\...` collapsing to `C:Users...`.

Applied at all judging sites:

- `is_agent_readonly_shell_command` and `is_parallel_readonly_command` (command_safety.rs) — the subagent posture gate;
- both `shell_words::split` sites in shell.rs (`enforce_readonly_workspace_operands` and `hardened_readonly_argv`).

Only Windows-absolute-path-like words are touched — POSIX escapes (`echo a\ b`) and unix hosts are unaffected.

## Testing

- [x] New unit tests: verbatim/drive/UNC/device operands survive the POSIX split round-trip; the agent classifier admits verbatim-prefixed `git -C` commands; POSIX escapes and drive-relative spellings untouched
- [x] `tools::shell` module — 139 passed
- [x] `tools::subagent` module — 555 passed
- [x] Clippy `-D warnings` on `codewhale-tui` — 0 warnings

The definitive validation is the Windows CI job (Linux cannot reproduce verbatim-path canonicalization); a green `Test (windows-latest)` on this PR unblocks FEAT-019.

Paulo Aboim Pinto

No-Issue: this fix unblocks FEAT-019 (PR #5609); umbrella #5316 must remain open for the remaining decomposition FEATs.

Paulo Aboim Pinto

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,7 +136,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   placeholder while waiting for a branch-specific CNB mirror (#5547).
 - Fleet roster members in a selected Fleet now expose a visible edit affordance
   and a direct `m` model-picker shortcut, while the coordinator row remains
-  read-only (#5589).
+  read-only (#5604, covers #5589).
 - The goal-continuation quiet period (`[goal] continuation_delay_seconds`,
   added in #5508) now applies on every dispatch path. Previously the
   within-turn dispatch hook fired the next continuation prompt immediately
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -136,7 +136,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   placeholder while waiting for a branch-specific CNB mirror (#5547).
 - Fleet roster members in a selected Fleet now expose a visible edit affordance
   and a direct `m` model-picker shortcut, while the coordinator row remains
-  read-only (#5589).
+  read-only (#5604, covers #5589).
 - The goal-continuation quiet period (`[goal] continuation_delay_seconds`,
   added in #5508) now applies on every dispatch path. Previously the
   within-turn dispatch hook fired the next continuation prompt immediately
--- a/crates/tui/src/command_safety.rs
+++ b/crates/tui/src/command_safety.rs
@@ -405,10 +405,66 @@ const GITHUB_READONLY_PREFIXES: &[&str] = &[
     "gh workflow view",
 ];
 
+/// Normalize Windows absolute path spellings before any POSIX-style splitter
+/// (`shlex` / `shell_words`) or glob-charset gate in this module:
+///
+/// - `Path::canonicalize` on Windows embeds the verbatim prefix `\\?\C:\...`
+///   whose `?` trips the glob-charset gates and whose backslashes the POSIX
+///   splitters eat as escapes; strip it so the remaining spelling resolves to
+///   the same location (device `\\.\` paths are preserved verbatim);
+/// - double the backslashes of Windows-absolute-path-like words so the
+///   splitters round-trip the real path instead of `C:\Users\...` collapsing
+///   to `C:Users...`.
+///
+/// Words that do not look like Windows absolute paths are untouched, so POSIX
+/// escapes and unix hosts are unaffected.
+pub(crate) fn normalize_windows_command_paths(command: &str) -> String {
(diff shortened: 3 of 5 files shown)


Hmbown:
Thank you for this — I verified it and it is doing more work than the title suggests: **this PR unblocks the whole open-PR queue.**

`main` (6110159bd) is currently red on two required checks, and every open PR inherits both:

1. **`Test (windows-latest)`** — `tools::shell::tests::readonly_operands_are_workspace_bounded_and_symlink_aware` and `tools::subagent::tests::read_only_inspection_roles_execute_pwd_and_absolute_git_log` fail identically on #5611 and #5609. Your diagnosis is right and it is the part #5606 missed: #5606 stripped the `\\?\` prefix *at classification time*, but `shell_words::split` had already eaten the separators, so the operand arrived as `\\?C:UsersrunneradminAppData...`. Normalizing before the split is the correct place.
2. **`Version drift`** — reproduced against `origin/main`:
   ```
   ::error::Feature commit 6110159bd0bb references #5604, but no release-note receipt exists in CHANGELOG.md docs/CHANGELOG_ARCHIVE.md.
   ```
   Your `(#5589)` → `(#5604, covers #5589)` edit is exactly the right fix, and it is the only PR in the queue carrying it.

Your run is green on both `Version drift` and `Test (windows-latest)` (only `Test (macos-latest)` was still running when I checked).

One non-blocking observation for the record, not a change request: `normalize_windows_command_paths` runs the `\\?\` strip and the backslash-doubling on all platforms, and its output feeds `hardened_readonly_argv`, which builds the argv that actually executes. So a POSIX operand containing a literal `\\?\` or `\\` prefix would be rewritten there too. In practice no real POSIX command looks like that, and keeping it platform-independent is what lets the classifier tests run on every runner, which is worth more than the edge case. Flagging it only so the tradeoff is written down.

Nice, well-tested fix — the `windows_verbatim_and_drive_operands_survive_posix_split` table covering `\\?\`, drive, UNC, and device spellings is exactly the right shape.

Hmbown:
Follow-up: **every check is now green** — `Test (macos-latest)` came back `pass` (42m35s), so the full matrix plus `Version drift`, `Lint`, `Safety gate` and the rest are all passing. `mergeStateStatus: CLEAN`.

For the record on why this one matters more than its diff size suggests — the two gates it fixes are currently red on **five** other open PRs (#5592, #5593, #5594, #5611, and #5609's Windows lane), all inheriting the breakage from `main` rather than causing it. This PR is the single change that clears both.

aboimpinto:
CI is fully green, including the previously failing **`Test (windows-latest)`** job (19m53s). The fix normalizes Windows verbatim/backslash path spellings at every judging site (both `command_safety` classifiers and both `shell.rs` split sites), so:

- `tools::shell::tests::readonly_operands_are_workspace_bounded_and_symlink_aware` ✅
- `tools::subagent::tests::read_only_inspection_roles_execute_pwd_and_absolute_git_log` ✅

Once merged, FEAT-019 (PR #5609) will be rebased and its Windows job should go green as well. Ready for review/merge.

Paulo Aboim Pinto

PR #5616, opened 2026-08-25
fix(tui): move git_status/git_diff off the async executor thread

rafaelcavalheri:
## Summary
- `GitStatusTool`/`GitDiffTool` in `crates/tui/src/tools/git.rs` called the blocking `std::process::Command::output()` directly inside async `execute()`, which can stall the tokio worker pool and hang the whole session with no error, timeout, or approval prompt (both tools are `ApprovalRequirement::Auto` and `supports_parallel() == true`, so several read-only tool calls firing concurrently can exhaust the worker pool at once).
- `crates/tui/src/tools/git_history.rs` (`log`/`show`/`blame` actions) already has the correct fix, wrapping the same kind of call in `tokio::task::spawn_blocking` (added in 346bfe3b6, scoped only to `git_history.rs`/`review.rs`).
- This PR mirrors that fix in `git.rs`: adds a `run_git_command_async` wrapper around the existing `run_git_command`, and switches `GitStatusTool::execute`/`GitDiffTool::execute` to call it instead of the blocking function directly.

## Test plan
- [x] `cargo check -p codewhale-tui --lib` — compiles cleanly
- [x] `cargo test -p codewhale-tui --lib tools::git::` — 6 passed

### DIFF
--- a/crates/tui/src/tools/git.rs
+++ b/crates/tui/src/tools/git.rs
@@ -80,7 +80,7 @@ impl ToolSpec for GitStatusTool {
         }
 
         let command_str = format_command(&git_ctx.working_dir, &args);
-        let output = run_git_command(&git_ctx.working_dir, &args)?;
+        let output = run_git_command_async(git_ctx.working_dir.clone(), args).await?;
 
         if !output.status.success() {
             let stderr = String::from_utf8_lossy(&output.stderr);
@@ -182,7 +182,7 @@ impl ToolSpec for GitDiffTool {
         }
 
         let command_str = format_command(&git_ctx.working_dir, &args);
-        let output = run_git_command(&git_ctx.working_dir, &args)?;
+        let output = run_git_command_async(git_ctx.working_dir.clone(), args).await?;
 
         if !output.status.success() {
             let stderr = String::from_utf8_lossy(&output.stderr);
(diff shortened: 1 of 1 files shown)


Hmbown:
Landed on the 0.9.12 integration branch as 868fb7f6f — `fix(tui): run git_status/git_diff off the async executor (#5616)` — with your authorship preserved as a `Co-Authored-By` trailer, which is how this repo credits taken PRs.

Your diagnosis was correct, and reviewing it surfaced something that makes the impact worse than the PR text: both tools declare `supports_parallel() == true`, and the engine drains an advertised-parallel batch from a single task (`crates/tui/src/core/engine/tool_execution.rs`), so an inline blocking git spawn doesn't just stall this one tool — it serializes the entire parallel batch. Your fix removes exactly that.

Your two hunks landed unchanged. Adaptation on our side, per lane convention for taken PRs:

- Extended the wrapper's doc comment (why the offload is not cosmetic, and that the wrapper must keep routing through `run_git_command` so `GIT_OPTIONAL_LOCKS=0` from #5617's fix stays attached — your wrapper already does this correctly; the comment pins it against future refactors).
- Two regression tests: `async_offload_matches_the_sync_path_byte_for_byte` (the wrapper is a pure offload) and `readonly_tools_do_not_rewrite_the_users_index` (end-to-end lock that the offload still suppresses opportunistic `.git/index` rewrites — composed with #5617, verified empirically on git 2.51.0).

Gates: rustfmt clean, clippy clean, full `codewhale-tui` lib suite 11399 passed / 0 failed / 13 ignored.

For what it's worth, your instinct to mirror the `git_history.rs` wrapper (346bfe3b6) rather than invent a new mechanism was exactly right — it's now character-identical in both files, which is where we want it. Thanks for the clean report and the minimal diff; both made this easy to take.

Issue #5618, opened 2026-08-25
Replace internal `git` CLI reads with gix (gitoxide)


LmeSzinc:
## Problem

codewhale shells out to the real `git` CLI for all internal read operations — repository badge probes, workspace context, `@status`/`@diff` mentions, and the model-callable Git tools. After #5617, two costs remain:

1. **Process spawn overhead**: the internal probes still spawn roughly **48 git processes per minute** while the TUI is idle in a git repo (≈190/min before the lock-reduction issue). Each spawn pays process creation + CLI startup for a computation that is a few milliseconds in-process.
2. **Unnecessary dependency on the git binary**: internal badge and context reads could work purely from the repository's on-disk format, without requiring git to be installed at all.

## Proposed solution

Migrate codewhale's **internal read-only** git operations from the `git` CLI to **gix** (gitoxide, pure Rust). Requirement-level scope (implementation follows the project's own conventions; no specific files are prescribed). This issue only makes sense after the lock-reduction issue is merged: the migration replaces the lock-free CLI reads with in-process reads and removes the event-driven refresh path's remaining process spawns.

- Add `gix` as a workspace dependency, pinned to an exact version, with the minimal feature set covering the read paths.
- Replace the following internal operations with gix in-process equivalents:
  - repository discovery / work-tree root resolution
  - HEAD / branch resolution (including detached-HEAD short hash)
  - working-tree status computation (staged / modified / untracked / conflicts)
  - ahead/behind counts vs upstream
  - worktree listing
  - remote origin URL and git config reads
  - commit log / show / blame (model-facing Git tool — may be a later phase)
- **Keep the `git` CLI for real write operations** (side-repo snapshot commits, worktree create/remove, branch deletion) and for agent-executed shell commands (`exec_shell` remains the model's surface and must stay a real CLI).

## Expected gains

- **Zero subprocess spawns for internal reads**: the idle ~48 processes/min (baseline after the lock-reduction issue) drop to **~0** — badge and context data become pure in-process computation.
- **No lock files, permanently**: gix reads never touch `.git/index.lock`, eliminating the commit-collision problem at the root, without environment-variable mitigations.
- **Faster refresh**: in-process status computation (~5-30ms) vs. process spawn + CLI startup per probe.
- **Works without the git binary**: badge and context reads succeed from the on-disk format even where `git` is not installed; write paths still report cleanly when git is missing.
[...]

Issue #5586, opened 2026-08-23
Decompose the mega files: lib.rs (18.7k), config.rs (12.3k), client.rs (11.1k), runtime_threads.rs (9.3k)


Hmbown:
## Combined Core execution: C09

[Core execution plan](<https://linear.app/shannon-labs/document/codewhale-core-execution-plan-a11e2e1c601d>) owns order and scope. This issue contributes to C09; the linked plan owns the complete packet and any outstanding candidate acceptance. Record this issue's path claims and implementation evidence here.

**Issue mapping:** [SHA-6291 / GitHub #5586](<https://linear.app/shannon-labs/issue/SHA-6291/decompose-the-mega-files-librs-187k-configrs-123k-clientrs-111k>)
  covers cohesive large-file decomposition across C03–C09; move whole behaviors
  and owning tests, with checked dependency boundaries rather than line-count
  targets. [SHA-6286 / GitHub #5580](<https://linear.app/shannon-labs/issue/SHA-6286/streamline-tui-monoliths-split-event-looprs-match-into-per-domain>) moves remaining
  event handling into domain reducers and leaves dispatch in `event_loop.rs`;
  preserve the existing reducer pattern and migrate touched arms completely.
  [SHA-6287 / GitHub #5581](<https://linear.app/shannon-labs/issue/SHA-6287/event-granularity-audit-surfaces-that-stall-at-turn-boundaries-despite>) supplies the per-call
  telemetry consumer requirements implemented in C11 with #6011: live token/time
  display, cache visibility and durable mid-turn accounting, with authoritative
  completion reconciliation and no double counting. Credit already-live cost,
  background cost, timing and context surfaces.

[SHA-5509 / GitHub #3957](<https://linear.app/shannon-labs/issue/SHA-5509/refactor-split-shared-modal-infrastructure-and-owned-views>) separates shared modal
  infrastructure from owned Config/SubAgents views and makes all provider
  credential entry use the same masking, validation and geometry primitive.
  Prove paste/delete/Enter/Esc/failure/retry and 80x24/120x32 wrapping; secrets
  must never reach screenshots, logs or errors.
  [SHA-5506 / GitHub #3954](<https://linear.app/shannon-labs/issue/SHA-5506/refactor-split-setup-wizard-constitution-runtime-provider-and>) splits setup by
  constitution, runtime, provider, persistence/report and navigation ownership
  where warranted; preserve constitution checkpoints and onboarding handoff.
[...]

PR #5628, opened 2026-08-26
Enterprise launch readiness: operator packet, Codewhale launch, #5585, #5617

### LINKED ISSUE #5585: Test setup_confirm_toast_names_secret_store_and_global_scope dies by stack overflow
`tui::ui::tests::setup_confirm_toast_names_secret_store_and_global_scope` SIGABRTs with 'has overflowed its stack' — reproduced on `main` and on `codex/v0912-integration-20260823` (macOS, nextest default stack). Pre-existing, not introduced by the 0.9.12 cycle (bisected: fails identically at 12553c5e1 and on main). Needs triage before release: either a genuinely deep recursion in the setup-confirm render path (product bug) or a test that needs a bigger stack / iterative rewrite. Found while running full slices for the dead-code sweep.

### LINKED ISSUE #5617: Reduce background git command runs and avoid git probes holding `.git/index.lock`
## Problem

While developing with codewhale inside a git repository, `git commit` occasionally fails because the repository is locked (`Unable to create '.../.git/index.lock': File exists`).

Root cause: codewhale's internal read-only probes run the real `git` CLI in the workspace, and `git status` / `git diff` opportunistically refresh the index — which briefly creates `.git/index.lock` in the **user's** repository. With the TUI open and idle, the most aggressive probe (the repository chrome badge) fires every **2 seconds** and runs **6 git commands** per probe:

- `git rev-parse --show-toplevel`
- `git symbolic-ref --short HEAD` (fallback: `git rev-parse --short HEAD`)
- `git status --porcelain`
- `git rev-list --left-right --count @{upstream}...HEAD`
- `git worktree list --porcelain`
- `git rev-parse --git-common-dir`

Combined with the 15-second workspace badge probe (`git rev-parse --abbrev-ref HEAD` + `git status --short`), this spawns roughly **190 git processes per minute** while the TUI is idle in a git repo — every `status` is a potential index-lock collision with the user's own `git commit`. The read-only shells used by the agent already set `GIT_OPTIONAL_LOCKS=0`, but the internal probes do not.

### 1. Make all status-refresh-related reads lock-free

Use `git --no-optional-locks status/diff` (git ≥ 2.15; equivalent to `GIT_OPTIONAL_LOCKS=0`), which skips the index refresh and therefore **never creates `.git/index.lock`** — without changing command output semantics:

- Set the env var once in the central git executor used by all internal read paths (workspace badge, `@status`/`@diff` mentions, the model-callable `Git` tools, review flows, GitHub CLI helper reads, and the snapshot side-repo runner — harmless for real writes).
- Add it to the few remaining status/diff call sites that bypass the central executor (the chrome probe and subagent claim verification).

`rev-parse` / `symbolic-ref` / `rev-list` / `config` / `remote get-url` / `worktree list` do not write the index and need no change.

### 2. Trim the periodic probe

The periodic probe keeps only the minimal command set needed for the status badge. The ahead/behind computation and the worktree listing move to the explicit refresh path (after events / when the worktree manager opens). The periodic probe drops from **6 commands to 3** (`rev-parse --show-toplevel`, `symbolic-ref --short HEAD`, `status --porcelain`).

### 3. Watch git metadata to trigger updates (event-driven refresh)

Introduce a small cross-platform file-watching dependency (pure Rust) that watches the repository's git metadata (`.git` — HEAD, refs, index, logs, worktrees, **excluding `objects/`**) and triggers a refresh when it changes, instead of waiting for the next timer tick.

**Trigger logic:**

- Any write to the git metadata directories (branch switch, `git add`, `git commit`, `git stash`, fetch, worktree add/remove) marks the badge stale.
- Events should be debounced so a burst of git writes (e.g. `git add -A` followed by a commit) coalesces into a single refresh. The debounce window, liveness cap, and event filtering are left to the implementer's judgment — the only hard requirements are that one burst yields exactly one refresh and that continuous events cannot starve the refresh forever.
- On debounce expiry the existing refresh path runs once. This also covers agent-executed `git` commands, which today rely on a separate post-tool refresh hook.
- A slow safety-net poll (e.g. every 30s) remains for the data the metadata watcher cannot see — working-tree edits made by an external editor (modified/untracked counts) — and for filesystems without reliable change events (network shares).

**Considerations:**

- The watch set must exclude `objects/`: every commit writes object files, so watching it turns each commit into a large event burst with no badge-relevant signal — the HEAD/index/refs events already carry the commit signal.
- OS change-event APIs are directory-based, and `.git` is small (a few dozen directories, essentially independent of project size): registration cost and steady-state overhead are negligible, with no watch/descriptor-limit risk even on very large repositories (unlike watching the whole working tree).
- Branch, staged, ahead/behind, and worktree-list changes are all driven by git metadata writes — where most badge staleness comes from — and become visible within ~1-2s instead of up to one poll interval later.
- What it cannot cover: external edits to working-tree files. Those stay on the safety-net poll (freshness within ≤30s). Watching the working tree itself is a possible later enhancement but scales with project size and hits OS watch limits on large repositories, so it is deliberately out of scope here.
- On filesystems without reliable change events (NFS/SSHFS/cloud-synced drives), the watcher degrades to polling — behavior is unchanged, no regression.

### 4. Fast-fail watch establishment

When the watcher tries to establish the git metadata watch and the watch path does not exist (the workspace is not a git repository yet, or git metadata is absent), **skip all remaining refresh work for this update cycle** — no probe, no retry — and treat the **next update cycle as a fresh attempt to establish the watch** (so a later `git init` is picked up automatically).

- The watcher state is simply "watch established" vs "not established": when not established, only the establishment attempt runs, and it fails fast with a single cheap path-existence check when the path is missing.
- This replaces the current behavior — a non-git workspace keeps spawning failing `git` probes every ~2s because the unavailable state is never cached as permanent — with one directory stat per update cycle.

### Idle (TUI open, no agent running, no changes)

| Metric | Before | After 1-2 (no watcher) | After 3 (watcher) |
| --- | --- | --- | --- |
| chrome probe rate | every 2s (30/min) | every 2s (30/min) | event-driven (0 idle) |
| Commands per periodic probe | 6 | 3 | — (recompute on `.git` events only) |
| chrome probe processes/min | ~180 | ~90 | ~0 |
| workspace badge processes/min | ~12 | ~12 | ~2 (30s safety poll) |
| **Total processes/min** | **~192** | **~102** (≈47% less) | **~2** (≈99% less) |
| `index.lock` creation by internal probes | possible (status refresh) | **never** (lock-free reads) | **never** |

Net: lock-free reads remove codewhale's contribution to `.git/index.lock` contention entirely — the direct fix for the reported `git commit` failures — while the badge data stays correct and fresh.

### Worst-case busy (agent turn with file writes)

The chrome/workspace probes are already suppressed while a turn is live, so busy-state git work is dominated by intentional turn bookkeeping, which the four changes do not alter:

- pre/post-turn side-repo snapshots: ~8 commands per turn
- pre-write-tool snapshot: ~4 commands per `write_file` / `edit_file` / `apply_patch`
- per-turn `turn_meta` collect: ~2-3 commands
- post-tool immediate refresh: ~2-3 commands per tool completion

A busy turn with ~5 write tools therefore runs roughly **40-60 git commands per turn** — unchanged before/after (side-repo snapshots never touch the user's `.git`, and these paths are already event-driven). The existing lever for busy-state git work remains the `[snapshots]` config (`enabled = false`).

Hmbown:
## What

Closes #5585
Closes #5617

Execute the app launch-readiness handoff as enterprise-grade product work, without colliding with the Crusoe/Composio/Aperture lane.

- Add `docs/ENTERPRISE.md` (and zh-Hans) as the operator/security review packet: local runtime, BYOK, first-party telemetry kill switches, local crash dumps, sandbox, air-gap controls, concurrent-session store lock, and Tailscale web.
- Publish it on codewhale.net at `/docs/enterprise` (en/zh), registered in the docs map, sitemap, and llms.txt.
- Point every localized README and `SECURITY.md` at that packet.
- Write process panic dumps and fatal-signal markers only under `$CODEWHALE_HOME/crashes` (otherwise `~/.codewhale/crashes`). The process panic hook no longer creates `~/.deepseek/crashes`.
- Record [#5630](https://github.com/Hmbown/CodeWhale/issues/5630): the runtime-store owner lock is one Codewhale per default store. The lock stays. Until the per-session root lands, the startup error names `CODEWHALE_RUNTIME_DIR`.
- `codewhale web --tailscale` publishes the loopback browser client on the current Tailscale tailnet via `tailscale serve`. HTTP stays on `127.0.0.1`. Not Funnel.
- Launch surface: the whale rises once (~640 ms). The wordmark writes **Codewhale** left to right in the current cyan, then settles. Caption and prompt stay readable. `low_motion` / `NO_ANIMATIONS` keep the settled mark. Not a splash; input is not delayed.
- ConfigToml and TUI ConfigFile parses now run on a dedicated 16 MiB stack thread so guided-setup save (and `/v1/config/reload`) cannot overflow a 2 MiB debug worker.
- Background git chrome probes pass `--no-optional-locks` and refresh at most every 10s, so an idle TUI no longer holds `.git/index.lock` against a user commit.

## Why this is not PostHog / Sentry

The handoff asked to “wire up PostHog” and “add Sentry.” That would violate the shipped first-party telemetry contract in `docs/TELEMETRY.md`. This PR documents that fact instead of adding vendors. There is no PostHog or Sentry dependency in the workspace.

## Not in this PR

- Cloudflare production deploy, device-flow login, BYOK vault UI, package publish
- Provider smoke tests and the running benchmark campaign
- M-Maciej’s per-session store rewrite for #5630
- Winget snapshot bump to 0.9.11
- Live Tailscale node e2e
- The `runtime_threads.rs` mega-file split (#5586)
- Replacing git CLI probes with gix (#5618)

The Crusoe/Composio/Aperture work stays on its own branch.

<div><a href="https://cursor.com/agents/bc-01a03fea-626b-77ba-bd79-9c4fc8f58ed3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a03fea-626b-77ba-bd79-9c4fc8f58ed3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Hmbown:
Closing as never-merge per the standing boundary in DIRECTION.md / TRUTH_LEDGER. The enterprise-launch scope (operator packet, #5585, #5617) was never gated for merge as a unit; its occupied local checkout is explicitly preserved as evidence and untouched. Any wanted pieces should land as individually-gated slices against current main. Reopen only by founder direction.

PR #5634, opened 2026-08-27
fix(runtime): scope the thread store per session (#5630)

Hmbown:
## Summary
- Closes #5630: interactive Runtime thread stores now default to `$CODEWHALE_HOME/sessions/<session-id>/runtime`, so multiple Codewhale processes on one machine no longer share `~/.codewhale/tasks/runtime`.
- The exclusive `RuntimeProcessOwnerLock` is unchanged. Two processes on the **same** session (or the same `CODEWHALE_RUNTIME_DIR`) still fail closed. The startup error names `CODEWHALE_RUNTIME_DIR` for an intentional shared root.
- The runtime API / `TaskManager::start_with_runtime_manager` path still uses `$CODEWHALE_HOME/tasks/runtime` unless overridden.

Design proposed by @M-Maciej. No PR from them existed at implementation time; authorship is preserved via `Co-authored-by`.

## Testing

- [x] focused `codewhale-tui` lib tests (see below)
- [ ] `cargo fmt --all -- --check`
- [ ] `cargo clippy --workspace --all-targets --all-features --locked` (warning-free under the CI allow list)
- [ ] `cargo test --workspace --all-features --locked`

Focused evidence:

- `runtime_threads::tests::session_scoped_runtime_default_lives_under_the_session_directory`
- `runtime_threads::tests::session_scoped_runtime_roots_do_not_share_the_process_owner_lock`
- `runtime_threads::tests::explicit_runtime_dir_override_beats_session_scope`
- `runtime_threads::tests::runtime_manager_store_has_one_lifetime_process_owner` (lock still exclusive)
- `task_manager::tests::task_and_runtime_roots_honor_explicit_codewhale_home`

## Checklist

- [x] Updated docs or comments as needed
- [x] Added or updated tests where relevant
- [ ] Verified TUI behavior manually if UI changes
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address

Made with [Cursor](https://cursor.com)

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5634" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #5638, opened 2026-08-27
fix(runtime): scope the thread store per session (#5630)

### LINKED ISSUE #5630: Bug 0.9.12 integration: runtime store owner lock
# Runtime store owner lock blocks multiple codewhale sessions on one machine

## Problem

The v0.9.12 integration tree (commit `80f026e7c`, "feat(runtime): unify
managed Chat with native runtime threads") introduces a machine-global
single-owner lock that hard-fails every codewhale after the first one on a
host: Effectively breaking multi session of codewhales.

```
error: This Runtime thread store is already active in another process;
close the other Runtime before retrying
```

`RuntimeProcessOwnerLock::acquire` takes an exclusive non-blocking `flock`
on `<store-root>/runtime-process.owner.lock` at startup and holds it for
the process lifetime. The store root defaults to
`~/.codewhale/tasks/runtime` — the same path for every session on the
machine — so the per-store guard becomes a **one-codewhale-per-machine**
guard. The failure is a hard startup error, not per-session isolation.

I run exactly the deployment the lifecycle-outbox work targets: many
codewhale sessions per machine under external supervision (a terminal
manager / automation harness). The outbox issue says it outright —
"many codewhale sessions on one machine". With this lock in the tree, the
second session on a host cannot start, so the supervised multi-session
deployment breaks on the same release that ships the supervision
features. Today's practical damage: a lane creation spawned three
sessions; one won the lock, two died at startup.

## Verification

Reproduced on a clean machine (no codewhale installed, no state) with a
pristine build of the upstream v0.9.12 integration tip `96bc9e79c`:

- instance 1 boots, creates `<home>/tasks/runtime/runtime-process.owner.lock`, runs;
- instance 2, started while 1 holds it, fails with the error above.

`CODEWHALE_RUNTIME_DIR` / `DEEPSEEK_RUNTIME_DIR` relocates the store and
does work around it (verified: a session with a per-session override
starts fine), which shows the root was meant to be relocatable — but
nothing scopes it per session, so every session collides on the default.

## Proposed solution

Scope the store (or at least the owner claim) per session rather than per
machine.

- default the store root per session — e.g. under the session's own
  state directory (the durable store already holds per-thread data, so a
  per-session root is the semantically honest default).

I'll send pr with it.

Hmbown:
## Summary

Closes #5630. The runtime process-owner lock stays exclusive; the default store root is now per interactive session (`$CODEWHALE_HOME/sessions/<id>/runtime`) so a second Codewhale on the same machine can start. `CODEWHALE_RUNTIME_DIR` still selects a shared root when that is intended. The runtime API server still uses `$CODEWHALE_HOME/tasks/runtime` unless overridden.

Credit: @M-Maciej reported the one-process-per-machine failure and proposed per-session store scoping.

Supersedes #5634 (same diff; that PR's lint gate rejected a bot co-author trailer).

## Test plan

- [x] Existing lock still fails closed for two processes on the same store root
- [x] Two interactive sessions with distinct session ids can start
- [ ] Hosted CI green (Test ubuntu/macos/windows, Lint, Safety gate)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Context-pressure warnings and critical alerts now remain visible in sticky UI
   status until compaction or explicit dismissal, instead of disappearing into
   scrolling turn metadata (#5620).
+- The Runtime thread store defaults to a per-session root
+  (`$CODEWHALE_HOME/sessions/<id>/runtime`) so multiple Codewhale processes on
+  one machine no longer share one owner lock (#5630). The exclusive lock is
+  unchanged; `CODEWHALE_RUNTIME_DIR` still selects a shared root when that is
+  intended.
 - Session token totals now include display-only per-model-call deltas while a
   turn is running, including input/output and cache-class counters; the
   authoritative `TurnComplete` totals still reconcile exactly once (#5581).
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -68,6 +68,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Context-pressure warnings and critical alerts now remain visible in sticky UI
   status until compaction or explicit dismissal, instead of disappearing into
   scrolling turn metadata (#5620).
+- The Runtime thread store defaults to a per-session root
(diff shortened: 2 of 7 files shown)


Hmbown:
Merging: hosted CI is green (Test ubuntu/macos/windows, Lint, Safety). Exclusive flock stays; interactive sessions default to `$CODEWHALE_HOME/sessions/<id>/runtime`. Credit to @M-Maciej for the report and per-session scoping proposal.

PR #5641, opened 2026-08-27
feat(runtime-api): rescue per-thread usage with CNY coverage

Hmbown:
## Summary

This is a clean, current-`main` rescue of contributor PR #5626.

- Preserves Ben Gao's original feature commit and authorship exactly.
- Adds `GET /v1/threads/{id}/usage` using the existing provider-aware usage ledger.
- Persists parent and routed-child session cost without double-counting.
- Preserves USD/CNY coverage and CNY unpriced-reason evidence across reloads.
- Adds a regression test for truncated CNY journal coverage.
- Retains the newer session-store ownership and post-compaction accounting already on `main`.
- Freezes live workflow events after a terminal cancel so a racing VM/child completion cannot append a duplicate cancel receipt (the ubuntu-latest failure on run 33050954290).

The source patches match the original contributor feature and maintainer follow-up. The only metadata repair is removal of an unrecognized Cursor co-author trailer that made the original PR's lint lane fail. Contributor credit remains attached to Ben's commit.

## Verification

- focused thread-usage endpoint coverage
- focused parent/child session-cost split and coverage
- focused CNY unpriced-reason persistence
- focused two-currency per-thread aggregation
- focused truncated-journal USD/CNY coverage
- `workflow_cancel_interrupts_vm_and_blocks_further_spawns` (10 local reruns)
- co-author credit validation
- rebased onto current `origin/main`

Supersedes #5626 after this rescue lands; do not close the contributor PR before merge.

No-Issue: Preserving-authorship rescue of existing contributor PR #5626.

### DIFF
--- a/crates/tui/src/runtime_api.rs
+++ b/crates/tui/src/runtime_api.rs
@@ -66,7 +66,7 @@ use crate::runtime_threads::{
     CompactThreadRequest, CreateThreadRequest, ExternalApprovalDecision,
     MAX_RUNTIME_EVENT_REPLAY_TAIL, RuntimeThreadManager, RuntimeThreadManagerConfig,
     SharedRuntimeThreadManager, StartTurnRequest, SteerTurnRequest, ThreadDetail, ThreadListFilter,
-    ThreadRecord, TurnItemKind, TurnRecord, UpdateThreadRequest, UsageGroupBy,
+    ThreadRecord, TurnItemKind, TurnRecord, UpdateThreadRequest, UsageGroupBy, UsageTotals,
 };
 #[cfg(test)]
 pub(super) use crate::runtime_threads::{RuntimeTurnStatus, TurnItemLifecycleStatus};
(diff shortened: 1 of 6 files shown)


Hmbown:
Rebased onto current `main` and fixed the ubuntu-latest failure from run 33050954290 / job 98446121611.

Root cause: `tools::workflow::tests::workflow_cancel_interrupts_vm_and_blocks_further_spawns` saw 6 events after a second cancel vs 5 after the first. A racing VM/child completion could still `push_event` after the run was already `Cancelled`, so the second cancel looked like a duplicate terminal receipt. The test coverage is intact; live recording now stops once status is no longer `Running`.

Feature authorship remains Ben Gao (`a76f9e3b85`). HEAD is `9a5649e895`.

Hmbown:
Windows CI on `8d207ba` failed the same cancel test, but at the first-child wait (`expect("workflow should spawn at least one child before cancel")`, 3.08s) rather than the duplicate-terminal assertion. Ubuntu on that SHA was green. Pushed `7b2b8d085` to keep the coverage and give the first-child poll 15s (still returns as soon as a child is admitted).

PR #5629, opened 2026-08-26
fix(tui): persist context pressure warnings

wuisabel-gif:
## Summary

Addresses the approved display-only slice of #5620.

The existing context-pressure warning/critical signal is now promoted to persistent UI status:

- warning, high, and critical pressure remain visible in sticky status instead of disappearing into scrolling turn metadata;
- pressure updates can replace an earlier pressure status but do not clobber unrelated sticky error/status chrome;
- the warning clears when compaction starts or when the idle operator presses `Esc` with no composer/menu activity;
- existing turn metadata text, compaction thresholds, auto-compaction behavior, and model-facing guidance are unchanged.

## Tests

Passed locally:

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p codewhale-tui --lib 'tui::ui::tests::context_pressure_warning' --locked` (4 passed)
- `cargo clippy -p codewhale-tui --all-targets --locked -- -D warnings`

Coverage verifies persistent warning visibility across later status activity, explicit dismissal, compaction clearing, and silence below the pressure threshold.

No provider credentials or network access are required.

No-Issue: This PR addresses the approved persistent human-visible warning slice without automatically closing the broader #5620 issue.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5629" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #5661, opened 2026-08-27
fix(tui): make context pressure an agent directive

### LINKED ISSUE #5620: Context pressure warning is transient and the agent does not proactively react to it
# Bug: Context pressure warning is transient and the agent does not proactively react to it

## Metadata

- **Date observed:** 2026-08-26
- **Component:** CodeWhale runtime / agent turn metadata
- **Severity:** Medium (silent context degradation; no crash, but defeats a safety signal)
- **Environment:** Linux (x360), CodeWhale CLI/agent session, `Full Access` posture, sandbox disabled

---

## Summary

When a session's context window approaches its limit, CodeWhale emits a
**"context pressure" warning**. That warning is directed at the *agent*
(imperative wording: "prefer /compact, narrow scope, or finish the current
task"), but two problems combine to make it effectively invisible:

1. **It scrolls out of view almost immediately.** The warning is rendered as
   inline text in the per-turn metadata (`<turn_meta>`), which is pushed up the
   screen by the next tool output / assistant message. The user cannot reliably
   read it — in the session where this was observed, the user noticed it only
   "by luck", in their own words.

2. **The agent does not react to it.** The agent received the warning and simply
   continued its current task (writing a session-save file), and only
   acknowledged the warning *after* the user manually asked "did you receive a
   context pressure warning from codewhale". There is no mechanism or
   instruction that causes the agent to surface the warning, recommend
   compaction, or narrow its output scope when the warning is present.

Net result: the one signal that is supposed to prevent context overflow / prompt
degradation is neither reliably seen by the human nor acted upon by the agent,
so it fails its purpose silently.

---

## Observed behaviour (exact)

The warning text was:

```
Context pressure: warning — ESCALATED: prefer /compact, narrow scope, or finish the current task
```

It appeared in the per-turn `<turn_meta>` block, alongside `Current local date`,
`Current workspace`, `Current permission posture`, and `Current sandbox posture`.
It is not a separate, persistent UI element; it is one more line in the flowing
conversation log.

Sequence of events in the observed session:

1. A long, multi-item coding session filled the context window.
2. At some point the `Context pressure: warning — ESCALATED: …` line began
   appearing in the turn metadata.
3. The agent continued working normally (completed a "save session" request),
   giving no visible indication that it had seen or was acting on the warning.
4. The user, separately, happened to notice the line in the scrollback and asked
   the agent: *"did you recieve a context pressure warning from codewhale"*.
5. The agent answered yes, and only then suggested running `/compact`.

---

## Expected behaviour

Any of the following (ideally a combination):

- **Persistent, non-scrolling presentation.** The warning should be surfaced in
  a way that does not disappear into scrollback — e.g. a sticky banner, a
  dedicated system message, a status-bar indicator, or a repeated prompt — so
  the user can see it without hunting through the log.

- **Explicit agent reaction.** The agent should be instructed (in its system
  prompt / runtime guidance) that the "context pressure" warning is an
  *actionable signal*, not decorative text. On seeing it the agent should, at a
  minimum:
  - stop and tell the user the warning is active,
  - recommend `/compact` (or otherwise propose narrowing scope / finishing the
    current task),
  - and immediately economise its own output (shorter replies, fewer tool
    calls, no speculative investigation).

---

## Why this matters (impact)

- The warning's entire purpose is to avoid running out of context (which causes
  prompt truncation, lost detail, or degraded behaviour). If it is neither seen
  nor acted upon, the failure is *silent* — the user only discovers the problem
  after the context has already degraded.

- The wording is agent-directed ("prefer /compact …"), which means the design
  intent is that *the agent* respond. If the agent is not actually wired to
  respond, the warning is effectively a no-op.

- The user is left with the burden of noticing a transient line and then
  manually interrogating the agent ("did you receive the warning?"), which is
  backwards: the tool already knows the condition and should surface it
  proactively.

---

## Steps to reproduce

1. Start a CodeWhale agent session and drive a long task (many tool calls /
   large file reads) so the context window approaches its limit.
2. Observe the `Context pressure: warning — ESCALATED: …` line in the turn
   metadata.
3. Note that (a) the line scrolls out of view with the next output, and (b) the
   agent continues without mentioning or acting on it.
4. Confirm that the only way to learn about the warning is to scroll back and
   spot it, or to ask the agent directly.

---

## Possible causes / analysis

- **Presentation:** the warning is emitted as inline text in the conversation
  stream rather than as a dedicated, persistent UI affordance (banner/status
  bar/dialog). Anything inline will scroll away under normal tool output.

- **Agent wiring:** the agent's instructions do not appear to include any rule
  for reacting to the "context pressure" warning. The warning text reads like a
  directive ("prefer /compact …") but there is no corresponding instruction in
  the agent's system prompt, so the agent treats it as informational and carries
  on.

- **No acknowledgement channel:** there is no defined way for the agent to
  surface the warning *and* for the runtime to keep it visible (e.g. pin it) —
  so even a well-behaved agent would have to embed the message in its own reply,
  which is easy to miss.

---

## Suggested fixes

1. **Make the warning persistent.** Render it as a sticky banner / system note /
   status-bar indicator that remains visible (or is re-emitted) until the user
   acts (e.g. runs `/compact` or explicitly dismisses it).

2. **Add an explicit agent directive.** In the agent's system/instruction text,
   specify: *"When a `Context pressure` warning is present, stop and tell the
   user, recommend `/compact`, and reduce your output scope."* This turns the
   existing wording into an actually-enforced behaviour.

3. **Optional — provide a structured signal.** If feasible, pass the context
   pressure level to the agent as a structured field (e.g. a distinct
   `context_pressure` key) rather than free-form prose, so the agent can detect
   it reliably instead of pattern-matching prose that scrolls by.

---

## Additional context

- The session in which this was observed had already been saved to a
  `context-transfer-*.md` file, so `/compact` was safe and lossless — but the
  agent only recommended it after the user asked, rather than proactively.
- The same transient-presentation problem likely applies to any other runtime
  warnings emitted in the same inline `<turn_meta>` region (worth checking
  whether other warnings share the "scrolls out of view" flaw).

Hmbown:
## Summary

Second slice of #5620. Isabel's display-only sticky warning is already on `main` (#5629). The remaining failure was that the model-facing line still read like decorative turn metadata, so the agent kept working until the user asked whether the warning existed.

This PR:

- names the standing reaction in `BASE_PROMPT` (frozen KV-cache prefix): when turn metadata includes a Context pressure warning or critical line, tell the user immediately, recommend `/compact`, and reduce scope;
- makes the live `turn_meta` line an explicit `ACTIONABLE: tell the user now` directive;
- keeps the line byte-stable within each pressure band (no percentages, token counts, or headroom), so ordinary turns do not bust the prefix cache.

Does not close #5620: structured `context_pressure` fields and prompt-injection economisation remain follow-up.

Credit: report and recovery notes from @ronohara; display slice already landed as #5629 by @wuisabel-gif.

## Testing

- [x] `cargo fmt` on the touched files
- [x] `git diff --check`
- [x] `scripts/dev-test.sh tui context_pressure_message_emits base_prompt_carries_constitutional_core` — 2 passed

No provider credentials or network access.

## Checklist

### DIFF
--- a/.github/AUTHOR_MAP
+++ b/.github/AUTHOR_MAP
@@ -26,6 +26,7 @@ Mr-Moon121 = Jeffrey Luna <56358802+Mr-Moon121@users.noreply.github.com>
 lerugray = Ray Weiss <15095329+lerugray@users.noreply.github.com>
 lerugray@gmail.com = Ray Weiss <15095329+lerugray@users.noreply.github.com>
 reidliu41 = reidliu41 <61492567+reidliu41@users.noreply.github.com>
+ronohara = ronohara <4027321+ronohara@users.noreply.github.com>
 reid201711@gmail.com = reidliu41 <61492567+reidliu41@users.noreply.github.com>
 ousamabenyounes = Ben Younes <2910651+ousamabenyounes@users.noreply.github.com>
 benyounes.ousama@gmail.com = Ben Younes <2910651+ousamabenyounes@users.noreply.github.com>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Context-pressure warnings and critical alerts now remain visible in sticky UI
   status until compaction or explicit dismissal, instead of disappearing into
   scrolling turn metadata (#5620).
+- Context-pressure turn metadata is now an explicit agent directive: tell the
+  user immediately, recommend `/compact`, and reduce scope. The constitution
+  names that standing reaction; the live line stays byte-stable within each
+  pressure band so ordinary turns do not bust the prefix cache (#5620).
 - The Runtime thread store defaults to a per-session root
   (`$CODEWHALE_HOME/sessions/<id>/runtime`) so multiple Codewhale processes on
   one machine no longer share one owner lock (#5630). The exclusive lock is
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -75,6 +75,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Context-pressure warnings and critical alerts now remain visible in sticky UI
   status until compaction or explicit dismissal, instead of disappearing into
   scrolling turn metadata (#5620).
+- Context-pressure turn metadata is now an explicit agent directive: tell the
+  user immediately, recommend `/compact`, and reduce scope. The constitution
+  names that standing reaction; the live line stays byte-stable within each
+  pressure band so ordinary turns do not bust the prefix cache (#5620).
 - The Runtime thread store defaults to a per-session root
   (`$CODEWHALE_HOME/sessions/<id>/runtime`) so multiple Codewhale processes on
   one machine no longer share one owner lock (#5630). The exclusive lock is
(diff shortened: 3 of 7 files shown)


Hmbown:
Closing as never-merge per the standing boundary in DIRECTION.md / TRUTH_LEDGER: this change touches prompt authority (context pressure as an agent directive), which is a rail-protected surface — the same reason issue #5620 deliberately remains open as roadmap rather than closing on this implementation. Branch preserved; reopen only by founder direction.

PR #5664, opened 2026-08-27
perf: trim process startup, diagnostic dispatch, and foreground command latency

AdityaVG13:
perf: trim process startup, diagnostic dispatch, and foreground command latency

Three clusters of changes, each targeting waste found by profiling rather than guesswork: diagnostic subcommands built a 45-thread tokio runtime they never used; the models.dev catalog was parsed repeatedly per process and twice per byte on the boot path; and every foreground shell call carried a fixed 100 ms poll tick no matter how fast the command finished. All measurements below come from interleaved same-window batch pairs on one aarch64 Linux machine, with attribution from strace censuses and mid-boot sampling.

| Area | Before | After | Delta |
|---|---|---|---|
| `doctor` wall, quiet host, n=60 | 45.4 ms | 37.6 ms | -17.2% |
| rustc process launches per doctor run | 2 | 1 | one libLLVM load avoided |
| thread spawns per doctor run | 45 | 27 | -18 |
| `eval` wall, under load, 3 interleaved pairs | baseline | baseline | -8.1% / -10.4% / -6.5% (NEW wins 3/3) |
| `setup --status` wall, under load, 3 pairs | baseline | baseline | -17% / -21% / -16% (NEW wins 3/3) |
| boot cache-load stage | 44.0 / 52.7 / 48.6 / 51.7 ms | 13.9 / 14.1 / 13.9 / 14.4 ms | ~-70% (NEW wins 4/4) |
| interactive time to first paint | ~229 ms | ~187 ms | -42 ms |
| provider cutline pass, first paint medians | 195.7 / 196.9 / 192.9 / 197.2 ms | 185.0 / 190.9 / 190.3 / 193.8 ms | -2 to -6 ms (NEW wins 4/4) |
| foreground command completion detection, p50 | ~100 ms floor | ~10 ms | ~10x for instant commands |
| foreground bash `true`, tail (n=150, real MCP serve) | p95 16.2 / p99 21.5 / p100 35.9 ms | p95 13.9 / p99 18.6 / p100 21.6 ms | p95 -14%, p99 -13%, p100 -40% |

## Runtime sizing for read-only diagnostics

A strace census on the doctor path showed 88.7% of syscall time in futex behind 45 thread spawns: `build_runtime()` built one worker per CPU with 16 MiB stacks each for read-only commands that use none of it. `doctor`, `eval`, `setup --status`, and the sessions/diagnostics family now cap workers at 2 through a structured match mirroring `telemetry_command_is_read_only`. Interactive sessions and servers keep default sizing.

## Parse the bundled catalog once

`bundled_models_dev_catalog()` re-parsed the ~50 KB snapshot at every call site: client routing, provider picker, provider lake, and fleet identity each paid an independent serde parse. It now returns a `&'static` catalog backed by `OnceLock`. The asset is an `include_str!` constant, so sharing the parsed form is immutability-safe. One existing assertion is adapted to the new signature and the build-time deep-equality guard keeps its exact semantics.

## Capture the rustc banner during the probe

`RustC::resolve()` proved presence by running `rustc --version` and discarding stdout; the diagnostics path then launched a second rustc process to read the same banner, and each launch loads libLLVM. The probe now records the banner behind a `OnceLock` and diagnostics consume it after an availability check. The reported `rust:` line is byte-identical to the toolchain's own `rustc --version` output.

## Single-parse disk cache, v2 format

The persisted catalog stored a ~5 MB body JSON-escaped inside a JSON envelope, and the synchronous boot path parsed it twice plus a full-body string copy. Mid-boot sampling caught `serde_json::visit_map` inside the largest silent window of startup, alongside a 17.7 ms zero-syscall compute burst after config load. v2 is a one-line JSON metadata header followed by the verbatim body: one small header parse, one body parse, zero body copies. Every refresh writes v2 and v1 envelopes still load through a fallback, so the first run after upgrade migrates the cache with no manual step.

## Memoize provider resolution in the cutline pass

`apply_provider_model_cutlines` called `ApiProvider::parse` per offering. Parse scans every provider and its alias list with case-insensitive compares, and the live snapshot carries thousands of rows. Each distinct provider string is now resolved once through a map.

## Adaptive poll cadence for foreground shell completion

The foreground wait loop polled child status on a fixed 100 ms tick, so a command finishing in 2 ms was only noticed at the next tick. A simulated 8-tool turn spent ~400 ms of its ~526 ms wall in poll quantization. The tick now starts at 10 ms and doubles to the old 100 ms cap in all three wait loops, so instant commands are detected about 10x sooner while long-running commands reach the same steady state after one doubling step. A first A/B that compared only medians missed the win and was reverted; the tail distribution above is what justified re-landing it, so the revert and reapply are squashed away in this branch's history.

## Method

Latency claims come from interleaved OLD/NEW batch pairs on the same host in the same window, n=60 on a quiet host for doctor and three batch pairs under synthetic load where the host was busy. Goldens: `--help` and `--version` output byte-identical, the `rust:` line matches the toolchain banner, eval behavior identical across a 6-step run.

## Testing

`cargo fmt` clean. Config crate: 603 passed. TUI modules touched by this series (`dependencies`, `catalog`, `models_dev_live`, `provider_lake`, `shell`): 270 passed at the branch tip.

## Reviewer note, unrelated to this series

`runtime_api::tests::reload_config_preserves_profile_selected_named_custom_route` overflows its stack in the full lib-test run. It reproduces identically on a pristine checkout of `main` in a throwaway worktree, so it is pre-existing and untouched here.

### DIFF
--- a/crates/config/src/catalog.rs
+++ b/crates/config/src/catalog.rs
@@ -32,6 +32,7 @@
 //! [`ProviderCatalogCache`] tests).
 
 use std::collections::BTreeMap;
+use std::sync::OnceLock;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use serde::{Deserialize, Serialize};
@@ -186,16 +187,30 @@ impl CatalogOffering {
 /// honesty rule on omitted pricing (`UnknownOrStale`, never a fabricated zero).
 pub const BUNDLED_MODELS_DEV_JSON: &str = include_str!("../assets/models_dev.bundled.json");
 
+/// Parse-once cache for the committed bundled Models.dev snapshot.
+///
+/// The bundled asset is compile-time constant (`include_str!`), so its parsed
+/// form is immutable and safe to share process-wide. Before this cache, every
+/// call site parsed the full snapshot independently — the client route path,
+/// pickers, provider lake, and fleet identity each paid a full serde parse of
+/// ~50KB on their own first use (perf-attributed during the 0.9.x perf
+/// gauntlet: `ModelsDevCost` serde frames in startup profiles).
+static BUNDLED_MODELS_DEV_CATALOG: OnceLock<ModelsDevCatalog> = OnceLock::new();
+
 /// Parse the committed bundled Models.dev snapshot.
+///
+/// The first call parses; later calls return the shared parsed catalog.
 ///
 /// # Panics
 /// Panics only if the committed asset is not valid Models.dev JSON. The
 /// `tests::bundled_asset_parses` guard makes that a build-time failure, so this
 /// never panics in shipped builds.
 #[must_use]
-pub fn bundled_models_dev_catalog() -> ModelsDevCatalog {
-    ModelsDevCatalog::parse_json(BUNDLED_MODELS_DEV_JSON)
-        .expect("committed bundled Models.dev asset must be valid JSON")
+pub fn bundled_models_dev_catalog() -> &'static ModelsDevCatalog {
+    BUNDLED_MODELS_DEV_CATALOG.get_or_init(|| {
+        ModelsDevCatalog::parse_json(BUNDLED_MODELS_DEV_JSON)
+            .expect("committed bundled Models.dev asset must be valid JSON")
+    })
 }
 
 /// Bundled-layer [`CatalogOffering`] rows from the offline snapshot (#4188).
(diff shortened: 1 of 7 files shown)


Hmbown:
Thank you, Aditya. I verified all eight patches in this PR are already on `main` through #5667 (`6ef5f87d8182eb38a18cfff99e58bbf7ef01b20c`). Each original commit has an identical stable patch ID to its landed counterpart, and the landed commits preserve `AdityaVG13 <adityavgcode@gmail.com>` as author.

Landed commits: `dae07dd0d`, `cab71f543`, `ffbb0b519`, `6f6e5bbb7`, `9fa638e6a`, `0d8be2001`, `3364a8017`, `4fc1d594f`.

I’m closing this duplicate PR because the contribution is already shipped on `main`; merging this stale head again would duplicate the patches.

PR #5665, opened 2026-08-27
perf(tui): single-pass token accounting on per-turn pressure paths

AdityaVG13:
perf(tui): single-pass token accounting on per-turn pressure paths

The per-turn metadata build and the compaction decision each re-walked the full transcript more than once per step, and the streaming render re-processed the whole accumulated message on every chunk. This series makes those walks single-pass with byte-identical behavior, pins the equivalence with tests, and fixes one cache-invalidation bug found along the way.

| Path | Before | After |
|---|---|---|
| Compaction decision per step under pressure | 2 full-transcript estimate walks | at most 1 |
| `<turn_meta>` build per user or steer message | full-history deep clone, then estimate | estimate in place, arithmetic delta |
| Billed input above threshold | full estimate walk | skipped, guards preserved |
| Session restore | stale estimate could serve capacity checkpoints | cache invalidated, recomputed |
| Streaming render per chunk, no math present | full-content copy and transform | one byte scan, zero copies |

## One estimate per compaction decision

`compaction_decision_with_billed` called the same pure estimator twice per step once pressure existed: once in the pressure gate and once in the prune projection. It is now computed at most once and threaded to both consumers. Guard order (gate, prune projection, too-few-messages, retained-floor) is untouched and decisions are identical for all inputs. When provider-billed input alone proves pressure, the estimate walk is skipped entirely without bypassing the guards that keep a compaction loop from re-firing on every step.

## No full-transcript clone in the per-turn metadata build

The builder deep-cloned the entire message history just to append the composer text before estimating. It now estimates the history in place and adds the one-message delta arithmetically. The subtle part: the estimator inflates the summed byte count by a factor of 1.5 rounded up as a whole, so the delta is `floor(own * 3 / 2)` plus one exactly when the running byte sum is even and the composer text contributes an odd count. This is proven against a clone-and-estimate reference over 80,000 parity combinations and pinned per case by a test that also covers replayed reasoning, the arm where a naive helper undercounts.

## Session restore invalidates the estimate cache

Session sync assigned the restored message list directly to the session field, bypassing the revision bump the token-estimate cache keys on, so a restored conversation could be priced from a stale estimate. The bump is restored and a test pins that the cache recomputes after a restore.

## Zero-copy LaTeX pre-pass

The streaming render ran the math transform over the full accumulated content on every chunk, before the incremental markdown cache that itself only renders deltas. The transform now returns a `Cow<str>`: one byte scan for the three opening delimiters decides between borrowing the input untouched, the overwhelming case for streamed prose, and running the transform. Output is byte-identical either way. A test pins the borrowed path and that the `\(` form is not missed by the fast scan.

## Testing

`cargo fmt` clean. `compaction::` and `latex_render::`: 62 passed. Engine preview, pressure, and restore tests: 14 passed. `cargo check --all-targets`: clean, no warnings.

### DIFF
--- a/crates/tui/src/compaction.rs
+++ b/crates/tui/src/compaction.rs
@@ -283,7 +283,7 @@ pub(crate) fn is_compaction_checkpoint_message(message: &Message) -> bool {
     user_text_of(message).is_some_and(|text| is_compaction_summary_text(&text))
 }
 
-fn estimate_tokens_for_message(message: &Message, include_thinking: bool) -> usize {
+pub(crate) fn estimate_tokens_for_message(message: &Message, include_thinking: bool) -> usize {
     message
         .content
         .iter()
@@ -340,14 +340,14 @@ pub fn estimate_tokens(messages: &[Message]) -> usize {
         .sum()
 }
 
-fn message_has_tool_use(message: &Message) -> bool {
+pub(crate) fn message_has_tool_use(message: &Message) -> bool {
     message
         .content
         .iter()
         .any(|block| matches!(block, ContentBlock::ToolUse { .. }))
 }
 
-pub fn estimate_text_tokens_conservative(text: &str) -> usize {
+pub(crate) fn estimate_text_tokens_conservative(text: &str) -> usize {
     text.chars().count().div_ceil(3)
 }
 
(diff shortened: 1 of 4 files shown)


Hmbown:
Thank you, Aditya. I verified both patches in this PR are already on `main` through #5667 (`6ef5f87d8182eb38a18cfff99e58bbf7ef01b20c`). The original commits and landed commits have identical stable patch IDs, and the landed commits preserve `AdityaVG13 <adityavgcode@gmail.com>` as author.

Landed commits: `692cf1efb` and `5edb8c3f9`.

I’m closing this duplicate PR because the contribution is already shipped on `main`; merging this stale head again would duplicate the patches.

PR #5689, opened 2026-08-28
fix: show all DeepSeek catalog models in configured picker

musichen:
## Summary

Hi @Hmbown - this fixes two related DeepSeek picker issues in the default `configured` view:

- The picker rendered the catalog metadata family (`deepseek-thinking`) as a model-like heading.
- It showed only the currently selected DeepSeek model, hiding the other official catalog models until the operator switched to the broader catalog view.

The picker now uses a stable `DeepSeek` heading for first-party DeepSeek rows and keeps the active provider's catalog rows visible in `configured`. Other providers remain subject to the existing configured/enabled filter.

`ModelPickerView::visible_model_rows` is the visibility boundary for the default `configured` view. Previously, that path delegated to `model_row_visible_by_default`, which admitted only `auto` and rows marked `enabled`; consequently, catalog entries for the active DeepSeek route were hidden until each was explicitly selected or the user switched to the broad catalog view.

This change carries the active provider into that visibility decision. It keeps every catalog row for the selected provider visible, while preserving the existing conservative filter for other providers. The picker therefore exposes the complete first-party DeepSeek choice set without turning the configured view into an unrestricted cross-provider catalog.

The display change is intentionally separate: `catalog_family_for` now emits the provider-facing `DeepSeek` heading for first-party DeepSeek rows rather than surfacing mutable Models.dev family metadata such as `deepseek` or `deepseek-thinking`. The two focused regressions cover both contracts.

## Reproduction

1. Configure the built-in `deepseek` provider with `deepseek-v4-pro` as the default model.
2. Launch CodeWhale and run `/model`.
3. Before this change, the configured view shows a raw `deepseek-thinking` family heading and only `deepseek-v4-pro`.
4. With this change, the configured view shows a `DeepSeek` heading and all official DeepSeek catalog models:
   - `deepseek-v4-pro`
   - `deepseek-v4-flash`
   - `deepseek-v4-flash-vision-exp`

### Before

Before: raw deepseek-thinking heading and one visible model

### After

After: DeepSeek heading and all three official DeepSeek models

## Validation

```text
cargo fmt --all -- --check
cargo test -p codewhale-tui deepseek_picker_heading_hides_legacy_family_metadata --lib
cargo test -p codewhale-tui configured_view_keeps_active_provider_catalog_models_visible --lib
cargo build --bin codewhale
```

Manual verification used the built local binary and `/model`; the attached After screenshot is from that run.

No-Issue: picker visibility defect with no open tracking issue; the picker showed the catalog metadata family as a heading and hid unselected first-party models.


musichen:
@Hmbown, great work on the CodeWhale coding agent. I recently started supporting the project and wanted to contribute this fix. Thank you for building it!

PR #5711, opened 2026-08-29
feat(runtime-api): rehydrate persisted goals and run host-managed continuation loop

gaord:
## Summary

No-Issue: runtime goal records were durable but never reached hosted engines; this wires the persisted record into the engine and adds the host-side re-arm loop.

Restore a thread's durable goal into the engine on load so the goal loop, prompt surface, and `update_goal` tool operate on the persisted record from the first turn, preserving accumulated usage and continuation counters. Adds `GoalState::from_persisted` / `new_shared_goal_state_from_persisted`, which rehydrate without resetting counters (unlike `sync_from_host_status`).

Host-managed runtime engines never self-continue, so the runtime host owns the durable loop: each completed turn settles the goal (accrues spend, mirrors the model's terminal complete/blocked/paused decision) and arms the next pass after the configured quiet period while still Active. Goal upsert also injects the objective into a cached engine and dispatches the kickoff turn when idle; delete/complete/block sync the terminal status back into the engine.

### Review follow-up (8ec0834a5)

- **Counter authority**: the engine snapshot is now the authority for the durable `continuation_count` (it counts every intra-turn pass via `record_continuation`), replacing the flat per-turn `+1` that diverged from the engine's own `ContinuationLimit` gate.
- **Engine cap is terminal at the host**: a snapshot at or beyond `goal_max_continuations` mirrors the engine's stop into the record as `Paused` and does not re-arm — no more passes that can only trip the same gate.
- **No `update_goal`, no loop**: a finished turn whose tool catalog lacked `update_goal` (`allowed_tools` restrictions, `isolated_chat`) no longer re-arms; the model has no terminal path on such threads, matching the engine's own precondition in `goal_continuation_message_if_needed`.
- **Kickoff labels**: index-0 items now persist `Goal kickoff (host-driven)` instead of `continuation pass #0`.
- **Comments**: reworded the two "only continuation dispatcher" comments (`turn_loop.rs`, `goal_loop.rs`) to name both dispatchers and their split (turn loop = intra-turn passes, host settle = cross-turn re-arm); documented `spawn_goal_continuation`'s cancellation semantics and the mail-wake/goal-pass race at the settle site.
- **Tests**: two `runtime_threads` tests pin the loop (below).

### Contract change (behavioral, intentional)

`PUT /v1/threads/{id}/goal` now **dispatches a kickoff turn** on an idle thread (and therefore spends provider tokens) instead of only persisting the record. That is the point of the endpoint for GUI clients, but existing runtime-API callers should know a PUT on an Active goal starts work. Status-sync endpoints (`DELETE`/complete/block) remain record-only plus an engine-side status sync.

### Provenance

Both the kickoff and continuation passes are dispatched as `UserInputProvenance::Runtime`, so they cannot authorize work (`can_authorize_work` is `ExternalUser`-only): approvals, permission elevations, and any `/goal`-style directive parsing inside the objective text are not granted by the host loop. The engine still executes tools under the thread's normal permission posture (approvals surface through the standard runtime approval flow); only *user-attributed* authority is withheld. This is deliberate — a host-driven pass is runtime text, not the user's request.

### One-scheduler contract

The host loop cannot race the engine's in-process scheduler: hosted engines never self-dispatch goal continuations (`host_managed_engine_does_not_self_dispatch_goal_continuation`, `engine/tests.rs:3721`), so there is exactly one `ContinueGoal` scheduler per session and it is inert here. The host supplies the durable turn claim that test says the host owes, through the same `start_turn_with_source` machinery Agent Mail wakes use, so the single-turn-loop guard (`single_turn_loop.rs`) is untouched.

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features --locked` (CI allow-list form from CONTRIBUTING.md)
- [x] `cargo test -p codewhale-tui --lib --locked runtime_threads::`

Real receipts (N > 0):

```
test result: ok. 164 passed; 0 failed; 2 ignored; 0 measured; 11506 filtered out; finished in 13.34s
```

(the two new goal-loop tests are inside that module run; filtered they report `test result: ok. 2 passed; 0 failed; ... 11360 filtered out; finished in 1.79s`)

Note: local clippy on this toolchain (rust 1.96) also flags 3 pre-existing `nonminimal_bool` findings in `crates/tui/src/tools/subagent/mod.rs:4931` that came in from `main` via the maintainer merge — they reproduce on `1aa69f456` without this branch's commits and are left for the main-side lane. The full workspace suite runs in CI's Test lane on the PR head.

## Checklist

- [x] This PR adds a new layer/module/abstraction — it names or deletes the layer it replaces (the two dispatcher comments now name the host settle alongside the turn-loop hook)
- [ ] Verified TUI behavior manually if UI changes (no UI changes; runtime-API behavior covered by tests)
- [ ] Harvested/co-authored credit uses a GitHub numeric noreply address (no harvested credit in this PR)


Hmbown (changes requested):
Thanks @gaord — this is a careful read of a genuinely under-served seam. Today `runtime_threads.rs` passes `goal_objective: None` on every `Op::SendMessage` (main `runtime_threads.rs:7018`) and builds every cached engine with `new_shared_goal_state()` (`:7658`), so a `PUT /threads/{id}/goal` record is durable but never actually reaches the engine's prompt surface, and `handle_send_message` (main `engine.rs:4916-4922`) would clear any injected goal on the next ordinary message exactly as your comment says. Rehydrating from the record (`GoalState::from_persisted`) instead of `sync_from_host_status` is the right call — the counters are the durable history, not a re-declaration. I verified that every symbol you lean on (`Op::SetGoalStatus`, `UserInputProvenance::Runtime`, `thread_goal_status_projection`, `GoalSnapshot::from_thread_goal`, `render_continuation_prompt`, `Event::GoalUpdated`) exists on main, and CI is green on Lint + Test across all three OSes, so this is a review of behaviour, not of compile state.

I'm requesting changes for a handful of concrete things. None of them are structural; the shape of the PR is sound.

### 1. The single-turn-loop contract holds, but two comments now lie — please fix them in this PR

`AGENTS.md` ("Current contracts") says there is exactly one turn loop, `Engine::run_turn` in `crates/tui/src/core/engine/turn_loop.rs`, guarded by `crates/core/tests/single_turn_loop.rs`. Your host loop does **not** add a `run_turn`, and it dispatches through the same durable `start_turn_with_source` claim that Agent Mail wakes use, so the guard passes (Test is green) and the contract is intact. It is also consistent with what the engine already promises: `reconcile_non_completed_goal_turn` says "Hosted engines leave lifecycle decisions to their durable host", and `host_managed_engine_does_not_self_dispatch_goal_continuation` (`engine/tests.rs:3721`) pins that a hosted engine "must wait for an explicit durable turn claim". You are supplying the claim that test says the host owes. Good.

But two doc-comments on main now state the opposite of what this PR makes true, and the PR checklist item "names or deletes the layer it replaces" exists for exactly this:

- `turn_loop.rs:4823-4825`: "This within-turn hook is the only goal-continuation dispatch site for every session".
- `goal_loop.rs:~220`: "for host-managed turns, where the turn loop is the only continuation dispatcher".

Please reword both to name the two dispatchers and their split: the turn loop owns *intra-turn* passes (bounded by the step budget), and `RuntimeThreadManager::settle_thread_goal_after_turn` owns the *cross-turn* re-arm for hosted engines. A reader who trusts those comments today would delete your loop as dead.

### 2. Continuation counting diverges between engine and store, and it can spend real provider calls

The engine increments `continuation_count` once per **intra-turn** pass (`turn_loop.rs:4870-4874`, `state.record_continuation()`), and `goal_continuation_allowed` stops with `ContinuationLimit` against that in-memory count. Your `settle_thread_goal_after_turn` increments the **durable** `continuation_count` once per **turn**, and `run_goal_continuation` gates on the durable count. With `[goal] max_continuations = 20` and, say, 15 intra-turn passes in the first turn:

1. Turn 1 ends `Completed` with the goal still `Active`; engine count = 15, durable count = 1.
2. Host arms pass #2. The cached engine still has count 15, runs one model step, and (once it reaches 20) `goal_continuation_allowed` returns `Stop` and emits only a status line — the goal stays `Active` in `SharedGoalState`.
3. The turn completes `Completed` + `Active`, so the host increments durable to 2 and arms again. Repeat until the durable count reaches 20.

That is up to ~18 real provider requests (default `continuation_delay_seconds` is 0, so back-to-back) that can do nothing. Two asks:

- In `settle_thread_goal_after_turn`, make the engine snapshot the authority for the counter: when `latest_goal_snapshot` is present, write `snapshot.continuation_count` into the record rather than `+1`. The engine already counted the passes this turn actually ran.
- Treat the engine's `ContinuationLimit` stop as terminal at the host: if the snapshot shows the cap reached (or better, if `GoalUpdated` carried a `pause_reason` of `Backoff`), mirror it to `Paused`/`UsageLimited` as appropriate and do **not** re-arm. Today the intra-turn stop only emits `Event::status`, so you may need to look at `snapshot.continuation_count >= max_continuations` directly.

### 3. A thread without `update_goal` in its registry loops forever

`goal_continuation_message_if_needed` returns `None` when the registry lacks `update_goal` (`turn_loop.rs:4818-4820`), so on such a thread (an `allowed_tools` restriction, or `isolated_chat`, which sets `allowed_tools: Some(Vec::new())`) every pass completes `Completed` with the goal still `Active`. With the default `max_continuations = 0` (unlimited) and delay 0, your host loop then re-arms indefinitely, one provider call per pass, and the model has no tool with which to ever say "complete" or "blocked". On main this thread simply never runs a goal; after this PR it spends until someone notices. Please gate the re-arm on the same precondition the engine uses — skip arming (and log) when the turn's tool catalog did not include `update_goal` — or, at minimum, stop re-arming after N consecutive passes that produced no `GoalUpdated` event.

### 4. Kickoff prompt and provenance

`activate_thread_goal` dispatches the raw objective string as the kickoff message with `UserInputProvenance::Runtime`. Two things to confirm or adjust:

- `Runtime` provenance cannot authorize work (`ops.rs:101`, `can_authorize_work` is `ExternalUser` only). That is the safe choice for continuation passes, but for the kickoff it means the objective is delivered as runtime text rather than as the user's request. Please state in the PR body what that gates in practice (explicit `/goal` directives inside the objective, approvals, anything else you found) so a reviewer can agree it is intentional.
- The persisted item summary for the kickoff reads `Goal continuation pass #0 (host-driven)`. A kickoff is not a continuation; please special-case index 0 to something like `Goal kickoff (host-driven)` so transcripts and the events feed read truthfully.

### 5. `sync_engine_goal_status` on delete

`delete_thread_goal` calls `sync_engine_goal_status(id, Active, clear = true)`, which in the engine becomes `sync_from_host_status(None, None, Active)` and publishes an empty snapshot — correct. But `run_goal_continuation` may already be sleeping through a quiet period for that thread; it re-reads the store on wake and bails on `None`, so this is fine as written. Worth one sentence in the doc-comment on `spawn_goal_continuation` saying that the sleep is not cancelled and the re-read is the cancellation.

### 6. One cheap test, please

`AGENTS.md` asks us not to add tests by default, and to add one when it cheaply protects a high-risk behaviour. An autonomous loop that spends provider tokens across restarts is that. One `runtime_threads` test that (a) upserts a goal on an idle thread and asserts exactly one kickoff turn is claimed, (b) completes it with a `Completed` status and a still-`Active` snapshot and asserts exactly one continuation is armed, and (c) completes a pass with `Failed` and asserts nothing is armed — would pin the contract from item 2/3 and the failure path you already handle. Please quote the real `test result: N passed` line in the PR body (with `N > 0`) rather than ticking the box; the body currently has clippy and `cargo test` unchecked even though CI ran them green.

### 7. The red `link` check

The only failing check is `PR closes an issue`. It reads the PR **body** (not the title) and passes on either of:

```
Closes #1234           (or Fixes / Resolves — any GitHub closing keyword)
No-Issue: <one-line why>
```

`Refs #N` on its own does not satisfy it. I could not find an open issue that describes "runtime goals are stored but never reach the engine", so either open one and add `Closes #<n>`, or add a line such as `No-Issue: runtime goal records were never injected into hosted engines; this wires the persisted record and the host-side re-arm.` Editing the body re-triggers the check.

### Smaller notes (no action required)

- Settle runs after the active-turn cleanup and before `spawn_agent_mail_safe_boundary_delivery`, so a mail wake and a goal pass can race for the same claim. Since settle runs after every terminal turn, a goal pass that loses the race simply re-arms after the mail turn; I read that as self-healing rather than a bug, but a one-line comment near the two spawns would save the next reader the trace.
- `from_persisted` sets `started_at: Instant::now()`, so any wall-clock derived elapsed time restarts at rehydration while `time_used_seconds` carries the history. That looks intended; just noting it.

Happy to re-review quickly once items 1, 2, 3 and 7 land — the rest can ride along. Thank you again for taking this on; the rehydration half is exactly what the runtime API needed.

Hmbown:
Thank you, @gaord. This is a substantial, well-explained slice and it follows on nicely from #5688. First-pass notes from a read of the diff against `main` (not run locally on my side):

**What we confirmed**
- CI is fully green on the current head: Lint, Test on macOS/Ubuntu/Windows, Safety gate, Mobile runtime smoke, Version drift, Signed-off-by.
- Every engine surface you lean on exists on `main` today: `Op::SetGoalStatus` (`core/ops.rs:180`), `UserInputProvenance::Runtime` (`core/ops.rs:73`), `Event::GoalUpdated` (`core/engine.rs`), and `GoalSnapshot::from_thread_goal` / `thread_goal_status_projection` / `render_continuation_prompt` (`tools/goal.rs:584-631`). The cap and quiet period reuse `goal_max_continuations()` / `goal_continuation_delay_seconds()` from config instead of adding a second knob. Good.
- `start_turn_with_source` now passes the persisted goal on every `Op::SendMessage`, which fixes a real gap (passing `None` cleared an injected goal on any ordinary message).

**What we need before it can move**
1. Tests. The checklist's "Added or updated tests" is unchecked and the diff adds none. The pieces that most need coverage: `GoalState::from_persisted` keeps `tokens_used` / `continuation_count` (where `sync_from_host_status` resets them); `settle_thread_goal_after_turn` mirrors a `complete` / `blocked` snapshot into the store and does not arm another pass; a `Completed` turn with an Active goal arms exactly one continuation and respects the cap; a failed or interrupted turn leaves the goal Active without arming. `runtime_threads.rs` already has a large in-file test module to follow.
2. Please run and tick `cargo clippy --workspace --all-targets --all-features --locked` and `cargo test --workspace --all-features --locked` locally. CI's Lint is green, so this is mostly the receipt.
3. Call out the contract change in the PR body: `PUT /v1/threads/{id}/goal` now dispatches a kickoff turn on an idle thread (and spends tokens) instead of only persisting the record. That is arguably the point, but it is a behavior change for existing runtime-API callers and should be stated up front, ideally with a test.
4. The one-scheduler contract. `core/engine.rs:1025` says the goal loop has a single `ContinueGoal` scheduler and no second one. Your host-managed loop goes through the standard `start_turn_with_source` durable-turn path rather than driving the engine directly, which looks like the right shape, but it will get a careful read at review. A sentence in the PR body on why the host loop cannot race the engine's in-process scheduler (for example, because runtime engines never self-continue) will speed that up.

Hmbown:
Closing: superseded by #5816 (re-land addressing review findings), which is now merged. The branch also conflicts with current main.

PR #5712, opened 2026-08-29
feat(cli): cloud-dispatch remote runner — sandbox to forge PR

Hmbown:
## What

The leftover that makes `/dispatch` real: a **confirmed dispatch actually
runs a cloud agent in a sandbox and opens the forge PR** (follow-up to #5701
flagged in the ops ledger: runner behind `confirm_job`, real sandbox
receipts in the status card, teardown-on-cancel).

New lifecycle, driven by `dispatch_runner` (crates/tui) and persisted on the
existing `/jobs` cloud surface:

```
proposed → launching → running → openingpr → done
                          │            │
                          └── failed ───┘   (+ canceled from any active state)
```

- `launching → running`: create the sandbox (labeled with job id + forge),
  wait until it accepts work.
- `running`: clone the target forge repo in the sandbox, then run **one**
  `codewhale exec --auto "<prompt>"` turn — the same single-`Engine::run_turn`
  path every local non-interactive caller uses. No second engine; the runner
  itself runs no model turn.
- `openingpr`: collect `git format-patch` against the clone's default branch,
  apply it locally on a fresh shallow clone, push with a **plain** push
  (`--force` is never passed; a moved branch fails closed instead of
  rewriting history).
- `done`: open the PR — `github` via the repo's existing `gh` seam
  (`gh pr create`, body via `--body-file` so the prompt never hits argv/ps);
  `gitee` via Gitee API v5 `POST /repos/{slug}/pulls` (service slot `gitee`);
  `cnb` via CNB OpenAPI `POST /{slug}/-/pulls` (service slot `cnb`).
- teardown on **done / failed / canceled** — `--cancel` / `/dispatch cancel`
  / `/jobs cancel cloud_*` flip the record and tear a live sandbox down; a
  runner in flight stops at its next checkpoint and tears down too.

Surfaces: the status card and job card now show real receipts (sandbox id,
PR URL when opened, head sha, whole-minute runtime labeled as Codewhale
bookkeeping, not a bill), with zero provider branding anywhere ("Codewhale
cloud agent" language only). The CLI stays attached to the runner so a
sandbox is never orphaned by an early exit; the TUI detaches and streams
progress through the job record. Sandboxes remain Codewhale-internal
infrastructure: membership-gated, fail-closed, no provider setup steps for
users (docs rewritten to say exactly that).

## How

- `crates/tui/src/dispatch_runner.rs` (new): `run_confirmed_job` state
  machine (only `launching`/`running` jobs can run — a `proposed` job is
[...]

Issue #258, opened 2026-05-01
Release binary missing companion deepseek-tui on macOS arm64


whereiszebra:
Description                                                                   
                                                                                
  When running the downloaded release binary deepseek-macos-arm64 on macOS    
  (Apple Silicon), the CLI immediately errors out because it cannot find a      
  required companion binary deepseek-tui.                                       
                                                                                
  Steps to reproduce                                                          
                                                                              
  1. Download the release binary deepseek-macos-arm64 from the GitHub Releases  
  page
  2. Make it executable: chmod +x deepseek-macos-arm64                          
  3. Run: ./deepseek-macos-arm64                                              
                                                                                
  Expected behavior
                                                                                
  The CLI launches successfully and enters the TUI interface.                   
                                                                              
  Actual behavior                                                               
                                                                              
  The following error is printed and the process exits immediately:             

  error: deepseek-tui binary not found at /path/to/deepseek-tui.                
  Build workspace default members to install it, or set DEEPSEEK_TUI_BIN to its
  absolute path.                                                                
                                                                              
  Environment                                                                   
                                                                              
  - OS: macOS 15 (Darwin 25.3.0, Apple Silicon arm64)                           
  - DeepSeek CLI version: deepseek-macos-arm64 (downloaded from Releases)
  - Model: N/A (cannot launch)                                                  
  - Shell: zsh                                                                
                                                                                
  Logs or screenshots                                                         

  error: deepseek-tui binary not found at                                       
  /Users/mac/project/githubclone/dstui/deepseek-tui.
  Build workspace default members to install it, or set DEEPSEEK_TUI_BIN to its 
  absolute path.

whereiszebra:
Release page does not document that both deepseek-macos-arm64 and    
  ▎ deepseek-tui-macos-arm64 must be downloaded together

Issue #251, opened 2026-05-01
v0.7.9: surface live sub-agent progress to the parent agent


Hmbown:
## Summary

The parent agent does not get enough live visibility into what spawned sub-agents are actually doing. In practice, `agent_result` / `agent_wait` often reduce an in-flight child to a bare `Running` status, so the parent falls into repeated "let me wait longer" loops even when the runtime already knows much more.

This should be part of `v0.7.9` because it makes sub-agent orchestration feel opaque and slows down the main agent's ability to make good decisions.

## Screenshot-backed symptoms

Observed in a live DeepSeek-TUI session on 2026-04-30:

- The main transcript shows repeated parent reasoning like "Both sub-agents are still running" and "Let me wait a bit longer."
- `agent_result` returns `status: Running` without a useful progress summary.
- `agent_wait` keeps returning control without telling the parent what meaningful work has happened since the last check.
- Meanwhile, the UI already knows more detail:
  - the bottom status rail shows progress like `Sub-agent agent_5ab4ea01: step 15/100: finished tool 'exec_shell'`
  - the right sidebar shows aggregate agent counts
- Result: the human can see signs of progress, but the parent agent still behaves as if all it knows is "Running".

## Why this is a separate issue

This is related to the broader sub-agent/swarm lifecycle issues, but it is not the same as:

- sidebar vs transcript count drift
- seeded pending slots
- background-job cancellation semantics

This issue is specifically about **parent-visible transparency** for spawned sub-agents: the main agent should receive useful structured progress, not just an opaque running bit.

## Current source pointers

- `crates/tui/src/tools/subagent/mod.rs`
  - `SubAgentResult` currently only carries `agent_id`, `agent_type`, `assignment`, `model`, `nickname`, `status`, `result`, `steps_taken`, `duration_ms`
  - `summarize_subagent_result()` collapses all running work to the single string `Running`
  - `agent_result` metadata for a running child is only `{ "status": "Running" }`
- `crates/tui/src/core/events.rs`
  - runtime already emits `AgentProgress`, `AgentList`, and mailbox events with richer progress timing
- `crates/tui/src/runtime_threads.rs`
  - UI status messages already surface more detailed sub-agent progress than the parent tool results do

## Acceptance criteria

- `agent_result` for a running child returns a richer progress snapshot than plain `Running`.
- The parent agent can see, at minimum, some structured combination of:
  - current or latest step count
  - last activity timestamp or elapsed time
  - current tool being run, if any
[...]

PR #5721, opened 2026-08-29
feat(cli): Codewhale-account machine tokens (CODEWHALE_API_KEY)

Hmbown:
The CI/account identity path, inherited from a dead agent's lane and finished by the integrator: with `CODEWHALE_API_KEY` set, the CLI authenticates as the account with no local session file and no browser.

The design follows the control-plane contract's asymmetry rather than softening it (the cwc side landed today as cwc#296 — `cwc_account_machine_keys`, whole-token SHA-256, deny-by-default routes; a machine key reaches exactly two read-only routes and can never mint/list/revoke, so a leaked key cannot bootstrap a successor that outlives its own revocation). The CLI states that rule locally so a CI failure reads as a credential problem, a configuration problem, or a CLI bug — never as an ambiguous 403.

Load-bearing rules, each held by a test:
- The token value never leaves `crates/cli/src/cloud/machine.rs`: no Display, Debug prints only the 32-char non-secret head (`cwc_key_` + the 24-hex id), printed exactly once at mint.
- A machine credential never silently downgrades to the interactive session (falling back would run CI as the wrong identity); managing keys requires interactive login, with the refusal explaining WHY (minting a replacement).
- Bounded retries with hostile-Retry-After clamping; terminal escapes scrubbed from server prose; unknown bodies classify by status.

Surfaces: `codewhale account api-keys` mint/list/revoke/whoami over the account routes (management via the signed interactive session; whoami/agent via the machine bearer), exit codes separated (auth vs input vs unconfigured-agent vs unavailable), `CODEWHALE_API_BASE` override.

This is the CLI half the dispatch slice (#5712) injects into sandboxes and the identity the strategic spine describes — one token, every host.

Verification on the merged tree (current main + this branch): CLI suite **321/321 passed** (was 277 before the module; includes the 620-line machine test module + 386-line cloud integration tests); fmt clean; CI-exact clippy exit 0; feature release-note receipts OK.

No-Issue: n/a (CI infrastructure; the account-key design is documented in docs/GITHUB_APP.md)


Hmbown:
Rebuilt to strip the assistant co-author trailer from the feat commit — exactly what the credit check instructs ('contributor trailers are for humans'). Author and content byte-identical; only the bot trailer line removed. Tree diff vs the previously tested head is empty.

Hmbown:
Triaged the CodeQL failure (release triage, 2026-08-31).

**All 8 `rust/cleartext-logging` alerts (#251–#258) — dismissed as `used in tests`.**

Every one is in `crates/cli/src/cloud/tests.rs`, which is reached only through `crates/cli/src/cloud.rs:1195`:

```rust
#[cfg(test)]
mod tests;
```

so the file never compiles into a release binary. The flagged "log file" writes are `assert!` failure messages (`"{output}"`).

The tainted value is `MACHINE_TOKEN`, a hardcoded fixture literal at `tests.rs:1049`, not a credential.

Worth calling out: alert #253 targets line 1120, which is the assertion that the secret half is *absent*:

```rust
assert!(
    !output.contains(&MACHINE_TOKEN[32..]),
    "secret half leaked: {output}"
);
```

CodeQL flagged the leak guard as the leak. Dismissing rather than rerouting `run_account(...)` away from the assertion sink — the assertion messages are what make a failure diagnosable, and they only ever run in a test binary.

This clears the last non-required red check on this PR; the six required checks were already green.

PR #5717, opened 2026-08-29
refactor(tui): adopt command shapes in project group (FEAT-021)

aboimpinto:
## Summary

FEAT-021 converts the TUI project command group (`/init`, `/lsp`, `/share`, `/goal` — no aliases) to the external command shapes introduced by FEAT-014 and hosted by FEAT-015, following the pattern FEAT-018 (utility) established. The command files remain under `codewhale-tui`; this PR changes their execution boundary without physically moving them.

**Baseline:** rebased onto `main` (`20e9574`) per maintainer request — the `codex/v0912-integration-20260823` base will not land (232 ahead / 324 behind main). The 7-commit series now carries 8 commits directly on `main`.

This PR:

- adds the contract-owned `CommandProjectContext` facet with **exact-minimum typed delegates**: `/lsp` status/set, `/share` session projection (emptiness, length, model, mode label), `/goal` goal projection (objective, status, pending controls, time, tokens, session-derived effective values) — portable DTOs (`ProjectGoalState`, `ProjectShareProjection`, `ProjectGoalStatus`);
- converts the complete `/project` dispatch and render paths to portable contract values — byte-identical output preserved;
- registers the group via `ContextualCommand::from_contract` on **main's model** (no capability bitmask on main yet — the FEAT-019/020 capability machinery lives only on the integration branch): each command is a `CommandHandler::Contextual(fn)` whose handler destructures exactly its required facets with safe missing-facet errors — `/init` → `WORKSPACE` only, `/lsp` → project, `/share` → project, `/goal` → project + presentation;
- routes the cross-group `/lsp` bridge (`config::config::lsp_command`) exclusively through the TUI adapter (FEAT-020 D1 pattern); the portable handler never imports the config group;
- resolves all catalog messages through the presentation facet for `/goal` (`goal_control_accepted`, `goal_status_idle_hint`) and the metadata bridge for all four description keys;
- removes `project` from **both** migration-frontier representations atomically (frontier now `config, core, debug, memory, plugins, session, skills` — memory/plugins remain pending on main);
- keeps the temporary TUI-owned `CommandResult` and the exact `AppAction` variants the group emits (`SendMessage`, `ShareSession`, `SetGoalStatus`, `SetGoalObjective`) bounded for FEAT-037.

Tracking: EPIC-006 / FEAT-021 in umbrella [Hmbown/CodeWhale#5316](https://github.com/Hmbown/CodeWhale/issues/5316).

## Dependency boundary

The intended dependency direction remains acyclic:

```text
[...]

PR #5724, opened 2026-08-29
fix(sandbox): match the read deny-list against a rule's resolved path, greening shared macOS/Windows CI

Hmbown:
## Summary

Restores a green shared base: hosted `Test (macos-latest)` and `Test (windows-latest)` are red on `main` at `20e957406` and every open PR riding it (#5712, #5719, #5720, #5721, #5703, #5722) inherits the same failures.

**macOS (6 failures, `sandbox::read_guard::tests`)** — hosted runners give `tempfile` a `/var/folders/...` directory that is really `/private/var/folders/...`. The deny-list compared the *canonicalized* candidate (and the already-canonical cwd) against a *literal* rule, so the six symlink / denied-tree tests never matched. Locally this only reproduces when `TMPDIR` is the macOS default (the prior verification ran with `TMPDIR` on a plain volume).

This is a product bug, not a test bug: on macOS `/etc`, `/var`, and `/tmp` are symlinks into `/private`, so `read_file /private/etc/sudoers` walked around the built-in `/etc/sudoers` rule (the OS-level Seatbelt setter already canonicalized its own copy; the in-process matcher did not). Fix: a subtree rule now remembers its resolved spelling (`DenyRule::subtree`) and `check` matches a candidate against either spelling.

**Windows (1 failure)** — `root_parent_traversal_does_not_escape_above_root` asserted a Unix `/etc` while `normalize_lexically("/../../etc")` correctly resolves against the cwd's drive (`D:\etc` on the hosted runner). The test now spells the traversal from the current drive root and keeps the Unix literal under `cfg!(unix)`.

Defeat tests added (fail on the unpatched tree regardless of `TMPDIR`):
- `rule_spelled_through_a_symlinked_root_matches_the_resolved_spelling` (unix): rule written through a symlinked directory; direct read, symlink read, and relative-from-cwd read are all refused; the real directory's innocent sibling stays readable.
- `macos_private_spelling_of_a_machine_wide_rule_is_refused` (macOS): `/private/etc/sudoers` and `/private/etc/ssh/…` are refused by the `/etc/*` defaults; `/private/etc/hosts` stays readable.

Not in this PR (branch-specific, not shared-main): the Windows `runtime_api::tests::events_endpoint_respects_since_seq_cursor` timeout appeared only on #5712's Windows run (8.91 s against an 8 s `ci_scaled` budget; #5712 does not touch `runtime_api`), and #5722's Ubuntu `fleet::executor` failure is its own.

[...]

PR #5729, opened 2026-08-30
fix(sandbox): canonicalize deny-rule paths at build; unbreak macOS/Windows read_guard baseline

aboimpinto:
## Summary

Fixes the `sandbox::read_guard` baseline failures that are currently red on `main`'s own CI (macOS + Windows) and block every PR targeting `main` — including [FEAT-021 #5717](https://github.com/Hmbown/CodeWhale/pull/5717), whose only red checks are these exact tests.

Two independent defects in `crates/tui/src/sandbox/read_guard.rs`:

**1. macOS — deny rules were lexical while candidates were canonical (real deny-bypass gap).**

- `ReadDenylist::build` stored rule and exempt paths with `normalize_lexically` only.
- `check()` compares `canonicalize_best_effort`d candidates against those rules via `path_is_within` (component-wise).
- On macOS the system redirects `/var` (and `/tmp`) to `/private/...`. A temp-dir deny rule stayed `/var/folders/...` while the canonical candidate was `/private/var/folders/...` → first component mismatch → the deny was silently missed. This is why the symlink/`..` evasion tests (`dot_dot_through_a_symlink…`, `symlink_chains…`, `relative_read_from_inside_a_denied_tree…`, `symlinked_parent_directory…`, `symlink_pointing_into_a_denied_tree…`, `denial_message_names_the_rule…`) fail on macOS but pass on Linux.
- Consequence in production: any user-listed deny under a symlinked root on macOS could be bypassed.

**Fix:** canonicalize every rule and exempt path at build time (`canonicalize_best_effort(&normalize_lexically(&path))`), so both sides of `path_is_within` compare canonical paths. `subtree_paths()` now hands the OS wrappers the canonical form, which is also the correct form for Seatbelt on macOS. Non-existent built-ins (`~/.ssh`, …) resolve to their canonical parent + suffix as before. `canonicalize_best_effort` never fails to produce at least the root-canonicalized path, so no rule can be lost.

**2. Windows — `root_parent_traversal_does_not_escape_above_root` asserted Unix path semantics everywhere.**

- The test asserted `normalize_lexically(Path::new("/../../etc")) == PathBuf::from("/etc")`. On Windows a leading `/` is **not absolute** (drive-relative), so the path joins `current_dir()` and normalizes to the drive root — the assertion failed.
- On macOS it also failed because `subtree_paths_feed_the_os_wrappers_and_omit_the_filename_rule` compared against the lexical form (now canonical).

**Fix:** per-platform assertions — Unix keeps `/../../etc` → `/etc`; Windows uses a drive-rooted path `C:\..\..\etc` → `C:\etc` (the root never pops above itself). The subtree_paths test now asserts the canonical form.

## Testing

- [x] `cargo test -p codewhale-tui --lib --locked sandbox::read_guard` — 20 passed / 0 failed (Linux)
- [x] `cargo test -p codewhale-tui --lib --locked sandbox::` — 74 passed / 0 failed (Linux)
- [x] `cargo check --workspace --locked` — clean
- macOS/Windows behavior verified by reasoning (canonicalize resolves the `/var → /private/var` redirect; the per-platform test expectations follow the respective path models); the previous CI matrix on `main` reproduced the failures, so a green macOS/Windows run on this PR is the acceptance signal.

Note: the only remaining clippy error in this workspace (`too_many_arguments` in `runtime_threads.rs:2562`) is pre-existing on `main` with this toolchain and unrelated to this change.

Paulo Aboim Pinto

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5729" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tui/src/sandbox/read_guard.rs
+++ b/crates/tui/src/sandbox/read_guard.rs
@@ -161,11 +161,19 @@ impl ReadDenylist {
     ///   from the built-in defaults only.
     #[must_use]
     pub fn build(include_defaults: bool, extra: &[PathBuf], exempt: &[PathBuf]) -> Self {
+        // Every rule/exempt path is canonicalized (after lexical
+        // normalization) so `check()`'s canonicalized candidates compare
+        // against a canonicalized rule. On macOS this matters because the
+        // system redirects `/var` (and `/tmp`) to their real `/private/...`
+        // locations: a lexical-only rule would silently miss the canonical
+        // candidate and the deny could be walked around (the read_guard
+        // symlink/`..` baseline failures on the macOS CI runner).
         let exempt_normalized: Vec<PathBuf> = exempt
             .iter()
             .cloned()
             .map(expand_home_prefix)
             .map(|p| normalize_lexically(&p))
+            .map(|p| canonicalize_best_effort(&p))
             .collect();
 
         let mut subtrees = Vec::new();
@@ -182,7 +190,7 @@ impl ReadDenylist {
                     .is_some_and(|name| name == std::ffi::OsStr::new(".env"))
             });
             for (raw, label) in default_denied_subtrees() {
-                let path = normalize_lexically(&raw);
+                let path = canonicalize_best_effort(&normalize_lexically(&raw));
                 if exempt_normalized.iter().any(|e| path_is_within(&path, e)) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


aboimpinto:
Closing — superseded by the maintainer's own fix, merged in #5724 ("match the read deny-list against a rule's resolved path"). This PR was prepared in parallel and takes the same approach (canonicalize/resolve deny rules at build + per-platform root-parent assertion); the merged fix covers the same ground. Closing to avoid overlap.

PR #5749, opened 2026-08-30
feat(app-server): unix-socket transport + daemon/attach advertisement

Hmbown:
## What
App-server unix-socket transport and daemon/attach advertisement (desktop Phase 0 foundation).

## Evidence
- Adversarial review (2026-08-30): approve; daemon spawn → socket connect → round-trip → shutdown verified; socket permissions/traversal/stale-cleanup lens clean.

## Unproven
- Workspace CI on this PR.

No-Issue: desktop Phase 0 socket listener

---

> [!NOTE]
> **Medium Risk**
> Introduces a new local multi-client control plane with ownership and shutdown rules; mistakes could affect desktop daemon lifecycle or leave stale sockets, but scope is loopback-only with peer-uid checks and broad integration tests.
> 
> **Overview**
> Adds **Desktop Phase 0** local daemon support: `codewhale app-server --socket` (optional `--socket-path`) listens on a user-private Unix domain socket and speaks the same newline-delimited JSON-RPC as `--stdio`, after a mandatory **`daemon/attach`** handshake.
> 
> **Transport behavior:** Socket path resolution follows explicit override → `CODEWHALE_HOME/run` → XDG runtime → macOS Application Support → `~/.codewhale/run`, with `0700` dirs, `0600` socket, stale-socket probing, and refusal to clobber live listeners or non-socket files. Connections are accepted only from peers matching the daemon uid. Clients attach as **owner** (`claim`) or **guest** (`attach`); only the owner may `shutdown`, and the owner claim is released when the connection closes (including mid-turn). New JSON-RPC errors cover attach-required, already claimed, version skew, and not-owner shutdown. `capabilities` and `healthz` report transport `unix-socket` and advertise `daemon/attach` on this path only. Windows fails typed with a reserved named pipe—no TCP fallback.
> 
> **CLI & docs:** `app-server` gains `--socket` / `--socket-path` with conflicts against other transports; `docs/RUNTIME_API.md` documents the handshake and ownership model. Acceptance tests in `crates/app-server/tests/daemon_socket.rs` cover attach, guests, version skew, and stale cleanup.
> 
> CHANGELOG entries in this diff also note compaction survival and a TUI session-id duplicate fix; those are not part of the app-server socket code changes in this PR.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fc4a2ec4fb0f7bbf52aed2cc7b61e65f3a447c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hmbown:
Holding this in the release queue. The unresolved **P1 on `crates/app-server/src/daemon_socket.rs:592` is still unfixed at head** — the only commits since the reviewed `67aee8935f` are `9c7adb16bf` (merge from main) and `844a225f13` (changelog receipt).

The shape is exactly as reported:

```rust
let exit = run_stdio_loop(&context.state, lines, writer, policy).await;
drop(guard);
```

`ConnectionGuard` is held across the entire loop and released only after it returns. `run_stdio_loop` sets `stdin_open = false` on EOF and keeps awaiting the in-flight dispatch, so an owner who disconnects mid-`thread/message` keeps the claim for as long as the turn runs. If that turn is stalled and emits nothing further, a relaunched desktop gets `daemon_already_claimed`, and falling back to guest attachment leaves it unable to shut the daemon down either — a user-visible lockout with no way out but killing the process.

Release the owner claim when EOF is observed, not when the active turn completes.

Two other notes so this is not re-litigated later:

- The red `Test (windows-latest)` is a **flake**, not a real failure: `13785 tests run: 13784 passed, 1 failed`, the failure being `runtime_threads::tests::terminal_turn_cancels_pending_dynamic_tool_exactly_once` with `deadline has elapsed`. This PR touches only `crates/app-server` and `crates/cli`, nothing under `runtime_threads`. I have re-run the job.
- This PR and #5751 both add `CHANGELOG.md` hunks and will conflict with each other. Whichever lands second needs a one-line changelog resolution — no code conflict.

This is a new local IPC surface with uid checks, so it is worth getting the ownership lifecycle right before it ships rather than after.

Hmbown:
P1 fixed in `bab91b1a7`.

`run_stdio_loop` now takes an opaque `input_claim: Option<C>` and drops it at the EOF arm rather than on return:

```rust
match line? {
    None => {
        stdin_open = false;
        // Release the claim here, not after `dispatch` resolves.
        drop(input_claim.take());
    }
```

The socket transport moves its guard in (`Some(guard)`); process stdio has no claim and passes `None::<()>`. Opaque rather than a concrete `ConnectionGuard` because that type is private to `daemon_socket` and `lib.rs` should not learn about it — only that *something* must be released when input closes.

```
cargo nextest run -p codewhale-app-server
Summary [0.213s] 104 tests run: 104 passed, 0 skipped
```

**No regression test, deliberately.** Reproducing this needs a stub runtime that holds a turn open while the socket closes, and the `daemon_socket` suite has no turn scaffolding — the four existing tests all attach and shut down without running one. That is more machinery than the three-line ownership move warrants, so the invariant is stated on the parameter instead of implied by a test. Flagging it rather than letting the green suite imply coverage that is not there.

Also worth recording: the earlier red `Test (windows-latest)` was a flake, not this PR — `13785 tests run: 13784 passed, 1 failed`, failing on `runtime_threads::tests::terminal_turn_cancels_pending_dynamic_tool_exactly_once` with `deadline has elapsed`, in code this PR does not touch.

PR #5751, opened 2026-08-30
feat(protocol): Op/EventMsg parity + compile-enforced guard

Hmbown:
## What
Protocol A1: Op/EventMsg parity between the Rust core and TS surfaces, with a compile-enforced guard so the parity cannot silently drift.

## Evidence
- Wave-2 journal wf_7d8a5023-bb6 coder/review/fix verdicts + independent re-verification (2026-08-30): parity guard test green; `cargo test -p codewhale-protocol -p codewhale-core` green.

## Unproven
- Workspace CI on this PR.

No-Issue: protocol extraction spec A1

---

> [!NOTE]
> **Medium Risk**
> Large, stable wire contract plus scheduling-id hardening and MCP secret redaction; runtime paths do not call the projections yet, but serde shape mistakes would affect future CLI/app-server consumers.
> 
> **Overview**
> **Phase A1** of the core/protocol extraction: `codewhale-protocol` now defines the full serializable **`Op`** and **`EventMsg`** surfaces (turn lifecycle, streaming, tools, compaction/purge, sub-agents, MCP, approvals, session sync, and related typed payloads), with **`OP_KINDS` / `EVENT_KINDS`** tables and serde round-trip tests. Older minimal JSON (e.g. `send_message` with only `content`) still deserializes with defaults.
> 
> The TUI adds **`protocol_parity`**, which exhaustively projects engine **`Op` / `Event`** into protocol types so a new engine variant without a wire twin **fails to compile**. Wire-boundary rules strip in-process handles (reply channels, hook executors, pause acks) and **drop `engine_schedule_id` on deserialized `continue_goal`** so external callers cannot hijack engine scheduling. **`McpSessionBoot`** redacts URL credentials and stdio argv secrets before they reach SSE/stream-JSON consumers.
> 
> The protocol crate **`deny`s stdout/stderr** from clippy. Changelogs also record compaction survival/history behavior and a fix for **phantom duplicate sessions** when TUI and engine disagreed on session id (not shown in the protocol files above).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0205f21f82acaec39f5620a3780a4217cdc0c8fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hmbown:
Holding this in the release queue. All six required checks are green, but the review thread has **two unresolved P1 findings and both are still unfixed at the head commit**. Flagging per the repo rule that a green rollup plus an unread review is a merge that ships known bugs.

The Codex review was against `57e2ae5a12`. The only commits on this branch since then are `5c01774cfc` (merge from main) and `743a0bff04` (changelog receipt) — no fix for either finding.

---

### P1 — MCP credentials reach a new outbound wire surface

`crates/tui/src/core/protocol_parity.rs:272` still projects the field raw:

```rust
command_or_url: server.command_or_url.clone(),
```

And the snapshot builder at `crates/tui/src/mcp.rs:4618` applies no redaction at all:

```rust
let command_or_url = server.url.clone().unwrap_or_else(|| {
    let mut command = server.command.clone().unwrap_or_else(|| "(missing)".to_string());
    if !server.args.is_empty() {
        command.push(' ');
        command.push_str(&server.args.join(" "));
    }
    command
});
```

So the string is verbatim either `https://user:token@host/mcp` / `https://host/mcp?api_key=…`, or a stdio command line with a token in `args`.

What makes this a new exposure rather than pre-existing: today `command_or_url` is only read by the local command palette (`command_palette.rs:333` and `:450`), i.e. shown back to the person who configured it. This PR clones it into `EventMsg::McpSessionBoot`, which every SSE / stream-JSON consumer receives and any frame-retaining log keeps. Same value, materially different blast radius.

Suggested shape: project a sanitized descriptor — scheme + host + server name, args elided — rather than the configured string. `crates/tui/src/route_receipt.rs` already has redaction precedent worth reusing.

### P1 — `engine_schedule_id` is engine-owned by comment only

`crates/protocol/src/op.rs:167-170`:

```rust
/// Engine-owned coalescing token; direct callers send `None`.
#[serde(default, skip_serializing_if = "Option::is_none")]
engine_schedule_id: Option<u64>,
```

The doc comment states the contract; nothing enforces it. `skip_serializing_if` only affects *output* — deserialization still accepts any `u64` from an out-of-process `ContinueGoal`. Since the IDs are predictable per-session counters, a guessed value is treated as an already-delayed internal token, consuming the pending schedule and bypassing the host-injected quiet period.

Either drop the field from the input schema, or force it to `None` on deserialize (a custom `deserialize_with`, or split the input and internal types) so the invariant is enforced rather than described.

---

Both are small relative to the 3,193-line diff, and neither is a design objection — the parity guard itself is good. Happy to merge as soon as they are addressed or explicitly dispositioned on the record.

For the record on the other red check: `Codewhale review` failed with `Model response incomplete: provider stop reason 'length'; the partial review was not accepted` after a 16-minute run — it is the output cap (`CODEWHALE_REVIEW_MAX_OUTPUT_TOKENS` is unset), not a finding and not a funding problem. That check is advisory and is not what is holding this.

Hmbown:
Both P1s fixed in `634118028`.

**1. `command_or_url` is redacted before the wire.**

```rust
fn redacted_command_or_url(raw: &str) -> String {
    if raw.contains("://") {
        return crate::client::redact_url_for_display(raw);
    }
    // No `match` here on purpose: `projections_have_no_wildcard_arms` scans
    // this whole file for wildcard arms, and a `_ =>` would trip it even in a
    // helper that projects no engine variant.
    let trimmed = raw.trim();
    if let Some((program, rest)) = trimmed.split_once(char::is_whitespace)
        && !rest.trim().is_empty()
    {
        return format!("{program} …");
    }
    trimmed.to_string()
}
```

URLs reuse `client::redact_url_for_display`, which already masks userinfo and sensitive query keys — no new redaction logic. Stdio keeps the program name and elides the args, because a secret there can be positional and is not reliably recognizable by key.

Worth recording: my first version used `match` with a `_ =>` arm and **`projections_have_no_wildcard_arms` caught it** — the guard this PR adds did its job on the first change made after it landed.

**2. `engine_schedule_id` is now enforced, not just documented.**

Deserialization always yields `None`. This cannot affect the engine's own coalescing: `Op::ContinueGoal` is sent over an in-process channel at `engine.rs:2193` (`tx_op.try_send`) and never crosses serde, so anything deserialization sees came from outside. The `every_variant` fixture moves to `None` because the field is deliberately no longer round-trippable, and a new test proves a wire-supplied `3` is dropped.

```
cargo nextest run -p codewhale-protocol -p codewhale-tui
Summary [0.129s] 15 tests run: 15 passed, 12038 skipped
  PASS codewhale-protocol op::tests::wire_supplied_engine_schedule_id_is_dropped
  PASS codewhale-protocol op::tests::every_variant_round_trips_and_tags_by_kind
  PASS codewhale-tui core::protocol_parity::tests::mcp_command_or_url_is_redacted_before_it_reaches_the_wire
  PASS codewhale-tui core::protocol_parity::tests::projections_have_no_wildcard_arms
```

The four remaining P2s on this thread are unaddressed and I have not silently dropped them — flagging explicitly so they are a decision rather than an oversight: turn-ID preservation in completion projections, identifier catch-all arms, MCP capability-flag preservation, and the two `to_value`/`count` saturation notes from Copilot. Say the word and I will take them in a follow-up; none of them is a credential or trust-boundary issue, which is why they are not blocking here.

PR #5813, opened 2026-09-02
feat(tui): diff cards emphasise the changed words within a line

Hmbown:
No-Issue: internal 0.9.12 shell wave slice.

Rendering wave R5 (founder's list, rung 5 of the ladder: both crates already installed).

**What changed for the user.** In a diff card, when a line was replaced by another line, the words that actually changed are now emphasised (bold + reversed) on the red/green line, so a review card reads at a glance instead of forcing a word-by-word compare. Only 1:1 replaced runs get this; unequal runs and wholesale rewrites (less than half the words shared) render exactly as before.

**Why.** `diff_render.rs` painted whole lines −rose/+green and never used `similar`'s inline ops even though the crate was already the diff engine; review cards were hard to scan.

**How.** `render_diff_body_bounded` buffers a `-` run and the `+` run that follows it; equal-length runs are paired and each pair goes through `TextDiff::from_unicode_words` (unicode word boundaries keep punctuation out of the emphasis). Emphasis flags are carried across the wrapper's breaks; line text is byte-identical to before. `similar` gains its `unicode` feature — `unicode-segmentation` was already in the tree, no new crate. syntect is not used here: colouring hunk lines by extension was not a small, obviously-safe addition to this renderer.

**Evidence.**
- `scripts/dev-test.sh tui diff_render` — `test result: ok. 9 passed; 0 failed; 0 ignored` (3 new tests: replaced-pair emphasis, unequal/rewrite fall-through, emphasis across wrapping).
- `scripts/dev-test.sh tui` — `11854 tests run: 11854 passed, 13 skipped`; no golden changed.
- `cargo clippy -p codewhale-tui --all-targets -- -D warnings` with CI's allow flags: clean.

Refs: 0.9.12 runlog, rendering wave R5.

---

> [!NOTE]
> **Low Risk**
> TUI-only diff presentation with bounded allocation and fallbacks to unchanged line rendering; no security or persistence impact.
> 
> **Overview**
> **Diff cards** now highlight only the words that changed when a deleted line is replaced by an added line of the same length, instead of painting the whole row red/green.
> 
> The renderer buffers equal-length `-`/`+` runs, pairs lines, and uses `similar`’s **unicode word diff** (`TextDiff::from_unicode_words`) to build emphasis segments. Changed tokens get **bold + reversed** styling; line text stays the same. Emphasis is skipped for unequal runs, runs longer than 64 lines, pairs with low shared-word ratio (treats them as rewrites), and lines that differ only in syntax. `\ No newline at end of file` markers are peeled off so pairing still works. Wrapping maps emphasis per character with a safe fallback to plain lines if mapping fails.
> 
> `similar` is enabled with the **`unicode`** feature in `codewhale-tui` (lockfile pulls in `unicode-segmentation`). New unit tests cover pairing, unicode words, wrapping, and newline markers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 812e2f5738089ef5f46c08eb9395824d26b9e8f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hmbown:
Review findings addressed in the follow-up commit:

1. **Emphasis on wrapped, indented lines** — chunks now map each non-whitespace character back to the source by position (the invariant `wrap_text` keeps: it only drops, collapses, or re-inserts whitespace). Whitespace joins an emphasised run only when both neighbours are in it. `debug_assert!` pins the order; any mismatch falls back to whole-line styling. New test `emphasis_lands_on_the_right_chunk_of_a_wrapped_indented_line` pins the change on the third wrapped chunk with the exact text before the emphasis (`"        gamma, "`).
2. **Threshold** — `INTRALINE_MIN_RATIO` is now computed over word tokens only (tokens bearing an alphanumeric char). Test `lines_sharing_only_syntax_are_not_emphasised` (`let alpha = beta;` vs `let gamma = delta;`) stays plain.
3. **Tests** — multiple pairs in one run; a rewrite next to an edit in one run; unicode word boundaries (`lait`→`thé`); the wrapping test now asserts emphasis per chunk plus the text before it.
4. **`\ No newline at end of file`** — lifted out of the run and painted after it; test `no_newline_marker_does_not_break_pairing`.
5. **Allocation claim** — both: the run is capped at `INTRALINE_MAX_RUN = 64` replaced pairs (beyond that, plain line-by-line), and the `render_diff_body_bounded` doc now states the bound as the retained preview + one wrapped source line + the current run's borrowed slices and word segments for at most 64 pairs.

Evidence: `scripts/dev-test.sh tui diff_render` → `test result: ok. 15 passed; 0 failed`. `scripts/dev-test.sh tui` → `11860 tests run: 11859 passed, 1 failed` — the one failure is `chatgpt_oauth::tests::callback_is_accepted_on_either_loopback_family`, untouched by this PR and passing alone (`1 passed; 0 failed`); loopback port contention under six concurrent builds. clippy `-D warnings` with CI's allow flags: clean.

PR #5868, opened 2026-09-04
feat: send x-opencode-session header for OpenCode Go/Zen providers

huangxianzhan:
### What

OpenCode Go (https://opencode.ai/docs/go/) asks clients to send a stable \x-opencode-session\ header so it can optimize prompt caching and attribute traffic to a conversation. Codewhale requests were missing it, and the client UA (Mozilla/5.0 compatible codewhale/...) gets classified as a plain browser on the gateway; traffic without the header may error starting 2026-09-06.

### Changes

- Inject one process-stable UUID v4 into \x-opencode-session\ in \uild_default_headers\ for \OpencodeGo\/OpencodeZen\ providers only.
- Other providers are untouched; a user-configured \x-opencode-session\ via the existing \http_headers\ config always wins (extra-header loop overwrites).
- Three unit tests: header present and stable per process; user-configured value wins; absent for non-OpenCode providers.

### Verification

- \cargo check -p codewhale-tui\ passes
- \cargo test -p codewhale-tui --lib session_header\: 3 passed, 0 failed
- End-to-end capture against a local mock gateway shows the header on every request; live opencode.ai models/chat/SSE behave normally.

### DIFF
--- a/crates/tui/src/client.rs
+++ b/crates/tui/src/client.rs
@@ -1668,6 +1668,25 @@ fn build_default_headers(
             HeaderValue::from_static("Codewhale"),
         );
     }
+    // OpenCode Go / OpenCode Zen gateways (https://opencode.ai/docs/go/)
+    // ask clients to send a stable `x-opencode-session` header so the
+    // service can optimize prompt caching and attribute traffic to a
+    // conversation. Generate one UUID v4 per process so every request a
+    // session makes to the gateway shares a single ID ("one stable ID per
+    // conversation"). A user-configured header of the same name wins: the
+    // extra-header loop below overwrites this value.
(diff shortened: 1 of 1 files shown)


Hmbown:
Closing with credit: this change landed on main as 2ad300fff4 (`feat(client): x-opencode-session header for OpenCode Go/Zen`, see CHANGELOG) — harvested from your PR with Co-authored-by attribution, including the process-stable UUID v4 design, user-header precedence, and your three tests. Merging the PR directly would double-apply. Thank you for the careful work — the LF/CRLF cleanup note was especially helpful. If you spot drift between your version and what landed, please say so and we'll reconcile.

PR #5870, opened 2026-09-04
Fix: Tools: atomic commit splitting — order unrelated changes by dependency

### LINKED ISSUE #3999: Tools: atomic commit splitting — order unrelated changes by dependency, reject cycles
## Problem

When the agent (or a subagent fleet) lands a batch of unrelated changes in one working tree, it tends to produce one large commit. That is worse for review, bisection, and revert: a reviewer cannot tell which hunk belongs to which logical change, `git bisect` lands on the mega-commit for unrelated regressions, and reverting one concern means reverting them all.

Today the agent has `git` tools (`crates/tui/src/tools/git.rs`: `git_status`, `git_diff`, `git_log`) but no structured way to say "split this working tree into atomic commits, ordered so each one builds on the last." It either commits everything at once or hand-rolls the split with multiple `git add -p` shell round-trips.

## Scope

Add a first-class `commit` (or `commit_split`) tool that:

1. Reads the working tree through the existing `git` tool surface (`git_status` porcelain, per-file diff, hunk-level granularity).
2. Groups hunks into logical commits by concern (heuristic: shared symbols/modules/import paths belong together; source files rank above tests/docs/configs; lock files are excluded from analysis and always go with the dependency change that moved them).
3. Orders the commits by dependency so each commit is self-consistent (a commit that defines a symbol comes before the commit that uses it).
4. **Rejects cycles before anything is written**: if the dependency graph has a cycle, the tool fails cleanly with a diagnostic and writes nothing.
5. Produces each commit through the normal approval/write gate — the model proposes the split, the user approves, commits land atomically.

## What we already have — extend, do not rebuild

- The `git` tool surface already exists and is the data source: `git_status_porcelain` (`github.rs:448`), `git_status` / `git_diff` (cached + path-scoped, `tools/git.rs:405,451`), `git_log` (`git_history.rs`). The commit-splitter consumes these, it does not re-implement git plumbing.
- Symbol/import dependency can lean on the LSP references surface (#3975) once it lands — "this hunk defines `foo`, which other hunks reference `foo`?" is exactly an LSP `references` query. For the first cut, a regex/import-based heuristic is enough; LSP-backed dependency ordering is a natural follow-on.
- Approval gating already exists: `pr_attempt_*` and approval-required `github_*` tools (`registry.rs:690-721`) are the model for "model proposes, user approves, action lands." Commit creation should go through the same gate (`permissions.toml` ask-rules + `execpolicy`).
- The existing `commit_all` test helper (`git.rs:341`, `git_history.rs:556`) shows the commit path; the splitter is a structured multi-commit driver over the same plumbing.

## Likely files

- new `crates/tui/src/tools/commit_split.rs` (grouping, dependency ordering, cycle detection, commit driver)
- `crates/tui/src/tools/git.rs` (expose hunk-level diff if not already present)
- `crates/tui/src/tools/registry.rs` (register the tool, approval-required)

## Acceptance criteria

- A working tree with N unrelated concerns is split into N commits, each containing only its own hunks.
- Commits are ordered so each is self-consistent (no commit references a symbol defined in a later commit), given a detectable dependency relationship.
- A working tree whose changes form a dependency cycle is rejected with a diagnostic and **nothing is committed**.
- Source files rank above tests/docs/configs in grouping; lock files are excluded from grouping analysis and ride with their dependency change.
- Each commit passes through the normal approval/sandbox gate before landing.
- Commit messages are conventional and describe the single concern of that commit (not a generic "wip" / "updates").
- Tests cover: clean split, ordering, cycle rejection, lock-file handling, approval gating, and a fixture where regex-based dependency detection disagrees with reality (graceful fallback).

## Notes

This is bundled as one issue because grouping, ordering, and cycle-rejection are one algorithm — splitting them would produce three issues that each only make sense once the others land. It is a reliability/review-quality feature (better bisection, revert, and review) more than a raw capability, so it lives in the tools + reliability lane. The LSP-references integration (#3975) is a natural enhancement once both exist, but the first cut should ship with a regex/import heuristic so it does not block on LSP nav landing.

goransh-walia:
This PR addresses #3999.

Tools: atomic commit splitting — order unrelated changes by dependency, reject cycles

Generated with AI assistance and validated against the original
file before submission (syntax check + change-scope check).
Please review carefully — happy to adjust based on feedback.

Closes #3999

---

## Maintainer follow-up

Thanks @goransh-walia for the original implementation and the cycle-rejection
tests, which this rework keeps. Per the review, the planner has been reworked
into a **propose-only** tool and pushed to this branch (head `f94d4aa950`,
rebased onto `main` at `f9746854c`; author commit preserved below the
maintainer commit):

- The tool is now the `Git` action `commit_plan` (`GitCommitPlanTool`),
  routed like status/diff/log/show/blame. It returns a dependency-ordered
  split plan with proposed messages and **writes nothing** — no `git add -N`,
  no `git apply --cached`, no `git commit`, no index or object-store
  mutation. Landing commits stays on the ordinary `git add` / `git commit`
  shell path, where the approval gate already applies.
- The non-existent `ApprovalRequirement::Manual` variant is gone; the tool
  declares `ReadOnly` + `Sandboxable` and classifies read-only end to end
  (envelope `Bounded`, approval policy Safe/Benign, hooks gate `safe`,
  history activity `File`).
- Grouping, dependency ordering, and cycle rejection are kept and covered by
  tests, including a propose-only proof that the index, `HEAD`, and the
  untracked list are untouched after planning.
- Changelog entry added under `[Unreleased]` crediting @goransh-walia.

Gates on the final tree: `cargo fmt --all` clean; `cargo clippy --workspace
--all-targets --all-features --locked -D warnings` clean; targeted
`tools::git git_tool commit_plan canonical_action` 49 passed / 0 failed /
0 ignored; full `cargo test -p codewhale-tui --lib --locked` 11880 passed /
0 failed / 13 ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   per-World session container before each command — full tree first, then
   only changes and deletions — so remote builds and tests run on the files
   just edited locally and their outputs persist across commands.
+- `Git` grows a `commit_plan` action: a propose-only planner that splits the
+  working tree into ordered atomic commits (#3999). It groups whole files —
+  lock files ride with their manifest, tests ride with the source they name —
+  orders the groups so a commit that defines a symbol lands before the commit
+  that uses it, and refuses the whole plan when that dependency graph has a
+  cycle. It reads `git diff HEAD` plus the untracked-file list and writes
+  nothing: no `git add -N`, no `git apply --cached`, no `git commit`, so
+  staging and committing stay with the ordinary `git add` / `git commit` shell
+  path where the approval gate already applies. Thanks
+  [@goransh-walia](https://github.com/goransh-walia) for the original
+  implementation (PR #5870, fixes #3999).
 
 ## [0.9.12] - 2026-09-03
 
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -117,6 +117,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   per-World session container before each command — full tree first, then
   only changes and deletions — so remote builds and tests run on the files
   just edited locally and their outputs persist across commands.
+- `Git` grows a `commit_plan` action: a propose-only planner that splits the
+  working tree into ordered atomic commits (#3999). It groups whole files —
+  lock files ride with their manifest, tests ride with the source they name —
+  orders the groups so a commit that defines a symbol lands before the commit
+  that uses it, and refuses the whole plan when that dependency graph has a
+  cycle. It reads `git diff HEAD` plus the untracked-file list and writes
+  nothing: no `git add -N`, no `git apply --cached`, no `git commit`, so
+  staging and committing stay with the ordinary `git add` / `git commit` shell
+  path where the approval gate already applies. Thanks
+  [@goransh-walia](https://github.com/goransh-walia) for the original
+  implementation (PR #5870, fixes #3999).
 
 ## [0.9.12] - 2026-09-03
 
--- a/crates/tui/src/hooks/executor.rs
+++ b/crates/tui/src/hooks/executor.rs
@@ -2517,7 +2517,7 @@ fn tool_category_for(tool_name: &str, tool_args: Option<&str>) -> &'static str {
         "Git" | "git" => match action.as_deref() {
             // Every shipped Git action is read-only today; classify by action
             // anyway so adding a mutating one cannot silently inherit `safe`.
-            Some("status" | "diff" | "log" | "show" | "blame") => "safe",
+            Some("status" | "diff" | "log" | "show" | "blame" | "commit_plan") => "safe",
             _ => "other",
         },
         // `Run` executes test/verifier commands — closer to shell than safe.
@@ -5819,6 +5819,10 @@ command = "echo project"
         assert_eq!(tool_category_for("apply_patch", None), "file_write");
         assert_eq!(
             tool_category_for("Git", Some(r#"{"action":"log"}"#)),
+            "safe"
+        );
+        assert_eq!(
+            tool_category_for("Git", Some(r#"{"action":"commit_plan"}"#)),
             "safe"
         );
         assert_eq!(tool_category_for("web.run", None), "other");
--- a/crates/tui/src/tools/canonical_action.rs
+++ b/crates/tui/src/tools/canonical_action.rs
@@ -37,6 +37,7 @@ pub(crate) const CANONICAL_ACTION_ALIASES: &[(&str, &str, &str)] = &[
     ("Git", "log", "git_log"),
     ("Git", "show", "git_show"),
     ("Git", "blame", "git_blame"),
+    ("Git", "commit_plan", "git_commit_plan"),
     ("Run", "tests", "run_tests"),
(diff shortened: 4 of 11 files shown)


Hmbown (dismissed):
Kind: request-changes. The issue #3999 ask is legitimate, but this can't merge as-is. Blocking: (1) `ApprovalRequirement::Manual` (your `git.rs:254`) doesn't exist — the enum is Auto/Suggest/Required (`crates/tools/src/lib.rs:44-52`), so this doesn't compile; the PR's CI never ran the build. (2) This adds a second commit authority: a model-visible tool that heuristically groups changes and directly runs `git apply --cached` + `git commit -m` with generated messages (`git.rs:918`, `:923-1008`). Please make it propose-only (return the split plan + messages) and let commits flow through existing write paths, or route via `git_tool.rs` like status/diff. (3) It declares only `Sandboxable`, no `WritesFiles` — every write tool declares it, and the execution envelope fail-closes on under-declared tools (`execution_envelope.rs:310-341`). (4) Data safety: no `git reset` before `git apply --cached` (user-staged hunks get swept in), no rollback on partial failure, and `git add -N` runs before the dry_run check so dry-run mutates the index (`git.rs:276-295`). (5) No DCO sign-off. Happy to re-review a plan-only redesign — the cycle-rejection tests are a good start.

Hmbown:
Thank you @goransh-walia for this contribution and for the cycle-rejection tests, which the rework keeps. A maintainer follow-up push is now on this branch (head `f94d4aa950`): the planner is propose-only per the review — it returns the ordered atomic-split plan and writes nothing, with commits landing through the ordinary `git add` / `git commit` approval path. Details in the Maintainer follow-up section of the PR description. Authorship of your original commit is preserved.

Hmbown:
**Note: this comment is written by Claude Opus 5, posting through @Hmbown's account at his request. The automation that touched your branch was Claude too, so it seemed right that the explanation come from it directly rather than in Hunter's voice. He's read this and signed off on it.**

@goransh-walia — thank you for this one. It's your first PR to the repo and you picked a hard one. #3999 sat open a while because the analysis is the difficult part, and you did it: the hunk parser, symbol extraction, the relatedness and lockfile heuristics, dependency ordering, and the cycle rejection with tests. That's the substance of the feature and it's shipping as you wrote it.

You're owed a straight account of what happened here. After Hunter's review, automated agents — me, essentially — pushed a rework commit directly onto your branch instead of leaving the feedback for you to act on, and force-pushed a rebase over it. Your original commit came through byte-identical with your authorship intact, but that was more than should be done to someone's PR without asking first. That's on the automation and on the humans who pointed it at your branch, not on you. Apologies for the surprise.

What the rework actually changed: it removed `apply_hunks_and_commit` — the stage that rebuilt a patch from parsed hunks, `git apply --cached`'d it into the index, and spawned `git commit`. Two reasons. A patch reconstructed from hunk text silently drops renames, mode changes, and binary files. And it gave the tool its own commit authority separate from the normal approval gate, which is a line this codebase holds everywhere. The tool now returns the ordered split plan with the exact `git add` / `git commit` commands per group, landing through the ordinary approval path — which is what point 5 of #3999 asked for. Every one of your analysis functions is untouched; the only other deletion was `parse_range`, which went dead once nothing rebuilds patches.

Your commit keeps your authorship and you're credited in the CHANGELOG. If you'd rather take the rework back over or shape it differently, say so and it's yours. Either way — strong first contribution, and more would be welcome.

PR #5883, opened 2026-09-04
fix(tui): derive local output budget from route window

### LINKED ISSUE #5820: Ollama provider: input budget collapses to 1024 tokens on 32K local models (default output reservation 64K clamps the window)
# Issue: Ollama provider — input budget collapses to 1024 tokens on 32K-window local models (default output reservation 64K clamps the window)

## Environment

- OS: Windows 10 Pro 22H2 (zh-CN)
- Codewhale: 0.9.11 (96d13a0bc3f4)
- Ollama: 0.32.3 local, http://127.0.0.1:11434
- Model: qwen2.5:7b (native 32K context), NVIDIA RTX 4060 Ti 8GB
- Config: custom `[providers.ollama]` route with `context_window = 32768`

## What happened

1. Launch `codewhale --provider ollama` (TUI) and send any message ("hello").
2. Immediately fails with:
   - `Emergency context compaction failed to reduce request below model limit (estimate ~6225 tokens, budget ~1024)`
   - then `Network error: SSE stream request failed after HTTP/1.1 fallback: error sending request for url (http://127.0.0.1:11434/v1/chat/completions)`
3. The same happens when switching to the ollama route via `/model` inside an already-running session (the session shows "compacting context" and never recovers).

## Root cause (source-level analysis)

For Ollama with an uncatalogued model (`qwen2.5:7b` has no catalogue row):

- `effective_max_output_tokens_for_route` (`crates/tui/src/route_budget.rs`) falls back to `window / 2` where the assumed window is `128_000` → default output reservation = **64000**.
- With the configured 32K window, `clamp_output_cap` (`crates/tui/src/context_budget.rs`) clamps that to `32768 - 2048 = 30720`.
- Remaining input budget = `32768 - 30720 - 1024 = 1024` — exactly `MIN_INPUT_BUDGET_TOKENS`.
- The built-in system prompt alone is ~4676 tokens, so emergency compaction can never reach the 1024 budget and always fails.

The actual wire request also carries `max_tokens: 30720`; with ~6163 estimated input tokens that exceeds the 32K window.

## Evidence

`codewhale --provider ollama exec --output-format stream-json "hi"` metadata reports:

```
"codewhale_max_output_tokens": 30720
"codewhale_max_output_tokens_source": "route-declared"
"estimated_system_tokens": 4676
"estimated_request_tokens": 6163
```

Two failure screenshots attached (2026-08-28 and 2026-09-02).

## Workaround

Launch the ollama profile with `CODEWHALE_MAX_OUTPUT_TOKENS=8192` → input budget becomes ~23552 and everything works. Note: setting this globally also clamps the default deepseek route, so it has to be applied per-launch, which is awkward.

## Suggested fix

1. For uncatalogued models, do not derive the default output reservation from a fabricated 128K window (`128K / 2 = 64K`). Cap the automatic default by the actual window, e.g. `min(8192, window / 4)`, or use a conservative absolute ceiling for small windows (<= 32K).
2. Consider raising `MIN_INPUT_BUDGET_TOKENS` relative to the window size — 1024 is smaller than the built-in system prompt on any route, which makes the emergency path mathematically impossible to satisfy on small-window self-hosted models.
3. `/model` hot-switch inside a running session should re-apply per-route budget math (and ideally honor `CODEWHALE_MAX_OUTPUT_TOKENS` for the new route), so a route switched at runtime does not hit the same trap.

---

（中文摘要：Ollama 路由 + 未收录模型（qwen2.5:7b）时，默认输出预留按 128K/2=64K 计算，在配置的 32K 窗口上被 clamp 到 30720，输入预算被压到 1024 下限（系统提示就有 4676 token），紧急压缩永远无法成功；wire 请求 max_tokens=30720 叠加 ~6163 输入也超窗口。建议：未收录模型的默认输出预留按实际窗口取保守值、MIN_INPUT_BUDGET_TOKENS 随窗口调整、/model 热切换时重新应用路由预算。）

Image

Image

dajiaohuang:
## Summary

- derive the automatic output reservation from a route's declared context window when the model has no static catalogue row
- preserve explicit operator overrides, concrete route output limits, compatibility caps, and existing window clamping
- add a synthetic 32K Ollama regression covering the reported budget collapse and override precedence

Fixes #5820.

## Behavior

For an uncatalogued model on a 32,768-token Ollama route, the automatic output cap is now 16,384 instead of being derived from the 128K fallback and clamped to 30,720. The resulting input ceiling is 15,360 rather than 1,024. No live Ollama process, model, credentials, or provider request is used by the test.

## Validation

- regression test against the unchanged baseline fails as expected: output cap 30,720 vs expected 16,384
- `cargo test -p codewhale-tui route_budget::tests --lib --locked` with `--cap-lints warn`: 24 passed
- `cargo check -p codewhale-tui --locked`: passed with the repository's strict warning policy
- `rustfmt --check --edition 2024 crates/tui/src/route_budget.rs`
- `git diff --check`

The lint cap was used only for lib-test compilation because three unrelated Windows conditional-test variables are denied as unused at baseline. The normal non-test `cargo check` passed without that cap.

### DIFF
--- a/crates/tui/src/route_budget.rs
+++ b/crates/tui/src/route_budget.rs
@@ -24,11 +24,15 @@ pub(crate) fn route_context_window_tokens(
     model: &str,
     route_limits: Option<RouteLimits>,
 ) -> u32 {
+    route_context_limit_tokens(route_limits)
+        .unwrap_or_else(|| provider_capability(provider, model).context_window)
+}
+
+fn route_context_limit_tokens(route_limits: Option<RouteLimits>) -> Option<u32> {
     route_limits
         .and_then(|limits| limits.context_tokens)
         .and_then(|tokens| u32::try_from(tokens).ok())
         .filter(|tokens| *tokens > 0)
-        .unwrap_or_else(|| provider_capability(provider, model).context_window)
 }
 
 /// Provider/offering output cap, when the resolved route reports one.
@@ -230,7 +234,14 @@ pub(crate) fn effective_max_output_tokens_for_route(
     model: &str,
     route_limits: Option<RouteLimits>,
 ) -> u32 {
-    let requested_cap = effective_max_output_tokens(model);
+    let requested_cap = explicit_max_output_tokens_override().unwrap_or_else(|| {
+        route_context_limit_tokens(route_limits)
+            .filter(|_| context_window_for_model(model).is_none())
+            .map_or_else(
+                || effective_max_output_tokens(model),
+                |window| (window / 2).min(API_MAX_OUTPUT_TOKENS),
+            )
+    });
     let compatibility_source = output_ceiling_source(provider, model);
     let compatibility_cap = compatibility_source.clamp_tokens();
     let route_cap = route_output_limit_tokens(route_limits);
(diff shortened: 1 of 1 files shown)


Hmbown:
Thank you for this — you diagnosed #5820 independently and correctly. v0.9.12 shipped the same window/2 reservation for uncatalogued local routes (with the explicit operator cap still winning) as part of a broader route-budget fix, so this branch no longer applies to main. If you see any remaining gap on a real Ollama route after updating, please reopen against main and we'll take it from there.

PR #5895, opened 2026-09-05
fix(computer-use): scope HarmonyOS cleanup to owned temporary files

### LINKED ISSUE #5894: Prevent HarmonyOS file-read cleanup from deleting unrelated temporary files
During review of #5890, the computer-use transport revealed a cleanup bug already present before the packaging move.

At v0.9.12, `hdcExec().readFile()` creates its local file directly under `os.tmpdir()`, then calls recursive `rm(path.dirname(tmp))` after a successful read. That directory is the process-wide temporary directory, so unrelated temporary files can be removed. Failed reads also leave the downloaded file behind.

Evidence: static inspection of `plugins/computer-use/src/transport.mjs` at release tag dcd4c200f72f0c1ffd60d8e7f6850313db879fc5; now located at `crates/tui/plugins/computer-use/src/transport.mjs:120-125`. No destructive live-device reproduction was performed.

For v0.9.13: allocate a unique owned directory, scope cleanup to that directory in a finally block, preserve unrelated sentinel files, and verify cleanup after successful reads, transfer failures, and read failures using an isolated fixture. Track real HarmonyOS device acceptance separately from the local filesystem regression proof.

Hmbown:
A successful HarmonyOS file read downloaded directly under the system temporary directory and recursively removed its parent, potentially deleting unrelated temporary files. Failed transfers also left partial downloads behind.

Download into a unique owned directory and remove only that directory in a finally block. This also cleans up failed transfers and failed reads.

Fixes #5894

Validation: computer-use suite 34/34 passed; local file-read smoke returned the downloaded bytes and removed the owned directory. Three isolated filesystem cases preserve an unrelated sentinel and check cleanup on success, transfer failure, and read failure. Running those tests against the prior code reproduces unrelated-file deletion and leaked partial downloads entirely within the fixture. No real HarmonyOS device acceptance is claimed.

Depends on the bundle relocation in #5890; the defect predates that move. This is v0.9.13 work and does not alter the published v0.9.12 tag or assets.

---

> [!NOTE]
> **Low Risk**
> Localized temp-file lifecycle fix in the computer-use HDC transport with regression tests; no auth, data model, or API surface changes.
> 
> **Overview**
> **HarmonyOS `hdc` remote reads** no longer download into the system temp root and recursively delete the parent folder—a pattern that could wipe unrelated temp files and leave partial files on failed transfers.
> 
> **`hdcExec.readFile`** now pulls into a dedicated `mkdtemp` directory under `cu-hdc-`, reads `out`, and removes only that directory in a `finally` block (errors from cleanup are swallowed so they cannot mask transfer or read failures).
> 
> **Tests** add six parametrized cases (success, transfer failure, read failure × cleanup success/failure) that mock `pullFile` and optionally failing `fs.rm`, asserting sentinel files survive and the correct error or bytes are returned—including reproducing the old parent-delete behavior in isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 030ff761a6fc2a1cda14fb9223e524a16abc645c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hmbown:
Review follow-up is pushed at 29bdcf1dca. Cleanup remains confined to its owned temporary directory and is best effort, preserving both successful bytes and the original transfer/read error when directory removal fails.

37/37 plugin tests pass, including success, transfer failure, and read failure with both successful and failed cleanup. Hosted CI is running on this new head.

Issue #5901, opened 2026-09-05
feat(tui): list custom: theme overlays in the /theme picker


7jrxt42BxFZo4iAnN4CX:
### Summary

User theme overlays (`~/.codewhale/themes/<name>.json`, selected as `theme = "custom:<name>"`, introduced in commit `312a6e872d`) are not surfaced by the `/theme` **picker**. The picker only lists compiled `SELECTABLE_THEMES`; it never scans the themes directory, so custom themes are invisible in the picker and the app does not represent the active overlay correctly.

### Current behavior

1. Create an overlay (`~/.codewhale/themes/black-oc.json`, `base = "claude"`), set `theme = "custom:black-oc"`, restart — it renders fine (`resolve_user_theme`).
2. Open `/theme` (no argument): the overlay is absent; only built-ins are listed.
3. The app keeps only the **resolved base** `ThemeId`, not the `custom:<name>` string:
   - `tui/app.rs` stores `theme_id: ThemeId` and `ui_theme`, but no raw theme name.
   - On load, `app/init.rs` (`let (_, theme_id, ui_theme) = resolve_theme_setting(...)`) discards the normalized `custom:<name>` string.
4. Consequence in the picker: `open_theme_picker` uses `original = app.theme_id.name()` (e.g. `"claude"`), so the picker thinks the active theme is the *base*, not the overlay.
5. With a custom theme active, opening `/theme` and pressing **Enter without moving** persists the base name (`"claude"`), **silently overwriting** `theme = "custom:black-oc"`.
   - A guard exists in `theme_picker.rs` (`enter_without_navigating_preserves_a_custom_theme_selector`) but it only fires when the view is constructed with the `custom:` string — the real open path in `apply.rs` passes `app.theme_id.name()` (the base), so the guard doesn't help.
6. Selecting any built-in row also clobbers the custom theme (expected only if intended, but currently happens by accident even when merely opening/committing).

### Expected behavior

- The `/theme` picker lists user overlays from the themes directory (appended after compiled themes, or a "Custom" section), each as a `custom:<name>` row, with the same live preview.
- The currently active overlay is pre-selected / highlighted.
- Committing without a move keeps the active `custom:<name>` intact; selecting a built-in intentionally persists that built-in.

### Motivation

[...]

PR #5907, opened 2026-09-06
feat(tui): list custom themes in picker

wuisabel-gif:
## Summary

Addresses the approved #5901 slice.

The `/theme` picker now:

- discovers valid user-authored overlays from `$CODEWHALE_HOME/themes/`;
- appends custom overlay rows after compiled themes in stable order;
- preserves the normalized raw `custom:<name>` selector in `App` state;
- highlights the active custom overlay and previews its resolved colors;
- preserves a custom selector when the picker is opened and committed without navigation;
- continues to persist intentional built-in-theme selection normally;
- ignores invalid, unreadable, oversized, and symlinked overlay files without breaking the built-in picker.

The configuration guide and synchronized changelogs document the picker behavior.

## Tests

Passed locally:

- `git diff --check`
- `./scripts/sync-changelog.sh --check`
- user-theme discovery tests: 4 passed
- theme-picker tests: 20 passed with `RUST_MIN_STACK=16777216`
- custom live App-state test: passed
- config-command tests: 94 passed
- `cargo clippy -p codewhale-tui --all-targets --locked -- -D warnings`

No provider credentials or network access are required.

No-Issue: This PR addresses the approved focused #5901 picker slice without automatically closing the broader issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- The `/theme` picker now discovers valid user-authored `custom:<name>`
+  overlays, previews their colors, highlights the active overlay, and preserves
+  it when the picker is opened and committed without navigation (#5901).
+
 ## [0.9.12] - 2026-09-04
 
 Codewhale v0.9.12 puts computer use in the binary, opens two new routes —
--- a/crates/tui/CHANGELOG.md
+++ b/crates/tui/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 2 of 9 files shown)


Hmbown (commented):
Verified the clobber on main — `apply.rs` passed the *base* theme name as the picker's `original`, and `app/init.rs` dropped the normalized selector, so storing `app.theme_name` (and cloning it at `apply.rs:1981`) is the right fix; the `custom:midnight` preview/commit test and the live App-state test in `config.rs` are exactly the coverage this needed. Two asks: (1) are the new custom rows exercised at narrow width (the picker's list-when-narrow path at `theme_picker.rs:202` has infra; a one-line assertion would pin it), and (2) confirm custom row labels route through the same localized label helpers as built-ins. Thanks for skipping invalid/oversized/symlinked overlays and for leaving `prompts/text.rs` untouched. Housekeeping: commit is missing a `Signed-off-by` trailer — `git commit --amend -s` when convenient.

Issue #5896, opened 2026-09-05
Windows computer-use reports input success after PowerShell failure


Hmbown:
The computer-use Windows backend can return `{action_sent: true}` when its PowerShell subprocess has failed, and the targeted mouse-down path omits the actual button press.

Verified against 05241f4095462bcd62fff84709f0a811bafae7d4 using a temporary fake `powershell.exe` that captures commands and exits 1. No Windows GUI or live user input was exercised. Calling `create().left_mouse_down({target:{x:12,y:34}})` still returned `{action_sent:true}`.

Source findings in `crates/tui/plugins/computer-use/src/backends/win32.mjs`:

- `create()` registers the User32 type in a bootstrap PowerShell process, but `ps()` spawns a fresh process for each action. The action command does not carry the type definition.
- `withUser32()` returns `ps()` without rejecting a nonzero exit. Several input methods then report success unconditionally.
- In `left_mouse_down`, ternary/concatenation precedence means a supplied target sends only `SetCursorPos`; the `mouse_event(LEFTDOWN)` expression is in the no-target branch.

For v0.9.13: make every action self-contained in its PowerShell process, fail truthfully on subprocess errors, and verify that targeted mouse-down both moves and presses. Add command-capture and failure-path tests; keep native Windows desktop acceptance as a separate required receipt. Review other User32-backed actions for the same process-lifetime and success-reporting defect.

PR #5910, opened 2026-09-06
fix(computer-use): win32 input actions fail truthfully; targeted mouse-down presses

### LINKED ISSUE #5908: win32 computer-use: input actions report success on PowerShell failure; left_mouse_down drops the press
## Root cause

The Windows backend (`crates/tui/plugins/computer-use/src/backends/win32.mjs`) reports `{action_sent: true}` even when the underlying action never ran. Three independent defects:

1. **No in-process type definition.** `create()` only loads the `User32` P/Invoke type in the *bootstrap* PowerShell process. Every action runs in a *fresh* `powershell.exe` spawned by `ps()`, so `[User32]::mouse_event` / `SendInput` / `GetCursorPos` are undefined there and the call fails — but nobody checks.
2. **`withUser32` never inspects the exit code.** It just `return ps(script, opts)`, so a nonzero exit is ignored and every User32 action returns success unconditionally (see #5896).
3. **`left_mouse_down` precedence bug.** `target ? A : "" + B` evaluates `"\u201c" + B` on the no-target branch, so a *targeted* call only emits `SetCursorPos` and silently drops `mouse_event(LEFTDOWN)`.

## Impact

A parked/crashed agent, or any Windows session, believes input (clicks, drags, typing, key combos) was delivered when it was not. The model proceeds on false confirmations — a silent, high-impact correctness failure.

## Fix (in progress)

- Make `create()` accept an injectable runner (`{ exec }`) so the backend is testable without `powershell.exe`.
- Prepend the `User32` type definition (`USER32_DEF`) to every User32-backed script so each action is self-contained.
- `withUser32` now throws an `ExecError` on a nonzero exit instead of reporting success.
- Rewrite `left_mouse_down` so a targeted call both moves **and** presses.
- `cursor_position` routes through `withUser32` so it loads the type in-process and fails truthfully.
- Regression tests in `tests/win32.test.mjs` (mocked runner): type defined in-process, failure path rejects, targeted `left_mouse_down` moves+presses.

Relates to #5896.

EvanProgramming:
## Summary

Fixes #5908 (relates to #5896). The Windows computer-use backend reported `{action_sent: true}` even when the underlying PowerShell subprocess never delivered the input.

Three root causes, all in `crates/tui/plugins/computer-use/src/backends/win32.mjs`:

1. **User32 type never loaded in the action process.** `create()` only ran `Add-Type` in a bootstrap process, but every `ps()` spawns a *fresh* `powershell.exe`, so `[User32]::mouse_event`/`SendInput`/`GetCursorPos` were undefined there and the call failed silently.
2. **`withUser32` ignored the exit code.** It just returned `ps(...)`, so a nonzero exit still produced `{action_sent: true}`.
3. **`left_mouse_down` precedence bug.** `target ? A : "" + B` put the `mouse_event(LEFTDOWN)` in the no-target branch, so a targeted call only sent `SetCursorPos`.

## Changes

- `USER32_DEF` prepends the `User32` P/Invoke definition to every User32-backed action so each is self-contained in its own process.
- `withUser32` throws an `ExecError` on a nonzero exit — no more false success.
- `left_mouse_down` rewritten so a targeted call both moves the cursor **and** presses.
- `cursor_position` now routes through `withUser32` (loads the type, fails truthfully instead of lying about a position).
- `create({ exec })` accepts an injectable runner so the backend is testable on any host without `powershell.exe` (mirrors the harmonyos backend's pattern).

## Tests

New `crates/tui/plugins/computer-use/tests/win32.test.mjs` (runs on any host via a mocked runner):

- User32 actions define the type in their own process.
- A nonzero subprocess exit rejects the action (no false `action_sent: true`).
- `left_mouse_down({ target })` sends both `SetCursorPos` and `LEFTDOWN`.
- `cursor_position` loads the type in-process and never reports a fake position.

Verified: `node --test tests/win32.test.mjs` → 4/4 pass; `node --test tests/backends.test.mjs` → 10/10 pass (no regression).

### DIFF
--- a/crates/tui/plugins/computer-use/src/backends/win32.mjs
+++ b/crates/tui/plugins/computer-use/src/backends/win32.mjs
@@ -10,21 +10,9 @@ import crypto from "node:crypto";
 import { spawn } from "node:child_process";
 import { run, runOk, ExecError, tryJson } from "../exec.mjs";
 
-function ps(script, opts = {}) {
-  const encoded = Buffer.from(script, "utf16le").toString("base64");
-  return run("powershell.exe", ["-NoProfile", "-NonInteractive", "-EncodedCommand", encoded], {
-    timeoutMs: opts.timeoutMs ?? 25_000,
-    maxBuffer: 32 * 1024 * 1024,
-  });
-}
-
-async function psJson(script, opts = {}) {
-  const r = await ps(script, opts);
-  const out = r.stdout.trim();
-  const j = tryJson(out, null);
-  if (!j) throw new ExecError(`powershell did not return JSON: ${(r.stderr || out).trim().slice(0, 300)}`, r);
-  return j;
-}
+// One-shot PowerShell runner is defined per `create()` instance below so a
+// test can inject a fake runner; the shared bootstrap type and every action
+// script are assembled against that instance-local runner.
 
 const USER32 = `
 using System;
@@ -51,6 +39,14 @@ public static class User32 {
   }
 }`;
 
+// Prepend this to every User32-backed script. Each `ps()` spawns a FRESH
+// powershell.exe process, so the type must be (re)defined in-process — the
+// bootstrap process's Add-Type does NOT carry over. Defining it inline is what
+// makes every action self-contained (see issue #5896).
+const USER32_DEF = `Add-Type -TypeDefinition @'
+${USER32}
+'@ -ErrorAction SilentlyContinue; [User32] | Out-Null;`;
+
 function recordingsDir() {
   return process.env.CODEWHALE_CU_RECORDINGS_DIR || path.join(os.homedir(), ".codewhale-cu", "recordings");
 }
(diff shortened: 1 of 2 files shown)


Hmbown:
Closing with credit — this bug (#5896) was also fixed independently and has just merged as #5903 (self-contained `Add-Type` prelude per action, `psOk`/`psJson` throwing on nonzero exit/spawn failure, targeted mouse-down always pressing after the optional move). Your take arrived while ours was in CI, so rather than let it sit: thank you — and specifically, your **instance-local injectable runner is a cleaner test seam than the fake-powershell.exe-on-PATH approach we merged**, and we may adopt that pattern for future backend tests. If you'd like to rebase your DI seam on top of main as a follow-up refactor PR, it would be welcome. Also happy to review your other PR (#5911) — that one's untouched territory.

PR #5911, opened 2026-09-06
fix(tools): write_file preserves the existing file's line-ending style (CRLF)

### LINKED ISSUE #5909: write_file silently converts CRLF files to LF on overwrite (edit_file preserves, write_file doesn't)
## Root cause

`edit_file` preserves a file's line-ending style on mutation: it detects the prior ending via `contract_line_ending`, normalizes to LF internally, then restores with `restore_contract_line_endings` (see `crates/tui/src/tools/file.rs`).

`write_file` does **not**. Both write paths:

- `WriteFileTool::execute` (full replace) — writes `file_content.as_bytes()` verbatim.
- `WriteFileTool::execute_contract_write` (the hidden `write_file` / `File` contract path) — same.

So overwriting an existing **CRLF (Windows)** file with `write_file` rewrites every `\r\n` to `\n`. The original style is lost even though the model only intended to change content.

## Impact

Any `write_file` over a CRLF file on Windows (or any CRLF-committed repo) silently normalizes line endings to LF. This changes the byte content of files the user did not mean to touch, breaks diffs/blame, and is inconsistent with `edit_file`'s behavior for the same file.

## Reproduction

1. Create a file with CRLF endings (committed as CRLF).
2. Call `write_file` to replace its contents with LF- or CRLF-tagged content.
3. The on-disk file is now LF.

`edit_file` on the same file keeps CRLF.

## Fix (in progress)

In both write paths, when the file already existed, normalize the incoming content (`normalize_contract_line_endings`) and restore to the *prior* file's detected ending (`contract_line_ending(prior)` → `restore_contract_line_endings`). New files are written verbatim (no prior style to honor). Add a regression test asserting a CRLF file stays CRLF after `write_file`.

EvanProgramming:
## Summary

Fixes #5909. `write_file` silently converted an existing **CRLF** (Windows) file to **LF** on overwrite, while `edit_file` preserves the detected line ending — so the same file behaved differently depending on the tool used. Both write paths were affected:

- `WriteFileTool::execute` (full replace)
- `WriteFileTool::execute_contract_write` (the hidden `write_file` / `File` contract path)

## Root cause

Both paths wrote `file_content.as_bytes()` verbatim. Model-produced content is LF, so every `\r\n` in the on-disk file became `\n` — changing byte content the model never intended to touch, corrupting diffs/blame on CRLF-committed repos and on Windows.

## Changes

- New helper `preserve_prior_line_endings(content, prior)`: normalizes incoming content to LF internally, then restores the *prior* file's detected ending (`contract_line_ending`), exactly mirroring `edit_file`'s policy.
- Both write paths call it. A brand-new file (no prior content) is written verbatim — there is no style to honor, so creation behavior is unchanged.

## Tests

- E2E: `write_file` over an existing CRLF file with LF content leaves the file CRLF.
- Helper unit coverage: CRLF prior → CRLF out; LF prior → LF out; empty prior → verbatim (including CRLF input).

## Verification

- `cargo check -p codewhale-tui --lib --tests` passes.
- `cargo fmt -p codewhale-tui` clean (only the touched files changed).

### DIFF
--- a/crates/tui/src/tools/file.rs
+++ b/crates/tui/src/tools/file.rs
@@ -1385,26 +1385,25 @@ impl WriteFileTool {
             })?;
         }
         check_file_operation_cancelled(context)?;
-        crate::utils::write_atomic_workspace(&file_path, file_content.as_bytes()).map_err(
-            |error| {
-                ToolError::execution_failed(format!(
-                    "Failed to write {}: {error}",
-                    file_path.display()
-                ))
-            },
-        )?;
+        // Preserve the existing file's line-ending style on overwrite (see
+        // `preserve_prior_line_endings`); otherwise a CRLF (Windows) file is
+        // silently rewritten with LF line endings.
+        let written = preserve_prior_line_endings(file_content, &prior_contents);
+        crate::utils::write_atomic_workspace(&file_path, written.as_bytes()).map_err(|error| {
+            ToolError::execution_failed(format!("Failed to write {}: {error}", file_path.display()))
+        })?;
         check_file_operation_cancelled(context)?;
         context.note_file_read(&file_path);
         drop(mutation_guard);
 
         let outcome = if existed_before { "updated" } else { "created" };
-        let utf16_units = file_content.encode_utf16().count();
+        let utf16_units = written.encode_utf16().count();
         Ok(contract_mutation_result(
             context,
             &file_path,
             path_str,
             prior_contents.as_ref(),
-            file_content,
+            &written,
             outcome,
             format!("Successfully wrote {utf16_units} bytes to {path_str}"),
         )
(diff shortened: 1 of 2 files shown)


Hmbown (commented):
Verified the bug on main — both write paths wrote content verbatim (`crates/tui/src/tools/file.rs:1388`, `:1507`) while `edit_file` preserved endings, so parity here is exactly right, and thanks for updating the diff/summary to use `written` so "(no changes)" stays truthful. Two small follow-ups: (1) `normalize_contract_line_endings` (`file.rs:1568`) also rewrites bare `\r` to the detected ending, so a lone CR in the new content becomes CRLF when the prior is CRLF — same policy as `edit_file`, but a unit test would pin it; (2) the contract path (`execute_contract_write`, `file.rs:~1358`) is only covered via the shared helper — one E2E test there would close the loop. Housekeeping: the commit is missing its `Signed-off-by:` trailer; mind amending with `git commit --amend -s`? Nice work, and thanks for the well-written #5909.

EvanProgramming:
Thanks for the review — all three follow-ups are addressed in the updated commit (`5693fbd4d`):

1. **Lone-`\r` policy pinned**: `preserve_prior_line_endings_keeps_the_prior_style` now asserts that a bare CR in the incoming content normalizes like `edit_file` (lone `\r` → `\n` → re-emitted as the prior ending: CRLF prior → `\r\n`, LF prior → `\n`).
2. **Contract-path E2E added**: `contract_write_over_crlf_file_preserves_crlf_line_endings` drives `WriteFileTool::execute_contract_write` over an existing CRLF file and asserts the on-disk result stays `\r\n`.
3. **`Signed-off-by:` trailer** added via `git commit --amend -s`.

Verification: `cargo check -p codewhale-tui --lib --tests` passes and `cargo fmt -p codewhale-tui` is clean.

PR #5912, opened 2026-09-06
refactor(computer-use): injectable win32 runner for host-agnostic backend tests

EvanProgramming:
## Summary

Refactor-only follow-up, as invited on the closed #5910: the win32 input bug was already fixed and merged in #5903 — this PR does **not** re-touch that logic. It keeps main's merged behavior and adds the **instance-local injectable runner** seam that was called out as cleaner than the fake-powershell.exe-on-PATH fixture.

## Why

`crates/tui/plugins/computer-use/src/backends/win32.mjs` defined its PowerShell runner helpers (`ps`/`psOk`/`psJson`) at module scope, so backend tests either needed a real `powershell.exe` or a PATH fixture. Moving them into `create()` lets a test pass `create({ exec: { run } })` — the exact injection pattern the harmonyos backend already uses.

## Changes

- `create(opts = {})`: `ps`/`psOk`/`psJson` become instance-local and default to the production runner; `opts.exec.run` overrides it. No production behavior change when `create()` is called with no args.
- New `tests/win32.test.mjs` exercising the win32 backend on **any host** (no Windows, no fake powershell):
  - actions run through an injected runner;
  - a nonzero exit rejects the action (fail-truthful behavior from #5903 is pinned);
  - targeted `left_mouse_down` moves **and** presses, with the `User32` type defined in-process;
  - every User32 action carries the type prelude.

## Verification

`node --test tests/win32.test.mjs` → 4/4 pass; `node --test tests/backends.test.mjs` → 10/10 pass (no regression). All on macOS.

Refs #5903, #5910.

### DIFF
--- a/crates/tui/plugins/computer-use/src/backends/win32.mjs
+++ b/crates/tui/plugins/computer-use/src/backends/win32.mjs
@@ -10,30 +10,6 @@ import crypto from "node:crypto";
 import { spawn } from "node:child_process";
 import { run, runOk, ExecError, tryJson } from "../exec.mjs";
 
-function ps(script, opts = {}) {
-  const encoded = Buffer.from(script, "utf16le").toString("base64");
-  return run("powershell.exe", ["-NoProfile", "-NonInteractive", "-EncodedCommand", encoded], {
-    timeoutMs: opts.timeoutMs ?? 25_000,
-    maxBuffer: 32 * 1024 * 1024,
-  });
-}
-
-/** ps() but truthful: timeout, nonzero exit, and spawn failure all throw. */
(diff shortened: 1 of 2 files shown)


Hmbown (approved):
Verified locally on the PR head: the plugin suite passes (47/47 via `npm test`), the `create({ exec })` seam matches the harmonyos backend's shape, and `create()` with no args is behavior-identical to main. The four new tests pin exactly the #5903 guarantees (in-process User32 prelude, nonzero exit rejects, targeted `left_mouse_down` moves and presses) without a Windows host. Merging once CI finishes.

PR #5913, opened 2026-09-06
test(tui): stabilize the forced-idle-timeout worker-reuse test on Windows (#5898)

### LINKED ISSUE #5898: Stabilize the Windows worker-idle-timeout regression test
The Windows worker-idle-timeout regression test is intermittently red on otherwise passing release follow-up CI.

Evidence:

- Packaging PR #5890, head 05241f4095462bcd62fff84709f0a811bafae7d4, run [33963457588](https://github.com/Hmbown/Codewhale/actions/runs/33963457588/job/101299329493): 13,925 passed, 1 failed, 13 skipped. `task_manager::tests::forced_idle_timeout_releases_the_worker_for_later_tasks` expected the second task to be Completed but observed Failed (task_manager.rs:4114).
- Website PR #5889, head 26e0355128b4c4cc5304dcd51affed0bf3c7372e, run [33964195670](https://github.com/Hmbown/Codewhale/actions/runs/33964195670/job/101301342095): the same test passed; 13,926 passed, 13 skipped. Neither PR modifies task_manager.rs or Cargo.lock.

The test uses a 150 ms idle budget and 400 ms wall budget. Its second task goes through MockExecutor, which emits several persisted events and sleeps for 50 ms. Scheduling/storage contention is a plausible explanation; the failed record’s terminal reason was not printed, so this is not yet a confirmed root cause.

For v0.9.13: include the full terminal record in failure diagnostics, reproduce under contention on Windows, and make the worker-reuse test deterministic without weakening production timeouts or removing the proof that a forcibly timed-out task releases its worker. Preserve the rule that failures are visible rather than silently retried.

Hmbown:
Closes #5898.

**Root cause (from the failing CI log, job 101421344166):** the panic was `left: Failed, right: Completed` in **3.177s** — not a wait-budget timeout. The follow-up task ("run after hang") inherits the same `short_for_tests` budgets that exist to force-terminalize the *stuck* task (`idle_progress` 150ms, `wall_time` 400ms, `cancel_grace` 50ms), and the mock executor's follow-up path performs real awaits. Under a ≥150ms scheduler stall mid-flight — before its events queue, so the drain-on-idle-interrupt rescue cannot retract it — the guard interrupts with IdleTimeout, the executor observes cancellation and returns Canceled, and `preserve_timeout_reason` rewrites it to the timeout reason → `Failed`.

**Fix (test module only):** the follow-up branch of the test executor now completes with zero await points (one synchronous `events.try_send` to keep the released worker's event pipeline exercised, then immediate Completed). `run_task`'s biased select polls the executor future first and the first evaluate runs at elapsed ≈ 0, so an await-free future always finishes before any interrupt can be recorded — deterministic under any load, closed by construction rather than by margin. Both asserts now print the full terminal record on failure, and a new assert pins the follow-up's `terminal_reason` to `"completed"`.

**Evidence:** this exact test blocked #5899 and #5905 Windows matrices today. Local: task_manager suite 44/44; forced_idle_timeout ×5 single-threaded 5/5; ×10 under 8 CPU-burner starvation 10/10; full suite ×3 under the same load 3×44/44; `cargo fmt` clean. The hang branch (`std::future::pending`) is unchanged, so the test still proves a forcibly idle-timed-out task releases its worker.

---

> [!NOTE]
> **Medium Risk**
> Changes production `ExecutionGuard` timeout reason selection in `run_task` and engine turns; behavior is more accurate but could alter which terminal reason users see when both limits elapse under scheduler delay.
> 
> **Overview**
> Fixes flaky **forced idle timeout / worker reuse** coverage on loaded CI (including Windows) by tightening timeout attribution and making the follow-up test executor deterministic.
> 
> **`ExecutionGuard::evaluate`** no longer picks wall vs idle timeout from check order alone. It compares **`Instant` deadlines** (`checked_add` on wall and idle budgets) so a starved watchdog still reports **whichever limit actually expired first**; when both are past due, **idle wins** unless late progress pushed idle past wall (ties keep wall precedence per #5898).
> 
> The **`PromptRouterExecutor`** test follow-up path stops delegating to **`MockExecutor`** (multiple `await`s plus sleep). It now finishes **without await points**—one synchronous **`try_send`** and an immediate **`Completed`**—so **`run_task`’s biased poll** completes before short test budgets can interrupt and rewrite the result via **`preserve_timeout_reason`**.
> 
> Adds a unit test for dual-expired guard behavior, extends the progress/wall guard test, and improves integration asserts (including **`terminal_reason: completed`** on the follow-up task).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ad5c76d7c3f768865f88e65f19c263b9f768673. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/tui/src/task_manager.rs
+++ b/crates/tui/src/task_manager.rs
@@ -529,16 +529,28 @@ impl ExecutionGuard {
 
         let wall_elapsed = now.saturating_duration_since(self.started_at);
         let idle_elapsed = now.saturating_duration_since(self.last_progress_at);
+        // A limit whose deadline does not fit in `Instant` can never fire.
+        let wall_deadline = self.started_at.checked_add(self.limits.wall_time);
+        let idle_deadline = self.last_progress_at.checked_add(self.limits.idle_progress);
         let pending = if shutdown {
             Some(TaskTerminalReason::Shutdown)
         } else if cancel {
             Some(TaskTerminalReason::Canceled)
-        } else if wall_elapsed >= self.limits.wall_time {
-            Some(TaskTerminalReason::WallTimeout)
-        } else if idle_elapsed >= self.limits.idle_progress {
-            Some(TaskTerminalReason::IdleTimeout)
         } else {
-            None
+            // Attribute the timeout to the limit that was crossed first, not
+            // to the one this tick happens to check first. When the watchdog
+            // is starved past both deadlines (a >=250 ms scheduler stall on a
+            // loaded CI runner is enough with the test budgets), the idle
+            // limit that expired earlier is still the truthful reason; a tie
+            // keeps the wall limit's precedence (issue #5898).
+            match (wall_deadline, idle_deadline) {
+                (Some(wall), Some(idle)) if now >= wall && wall <= idle => {
+                    Some(TaskTerminalReason::WallTimeout)
+                }
+                (_, Some(idle)) if now >= idle => Some(TaskTerminalReason::IdleTimeout),
+                (Some(wall), _) if now >= wall => Some(TaskTerminalReason::WallTimeout),
+                _ => None,
+            }
         };
         if let Some(reason) = pending {
             return GuardAction::Interrupt { reason };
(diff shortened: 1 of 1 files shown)


Hmbown:
Second Windows flake in the same family, fixed on this branch: `heartbeat_status_does_not_refresh_idle_timeout` failed on #5911's matrix with `left: Some("wall_timeout"), right: Some("idle_timeout")`. Root cause is in the guard, not the test: `ExecutionGuard::evaluate` checked the wall limit before the idle limit, so a scheduler stall that carried the watchdog past both test budgets (150 ms idle, 400 ms wall) reported the wall limit even though idle expired first. The guard now attributes the timeout to whichever deadline was crossed first (tie keeps wall precedence). New unit test covers both orderings under a starved tick. Local: task_manager::tests 45/45. Rebased onto current main.

Issue #5929, opened 2026-09-06
Parallel-execution flakes in the codewhale-tui lib suite (tracking): six tests that fail under load and pass in isolation


Hmbown:
Every one of these failed exactly once tonight (2026-09-06) in a full `RUST_MIN_STACK=16777216 cargo test -p codewhale-tui --lib` run or a CI leg, passed in isolation, and passed on the next full run. None was touched by the PR that hit it. Each costs a rerun or a 25-minute matrix.

| Test | Where seen | Symptom |
|---|---|---|
| `hooks::executor::tests::background_hook_receives_the_same_stdin_payload_as_foreground` | #5921 ubuntu-latest | `valid JSON: EOF while parsing a value` at executor.rs:4649 — the background hook read an empty stdin |
| `session_control_acceptance::tests::tui_and_api_listings_agree` | #5920 local full run | touches shared-home session files |
| `remote_control::tests::separate_predispatch_crashes_on_one_run_get_distinct_recovery_turn_ids` | #5921 and #5924 local full runs (also fails on unmodified main under load) | |
| `model_inventory::…ollama_default_prefers_live_local_tags…` | #5924 local full run | live local tags probe under load |
| `client::tests::deepseek_anthropic_translate_uses_messages_endpoint` | #5920 local full run | |
| `tui::clipboard::tests::tmux_load_buffer_w_reaches_attached_client_with_default_passthrough_disabled` | #5920 local full run | |

Related, already fixed or filed: #5898 (idle-timeout race, fixed by #5913), #5919 (`/pin` blocks on the Windows message pump, #5922).

Acceptance: each test either isolates its shared state (temp home, no wall-clock races, no live probes) or is marked with the reason it cannot run in parallel; three consecutive green full-suite runs under `cargo nextest` default parallelism on macOS and ubuntu.

Hmbown:
Two more from tonight's full-suite runs on the #5935 branch, both passing in isolation and not touched by that PR:

- `runtime_api::tests::steer_and_interrupt_endpoints_work_on_active_turn` — this one was real once: the first cut of #5935 drained a process-wide notice queue, so another test's parked notice surfaced as an unexpected event. Fixed by keying the drain per workspace; listed here because the same shape (global static shared across tests) is worth grepping for.
- `runtime_threads::tests::engine_error_remains_failed_after_nominal_turn_complete` — failed once in a full run, 1/1 in isolation.

Hmbown:
[...]

PR #5946, opened 2026-09-06
feat(fleet): surface worker deliverables via summary and saved-session reply

gaord:
## Summary

Two related changes so a completed Fleet task no longer reports a meaningless receipt when it produced only text:

1. Bounded deliverable excerpt: the terminal `codewhale exec` `metadata` receipt carries `visible_final_answer_excerpt` — a bounded, secret-redacted excerpt of the **final assistant reply** — next to the real pre-bound `visible_final_answer_chars` count. The executor reads that receipt (never the streamed `content` deltas) and attaches the excerpt to `Completed.summary`. When a task has no scorer and no file artifact, `verify_task_result` surfaces that deliverable in the receipt notes instead of "no verifiable output".

2. Saved-session reply: `codewhale exec` now emits the real saved-session id in its `session_capture` stream event under `saved_session_id` (the log fingerprint stays redacted). The executor captures it, `FleetReceipt` persists it, and the runtime API exposes it so a client can resolve the worker's full final assistant reply via GET /v1/sessions/{id}.

## Testing

- [x] cargo test -p codewhale-tui --lib fleet::executor fleet::task_spec fleet_receipt_json
- [x] cargo test -p codewhale-protocol fleet

Note: main currently has pre-existing nonminimal_bool clippy errors (subagent/mod.rs, apply.rs, fleet_roster.rs), unrelated to this change.

---

## Maintainer follow-up

All credit for the feature and the original implementation goes to @gaord — the original commit is preserved below the follow-up. Per the review, the saved-session half was kept as-is and the summary half was reworked (head `4de6dc17f`, branch rebased on current `main`):

1. The excerpt is now taken from the right end of the stream: the terminal `codewhale exec` `metadata` receipt carries `visible_final_answer_excerpt`, a bounded, secret-redacted excerpt of the final assistant reply, next to the real pre-bound `visible_final_answer_chars` count.
2. The executor's frame accumulator is deleted — no assistant text is streamed into per-worker memory anymore.
3. Exactly one bound-and-redact helper remains, `exec_stream_final_answer_excerpt` in `crates/tui/src/lib.rs` (4,000 chars); the executor- and task_spec-side duplicates are gone.
4. Every worker stream line is parsed exactly once: `parse_exec_terminal_*` take `&serde_json::Value` and the incremental reader maps route evidence, final answer, session capture, and ledger payload from one parse.
5. `session_capture.session_id` is renamed `saved_session_id` (protocol `FleetReceipt.session_id` too, plus every consumer: runtime API receipt JSON, manager, task_spec, tests). This deliberately retires the old "no raw id anywhere in the capture" invariant: the raw recoverable id now lives in exactly one field, `session_capture.saved_session_id`, while the `metadata` receipt stays fingerprint-only and `metadata.resume_command` names that field instead of a fake redaction.
6. Docs updated in `docs/AGENT_RUNTIME.md` and `docs/zh_hans/AGENT_RUNTIME.md`.
7. The excerpt now also surfaces on FAILED outcomes (a worker that fails after writing most of a report keeps its text on the terminal event and the receipt), and lifecycle/event labels show a short excerpt while payloads and receipts keep the full one.

Gates (all run on `4de6dc17f`): `cargo fmt --all` clean; `cargo clippy --workspace --all-targets --all-features --locked -D warnings ...` clean; targeted tui tests 100 passed / 0 failed; `cargo test -p codewhale-protocol --locked` 85 passed / 0 failed; full `cargo test -p codewhale-tui --lib --locked` 11876 passed / 0 failed / 13 ignored.

---

## Contributor follow-up (Devin round 2)

Addressed on top of `4de6dc17f` (head `f94684cd6`, merged current `main` first):

1. **Excerpt is now genuinely the final assistant reply** (Devin: "Progress text pollutes final replies"). In a multi-step turn `ExecSummary::output` accumulates every streamed delta, including pre-tool commentary from earlier steps, so the terminal metadata could present progress text as the deliverable. New helper `exec_stream_final_answer_text` derives the answer from the last assistant-like message of the persisted session (`latest_messages`); the cumulative output is only a fallback when the session carries no assistant text. Covered by `exec_stream_final_answer_text_is_the_last_assistant_reply` (pre-tool commentary + tool result + distinct final answer) and `exec_stream_final_answer_text_requires_assistant_text`.
2. **Branch CHANGELOG hunks removed** (Devin: repo policy reserves changelog updates for merge-time commits on `main`). They also collided with main; the merge resolution keeps main's side.
3. **PR description updated** to name the current data flow (terminal-metadata excerpt), which the old Summary section predated.

Gates on `f94684cd6`: `cargo fmt --all -- --check` clean; `cargo test -p codewhale-tui --lib -- exec_stream_final_answer` 3 passed / 0 failed; `cargo test -p codewhale-tui --lib -- fleet::executor fleet::task_spec fleet_receipt` 49 passed / 0 failed (one load-flake rerun per #5929, green on rerun and in isolation); `cargo test -p codewhale-protocol --lib -- fleet` 24 passed / 0 failed.

CI note: the `Test (windows-latest)` leg on `d3b333b06` failed two `mcp_boot_*` engine tests — this branch does not touch that code, they pass on main's latest run, and the symptoms match the #5929 load-flake tracking issue (same class as the buildkite retrigger commit already on the branch).

No-Issue: fleet worker deliverable surfacing tracked by this PR

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5946" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/github/devin-review-dark.svg?v=3">
      Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   answerable work and no longer inflates the `blocked` chip; the receipts
   roster and the wire `state` gain `parked` (#5906, #5921).
 
-
 - `codewhale account keys set|remove|list` no longer carry a hardcoded
   eight-provider list. Provider ids come from the control plane's public
   catalog (`GET /api/model-providers`), are validated locally against
@@ -53,6 +52,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   maps a catalog row onto the local runtime provider through the catalog's own
   `runtimeProvider` field, so a newly supported provider needs no CLI release.
 
+- `/mcp` lists the servers that need a login first, as their own
+  `Needs login` group above `Needs attention`, and opens with the cursor
+  already on the first such row so the Enter the screen advertises runs
+  `/mcp login <server>` straight away; translated in all 15 packs. A
+  snapshot test pins the footer shape the chip landed with (`MCP · N
+  connected · N ◆ auth required · N failed`) so an expired login never
+  regresses into the failed count (#5926).
+
 ### Fixed
 
 - The posture bar states how long the session has been working and how long
@@ -60,8 +67,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tool, a sub-agent or the operator; the 0.9.12 shell had dropped the overall
   working-time indicator from the place a glancing user checks (#5914).
 
+- An MCP token refresh that fails to parse the provider's answer keeps the
+  endpoint's receipt — status line, content type, and a 200-byte excerpt
+  with every credential-shaped value (`access_token`, `refresh_token`,
+  `client_secret`, `id_token`, bearer schemes) masked before the cut —
+  instead of rmcp's bare `Failed to parse server response`, so a provider
+  outage answering an HTML 502 reads differently from a parser defect, and
+  the login remedy stays named (#5926; remedy wording landed in #5959).
+
 ### Added
 
+- The `rusty-alloc` cargo feature on `codewhale-tui` and `codewhale-cli`
+  opts the binaries into the `rusty_alloc` global allocator (the mimalloc
+  v2.4.5 architecture remade in pure Rust — no C compiler or build script
+  on that path) instead of the default mimalloc. It is off by default and
+  the default build is unchanged; build with
+  `cargo build -p codewhale-tui --features rusty-alloc` (#5872).
 - The `/theme` picker now discovers valid user-authored `custom:<name>`
   overlays, previews their colors, highlights the active overlay, and preserves
   it when the picker is opened and committed without navigation (#5901).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Fleet workers now surface their deliverable. The terminal `codewhale exec`
+  `metadata` receipt carries `visible_final_answer_excerpt`, a bounded,
+  secret-redacted excerpt of the final assistant reply next to the real
+  `visible_final_answer_chars` count; the Runtime executor attaches it to
+  `Completed.summary` and, for a task with no scorer and no file artifact,
+  to the receipt notes instead of "no verifiable output" — a worker that
+  fails after writing most of a report keeps the text too. `session_capture`
+  now carries the raw `saved_session_id` (the `metadata` receipt stays
+  fingerprint-only), `FleetReceipt.saved_session_id` persists it, and the
+  runtime API exposes it so a client can resolve the worker's full final
+  reply via `GET /v1/sessions/{id}` (#5946, thanks @gaord).
+
 ### Changed
 
 - `/statusline` drives the bottom chrome again. Since the 0.9.12 shell
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,19 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+### Fixed
 
-- Fleet workers now surface their deliverable. The terminal `codewhale exec`
-  `metadata` receipt carries `visible_final_answer_excerpt`, a bounded,
-  secret-redacted excerpt of the final assistant reply next to the real
-  `visible_final_answer_chars` count; the Runtime executor attaches it to
-  `Completed.summary` and, for a task with no scorer and no file artifact,
-  to the receipt notes instead of "no verifiable output" — a worker that
-  fails after writing most of a report keeps the text too. `session_capture`
-  now carries the raw `saved_session_id` (the `metadata` receipt stays
-  fingerprint-only), `FleetReceipt.saved_session_id` persists it, and the
-  runtime API exposes it so a client can resolve the worker's full final
-  reply via `GET /v1/sessions/{id}` (#5946, thanks @gaord).
+- A Fleet task that selects a roster member with `worker.agent_profile` now
+  runs with that member's posture. The launch-time resolver only consulted the
+  resolved member when the legacy `worker.role` label was absent, so a task
+  labelled `manager` that selected `member:reviewer` ran as a write-capable
+  manager instead of a reviewer and was never leased. The member's canonical
+  slot now wins whenever one resolved; the label remains the posture only when
+  no member resolved at all. To keep the fix from widening authority in the
+  mirror case (a read-only label on a write-capable member), a spec whose
+  `worker.role` names a different posture than the selected member's role is
+  rejected at run creation with a message naming both postures; casing and
+  legacy aliases of the member's own role are still accepted (#5945, thanks
+  @gaord).
 
 ### Changed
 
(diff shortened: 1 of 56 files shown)


Hmbown (changes requested):
Reviewed at `8f76be063`. The problem is real — "worker exited successfully but produced no verifiable output" on a task that wrote a full report is a bad receipt, and it should be fixed. I'm requesting changes on the *mechanism*, not the goal: as written the excerpt is taken from the wrong end of the stream, the accumulator is unbounded, and the character count in the receipt is not the number it claims to be.

## Tests I ran

Detached worktree, `RUST_MIN_STACK=16777216`:

- `cargo test -p codewhale-tui --lib -- fleet::executor fleet::task_spec` — **38 passed; 0 failed**
- `cargo test -p codewhale-tui --lib -- fleet_receipt` — **6 passed; 0 failed**
- `cargo test -p codewhale-protocol --lib -- fleet` — **24 passed; 0 failed**

All green. Everything below is from reading, not from a failing test.

## 1. The excerpt is the beginning of the run, not the deliverable

`observe_worker_stream_line` accumulates **every** `content` event (executor.rs, new block), and `bounded_worker_summary` then keeps the **first** 4,000 chars. For a report task the first 4,000 characters of the stream are the model thinking out loud on its way to the answer; the deliverable is at the end. So the receipt will usually show the opening of the run and call it "the deliverable".

The exec side already has the right value. In `exec_agent.rs` the terminal `metadata` event carries `visible_final_answer_chars: summary.output.chars().count()` — `summary.output` **is** the final assistant reply, already computed, already bounded by the output ceiling. Emitting a bounded excerpt of `summary.output` from exec (or the char count plus a bounded tail) would give the executor the real deliverable with no accumulation at all, and would delete the accumulator, `bounded_worker_summary`, and the double parse below in one move.

## 2. The accumulator is unbounded

```rust
Some("content") => {
    if let Some(content) = value.get("content").and_then(Value::as_str) {
        answer.push_str(content);
    }
}
```

No cap. `WorkerStream::answer` grows to the full size of the worker's streamed text — megabytes for a long run — and is held for the process's lifetime, per worker, across the whole fleet, so that 4,000 characters of it can be used at terminal. If you keep the accumulate-in-executor approach, stop appending once `answer.len()` passes the bound you actually need (plus a `truncated: bool` so the ellipsis stays honest).

## 3. The character count in the receipt note is wrong for exactly the cases it matters

`task_spec.rs`, new arm:

```rust
"worker produced {} characters of deliverable: {}",
summary.chars().count(),
bounded_receipt_excerpt(summary),
```

`input.summary` has already been through `bounded_worker_summary`, which truncates to 4,000 and appends `"..."`. So every deliverable longer than 4,000 characters reports **"worker produced 4003 characters of deliverable"** — a fixed number presented as a measurement. This repo is unusually careful about receipts being true; a count that silently saturates is the kind of thing someone will later debug for an hour. Either carry the real pre-truncation count alongside the excerpt, or drop the count and just show the excerpt.

## 4. Second copy of the bound-and-redact helper

`bounded_worker_summary` (executor.rs) and `bounded_receipt_excerpt` (task_spec.rs) are the same function twice: `redact_secrets`, take N chars, append `"..."`. The text is consequently redacted twice on the way to the note. One helper, one call site, one cap.

## 5. Every worker stream line is now JSON-parsed twice

`observe_worker_stream_line` does `serde_json::from_str::<Value>(line)` and matches on `value["type"]`, then immediately calls `map_exec_stream_line(line)` (executor.rs:366) which does `serde_json::from_str` and matches on `value["type"]` again. That is a second copy of the parse *and* of the dispatch, on the hot path for every line of every worker. Parse once and pass the `Value` down, or fold the two new arms into `map_exec_stream_line` — `"content"` is already a match arm there.

## 6. `session_id` now means two different things in the same event schema

After this PR the exec stream emits both of these:

- `metadata.session_id` — still `exec_stream_session_ref(id)`, i.e. the **redacted fingerprint** (`exec_agent.rs`, `ExecStreamMeta`)
- `session_capture.session_id` — the **raw recoverable id** (this PR)

Same field name, opposite meaning, same stream. And `metadata.resume_command` still renders the literal `codewhale exec --resume <redacted-session-id>` (`lib.rs:12202`), which is now a redaction that protects nothing while remaining useless to the caller. Please pick one: either name the new field something that cannot be confused with the fingerprint (`saved_session_id`), or make the three surfaces consistent.

Related: the PR deletes `assert!(!capture_json.contains(raw_session_id))`, which was the guard on a deliberate invariant. I don't think the id is secret — text mode already prints `session: {truncate_id(id)}` and `load_session_by_prefix` resolves it — so I'm not calling this a leak. But it is an intentional invariant being retired, and that deserves a line in the PR body rather than a deleted assertion.

## 7. Docs

`docs/AGENT_RUNTIME.md:144-148` documents the `codewhale exec --output-format stream-json` event vocabulary, and `docs/zh_hans/AGENT_RUNTIME.md:76` mirrors it. Both should mention that `session_capture` now carries `session_id`, since that is the field an external caller has to know about to resolve `GET /v1/sessions/{id}`. Right now the feature is undiscoverable outside this diff.

## Smaller things

- The summary is attached only to `Completed`. A worker that fails after writing most of its report loses the text entirely — arguably that is when you most want it.
- `"worker produced {} characters of deliverable: {}"` is a new user-visible string on a surface this PR touches, added as a raw literal. `crates/tui/locales/AGENTS.md` says "new and touched surfaces use typed `MessageId` keys". The neighbouring notes in `verify_task_result` are all raw English too, so this is consistent with its surroundings rather than a regression — flagging it so the decision is deliberate rather than inherited.
- `partial()` sets `failure_kind: None`, so this text does not reach `FleetAlertEvent::verifier_failed` and does not get shipped to alert sinks. Good — I checked, and it means the new content stays on the receipt surface.
- **DCO**: `0d7c29e40` has no `Signed-off-by:` trailer. `Check Signed-off-by` is advisory in `.github/workflows/dco.yml` so CI stays green, but CONTRIBUTING asks for it — `git commit --amend -s`.

The saved-session half of this (persist the id, expose it on the receipt, resolve the full transcript through the existing sessions API) is the right shape and I'd take it on its own. It's the summary half that needs another pass.

Hmbown:
Maintainer follow-up pushed to `feat/fleet-worker-deliverable` (head `4de6dc17f`): the saved-session half kept as authored, the summary half reworked per review — excerpt emitted at the exec terminal receipt (`visible_final_answer_excerpt` + real `visible_final_answer_chars`), executor accumulator deleted, one bound-and-redact helper, one parse per stream line, `saved_session_id` rename everywhere, docs in both languages, and the excerpt now survives FAILED outcomes too. Original commit preserved below the follow-up with full authorship — thank you @gaord for the feature and the original implementation!

gaord:
Follow-up pushed (`f94684cd6`, current `main` merged first) addressing the Devin round-2 findings:

- **Final-reply excerpt**: `visible_final_answer_chars`/`excerpt` now derive from the last assistant-like message of the persisted session instead of the cumulative `ExecSummary::output`, so pre-tool commentary no longer leaks into the deliverable receipt. The cumulative output remains only as the fallback when the session has no assistant text. Covered by two new tests (`exec_stream_final_answer_text_is_the_last_assistant_reply` with pre-tool commentary + distinct final answer, and the empty-cases test).
- **CHANGELOG**: branch hunks dropped per merge-time policy (they were also the merge conflict; main's side kept).
- **PR description** updated to name the current data flow.

Local gates: fmt clean; `--lib -- exec_stream_final_answer` 3/0; `--lib -- fleet::executor fleet::task_spec fleet_receipt` 49/0 (one load-flake rerun, #5929 pattern); `codewhale-protocol --lib -- fleet` 24/0. The windows leg's `mcp_boot_*` failures on `d3b333b06` match the #5929 flake class — untouched code, green on main's latest run.

Issue #5950, opened 2026-09-06
feat(tui): make 0.9.12 bottom chrome (posture bar + metrics line) configurable; /statusline is effectively dead; restore always-visible context


7jrxt42BxFZo4iAnN4CX:
### Summary

In the 0.9.12 shell redesign the bottom of the TUI became two hard-coded rows under the composer (posture bar + metrics line). They are not configurable, and the legacy `/statusline` / `tui.status_items` mechanism no longer controls what these rows render — toggling items in `/statusline` (cost, git branch, metrics) has **no visible effect**.

### Evidence (0.9.12 source)

- `status_items` is read by the renderer in exactly **one** place: `tui/ui/provider_routes.rs` (`should_fetch_provider_balance` gates the balance fetch on `StatusItem::Balance`).
- The posture bar and metrics line are built independently of `status_items`:
  - posture bar ← `TidelineFooterFacts` + `phase_strip::posture_items` (`tui/phase_strip.rs`);
  - metrics line ← `info_segments` (`tui/ui/frame.rs`).
  - Neither `phase_strip.rs`, `underwater.rs`, `ui/frame.rs`, nor `infoline.rs` reference `status_items` (0 hits).
- The `/statusline` picker (`tui/views/status_picker.rs`) still persists `app.status_items`, so users are shown toggles that do nothing except the (provider-dependent) balance gate.

### Symptoms

1. **Context indicator is gone from the always-on view.** The metrics line only shows the context reading (`ctx NN%`) when fullness is **≥ 50%** (`ui/frame.rs`: "Only displayed when context fullness >= 50%"); below that it stays silent, so for most of a session there is no context indicator at all. Before the 0.9.12 shell redesign the context utilisation was always visible (it was a fixed field of the old status line, and `/statusline` used to control it). Users who relied on watching context have lost that signal without a way to bring it back.

2. **Two unreliable placeholders on a custom OpenAI-compatible provider** eat space in the metrics line:
   - `high→effective unavailable` — reasoning effort rendered as `Unavailable` because the route is `ApiProvider::Custom`/Zai/Minimax without an immutable endpoint receipt (`App::effective_reasoning_effort_for_active_route`); it cannot be proven, but is still printed.
   - `cost: unknown` — deliberately added in 0.9.12 (`#5578`: "footer now honestly shows `cost: unknown` when nothing can be priced instead of hiding the line"); for a provider with no price table this is permanent noise.

[...]

Issue #5914, opened 2026-09-06
TUI: overall working-time indicator no longer visible during long sessions


Hmbown:
Founder report from live dogfooding (2026-09-05/06, 0.9.12 TUI, an operate session running many hours with sub-agents):

"it does seem kinda weird that we don't show how long the overall thing has been working anymore? just looking at the screen"

During a multi-hour agent session the screen no longer shows an overall elapsed/working indicator, so it is hard to tell from a glance whether the top-level agent is actively working, how long the current turn has been running, or whether anything is stuck. (Turn-level activity may exist somewhere, but the founder looking directly at the screen could not find the overall timer — that is itself the bug: it is not where the user looks.)

Acceptance:
- The header/status area shows the overall session working time and the current turn elapsed time in a place a glancing user checks.
- The indicator distinguishes actively-working vs waiting-on-something (tool, sub-agent, user input) states.
- Verify against recent status-bar / footer changes on main — this worked before, so find the change that lost it and decide whether to restore or replace it.

Evidence: founder direct observation during the 2026-09-05 operate session (no screenshot captured at report time). Related UX report in the same session: sub-agents parked by the runtime show as "waiting for input" indefinitely with no visible cleanup path (#5906).

PR #5962, opened 2026-09-06
feat(tui): /statusline composes the bottom chrome, ctx reads at every fullness (#5950)

Hmbown:
No-Issue: first slice of #5950 (ctx always visible, /statusline drives the metrics line, dead toggles retired); the row presets and auto-omission of unavailable segments stay open on #5950.

The 0.9.12 shell built the two rows under the composer without ever reading
`app.status_items`, so `/statusline` was a checklist that changed nothing —
`status_items` had exactly one reader in the whole renderer
(`should_fetch_provider_balance` in `tui/ui/provider_routes.rs`). And the
metrics line's context reading was gated at 50% fullness, which is most of a
session spent with no context signal at all.

## What changed

**1. `ctx NN%` at every fullness.** `info_segments` (`tui/ui/frame.rs`) no
longer drops the reading below 50%. The warning styling keeps the thresholds
it had: `ChromeInk::Info` below 80%, `ChromeInk::Failure` from 80% up. The
reading is still the row's floor — it sheds after everything else, including
the help hint.

**2. `/statusline` drives the metrics line.** Every segment in
`info_segments` is gated on its `StatusItem`, and one item that was not a
metrics-line fact is wired where it does live:

| `StatusItem` | What the toggle now shows/hides |
| --- | --- |
| `Model` | `InfoSegmentId::Model` (route identity) |
| `ContextPercent` | `InfoSegmentId::Context` (`ctx NN%`) |
| `Cost` | `InfoSegmentId::Cost` |
| `Balance` | `InfoSegmentId::Balance` — new segment, see below |
| `Cache` | `InfoSegmentId::Cache` |
| `Tokens` | `InfoSegmentId::OutputTokens` |
| `SessionMetrics` | `InfoSegmentId::Ttft` + `InfoSegmentId::Rate` |
| `Mode` | the posture bar's plan/act/operate chip (`tideline_footer_from_app`) |

`Balance` previously only authorised a background fetch whose result nothing
in the chrome painted. It now paints too: `InfoSegmentId::Balance`, label from
the existing `MessageId::FooterBalancePrefix` (already in all 15 locale packs
— no new user-visible strings in this PR), value from
`BalanceInfo::chip_label`, shed priority 5 so it outlives the cost, which is
right for a reading that is off by default and only on because its owner asked
for it by name. The same status item still gates the fetch, so the row can
only show a number this session asked for.

The posture bar's permission chip, working clocks (#5914 / #5920) and live
counts stay unconditional — they are the bar's own posture statement, and the
shed ladder and clock order in `phase_strip.rs` are untouched (the only change
in that file is a one-line `.filter()` on the already-optional `mode_chip`).

**3. Retired toggles, with receipts.** Eight items were checkboxes that could
not change anything, so they are gone rather than left lying:

| Retired | Receipt |
| --- | --- |
| `ReasoningReplay` | zero references outside `config.rs` — never rendered by anything |
| `RateLimit`, `LastToolElapsed` | the enum's own doc said "placeholder until wired" / "(reserved)"; never wired |
| `PrefixStability` | only reference was a test asserting it stays out of the default footer |
| `GitBranch` | `infoline.rs` module doc, 2026-09-02: "the repository and branch moved to the launch header and the git bottom view" — wiring it back would reverse that call and, since it was in `default_footer`, would put a branch on everyone's metrics line by default |
| `Status` | the posture bar's turn clock carries the activity word (`working 1m 15s`), and that clock is #5914/#5920's core statement, not a composable chip |
| `Agents` | the posture bar's live counts own the agent count, and each count is also the click target that opens the dock |

Old config files keep loading: `StatusItem::from_key` returns `None` for the
retired keys and `deser_status_items` skips them with a warning, which is the
same forward/backward-compatible path that already existed. A test covers a
pre-#5950 `status_items` list surviving the upgrade.

**4. No second config key.** `tui.status_items` already existed, is already
persisted by `config_persistence::persist_status_items`, and is already loaded
in `app/init.rs`. It is reused as-is; the issue's proposed `[tui.bottom]`
schema is deliberately *not* added. `default_footer()` gains
`ContextPercent` so the reading is on out of the box.

## Not in this slice

#5950 also asks for `compact`/`hidden` presets for small windows and for
omitting `→ effective unavailable` / `cost: unknown` on providers that cannot
price a turn. Neither is here. The cost half is now reachable — a custom
provider can turn the cost segment off in `/statusline` — but "omit the
segment automatically when the value cannot be proven" is a separate
behavioural change, and the row-count presets are a layout change.

## Verification

```
cargo fmt -p codewhale-tui                                     clean
cargo check -p codewhale-tui --lib --tests                     clean
RUST_MIN_STACK=16777216 cargo test -p codewhale-tui --lib -- \
  tui::ui::frame tui::phase_strip tui::views::status_picker \
  tui::infoline localization                                   119 passed, 0 failed
RUST_MIN_STACK=16777216 cargo test -p codewhale-tui --lib -- \
  config::tests config_persistence tui::ui::tests::default_footer \
  tui::ui::tests::should_fetch tui::widgets                     708 passed, 0 failed
```

New tests: `context_reading_paints_at_every_fullness` (10% and 60%, asserted
at 40, 80 and 160 columns), `context_reading_keeps_its_warning_threshold`
(10 / 79 / 80), `statusline_toggle_removes_its_segment_on_the_next_frame`
(drives the real `StatusPickerView` with key events and re-renders),
`every_metrics_segment_answers_to_a_status_item` (all items on paints every
segment; an empty list paints none), plus a config test for a pre-#5950
`status_items` list still loading.

No golden needed re-blessing: the `infoline_*` and `footer_*` goldens render
synthetic segments and default status items, and both stayed byte-identical.
The two `one_owner_tests` assertions that encoded the 50% silence were
updated to assert the reading paints once at every fullness.

Buffer-level evidence only: no run against a real provider and no live
terminal session.

https://claude.ai/code/session_0188XYyJaw9Mh9uSrqQBoqhm

---

> [!NOTE]
> **Low Risk**
> TUI presentation and config parsing only; retired keys are ignored rather than failing load, with no auth or data-path changes.
> 
> **Overview**
> **`/statusline` and `tui.status_items` actually control the bottom chrome again.** After the 0.9.12 shell split, most picker toggles did nothing because `info_segments` ignored the list; each remaining `StatusItem` now gates exactly one on-screen element (metrics segments for model, context, cost, balance, cache, tokens, session metrics; posture bar plan/act/operate for `mode`). Eight keys that never drove UI (`status`, `agents`, `reasoning_replay`, `prefix_stability`, `git_branch`, `last_tool_elapsed`, `rate_limit`) are removed from the enum but still parse safely from old configs.
> 
> **Context and balance behavior changes on the metrics line.** `ctx NN%` is shown at every fullness (not only ≥50%), with warning styling still from 80% up; `context_percent` is added to the default footer. **Balance** gets a new metrics segment when enabled—the same toggle still authorizes the prepaid balance fetch—plus updated shed order so balance outlives cost on narrow widths.
> 
> Docs, `config.example.toml`, and tests cover legacy `status_items` lists, picker integration, and frame invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f088718f15fb4bf916034a1b9715932658c30295. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `/statusline` drives the bottom chrome again. Since the 0.9.12 shell
+  redesign the posture bar and the metrics line were built independently of
+  `tui.status_items`, so every toggle in the picker except the balance fetch
+  was decoration. Each remaining item now shows or hides exactly one thing:
+  `model`, `context_percent`, `cost`, `balance`, `cache`, `tokens` and
+  `session_metrics` are metrics-line segments, and `mode` is the posture
+  bar's plan/act/operate chip. The `status`, `agents`, `reasoning_replay`,
+  `prefix_stability`, `git_branch`, `last_tool_elapsed` and `rate_limit`
+  items drove nothing and are retired; an existing `config.toml` still loads
+  and those keys are ignored (#5950).
+- The context reading is back on screen at every fullness. 0.9.12 painted
+  `ctx NN%` only from 50% up, which left most of a session with no context
+  signal at all; it now paints from 0% and keeps its warning colour from 80%
+  up (#5950).
+
 ### Added
 
 - The `/theme` picker now discovers valid user-authored `custom:<name>`
--- a/config.example.toml
+++ b/config.example.toml
@@ -1087,14 +1087,18 @@ turn_wall_clock_secs = 3600  # cumulative wall clock for one turn, excluding tim
 stream_max_content_mb = 10   # per-step cap on streamed content (0 = default, 1-512)
 stream_max_duration_secs = 1800 # per-step cap on one stream's duration (0 = default, 10-86400)
 osc8_links = true            # emit OSC 8 escapes around URLs (Cmd+click in iTerm2/Ghostty/Kitty/WezTerm/Terminal.app 13+); set false for terminals that misrender
-# Ordered footer chips shown in the TUI status line. Omit the key to use the
-# built-in default; set [] to hide all configurable chips. You can also edit
-# this interactively with `/statusline`.
-# Supported keys: mode, model, cost, balance (DeepSeek / DeepSeekCN only),
-# status, agents,
-# reasoning_replay, prefix_stability, cache, context_percent, git_branch,
-# last_tool_elapsed (reserved), rate_limit (reserved), tokens.
-# status_items = ["mode", "model", "status", "git_branch", "tokens", "cache"]
+# What the bottom chrome shows. Each key is one thing on screen: `mode` is the
+# posture bar's plan/act/operate chip, everything else is a segment of the
+# metrics line under it. Omit the key to use the built-in default; set [] to
+# strip the metrics line down to the help hint. You can also edit this
+# interactively with `/statusline`.
+# Supported keys: mode, model, context_percent, cost,
+# balance (prepaid providers only: DeepSeek, DeepSeekCN, OpenRouter,
+# SiliconFlow), cache, tokens, session_metrics.
+# Retired in 0.9.13: status, agents, reasoning_replay, prefix_stability,
+# git_branch, last_tool_elapsed, rate_limit — they drove nothing. Old files
+# keep loading; the retired keys are ignored.
+# status_items = ["mode", "model", "context_percent", "cost", "tokens"]
 # notification_condition = "unfocused" # unfocused | always | never
 #                                    "unfocused" = notify only after this terminal has been
 #                                    in the background for two seconds (default);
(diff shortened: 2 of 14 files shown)


7jrxt42BxFZo4iAnN4CX:
Nice slice — thanks for driving #5950.

Small coverage gap I noticed while reviewing: the new `mode` gate in `tideline_footer_from_app` is the one gate without test coverage. `tideline_tests.rs` builds `TidelineFooter` directly via the builder (so it bypasses the `status_items` filter entirely), and the `one_owner`/new tests all keep `status_items` at the default (Mode present) — none exercises removing `Mode` from the list to assert the chip actually disappears from the posture bar. Not a blocker; a one-line assertion (`status_items` minus `Mode` → `mode_chip: None`) would lock the behavior in.

PR #5983, opened 2026-09-07
fix(engine): never dispatch a tool call whose arguments were truncated

### LINKED ISSUE #5986: Engine: a tool call truncated at the provider's output limit is repaired into valid JSON and executed
## What happens

A tool call cut off by the provider's output limit can be silently completed into valid JSON and **executed**. For a `write`, the user's file is overwritten with a partial body while the model was still mid-argument.

## Root cause

`crates/tui/src/tools/arg_repair.rs` runs a five-stage repair ladder and returns `Ok(value)` identically regardless of which stage succeeded. Stages 1-3 (strict parse, control-char stripping, trailing commas) normalize text that was already structurally complete. Stage 4 `balance_braces` (`arg_repair.rs:114-132`) *invents* structure — and it counts `{`/`}` **without tracking string literals**, so a stream cut right after a complete string value needs exactly one appended `}` to parse.

The caller therefore cannot distinguish "the model finished writing this" from "we guessed the ending". `input_parse_error` stays `None` and dispatch sees a well-formed argument.

`output_limit_truncated` does exist (`crates/tui/src/core/engine/turn_loop.rs:1553`) but is first read at `:1888` and `:2528` — **after** dispatch at `:2648` — and the truncation notice is appended to the session *after* the tool result. Nothing on the dispatch path consults it.

## Reproduction

Argument text `{"path": "notes.md", "content": "first line"` (stream cut after a complete string value) parses to `{"path": "notes.md", "content": "first line"}` and dispatches. `notes.md` is written containing only `first line`.

## Why this matters

`AGENTS.md` lists data-loss handling as never-cut. There is no test covering it — `grep "output limit\|max_tokens" crates/tui/src/core/engine/tests.rs` finds nothing.

## Found by

Comparing against `refs/piagent`, whose agent loop branches on `stopReason === "length"` **before** executing anything, with a comment naming this exact defect class. Recorded in `codewhale-ops/FINDINGS-20260907-REFS-STUDY-AND-DEFECTS.md` §1.1.

Hmbown:
Closes #5986

## The defect

A tool call cut off by the provider's output limit could be silently completed into valid JSON and **executed**. For a `write`, that means the user's file is overwritten with a partial body while the model was still mid-argument.

`arg_repair::repair` reported success identically whether stage 1 parsed the text as-is or stage 4 *invented* the structure. `balance_braces` counts `{`/`}` without tracking string literals, so a stream cut right after a complete string value needs exactly one appended `}` to parse — `input_parse_error` stays `None` and dispatch sees a well-formed argument.

`output_limit_truncated` does exist (`turn_loop.rs:1553`) but is first read at `:1888`/`:2528` — **after** dispatch at `:2648` — so nothing on the dispatch path consulted it.

## The fix

`repair` now returns `Repaired { value, structure_synthesized }`. Stages 1-3 (strict parse, control-char stripping, trailing commas) are **intact** — they normalize text that was already structurally complete. Stages 4-5 (appending/discarding closers) are **synthesized** — they only fire on incomplete text.

`final_tool_input` and the `ContentBlockStop` handler route a synthesized parse into the **existing** `malformed_tool_arguments_input` path, so the model is asked to re-issue. No new gate and no new error path: the mechanism for "these arguments are unusable" already existed, truncation just wasn't reaching it.

The mid-stream delta parse deliberately ignores the flag (a partial buffer is *expected* to be incomplete there) and now says so in a comment.

## Evidence

```
cargo fmt -p codewhale-tui                  -> clean
cargo clippy -p codewhale-tui --lib         -> 0 errors
cargo test -p codewhale-tui --lib -- arg_repair final_tool_scenario
  -> test result: ok. 16 passed; 0 failed
```

The regression test **fails without the fix** — verified by removing the guard and re-running:

```
assertion `left == right` failed: a truncated write must not be dispatched as a completed argument
  left:  {"path": "notes.md", "content": "first line"}
  right: {"raw_arguments": "{\"path\": \"notes.md\", \"content\": \"first line\""}
```

Found by comparing against `refs/piagent`, whose agent loop branches on `stopReason === "length"` *before* executing anything. Recorded in `codewhale-ops/FINDINGS-20260907-REFS-STUDY-AND-DEFECTS.md` §1.1.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5983" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

---

> [!NOTE]
> **High Risk**
> Changes tool dispatch and file-mutating paths when streams truncate; wrong handling could still execute partial writes or break legitimate sloppy JSON (trailing commas).
> 
> **Overview**
> Fixes **#5986**: when a provider hits an output limit mid tool-argument stream, `arg_repair` could append missing `}` and produce valid JSON that was still **executed** (e.g. partial `write` content).
> 
> **`arg_repair::repair`** now returns `Repaired { value, structure_synthesized }`. Stages 1–3 (strict parse, control chars, trailing commas) stay **intact**; stages 4–5 (balance/strip closers) set **`structure_synthesized`**.
> 
> **Dispatch finalization** treats synthesized parses like malformed arguments: `final_tool_input` and new **`finalize_streamed_tool_input`** reject them via the existing `malformed_tool_arguments_*` path. Mid-stream delta mirroring still uses repaired values but **ignores** the flag.
> 
> **Stream end without `ContentBlockStop`** (truncated turn) now runs the same finalization + `ToolCallStarted` for in-flight tool blocks, so truncated calls never slip through on `tool.input` alone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35a6e7aafdf28f43952ea42c60bb88bda012817a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/tui/src/core/engine/dispatch.rs
+++ b/crates/tui/src/core/engine/dispatch.rs
@@ -274,35 +274,64 @@ pub(super) fn final_tool_input(state: &ToolUseState) -> serde_json::Value {
     if !state.input_buffer.trim().is_empty()
         && let Some(parsed) = parse_tool_input(&state.input_buffer)
     {
-        return parsed;
+        // Structure was synthesized to make this parse, so the argument text
+        // was cut off. Route it to the same malformed-arguments path as an
+        // outright parse failure rather than dispatching a completed guess.
+        if parsed.structure_synthesized {
+            return malformed_tool_arguments_input(&state.input_buffer);
+        }
+        return parsed.value;
     }
     state.input.clone()
 }
 
-pub(super) fn parse_tool_input(buffer: &str) -> Option<serde_json::Value> {
+/// A parsed tool-argument buffer, plus whether the parse only succeeded
+/// because the repair ladder synthesized structure (see
+/// `crate::tools::arg_repair`). Mid-stream callers mirroring partial state
+/// may ignore the flag; the caller making the final dispatch decision must
+/// not, because synthesized structure means the argument text was cut off.
+pub(super) struct ParsedToolInput {
+    pub(super) value: serde_json::Value,
+    pub(super) structure_synthesized: bool,
+}
+
+pub(super) fn parse_tool_input(buffer: &str) -> Option<ParsedToolInput> {
     let trimmed = buffer.trim();
     if trimmed.is_empty() {
         return None;
     }
     // Try the deterministic arg-repair ladder first (handles trailing commas,
     // unclosed braces, embedded control chars, etc.)
-    if let Ok(value) = crate::tools::arg_repair::repair(trimmed) {
-        return Some(value);
+    if let Ok(repaired) = crate::tools::arg_repair::repair(trimmed) {
+        return Some(ParsedToolInput {
+            value: repaired.value,
+            structure_synthesized: repaired.structure_synthesized,
+        });
     }
     // Fall back to existing strategies for code-fenced, double-encoded, and
     // segment-extraction patterns that the repair ladder doesn't cover.
     if let Some(stripped) = strip_code_fences(trimmed)
         && let Ok(value) = serde_json::from_str::<serde_json::Value>(&stripped)
     {
-        return Some(value);
+        return Some(ParsedToolInput {
+            value,
+            structure_synthesized: false,
+        });
     }
     if let Ok(serde_json::Value::String(inner)) = serde_json::from_str::<serde_json::Value>(trimmed)
         && let Ok(value) = serde_json::from_str::<serde_json::Value>(&inner)
     {
-        return Some(value);
+        return Some(ParsedToolInput {
+            value,
+            structure_synthesized: false,
+        });
     }
     extract_json_segment(trimmed)
         .and_then(|segment| serde_json::from_str::<serde_json::Value>(&segment).ok())
+        .map(|value| ParsedToolInput {
+            value,
+            structure_synthesized: false,
+        })
 }
 
 /// Decode a JSON container that a provider encoded as a string when the tool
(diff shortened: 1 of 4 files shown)


Hmbown:
Addressed the review finding on the ContentBlockStop state transition — and in verifying it, a second, worse path turned up.

**The flagged path** ( ContentBlockStop): synthesized parses were dropped from `final_tool_input` announcements but the state transition depended on where dispatch read its input. Dispatch reads `tool.input` directly, and for a synthesized buffer the parse-error marking only happened in the `final_tool_input` path — so the drop left the mid-stream mirror's value in place.

**The worse path found while tracing it:** a provider cutting the stream at its output limit omits the closing `ContentBlockStop` entirely. Those blocks never ran any final gate at all — the mid-stream mirror (which deliberately ignores `structure_synthesized`, since partial text is normal mid-stream) was the last writer of `tool.input`, and dispatch executed the truncated call.

`35a6e7a` extracts the finalization into `finalize_streamed_tool_input` and runs it for every block that never stopped, after the stream loop, with the same finalized `ToolCallStarted` announcement. New regression test drives a real turn cut mid-argument (`max_tokens`, no block stop) and asserts the malformed-arguments error instead of execution; a control test proves an omitted stop with complete arguments still dispatches. The regression test was verified failing on the previous head.

Evidence: `cargo nextest run -p codewhale-tui --lib -E 'test(core::engine::)'` → **508 passed, 0 failed**; new + arg_repair + final_tool_scenario slice → 18 passed; `cargo fmt -- --check` clean.

Issue #5988, opened 2026-09-07
Tests: two codewhale-tui tests overflow the 2MiB libtest thread stack; nextest's per-process isolation hides it from CI


Hmbown:
## What happens

```
cargo test -p codewhale-tui --lib -- mcp
...
test commands::groups::plugins::tests::mcp_review_discloses_host_authority_and_names_without_secret_values ...
fatal runtime error: stack overflow, aborting
error: test failed
  process didn't exit successfully: signal: 6, SIGABRT
```

Isolated with `--test-threads=1`: the abort happens while that test is running, so it is that test and not a neighbour.

## It is pre-existing, not a regression from in-flight work

Reproduced on clean `origin/main` (`3f3aa9ed7`) after removing an unrelated local change and rebuilding. Found while verifying #5974's fix; explicitly re-checked against a clean tree so it would not be misattributed.

## Why it matters

The abort kills the whole test binary, so **every test scheduled after it in that process is silently not run**. A filter that happens to include this test reports a process failure rather than a test failure, and a filter that excludes it looks green. That makes local `-- mcp` runs unreliable as evidence, and it is the kind of thing that hides other failures.

Worth checking whether CI's `cargo nextest run` masks this — nextest runs each test in its own process, so an abort there would be attributed to the single test and the rest of the suite would still run. If so, this reproduces locally but not in CI, which is the worst shape for a defect like this.

## Likely shape

A stack overflow in a test named for disclosure of host authority and names suggests unbounded recursion in a formatter/serializer walking a self-referential structure — likely a plugin/MCP review value that can contain itself, or a `Display`/`Debug` impl that recurses. Not investigated further.

Hmbown:
## Root cause found, and it is not specific to this test

**It is the test-thread stack size, and there is at least one more test with the same defect.**

`libtest` runs each test on a spawned thread whose stack defaults to 2 MiB (`RUST_MIN_STACK`), while a per-process runner gives the test the main thread's 8 MiB. Both known offenders pass at 8 MiB and abort at the default:

```
$ cargo test -p codewhale-tui --lib -- mcp_review_discloses_host_authority
fatal runtime error: stack overflow, aborting

$ RUST_MIN_STACK=8388608 cargo test -p codewhale-tui --lib -- mcp_review_discloses_host_authority
test result: ok. 1 passed; 0 failed
```

Second test with identical behaviour, found while verifying an unrelated fix:

```
[...]

Issue #5575, opened 2026-08-23
Fleet/subagent role posture has no single source of truth (the verifier contradiction was a symptom)


Hmbown:
## Problem
#5562's Bug 3 (verifier role contradicts its own description) was NOT a one-off typo: role posture is defined and interpreted independently in at least five places, and they have drifted more than once:

- `worker_profile.rs::for_role` — per-role write/shell/model default (Verifier = write-denied + Full shell ceiling, with a comment that says 'runs the test suite').
- `tools/workflow.rs::read_only_allowed_tools` — the dispatch-side allow-list for the same role (Verifier => File only).
- `fleet_setup.rs::ROLES` — the roster creation descriptions users read.
- `docs/SUBAGENTS.md` posture table + role bullets ("run tests / validation", examples that say 'run cargo test --workspace').
- `tools/subagent VERIFIER_AGENT_INTRO` — the injected role brief.
- Plus the runtime_permissions snapshot (shell="full" display) which audits as lying about the enforced envelope (C037).

The 0.9.12 fix made the DESCRIPTION truthful (bounded validation, no writes, unbounded shell refused) and the profile comment honest, but the architecture still allows drift: a role change in one table silently mis-describes the others.

## Ask (design)
One role-posture table as the source of truth (write ceiling, shell policy, explicit tool allow-list, description/intro text) consumed by: runtime profile, dispatch policy, roster UI, docs generation, and the permission snapshot. Any per-role change lands in exactly one place, and a guard test fails if a surface derives a posture the table does not describe.

## Reference
- #5562 (user report), C037 (audit), IMPROVEMENT-PLAN-0912 'Hold these lines' (honest chrome).
- 0.9.12 branch: description/intro/doc aligned at 360bff053; runtime profile deliberately unchanged (Full shell ceiling clamped by ChildAuthority + bounded verification seam).

Hmbown:
## Design note: one authority for delegated child tool calls

**Status:** design + inventory for v0.9.12. Build the regression wall before removing enforcement. The consolidation itself is next-cycle work unless scope is explicitly expanded.

### Finding

`crates/tui/src/tools/execution_envelope.rs` describes the capability-derived envelope as the centralized child-execution boundary. In the live dispatch path it is currently gate 9 of 11 child-specific gate clusters:

| Order | Current gate | Evidence |
|---:|---|---|
[...]

PR #5989, opened 2026-09-07
fix: five recorded defects — queue data loss, ACP schema, metrics path, fleet roles, goal-loop bound

### LINKED ISSUE #5969: `serve --acp` `initialize` response violates the ACP schema: `sessionCapabilities.list` is a boolean instead of an object — strict clients (JetBrains IDEA) fail to connect
# `serve --acp` `initialize` response violates the ACP schema: `sessionCapabilities.list` is a boolean instead of an object — strict clients (JetBrains IDEA) fail to connect

**Repo target:** `Hmbown/CodeWhale`

---

## Environment

- **Package / version:** `codewhale` npm wrapper 0.9.12, native binary `codewhale` 0.9.12 (Linux x64)
- **Command under test:** `codewhale serve --acp` (ACP v1 server over stdio, newline-delimited JSON-RPC 2.0)
- **Client:** JetBrains IntelliJ IDEA 2026.2 built-in ACP client (`com.intellij.platform.acp`, `ml-llm` AI Assistant), registered via `~/.jetbrains/acp.json`:

```json
{
  "codewhale": {
    "command": "/home/lujc/.npm-global/bin/codewhale",
    "args": ["serve", "--acp"]
  }
}
```

## Summary

Since 0.9.12, the `initialize` response of `codewhale serve --acp` advertises session capabilities as boolean literals:

```json
"sessionCapabilities": { "list": true, "load": true }
```

Per the [ACP schema](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/schema/schema.json), `SessionCapabilities.list` is typed as `SessionListCapabilities | null` — an **object** ("By supplying `{}` it means that the agent supports listing of sessions"), never a boolean. ACP clients with strictly-typed deserializers — notably JetBrains IDEA's built-in ACP client (Kotlin `agent-client-protocol` model, kotlinx.serialization) — throw a `JsonDecodingException` while parsing the handshake, treat the failure as non-recoverable, and kill the agent process. **Result: codewhale cannot be used from IDEA at all in 0.9.12.** The same client connects fine to 0.9.11 and to other ACP agents.

## Reproduction

```bash
printf '%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{}}}' \
  | timeout 15 codewhale serve --acp
```

### Actual behavior

The `initialize` response contains boolean capability values:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "protocolVersion": 1,
    "agentCapabilities": {
      "loadSession": true,
      "modelSelection": true,
      "promptCapabilities": { "image": false, "audio": false, "embeddedContext": true },
      "mcpCapabilities": { "http": false, "sse": false },
      "sessionCapabilities": { "list": true, "load": true }
    },
    "agentInfo": { "name": "codewhale", "title": "codewhale", "version": "0.9.12" },
    "authMethods": [ /* ... */ ]
  }
}
```

Note two schema deviations in `sessionCapabilities`:

1. **`"list": true`** — must be a `SessionListCapabilities` **object** (`{}` means "supported", absent/`null` means "not supported"). This is the one that breaks strict clients.
2. **`"load": true`** — not a defined property of `SessionCapabilities` in the current schema (`session/load` support is advertised via the top-level `loadSession` boolean, which codewhale already sets correctly). Currently harmless for lenient clients, but should be removed.

### Client-side failure (JetBrains IDEA 2026.2, `idea.log`)

```
WARN - AcpSessionLifecycleManager - Failed to initialize ACP session for chat 01fb1342-..., error:
kotlinx.serialization.json.internal.JsonDecodingException: Expected JsonObject, but had JsonLiteral
as the serialized body of com.agentclientprotocol.model.SessionListCapabilities at element: $.list
JSON input: true
	at com.agentclientprotocol.model.SessionListCapabilities$$serializer.deserialize(Capabilities.kt:130)
	at com.agentclientprotocol.model.SessionCapabilities$$serializer.deserialize(Capabilities.kt:183)
	at com.agentclientprotocol.model.AgentCapabilities$$serializer.deserialize(Capabilities.kt:248)
	at com.agentclientprotocol.model.InitializeResponse$$serializer.deserialize(Requests.kt:555)
	...

AcpSessionLifecycleManager - handleSessionConnectionFailed for agent=acp.codewhale, recoverable=false,
message=Failed to initialize ACP session. Error: Expected JsonObject, but had JsonLiteral ... at element: $.list
AcpServerProcessHandlerImpl - [codewhale] Sending SIGTERM (graceful shutdown)...
```

The client marks the failure `recoverable=false` and terminates the agent process ~60 ms after spawning it, so the user just sees a connection error in the IDE.

## Expected behavior

Advertise the capability with the schema-defined object shape:

```json
"sessionCapabilities": { "list": {} }
```

(or omit `list` entirely if `session/list` is not supported).

### Reference: working agents, same client, same request

Both connect successfully in IDEA 2026.2:

- **Kimi Code CLI 0.41.0** (`kimi acp`): `"sessionCapabilities":{"list":{},"resume":{},"close":{},"delete":{},"fork":{},"additionalDirectories":{}}`
- **Reasonix v1.38.1** (`reasonix acp`): `"sessionCapabilities":{"list":{},"resume":{},"close":{},"delete":{}}`

## Regression window

- IDEA logs from 2026-09-04 show a working codewhale ACP session (`0.9.11`; `initialize`/`session/new`/`session/prompt` all succeeded, see also `codewhale-acp-session-config-issue.md` filed against 0.9.11).
- After upgrading to `0.9.12` (2026-09-07), every connection attempt fails at the handshake. So the `sessionCapabilities` advertisement was introduced/changed in 0.9.12 with the wrong JSON shape.

## Suggested fix

In the `serve --acp` initialize handler:

1. Emit `sessionCapabilities.list` as an object: `"list": {}` (or use the schema/Rust-SDK type for `SessionListCapabilities` instead of a raw boolean).
2. Drop the non-schema `"load": true` field from `sessionCapabilities`; `loadSession: true` at the top level already covers it.
3. Consider validating the `initialize` response against the published ACP JSON schema in CI — this class of mistake is caught instantly by a schema check, and strict clients (JetBrains) break hard on it.

## Notes

- Framing: codewhale speaks newline-delimited JSON-RPC (Zed ACP style); sending `Content-Length`-headered framing only produces a `-32700` parse error line per header/body line.
- The failure is purely in handshake deserialization; the agent process itself is healthy and `codewhale --version` / interactive use are unaffected.

Hmbown:
Closes #5969

Five defects recorded with file:line evidence in `codewhale-ops/FINDINGS-20260907-REFS-STUDY-AND-DEFECTS.md`, each fixed in isolation and then adversarially audited as a composition before landing.

### 1. Concurrent sessions destroyed each other's unsent text — P0

The offline input queue used one global `checkpoints/offline_queue.json` for the whole sessions directory, and boot **cleared** it on session-id mismatch. A second instance therefore deleted the first's queued, user-authored text.

Keyed per session — the shape `session_manager.rs:1183-1198` already documents for checkpoints two hundred lines above. An existing global file is **adopted** into the current session (`write_atomic`, legacy file removed only on success), not discarded.

### 2. ACP `initialize` violated the schema, and a green test pinned it — #5969

`sessionCapabilities` emitted `{"list": true, "load": true}`. `list` must be a `SessionListCapabilities` **object**; `load` is not a field at all. Reported by @Lujc0523 — Codewhale is unusable from JetBrains because of it.

The unit test asserted the wrong shape, so it changed with the code: two field probes became a whole-object assertion, which also catches a resurrected `load` key.

### 3. `codewhale metrics` read a directory nothing has written since 2024

A private `deepseek_home()` resolved `$HOME/.deepseek` **unconditionally, with no existence check**, so the only reader of Codewhale's receipts read an empty tree and printed an all-zero rollup as truth — while `audit.rs:73` writes `~/.codewhale/audit.log` from ~24 call sites. Now primary with a *checked* legacy fallback, deleting the repo's third path-resolution policy.

### 4. Fleet role names resolved to different AUTHORITY per driver — part of #5575

`fleet_role_to_agent_type` was a second alias table `role.rs` and the docs knew nothing about:

| role string | durable Fleet | exact Fleet |
|---|---|---|
| `synthesizer` | Planner → write=false | Custom → **write=true, raw shell Full** |
| `smoke-runner` | Verifier → write=false | Custom → **write=true** |
| unknown | Worker → **write** | Custom → **write** |

`FleetRole::from_str` is now the only mapper, the second table is deleted, and an undeclared role **fails closed** to read-only `explore`.

**Privilege audit:** eight role strings traced through both drivers. The change is **monotonically narrowing** — no widening anywhere. All twelve built-in roster roles still resolve; only free-form names reach the fail-closed branch.

Note the user-visible break: an operator with a free-form role in an exact Fleet who expected write access is now refused. That is the boundary closing, loudly rather than silently.

### 5. The goal loop told the model it would be stopped, and nothing stopped it

The continuation prompt promised "repeated equivalent gap sets pause the loop", but `GoalPauseReason::NoProgress` was **never constructed anywhere**, `repeated_gap_count` had no reader, `last_gap_fingerprint` reset on resume, and `DEFAULT_MAX_GOAL_CONTINUATIONS` is 0. Implemented the promised stop rather than deleting the sentence, because the alternative leaves a real unbounded-spend path open.

## Cross-unit repairs

Three test breaks were flagged by the fixing agents in files they did not own, and are fixed here: the persistence-actor queue path, and two fleet posture assertions that **encoded the old privilege grant**. The second workflow failure is the fix working — a member declaring `permissions = "read_only"` can no longer have an undeclared role name hand its child a write scope.

## Evidence

```
cargo fmt                                          -> clean
cargo clippy -p tui -p cli -p config --lib         -> 0 errors
cargo test -p codewhale-tui --lib -- <the above>   -> 79 passed; 0 failed
cargo test -p codewhale-cli --lib -- metrics ...   -> 18 passed; 0 failed
```

Run with `RUST_MIN_STACK=8388608` to match nextest's per-process stack — see **#5988**, two pre-existing tests overflow the 2 MiB libtest thread stack and abort the binary locally while passing in CI. Not caused by this PR; reproduced on clean `origin/main`.

Test counts went **up** in every changed file; no assertion was weakened. The audit verified each regression test fails without its fix.

## Not done

Docs still describe the single global queue file (`ARCHITECTURE.md:239`, `OPERATIONS_RUNBOOK.md:41`, `CONFIGURATION.md:2885` + zh_hans mirror) and `FLEET.md:534` does not yet say undeclared roles fail closed. A follow-up should thread a session id through `PersistRequest::ClearOfflineQueue` so the queue-owner mutex can be deleted.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5989" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Changes span session persistence and Fleet write authority (privilege narrowing with a user-visible break for undeclared exact-Fleet roles), plus ACP wire format; fixes are well-tested but touch security-adjacent posture mapping.
> 
> **Overview**
> This PR bundles five defect fixes that were audited together: **offline queue data loss**, **ACP handshake schema**, **metrics reading the wrong state root**, **split Fleet role authority**, and **goal loops that never paused on repeated verifier gaps**.
> 
> **Offline input** moves from one shared `checkpoints/offline_queue.json` to per-session `checkpoints/<session_id>.offline_queue.json`, with legacy adoption, persistence-actor coalescing keyed by session, startup restore/clear scoped to the current session, and refusing to park queues without a session id.
> 
> **ACP `initialize`** advertises `sessionCapabilities` as `{"list": {}}` instead of boolean `list`/`load`, matching strictly typed clients (e.g. JetBrains, #5969).
> 
> **`codewhale metrics`** resolves audit log, sessions, and runtime events via shared `codewhale_config` resolvers (canonical `~/.codewhale` with existence-checked legacy fallback); unresolvable home is an error instead of a silent all-zero rollup.
> 
> **Fleet** deletes the duplicate `fleet_role_to_agent_type` table and routes both drivers through `runtime_role_for_member` (#5575): documented aliases via `member_role_alias`, undeclared role names **fail closed** to read-only `explore`/`Scout` while preserving the operator’s label on receipts; explicit `custom` remains the write-capable escape hatch.
> 
> **Operate goals** implement the promised stall stop: after `MAX_REPEATED_GAP_PASSES` (3) consecutive continuation passes with the same normalized gap set, the goal pauses with `NoProgress`; counting advances at most once per pass, not per tool call in one turn.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bd62b4ed5ce548fd0db75cc3d3a97bd68a5daec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/cli/src/metrics.rs
+++ b/crates/cli/src/metrics.rs
@@ -1,10 +1,15 @@
 //! `codewhale metrics` — reads the audit log and session/task stores and prints
 //! a human-readable usage rollup.
 //!
-//! Data sources:
-//! - `~/.deepseek/audit.log`   — one JSON line per event (approvals, credentials)
-//! - `~/.deepseek/sessions/`   — saved session JSON files (tool call history)
-//! - `~/.deepseek/tasks/runtime/events/` — runtime thread JSONL event streams
+//! Data sources, all resolved through the shared Codewhale state resolvers so
+//! the reader lands on the same files the writers use:
+//! - `~/.codewhale/audit.log`   — one JSON line per event (approvals, credentials)
+//! - `~/.codewhale/sessions/`   — saved session JSON files (tool call history)
+//! - `~/.codewhale/tasks/runtime/events/` — runtime thread JSONL event streams
+//!
+//! An install that never migrated off the DeepSeek-era `~/.deepseek` root still
+//! reads there, but only for a path that actually exists — see
+//! `resolve_state_file`.
 
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
@@ -27,17 +32,22 @@ pub struct MetricsArgs {
 }
 
 pub fn run(args: MetricsArgs) -> Result<()> {
-    let base = deepseek_home();
+    // `resolve_state_dir` is the shared read-path resolver already used by
+    // `doctor` and the session store; the runtime thread store hangs its event
+    // streams off `<tasks>/runtime`. Resolving the home is fallible, and a
+    // rollup of zeros is indistinguishable from real emptiness, so a home we
+    // cannot resolve is an error rather than a silent all-zero report.
+    let audit_log = resolve_state_file("audit.log")?;
+    let sessions = codewhale_config::resolve_state_dir("sessions")?;
+    let runtime_events = codewhale_config::resolve_state_dir("tasks")?
+        .join("runtime")
+        .join("events");
 
     // Collect data from every source; treat missing files as empty.
     let mut rollup = Rollup::default();
-    read_audit_log(&base.join("audit.log"), args.since, &mut rollup);
-    read_session_files(&base.join("sessions"), args.since, &mut rollup);
-    read_runtime_events(
-        &base.join("tasks").join("runtime").join("events"),
-        args.since,
-        &mut rollup,
-    );
+    read_audit_log(&audit_log, args.since, &mut rollup);
+    read_session_files(&sessions, args.since, &mut rollup);
+    read_runtime_events(&runtime_events, args.since, &mut rollup);
 
     if args.json {
         print_json(&rollup)?;
@@ -822,16 +832,26 @@ fn print_human(rollup: &Rollup) {
 // Helpers
 // ──────────────────────────────────────────────────────────────────────────────
 
-fn deepseek_home() -> PathBuf {
-    // This reader preserves the legacy DEEPSEEK_HOME/default-root precedence,
-    // but delegates every environment and platform-home decision to the shared
-    // runtime path authority.
-    codewhale_paths::codewhale_home_override()
-        .ok()
-        .flatten()
-        .or_else(codewhale_paths::legacy_deepseek_home_override)
-        .or_else(codewhale_paths::legacy_deepseek_home)
-        .unwrap_or_else(|| PathBuf::from(codewhale_paths::LEGACY_APP_DIR))
+/// Resolve a file that lives directly in the state root, preferring the
+/// canonical Codewhale root.
+///
+/// This is the file-shaped twin of `codewhale_config::resolve_state_dir` (and
+/// of `default_config_path`, which resolves `config.toml` the same way): the
+/// primary path wins whenever it exists, the legacy DeepSeek path is used only
+/// when it is the *only* one present, and with neither present the primary is
+/// returned so an empty rollup names the canonical location. An explicit
+/// `CODEWHALE_HOME` is an isolation boundary and never falls back.
+///
+/// The two roots are never unioned. `ensure_state_dir` may migrate legacy state
+/// by *copying* it (`StateMigrationKind::Copied` leaves the legacy tree in
+/// place), so summing both roots would double-count every migrated record.
+fn resolve_state_file(name: &str) -> Result<PathBuf> {
+    let primary = codewhale_config::codewhale_home()?.join(name);
+    if codewhale_config::codewhale_home_is_explicit() || primary.exists() {
+        return Ok(primary);
+    }
+    let legacy = codewhale_config::legacy_deepseek_home()?.join(name);
+    Ok(if legacy.exists() { legacy } else { primary })
 }
 
 /// Parse a timestamp from a JSON value field (tries RFC3339).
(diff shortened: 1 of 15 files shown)


Hmbown:
Both remaining review findings are addressed or shown unreachable:

**Offline-queue suffix check hiding a legitimate checkpoint** — unreachable: `validated_session_id` restricts session ids to `[A-Za-z0-9_-]` (no dots), and every checkpoint file is written through `validated_checkpoint_path`, so a real checkpoint can never end in `.offline_queue.json`. Files with dots in the stem in `checkpoints/` are either parked queues or foreign files, and skipping both is the intended behavior.

**Legacy adoption surfacing spurious NotFound in a same-session race** — real, fixed in `7bd62b4`: two instances of the same session can both adopt the legacy file, and the loser of the retire race now treats the vanished file as success, mirroring the tolerance `clear_offline_queue_state_for` already had.

Evidence: `cargo nextest run -p codewhale-tui --lib -E 'test(session_manager::)'` → 69 passed, 0 failed on the new head.

PR #5990, opened 2026-09-07
fix(test): derive the prune cut from recorded timestamps, not a fixed 6s

Hmbown:
No-Issue: windows-only test flake found while verifying an unrelated PR; sibling of the flake work in #5929/#5980 but not listed there.

`snapshot::repo::tests::prune_older_than_keeps_the_newest_and_drops_only_the_old_tail` fails intermittently on `windows-latest`:

```
assertion `left == right` failed: only the old tail should be removed
  left: 3
 right: 2
```

## Why

The fixture builds two old snapshots, sleeps 8s, then two new ones 1.1s apart, and cuts at a **hardcoded 6s**. That assumes `repo.snapshot()` is fast — `new:0` is only ~1.2s **plus one git subprocess** older than prune time, so on a loaded Windows runner that subprocess alone carries it past the 6s line and it gets pruned with the old pair.

The existing fixture guard could not catch this: it asserts on `before[0]` and `before[2]`, and **`before[1]` is the entry that drifts**.

## The fix

The cut is now derived from the timestamps the repo actually recorded — aim at the midpoint of the gap between the oldest survivor and the newest victim, leaving ~4s of slack in both directions instead of depending on wall-clock luck. The gap itself is asserted first, so a collapsed fixture says so plainly rather than failing later as a confusing count mismatch.

Behaviour under test is unchanged: two removed, `new:1` and `new:0` survive. **No production code is touched.**

```
cargo clippy -p codewhale-tui --lib          -> 0 errors
cargo test -p codewhale-tui --lib -- prune_older_than
  -> test result: ok. 3 passed; 0 failed
```

Found when it failed the windows leg of #5987 — a PR containing **zero Rust files** (TypeScript, CI config and `.gitignore` only), so it cannot have been caused there.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/5990" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Low Risk**
> Only adjusts a unit test fixture; snapshot pruning behavior in production is untouched.
> 
> **Overview**
> Fixes intermittent failures of **`prune_older_than_keeps_the_newest_and_drops_only_the_old_tail`** on slow Windows CI, where a **hardcoded 6s** `prune_older_than` cut could prune **`new:0`** along with the old pair when `repo.snapshot()` (git subprocess) ran long enough.
> 
> The test now **computes `max_age` from listed commit timestamps**: it checks the gap between **`before[1]`** and **`before[2]`** (the boundary between new and old pairs), uses the **midpoint** of that gap as the cutoff, then prunes with that derived age. The old wall-clock guard on **`before[0]`** / **`before[2]`** is removed because it did not catch drift on **`before[1]`**. Expected behavior is unchanged: two removals, **`new:1`** and **`new:0`** survive.
> 
> **Test-only** — no production snapshot/prune logic changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e95daf39ee4e1e07d3a3aae95d1431f8e8a39ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hmbown:
The fix is already incorporated on public main: the [prune fixture](https://github.com/Hmbown/Codewhale/blob/c200510e4b61524c997d65530114fe05365eac34/crates/tui/src/snapshot/repo.rs#L1496) derives its cutoff from the recorded survivor/victim timestamps and validates the gap. Closing this duplicate PR as incorporated; no fixed six-second cutoff is being restored.

PR #6052, opened 2026-09-11
fix(tui): align model-facing docs and tool gating

h3c-hexin:
## Summary

Five verified audit fixes for model-facing documentation and tool gating:

1. **finance bypassed the network policy** — the tool declares the `Network` capability but `execute` never consulted `context.network_policy`, so a `network.default = "deny"` session still reached `query1.finance.yahoo.com`. Both endpoint hosts (quote and chart) are now checked up front via `NetworkPolicyDecider::evaluate(host, "finance")` in the exact shape `web_search`/`speech` use: `Deny` and an undecided `Prompt` fail closed with the shared permission-error wording (including the `/network allow <host>` hint); no attached policy falls through permissively for back-compat. The chart fallback can no longer leak a request past a tightened session.
2. **verifier taught a retired name** — background-start metadata advertised `poll_with: ["exec_shell_wait", "task_shell_wait"]`, but `exec_shell_wait` is retired and cannot dispatch. Now `["task_shell_wait"]`, matching the shell timeout-recovery metadata precedent.
3. **notify `method = "off"` contract unpinned** — the promise that a configured `off` silences the tool is now regression-pinned: the emission is silenced before any sink write while the tool result still reports success to the model (nothing to retry), `settings()` installs the configured method alongside the gate, and the installed encoding round-trips every method variant. Known limitation (shared with the gate): install happens on the first `settings()` call, so a first-turn `notify` in a fresh process predates it.
4. **shell `background` description** now states that background tasks are not killed at `timeout_ms` and how to bound/stop them (`action=wait` timeout, `action=cancel`) — matching the dispatcher's actual behavior.
5. **fleet-manager skill** documents `codewhale fleet resume <run-id>` (orphaned-lease reconcile) and that `stop` requires `--all`, and names the concrete Runtime API evidence endpoints (`GET /v1/fleet/workers/{id}`, `.../receipts/{task}/evidence`, `.../events/replay`).

## Tests

- `finance_fails_closed_when_network_policy_denies_endpoint_host` — deny-listed host with a wiremock server asserting **zero** requests left before the policy check
- `finance_fails_closed_on_prompt_when_default_is_prompt` — undecided host fails closed with the approval hint, zero requests
- `configured_method_off_silences_the_tool_emission`, `configured_method_off_still_reports_success_to_the_model`, `settings_installs_configured_method_from_config`, `configured_method_round_trips_every_variant`
- existing verifier/notify/finance suites extended where noted

Verified on this branch: the affected suites (finance + notify + notifications + verifier + the retired-name guard) — 88 passed; `cargo fmt --all -- --check` clean.

## Credits

Original implementation by @asto18089 (co-authored); audit re-verified against the current tree before submission.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6052" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tui/assets/skills/fleet-manager/SKILL.md
+++ b/crates/tui/assets/skills/fleet-manager/SKILL.md
@@ -14,8 +14,8 @@ and leave a ledgered receipt or a safe escalation draft.
 ## Authority Boundary
 
 - Prefer typed fleet surfaces over shell spelunking: `codewhale fleet status`,
-  `inspect`, `logs`, `artifacts`, `interrupt`, `restart`, `stop`, and the
-  Runtime API endpoints.
+  `inspect`, `logs`, `artifacts`, `interrupt`, `restart`, `resume`, `stop`
+  (stop requires `--all`), and the Runtime API endpoints.
 - Do not read `.codewhale/fleet.jsonl`, host logs, or remote files directly
   unless the typed command or API is missing required evidence.
 - Do not send Slack, webhook, PagerDuty, email, or chat messages unless the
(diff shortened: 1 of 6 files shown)


Hmbown:
Thank you @h3c-hexin — the finance network-policy gap was real: a `network.default = "deny"` session could still reach Yahoo endpoints, and the fail-closed check with zero requests leaving before the policy gate is exactly right. The retired `exec_shell_wait` name in verifier metadata, the notify `off` contract pin, and the fleet-manager skill corrections all check out. Harvested into the 0.9.13 lane with @asto18089's authorship preserved; it'll auto-close with credit once it reaches main.

Hmbown (approved):
thank you so much!!!!

PR #6053, opened 2026-09-11
feat(tools): allow tool results to carry images via metadata.images

h3c-hexin:
## Summary

A small convention so a tool can hand image artifacts (e.g. computer-use screenshots) to vision-capable models: `ToolResult.metadata["images"]` holds an array of file paths; the engine turn loop attaches each readable, provider-accepted image (PNG/JPEG/GIF/WebP within the shared 5 MB limit) to the tool-result message as the same `<image path="…">`-bracketed `ContentBlock::ImageUrl` triplet that `image_attach` produces for user attachments — appended after the `ToolResult` block in the same user message.

Design points:

- The `ToolResult` struct and all wire builders are unchanged — the convention is documented on the `metadata` field, and images ride the existing `ImageUrl` path, so session persistence, compaction, and per-request stripping for blind routes need no changes.
- Capped at 2 images per tool result (`MAX_TOOL_RESULT_IMAGES`): each image is a base64 data URL in the request body, so an uncapped array could blow up a turn's token budget. Two covers the computer-use case (a before/after screenshot pair) with headroom.
- Missing, oversized, or non-image files are logged with `tracing::warn` and skipped; a bad image never fails the turn. Error-path tool results never attach.

This is the natural companion to the computer-use plugin shipped in 0.9.12: a screenshot tool can now return the capture path in metadata and the model actually sees it on the next step.

## Tests

Four new behavior tests (`forkguard_tool_result_images_reach_model_as_image_blocks`, `..._without_images_key_is_unchanged`, `..._bad_images_degrade_to_text_only_result`, `..._images_are_capped_per_result`) using an embedded 1×1 PNG fixture. Verified on this branch: 4 passed; `cargo fmt --all -- --check` clean.

Port note: adapted to the current turn-loop append sites (the `is_error: (!output.success).then_some(true)` shape and the hook-context wrapper).

## Credits

Original implementation by @asto18089 (co-authored).

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6053" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tools/src/lib.rs
+++ b/crates/tools/src/lib.rs
@@ -129,6 +129,14 @@ pub struct ToolResult {
     /// Whether the execution was successful.
     pub success: bool,
     /// Optional structured metadata.
+    /// Convention: a tool that produced image artifacts (e.g. a computer-use
+    /// screenshot) may list their file paths under the `"images"` key
+    /// (`{"images": ["<abs path>", ...]}`). The engine turn loop attaches each
+    /// readable, provider-accepted image (PNG/JPEG/GIF/WebP within the shared
+    /// size limit, at most a small capped number per result) to the
+    /// tool-result message as image content blocks so vision-capable models
+    /// can see it; unreadable or invalid images are skipped without failing
+    /// the turn.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub metadata: Option<Value>,
 }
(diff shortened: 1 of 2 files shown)


Hmbown:
Thank you @h3c-hexin and @asto18089 for the image-result work. I adapted it into the local 0.9.13 lane as `d727103ca7`, preserving your author/co-author credit and using the typed inline-image path already shared by native and MCP tools. It now checks full image decoding, MIME and size on admission and replay, with visible omission receipts and coverage for Chat, Responses and Anthropic message ordering. The focused suite passed 1,772 tests.

The adaptation deliberately does not add automatic file reads from `metadata.images`; tools return owned image bytes through `ToolResultContentBlock::Image` (or standard MCP image content). That keeps file access at the originating tool and avoids a second provider-message pipeline. Final runtime/CI qualification is still pending; the credited commit will auto-close this PR when it reaches main.

PR #6054, opened 2026-09-11
feat(execpolicy): deny-matching expressiveness for rulesets

h3c-hexin:
## Summary

Four additions to the deny-prefix matcher that let a ruleset express the vectors real exfiltration and sabotage commands use:

1. **cmd.exe single-letter slash flags** (`/f`, `/s`, `/q`) skip in any position/order like `-` flags. Only the single-letter shape skips — multi-character `/`-tokens are real POSIX paths (`/tmp`, `/etc`) and stay positional, so `cp /tmp/new_key ~/.ssh/authorized_keys` cannot hide its operand behind a skipped path (regression-tested).
2. **Mid-rule wildcard**: a rule token of exactly `*` matches zero or more consecutive command tokens regardless of shape, so a rule can anchor on a sensitive tail (`grep * ~/.ssh/id_rsa`, `dd * of=/dev/sda`) without enumerating every flag spelling. DFS bounded by the existing `seen` set; trailing `*` degrades to prefix semantics; leading `*` follows generic rules (documented + pinned). Each wildcard widens a rule's deny face — rule authors own that discipline; the engine stays permissive by design.
3. **`.exe` fold on the deny command word**: `cat.exe` and `C:\Windows\System32\cat.exe` now match a `cat` rule. One direction only (a rule naming `.exe` keeps requiring it; `catalog` never matches `cat`).
4. **Rooted absolute-path rules match exactly**: typed File path rules previously only matched after workspace-relative normalization, so a rule pinning an absolute location (a real home, `/root`, a Windows profile, a literal `~` spelling) could never match an outside-workspace call. A rooted-rule-only exact fallback (separator/case folding, no wildcards) fixes this; relative-rule semantics untouched; traversal spellings stay unmatchable (pinned).

Plus: `rulesets` becomes `Arc<RwLock<…>>` so `set_ruleset` through any clone is observed by every clone — a host cloning the engine into long-lived side executors no longer leaves them on a stale ruleset after a live permission update. `approved_for_session` stays deliberately clone-private (a remembered grant must not authorize a cloned executor's calls), documented and pinned.

## Tests

Ten new tests in `crates/execpolicy`: slash-flag interleavings + the POSIX-path regression guard, wildcard zero-match/trailing/leading/chained-segment semantics, `.exe` folding positives and negatives, absolute-path matching incl. tilde literals and the rooted-only fallback, traversal rejections, ruleset liveness across clones, and clone-private approvals.

Verified on this branch: `cargo test -p codewhale-execpolicy --locked` green (lib 199 + integration); the 13 execpolicy-related `codewhale-tui` tests pass; `cargo fmt --all -- --check` clean. Ported via a clean three-way merge against current main (the command_safety move and the redirection-preservation fix merge without conflicts).

## Credits

Original implementation by @asto18089 (co-authored), including three rounds of self-review with mutation-checked tests.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6054" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/crates/execpolicy/src/lib.rs
+++ b/crates/execpolicy/src/lib.rs
@@ -6,6 +6,7 @@ pub mod shell_expand;
 pub use approval_mode::ApprovalMode;
 
 use std::collections::HashSet;
+use std::sync::{Arc, RwLock};
 
 use anyhow::Result;
 use bash_arity::BashArityDict;
@@ -118,8 +119,11 @@ pub struct ToolAskRule {
     /// cannot silently authorize a later invocation with extra arguments.
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub command_exact: bool,
-    /// Optional workspace-relative file path matched exactly after
-    /// normalization.
+    /// Optional file path matched exactly. A workspace-relative rule
+    /// normalizes against the call's workspace; a ROOTED rule (leading `/`,
+    /// `~/`, or a Windows drive) matches the call path exactly after
+    /// separator and case folding, so it can pin locations outside the
+    /// workspace. Traversal segments never match on either channel.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub path: Option<String>,
     /// Optional absolute workspace root that limits this rule to one repo.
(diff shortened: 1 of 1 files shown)


Hmbown:
Thank you @h3c-hexin and @asto18089 — the deny matching and live ruleset sharing are harvested into the local 0.9.13 lane as `08018b8c4e`, with your authorship and co-author credit preserved. The adaptation keeps Windows command folding platform-specific, checks each complete decision against one policy generation, and publishes permission changes once so running clones cannot receive a delayed rollback. All 200 execpolicy tests and both approval-persistence regressions passed. Final release/runtime qualification is still in progress; this will auto-close with credit when the commit reaches main.

Issue #5529, opened 2026-08-21
Sub-agents cannot reliably execute: wall-time deaths lose uncommitted work, provider-route failures block dispatch, shell tooling needs workarounds


Hmbown:
## Summary

Sub-agents (workers/builders) cannot reliably run shell commands in this environment, which makes delegated execution work — the core Fleet value proposition — unusable. Three failure modes observed today alone:

1. **Wall-time budget deaths mid-task.** Two worker sub-agents (117 and 83 steps in) hit the 1800s wall-time limit and were killed with all uncommitted work. One had a complete, correct fix uncommitted (the sub-agent transcript scroll fix); the other produced nothing. The runtime's own lessons file (GROK-UX-LESSONS-20260820.md section 0) documents the same class of failure killing research sub-agents: a scout died at 60/60 steps with no report. Budget death must produce a final report from the checkpoint and force-commit uncommitted work, not silently discard it.
2. **Provider-route brittleness blocks dispatch entirely.** Fleet profiles pinned to xAI (`builder`, `reviewer` on grok-4.6) fail at start with "client could not be built" even after `codewhale auth external-consent --provider xai --mode read-only` succeeds and saves consent — fresh processes still fail, so the consent path in the shipped 0.9.10 binary does not resolve credentials the runtime itself just persisted. There is no fallback: the task never starts.
3. **Shell execution from sub-agents is unreliable/hostile to long commands.** Background jobs get parked to /jobs with completion arriving as runtime events that may or may not surface; heredocs with certain payloads fail to spawn at all ("nul byte found in provided data"); interleaved foreground/background waits make multi-step verification (edit → revert → test → restore) awkward enough that the parent ends up doing the work itself — defeating the point of delegation.

## Expected

- A sub-agent should be a first-class executor: reliable shell foreground execution with captured exit codes, straightforward file read/write (including binary-safe edits), and predictable background-job semantics.
- Budget exhaustion = graceful landing (final report + commit of finished pieces), never silent loss.
- Route failures (missing provider creds) should fall back or queue with a clear operator-facing signal, not fail the dispatch.
- The parent's own shell tooling should not need workarounds (python-script files instead of heredocs, split greps instead of single pipelines) for routine multi-step verification.

## Evidence

[...]

PR #6055, opened 2026-09-11
feat(subagent): rate-limit-adaptive launch scheduling

h3c-hexin:
## Problem

A swarm of sub-agents launched against one shared provider makes parallel 429s a steady state rather than an edge case. The launch gate is a fixed-capacity `Semaphore`, whose capacity cannot actually shrink (the only way is swapping the `Arc`, which silently fails while any child holds a permit — why `update_runtime_limits` today only applies launch-concurrency changes when the fleet is idle), and nothing observes provider throttling: every queued child retries in lockstep and the fan-out hammers the same rate limit.

## Fix

- **`DynamicGate`** replaces the `Semaphore`: runtime-adjustable capacity (allowed below the active-holder count; surplus holders finish naturally, new admissions block until `active < capacity`), FIFO waiters receiving pre-counted permits through a oneshot channel. Cancellation is safe on both paths — a waiter cancelled before the grant leaves a stale queue entry the granter skips, a waiter cancelled after the grant drops the dispatched permit whose `Drop` re-releases the slot — so a lost wakeup cannot occur (stress-tested with a timed drain assertion).
- **`RateLimitGovernor`** (owned by the manager, stamped onto every runtime at the single `spawn_background_with_assignment_options` chokepoint, inherited through the derive tree): a 60s sliding window drives AIMD — ≥2 rate-limited attempts or >30% ratio (with an attempts≥2 volume floor so a two-call fleet doesn't halve on one blip) halves capacity; ≥4 pauses new admissions (capacity 0; in-flight children unaffected); three consecutive successes add one unit back. A pause lifts only when the window drains; an external launch-concurrency change routes through the governor and cannot silently lift it. Queued children probe `recover_if_window_drained` every 5s, so the queue cannot freeze when the entire in-flight fleet finishes before any success arrives.
- **Retries**: `Retry-After` honored when present; otherwise full-jitter exponential backoff (250ms base, 120s cap) de-synchronizes a fan-out that was 429'd by the same response. `QuotaExhausted` is deliberately not reported — quota is a billing condition, not a transient throttle. The governor never delays an in-flight call; it only decides whether new launches may be admitted.
- **`update_runtime_limits`** applies launch-concurrency changes to the live gate immediately (previously deferred while children ran, for the semaphore-swap reason above).
- The queued-acquire future is pinned across probe iterations so a long pause doesn't leak one stale queue entry per tick per queued child.

## Tests

Governor suite (14 tests on this branch): window statistics, AIMD up/down boundaries, pause/recover/time-heal, gate cancel re-dispatch, queued-cancel branch, and a multi-task abort/pause stress test that must drain back to full capacity within a time budget (a lost wakeup, permit leak, or stale-entry slot swallow fails instead of hanging). Full `tools::subagent` slice: 602 passed; `cargo fmt` clean.

Port notes: adapted to current main — the roster-route tests' `acquire_owned` call sites converted to the new gate API; `running_count` is kept (still used by tests).

## Credits

Original implementation by @asto18089 (co-authored), hardened through five documented review follow-ups (governor wiring on the main path, pause liveness, lost-wakeup race, pinned queued acquire, direct stress tests).

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6055" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


Hmbown:
Triage (finish lane, 2026-09-16): rate-limit-adaptive launch scheduling overlaps the sub-agent execution cluster #6189 / #5529 / #6148 — linking so the dependency is visible. Conflicting with current main; will need a rebase and a decision whether #6148's JoinSet convergence supersedes this scheduling approach.

Hmbown (commented):
Design review (from the diff — I could not build it here): this is the right direction. The fixed-capacity `Semaphore` genuinely cannot shrink under load, and manager-level 429 observation is new capability, not a duplicate of per-request retry. The single-chokepoint governor stamping, `None`-means-report-nothing for manager-less runtimes, and the deadlock-aware gate (deep descendants bypass) all read carefully. Test changes are mechanical API adaptations, not weakened assertions.

Two things before this can land:

1. **Stale, not wrong.** The branch is `CONFLICTING` and red because main moved underneath it (`subagent/mod.rs` gained checkpointing, per-step context bounds, and provider fallback; `execpolicy` and the launch tests shifted too). Per our contributor rules the rebase-around-churn is on us, not you — I will rebase this onto current main and get it green. No action needed from you unless you want to do it first.
2. **One Rung-2 question:** `rate_limit_backoff_base` / `rate_limit_retry_delay` in `governor.rs` — does the `llm_client` retry path already own an equivalent backoff curve? If so the governor should reuse it rather than carry a second curve; if not, this one is fine where it is.

Leaving this as a comment (not approval) only because of the conflicts + reds; the design itself is sound.

Hmbown:
This is excellent work — the DynamicGate rationale (adjustable capacity below the active-holder count) and the AIMD recovery are exactly right, and the cancellation-path tests are the kind most PRs skip. The QuotaExhausted exclusion shows real judgment about billing vs. throttle.

Heads-up on timing: we have done a lot of subagent surgery on main since you opened this, so the launch paths this plugs into have moved and it conflicts through no fault of yours. We will harvest it onto current main ourselves once the in-flight refactor settles — the whole design deserves to land, not just pieces. Nothing needed from you, and thank you for building this.

Hmbown:
Thank you @h3c-hexin — the DynamicGate rationale (adjustable capacity below the active-holder count, pre-counted permits so a cancelled waiter can never swallow a wakeup) and the AIMD recovery with real Retry-After honoring are exactly right, and the cancellation-path tests are the kind most PRs skip. Harvested into main as 498490680 with your authorship preserved (plus @asto18089's co-authorship); it will auto-close with credit once the workflow sees it. The QuotaExhausted exclusion stays intact.

PR #6096, opened 2026-09-11
feat(commands): adopt capability shapes in the TUI session-export slice (FEAT-025)

aboimpinto:
# FEAT-025: adopt command shapes in the TUI session-export slice

Rewires `/export` (alias `/daochu`) through the portable command contract while the command file
stays in `codewhale-tui`. Structural migration only: no user-visible behaviour change.

Tracking: EPIC-006 / FEAT-025 in umbrella [Hmbown/CodeWhale#5316](https://github.com/Hmbown/CodeWhale/issues/5316).

## Review surface

Composition first, because 6,312 additions is the number that gets read:

| Class | Lines added | What it is |
| --- | ---: | --- |
| New contract boundary | 588 | `command-contract/facets.rs` (+221), `handler.rs` (+39), `commands/contract.rs` (+328) |
| New TUI adapter + host wiring | 445 | `session_export_host.rs` (+273), `commands/mod.rs` (+137), clipboard and minor wiring |
| Verbatim relocations | 988 | `secrets/redact.rs` (+599), `sanitize.rs` (+365), `secrets/lib.rs` plumbing (+24) |
| Reshaped in place | 1,147 | `groups/session/export.rs` (+1,108/-1,101, **net +7**) and its siblings |
| Verification | 3,113 | tests 2,669 (894 assertion/test lines), test support 178, goldens 233, gate script 33 |
| Build / attributes | 20 | `Cargo.toml`, `Cargo.lock`, `.gitattributes` |

**~1,030 lines are genuinely new production code.** The rest is moved code, reshaped existing code,
or the evidence that behaviour did not change. Net production growth is +1,271 lines, of which the
relocations contribute +223.

The 6.3k is not hidden, and it is also not the largest slice of this work: this is one command group,
and a mid-sized one. `session/` is 7,578 lines across 22 files, behind `core` (11,684) and `config`
(8,478), and the file this slice touches, `export.rs`, is 1,570 lines on `main`.

**Already rebased onto current `main` (`3016ef492`, 258 commits of drift) with zero conflicts.** If
other PRs land first we will resolve the conflicts from our side: this PR is not a source of merge
contention and is safe to merge last. If you would rather see this slice cut differently — split the
relocation out, split the test volume, cut the group another way — say so and we will follow your
lead.

This is one bounded slice of EPIC-006, which extracts the TUI command system into
`codewhale-commands` in two passes: adopt capability shapes in place, group by group, then move the
already-decoupled code. This PR does the first pass for session export only.

## What changes

**Contract** (`codewhale-command-contract`)
- New `CommandSessionExportContext`: seven operation-granular, `&self` delegates returning
  contract-owned DTOs (metadata, transcript projection, semantic restore points, turn handoff,
  clipboard mode/recovery/delivery, destination resolution/writing).
- New `SESSION_EXPORT` capability, declared by `/export` and nothing else. The TUI bridge exposes only
  that facet.
- D9 data minimisation at the type level: internal reasoning bodies, reasoning signatures, and
  inline/local image payloads are replaced by typed omission markers while the projection is built,
  so they never enter the DTOs.
- The raw-bits accessor is `#[cfg(test)]`; a canary pins the fact that bits 0-15 are now fully
  allocated (see "Reviewer notes").

**TUI** (`codewhale-tui`)
- `SessionExportAdapter` is the sole owner of concrete export machinery: metadata derivation,
  authoritative/visible-history projection, semantic restore-point projection, the existing
  `turn_handoff_markdown` renderer, clipboard mode/recovery/delivery, and protected destination
  resolution/writing.
- Host services that must outlive the physical move live in `commands/session_export_host.rs`,
  **outside** `groups/session`: the shared `last-copy.md` recovery writer (now reused by `/copy`) and
  the protected export-destination resolver/writer. Algorithms, check order, error text and platform
  behaviour are the baseline implementations relocated unchanged.
- The portable handler owns parsing, document rendering, redaction, restore-point correlation,
  operation sequencing and every user-visible string. It renders from an owned projection and redacts
  JSON in place, so no payload is copied twice.
- `/copy` and `/structcopy` only rewire their helper imports; neither is migrated.

**One shared sanitizer.** The pure ANSI/URL/text/secret-redaction primitives move to
`codewhale-secrets` (`redact` + `sanitize`) so portable command helpers have one acyclic
implementation with no TUI, client, or config dependency. Bodies are relocated verbatim;
`config::persistence` re-exports to keep its public API stable, and `tui::client` / `tui::osc8`
delegate so their callers cannot drift.

## Preservation evidence

Behaviour is preserved, and this is proven rather than asserted.

- **Four baseline-captured goldens.** Captured from the pre-change implementation in a scratch
  worktree by dispatching frozen fixtures through the *baseline* public `/export` seam: authoritative
  transcript with no restore repository (header, every content variant, omission markers, redaction,
  ANSI stripping); the turn handoff; the visible-history fallback over a workspace with recorded
  restore points; and correlation against recorded restore points (table, newest-first order, `tool`
  snapshots listed but not correlated, the ambiguous double match, the no-match line). Only genuinely
  clock-derived fields are normalised, with a unit test pinning the normaliser so it cannot quietly
  widen. Everything else is compared byte-for-byte. Re-capturing them from the implementation under
  test would make the comparison a tautology; the fixture header says so.
- **Verbatim relocations, checked by hand:** `redact.rs` is a byte-identical move of the 586-line
  block from `config::persistence`; `strip_ansi_impl`/`utf8_seq_len` are byte-identical to the former
  `osc8` bodies; `redact_url_for_display` is identical apart from `reqwest::Url` → `url::Url` (the
  same type); the parser and the protected-file/recovery writers are literal copies.
- **Exact behaviour retained:** read-only snapshot access (never creates a repository), overwrite
  refusal (and `--force`), private permissions, symlink rejection, unprotected-path errors,
  clipboard→recovery ordering, and all usage-error text.

## Least authority and extraction readiness

- `/export` declares exactly `SESSION_EXPORT`; no `PRESENTATION` facet, no new catalog string.
- A missing facet returns `Command capability unavailable: session_export`; there is no `.expect()`
  or panic on the facet path.
- A shared test helper destructures **all sixteen** `ContextParts` slots, so "exposes no unrelated
  facet" is a compile-time guarantee: a new facet fails to compile until it is classified, and every
  slot is asserted absent in one place.
- No `App`, clipboard, snapshot, filesystem, history, or turn-handoff type crosses the facet.
- Below `1 << 15` the backing storage stays `u16` and `NONE` semantics are unchanged.

**Extraction proven by compilation.** The handler plus its 21 portable tests were compiled in a
throwaway crate outside the TUI, with only a stand-in for the temporary FEAT-037 `CommandResult`.
Library and test targets build and all 21 tests pass with **no `codewhale-tui` in the dependency
graph**. The only adjustment needed was making the stand-in faithful to `CommandResult::error`
(`Error: {msg}` prefix) - which is the point: that outcome type is the one TUI-owned thing the slice
touches. Host-bound tests and fixtures stay at the `commands` root, outside the group that will move.

## Post-implementation audit

The slice was audited against the pristine pre-change tree before this PR, and again after the audit
fixes. No functional defect was found. The findings that changed the code are folded into the commits
above; the ones that did not are recorded here rather than silently dropped:

| Finding | Disposition |
|---|---|
| Complete-document parity was initially proven for only one state (authoritative transcript + no repository) | Fixed: the four goldens above cover the remaining states |
| "No unrelated facet" was a spot-check of six of sixteen slots | Fixed: exhaustive destructuring helper |
| Correlation compared the *rendered* role string against `"user"` | Fixed: `is_user_role` carries the exact `Role::User` comparison, so `Role::Unrecognized("user")` is not treated as a user turn |
| Stale `export::<helper>` seam comments in `/structcopy` | Fixed, and the audits now reject a stale reference |
| Extraction audits were substring scans; sanitizer purity was a manifest text match | Fixed: token-aware scans with their own unit test, and the purity check moved into the boundary gate as a real dependency-graph check |
| A relocated test dropped the repository's secret-scanner avoidance convention: the pre-move private-key test assembled its PEM header at runtime so no scanner could match it, and the moved version used a literal header (GitGuardian flagged 1 secret) | Fixed: restored the runtime assembly idiom rather than adding a scanner ignore; GitGuardian now scans the commits clean |
| The adapter's `Unreadable` restore arm has no adapter-level test | **Not fixed, recorded.** It is reachable only when the `git` binary cannot be resolved (`SnapshotRepo::list` maps a failing `git log` to an empty list), and `dependencies::Git::resolve` is process-cached, so any test would be flaky. Behaviour is unchanged from the baseline. |
| Peak memory: the projection is an owned copy of the transcript | **Accepted.** The facet must return owned data because `SharedCommandHost` lends `App` through a `RefCell`, and a `dyn` facet cannot return data borrowed from a temporary `Ref`. Removing it needs a host proxy that can lend a borrowed projection; the avoidable second copy was removed instead. |

## Scope disclosure

`crates/tui/src/tui/clipboard.rs` carries an unrelated, test-only bounded retry for a fork/exec
`ETXTBSY` race in the `wl-paste` regression. It is kept because the configured workspace gate was
intermittently red without it, and it is disclosed rather than hidden. There is no equivalent change
on `main` to reconcile against. Happy to split it into its own PR on request.

## Not fixed here

- `CommandCapabilities` widening to `u32` (see reviewer notes) - owned by the next slice.
- A borrowed/`Cow` export projection, which needs the host-proxy change described above.
- No scope expansion: FEAT-037's temporary allowance is still exactly `CommandResult`; no `AppAction`,
  callback, or deferred action was added. The root `session` migration frontier remains **pending**.

## Verification

Rebased onto current `main`; all gates re-run at the rebased head:

| Gate | Command | Result |
|---|---|---|
| Format | `cargo fmt --all -- --check` | PASS |
| Clippy | `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or` | PASS — 0 warnings |
| Tests | `sh scripts/with-hermetic-test-home.sh cargo nextest run --workspace --all-features --locked --profile ci` | PASS — 15403 run / 15403 passed / 21 skipped |
| Doctests | `sh scripts/with-hermetic-test-home.sh cargo test --workspace --all-features --locked --doc` | PASS |
| Release build | `cargo build --release --locked -p codewhale-cli -p codewhale-tui` | PASS |
| Contract tests | `cargo test -p codewhale-command-contract --lib --locked` | PASS — 59 / 59 |
| Migration gate | `python3 scripts/test_check_command_migration_manifest.py` + live check | PASS — frontier unchanged: `[config, core, debug, session]` |
| Boundary gate | `python3 scripts/test_check_command_crate_boundaries.py` + live check | PASS — no TUI edge from `codewhale-command-contract` or `codewhale-secrets` |
| CI wiring / locale parity | `python3 scripts/test_ci_migration_wiring.py` / `python3 scripts/check-tui-locale-parity.py` | PASS / 2211 of 2211 keys per pack |
| Doc build | `cargo doc -p codewhale-secrets --no-deps --locked` | PASS |
| Secret scan | GitGuardian (PR check) | PASS — 3 commits scanned, no secrets uncovered |

## Reviewer notes

1. **Two commits in this branch are a rebase, not new work.** Upstream moved 426 commits ahead while
   this slice was in progress. The branch is three logical commits on current `main`. Adopting the
   `codewhale-models` and `codewhale-localization` extractions needed four import-path updates, and
   the `CmdExportDescription` English string was reworded upstream (French unchanged); the tests pin
   the catalog values.
2. **`CommandCapabilities` has no free bits left.** `SESSION_EXPORT = 1 << 15` was the last one. The
   next slice needs its own exact-minimum facet, so it must widen the backing storage to `u32` (or
   justify reusing a bit, which would break the least-capability invariant this epic is built on). A
   canary test asserts the space is exactly full so this is a deliberate change, not an overflow.
3. **The remaining copy of the transcript is intentional** and explained above.
4. **The goldens pin host-derived values** (`Provider`, `Model`, `Mode`, the turn route line, and for
   the recorded states the snapshot id and timestamp cell). A future change to the default config or
   the handoff header will fail them by design; they must be re-captured from a verified baseline.

No-Issue: FEAT-025 is tracked in umbrella issue #5316, which remains open for the remaining EPIC-006 work.

Paulo Aboim Pinto


aboimpinto:
Resolved in `24994fdfd`, and noted here because the cause is worth recording.

The single finding was not a real credential. It came from moving the private-key redaction test out of `config::persistence` into `crates/secrets`: the original test deliberately assembled its PEM markers at runtime —

```rust
// Assemble the PEM markers at runtime so the source file never
// contains a literal private-key header for a scanner to match.
let begin = ["-----BEGIN RSA", " PRIVATE KEY-----"].concat();
```

— and the rewritten test used the literal header instead. The relocation preserved the algorithm but dropped the convention, which is exactly the class of thing a "verbatim move" is supposed to carry across.

Fixed by restoring the repository's own idiom (runtime assembly of the PEM markers and the token prefix) rather than widening `.gitguardian.yml`; the config was already correct and adding an ignore would have hidden the next occurrence. The scanner now reports the commits clean.

Paulo Aboim Pinto

aboimpinto:
**Windows failure fixed** in `e96013ab3` — `Test (windows-latest)` failed in "Run tests", and it was mine.

Cause: this PR adds four `include_str!()` goldens under `crates/tui/src/commands/fixtures/` that are compared **byte for byte** against a document the exporter builds with `\n`, but I did not declare them in `.gitattributes`. With `* text=auto` in effect, a Windows checkout gives those files CRLF, so the comparison failed on a pure `\r\n` diff. Only Windows can see it; Linux and macOS cannot.

This is precisely the class your own `.gitattributes` already documents — "fatal for the one that is compared byte for byte" — alongside the telemetry `v1.json` and Tideline `.txt` goldens that hit it before. I reproduced it locally by converting a fixture to CRLF: the golden failed with `first diff at byte 31`, and passed again once reverted.

Fix: declare the new fixtures `text eol=lf`, matching the existing entries for the other `include_str!()` inputs. The committed blobs were already LF, so nothing else changes; `git ls-files --eol` now reports `attr/text=auto eol=lf` for all four.

Paulo Aboim Pinto

aboimpinto:
**macOS failure fixed** in `54d9261d8` — again mine, and again only the platform CI could see it.

`adapter_preserves_protected_file_write_and_overwrite_refusal` and `adapter_rejects_directory_destination_and_parent_not_a_directory` failed with `refusing export through symlink component: /var`.

Cause: these two relocated tests call the facet operation `write_export_file` **directly**, handing it a raw `TempDir` path. The old tests in `groups/session/export.rs` drove `/export file …` through the command, and the command calls `resolve_export_path`, which canonicalizes the workspace first. The protected writer deliberately rejects any path with a symlink component — and on macOS `/var` is a symlink to `/private/var`, so every `TempDir` path is rejected. Linux `/tmp` has no symlink ancestor, so it never reproduced locally.

Fix: the adapter-level tests now build their workspace with a `canonical_workspace` helper that mirrors what `resolve_export_path` does for the real command, and the symlink test asserts against the intended symlink rather than passing on macOS for the wrong reason.

I also added `protected_writer_requires_a_path_free_of_symlink_ancestors`, which creates the symlinked root itself. That turns a macOS-only failure into one a Linux run catches, so this class cannot recur silently.

The production behaviour is unchanged: the symlink rejection is a verbatim relocation of the baseline, and `/export file /tmp/x.md` being refused on macOS is pre-existing baseline behaviour, not something this PR introduces.

Paulo Aboim Pinto

aboimpinto:
**Ready for review.** All checks are green on `54d9261d8` and the three platform failures CI caught have been fixed — each was mine, and each was invisible to the Linux-only local gate:

| Check | Failure | Cause | Fix |
|---|---|---|---|
| GitGuardian | 1 secret uncovered | the moved private-key test used a literal PEM header; the pre-move test assembled it at runtime so no scanner could match it | restored the repository's own idiom (no `.gitguardian.yml` change; the config was already right) |
| Test (windows-latest) | pure `\r\n` diff | the four new `include_str!()` goldens were compared byte for byte but not declared in `.gitattributes`, so `* text=auto` gave them CRLF | declared them `text eol=lf`, matching the other `include_str!()` inputs |
| Test (macos-latest) | `refusing export through symlink component: /var` | two relocated adapter tests called `write_export_file` directly with a raw `TempDir` path; the pre-move tests drove the command, which canonicalizes the workspace first, and macOS `/var` is a symlink | a `canonical_workspace` helper mirroring the real command, plus a new test that builds the symlinked root itself so the class now fails on Linux too |

Two of the three were **repository conventions that the relocation failed to carry over** rather than code defects, which is the useful thing to know when reviewing a "structural migration": the algorithm moved correctly, the surrounding discipline did not. Both are now documented in the code and pinned by tests.

Final state:

- 29 files, +6255 / −1905 against `main` `447223554`; three commits (contract + shared sanitizer, TUI wiring, tests + gates).
- fmt, Clippy (CI's lint set), nextest 15408/15408, doctests, release build, and the migration/boundary/CI-wiring/locale-parity gates all green locally and in CI.
- All four baseline-captured goldens still pass unchanged after the rebases, which is the evidence that the migration is behaviour-preserving.
- No bot review comments remain unaddressed; GitGuardian's notification has a reply.

Please take a look when you have time.

Paulo Aboim Pinto

Hmbown (approved):

Hmbown:
Triage (finish lane, 2026-09-16): this is the FEAT-025 slice of EPIC-006 tracked by #6145 (command contract: finish the FEAT-02x adoption or fold crates/command-contract). Cross-linking so it does not sit unparented — #6145 should decide whether this lands as-is, gets rebased, or is superseded by the fold decision. Still merges clean against main as of today; substantive review still needed (6.3k lines, structural).

aboimpinto:
## Reviewer's map (FEAT-025)

Rebased onto current `main` (`3016ef492`) — **zero conflicts** across 258 commits of drift. If any
other PR creates a conflict here, we will do the resolution work on our side. This PR does not need
to be merged first, and it is safe to merge it last.

### Where the 6,312 additions go

- **~1,030 lines of new production code** — the contract boundary and the TUI-side adapter/host wiring.
- **988 lines relocated verbatim** — `secrets/redact.rs` and `sanitize.rs`, byte-identical to the lines removed from `config/persistence.rs`, `tui/osc8.rs` and `tui/client.rs`.
- **1,147 lines reshaped in place** — mostly `groups/session/export.rs`, which changes +1,108/-1,101 for a **net +7**.
- **3,113 lines of verification** — tests, test support, four baseline-captured goldens, and the gate script. Only 894 of those are assertion or test-marker lines; the rest is fixture construction and expected-document text.

We are not hiding the size: this is one command group, and not even the largest one. `session/` is
7,578 lines across 22 files behind `core` (11,684) and `config` (8,478), and the file this slice
touches, `export.rs`, is 1,570 lines on `main`. The count is high because the shared sanitizer
relocation rides along and because this is the first FEAT in the sequence carrying byte-compared
goldens.

If you see a way to make this slice smaller — split the relocation out, split the test volume, cut
the group differently — say so and we will follow your lead.

### Suggested review order (~1,000 lines of real reading)

1. `crates/command-contract/src/facets.rs` (+221) — the boundary: seven operation-granular delegates returning contract-owned DTOs, and the least-authority shape.
2. `crates/command-contract/src/handler.rs` (+39) — `SESSION_EXPORT` on bit 15, envelope slot wiring, and the canary pinning that bits 0-15 are now fully allocated (which is why FEAT-026 must widen to `u32` first).
3. `crates/tui/src/commands/session_export_host.rs` (+273) — the TUI-owned adapter and host services; sole owner of the concrete export machinery.
4. `crates/tui/src/commands/groups/session/export.rs` — the **net** diff is what matters: the portable handler owns parsing, rendering, redaction, correlation and sequencing; every concrete concern stays behind the facet.
5. `crates/tui/src/commands/mod.rs` (+137) and `contract.rs` (+328) — registration and bridge wiring.
6. The four goldens in `crates/tui/src/commands/fixtures/` as the behaviour contract: captured from the **pre-change** implementation by dispatching frozen fixtures through the baseline `/export` seam, compared byte-for-byte with only clock-derived fields normalised. Re-capturing them from the code under test would make the comparison a tautology, and the fixture headers say so.

### Review order by commit

| Commit | Content | Size |
| --- | --- | --- |
| `24994fdfd` | contract facet + shared sanitizer relocation | +1,830/-748 |
| `ae48e2941` | route `/export` through the facet | +1,852/-1,135 |
| `54d9261d8` | goldens, regression/surface tests, gates | +2,630/-22 |

### How the relocation claim was checked

Removed lines from the origin files were compared against the added lines of the destination file,
whitespace-normalised and with comments excluded: `redact.rs` **438 of 438** non-trivial lines
identical, `sanitize.rs` **129 of 130** (the remainder is the merge seam between its two sources).
The relocated unit tests also carry the originals' conventions, including the runtime-assembled PEM
markers that exist so a scanner cannot match the source.

### Local verification on the rebased head

`fmt` clean · Clippy with the CI lint set clean (2m24s) · six project gates pass · doctests pass ·
release build passes (4m20s) · nextest workspace **15,833/15,837**.

The four nextest failures are **pre-existing and environmental**, not from this branch. All four were
reproduced identically on untouched `origin/main` in the same environment, and this branch touches
nothing under `sandbox/` or `tools/`:

- three `ghostty_*_keeps_full_motion` tests fail when the suite runs inside **tmux**, because `TMUX` sets `constrained_frame_rate` and those tests clear only the SSH markers;
- `enforced_readonly_python_queries_sqlite_...` fails when `TMPDIR` is under `/tmp`, because the sandbox mounts `--tmpfs /tmp` and then `--chdir`s into a temp workspace that no longer exists;
- `...python_cannot_reach_a_loopback_listener` trips an assertion that the refusal text must not contain `"connected"`, while the refusal echoes the command source, which contains `print('connected')`. Enforcement itself worked.

Separately, we removed a first-run `Secrets` scan finding by restoring the repository's own
convention rather than widening an ignore list: the moved private-key test now assembles its PEM
markers at runtime, as it did before the relocation.

### Not in this PR

`/export` behaviour is unchanged; `/copy` and `/structcopy` only rewire helper imports and are not
migrated; nothing moves out of `codewhale-tui` (that is Stage D, FEAT-038+). FEAT-026 is next and
must widen `CommandCapabilities` to `u32` before adding its own facet, because FEAT-025 consumes
bit 15.

Happy to answer anything, or to walk through the contract boundary directly if that is faster.

Paulo Aboim Pinto

PR #6111, opened 2026-09-12
feat(tui): add a file-scoped restore endpoint and gate the whole-tree rollback

gaord:
## Summary

Restores per-file revert, which the GUI had to withdraw because the engine
offered no file-scoped restore — and fixes the two defects in the existing
whole-tree rollback that the same investigation surfaced.

**Original problem: HengQuWorld/CodeWhale-VSCode#3.** The per-file Revert
control was removed because the only restore surface available to the GUI was
`POST /v1/snapshots/{id}/restore`, which restores the **whole workspace**: using
it to "revert one file" silently rolled back unrelated changes.

This is the engine half of that issue; the GUI half re-enables the control
against the new endpoint, behind a capability probe so it degrades with an
explanation on an engine that predates it.

No-Issue: engine half of HengQuWorld/CodeWhale-VSCode#3. That issue stays open
until the GUI half re-enables the control, so this PR deliberately carries no
closing keyword for it. A bare issue number in this body would resolve against
this repository, not that one.

### The issue's premise did not hold

The issue assumed the engine already had a file-scoped rollback (`patch_undo()`)
that only needed exposing through a thin adapter. It does not. `patch_undo()` is
scoped in *snapshot selection*, not in *effect*: it picks a recent `tool:`
snapshot, then calls `SnapshotRepo::restore()`, whose `git checkout <sha> -- :/`
checks out every path in the snapshot tree and deletes every working-tree file
the target does not contain. `revert_turn`, `patch_undo()` and the patch-undo
endpoint all call that one function, and no narrower primitive exists anywhere
in the engine. So the capability had to be built before it could be exposed.

### 1. A file-scoped restore primitive

`SnapshotRepo::restore_paths()` restores a pathspec rather than the tree, and
`path_differs_from_snapshot()` answers "would restoring just this file change
anything?" without consulting the whole-tree diff. `workspace_relative_path()`
is the single gate every caller-supplied path passes through, so a request
cannot name a path outside the work tree.

### 2. `POST /v1/threads/{id}/file-revert`

A thin adapter over (1), reusing the ownership rule `/undo` already applies:
only `tool:`/`pre-turn:` snapshots tagged with the thread's own session are
candidates, so a revert cannot reach into another conversation's work. It
reports the action it took (`modified` / `recreated` / `removed`), and "nothing
to revert" is a 409 rather than a silent success.

### 3. The existing whole-tree rollback is now trust-gated and cannot half-apply

- **No gate.** `POST /v1/threads/{id}/patch-undo` restores the whole workspace
  and had no gate of its own, so a client could drive a workspace-wide rollback
  from a thread in Ask or Auto-Review posture. `patch_undo()` refuses that
  (`yolo || trust_mode`); the endpoint now applies the same rule, read from the
  thread's own record so a caller cannot assert it.
- **Half-undo on failure.** The step reported failure inside `patch_result` and
  the turn was forked anyway — dropping the turn while its file changes stayed
  on disk, a workspace the transcript can no longer account for. The helper now
  returns `Result`: `Ok` means the fork may proceed, `Err` aborts the undo.

The gate is evaluated *after* a real target is found, so "there was nothing to
revert" still undoes the conversation, while "there is something to revert but
you are not trusted" refuses outright.

### 4. The undo summary describes its own effect

"Files affected" came from `git diff --stat` run in the user's workspace against
the user's `.git`, which reports the user's own uncommitted work: it could list
files the restore never touched, and report nothing when that work happened to
be committed. Both `/undo` and the endpoint now use
`SnapshotRepo::snapshot_diff_stat()`, captured before the restore runs.

## Testing

- [ ] `cargo clippy --workspace --all-targets --all-features --locked`

Ran instead of the full-workspace forms:

- `cargo clippy -p codewhale-tui --all-targets --all-features --locked` with the
  CI allow list → no findings in any file this PR touches. (11 errors remain in
  files this PR does not touch — `tools/subagent/mod.rs`, `model_inventory.rs`,
  `phase_strip/tideline_tests.rs`, `tui/ui/apply.rs`, `tui/views/fleet_roster.rs`
  — under clippy 1.96.0; pre-existing on `main`.)
- `cargo test -p codewhale-tui --lib snapshot::` → **68 passed** (+8)
- `cargo test -p codewhale-tui --lib patch_undo` → **9 passed** (+2)
- `cargo test -p codewhale-tui --lib revert_file` → **3 passed** (new)
- `file_revert_route_probes_as_available_and_404s_unknown_threads` → passes
  against a real bound server: GET on the route → 405, POST for an unknown
  thread → 404. That is the exact contract a client probes to decide whether the
  endpoint exists, so it is asserted rather than assumed.

## Checklist

- Adds `restore_paths()` as a narrower sibling of `restore()`. It does **not**
  replace it — the whole-tree path is still correct for `/undo`, which is
  supposed to roll back everything.
- Docs/comments updated; tests added for every new behaviour.
- No UI changes in this repository (the GUI half is a separate one).
- No changelog hunks, per CONTRIBUTING.

### DIFF
--- a/crates/tui/src/snapshot/repo.rs
+++ b/crates/tui/src/snapshot/repo.rs
@@ -772,6 +772,34 @@ impl SnapshotRepo {
         Ok(outcomes)
     }
 
+    /// `git diff --stat` between snapshot `id` and the current working tree,
+    /// computed inside the side repo.
+    ///
+    /// This is what restoring `id` *would* change, so it must be captured
+    /// before the restore runs — afterwards the work tree matches the snapshot
+    /// and the diff is empty by construction.
+    ///
+    /// It deliberately runs against the side repo rather than the user's. The
+    /// previous summary ran `git diff --stat` in the workspace with the user's
+    /// `.git`, which reports the user's own uncommitted work: it listed files
+    /// the restore had not touched, and reported nothing when that work
+    /// happened to be committed. Returns `None` when nothing differs.
+    pub fn snapshot_diff_stat(&self, id: &SnapshotId) -> io::Result<Option<String>> {
+        let diff = run_git(
+            &self.git_dir,
+            &self.work_tree,
+            &["diff", "--stat", id.as_str(), "--", ":/"],
+        )?;
+        if !diff.status.success() {
+            return Err(io_other(format!(
+                "git diff --stat failed: {}",
+                String::from_utf8_lossy(&diff.stderr).trim()
+            )));
+        }
+        let stat = String::from_utf8_lossy(&diff.stdout).trim().to_string();
+        Ok((!stat.is_empty()).then_some(stat))
+    }
+
     /// Return whether the current workspace matches the given snapshot's
     /// tracked file content.
     ///
@@ -1581,6 +1609,34 @@ mod tests {
             "user-work-in-progress",
             "the scoped restore must leave the unrelated edit alone"
         );
+    }
+
+    #[test]
+    fn snapshot_diff_stat_describes_what_a_restore_would_change() {
+        let tmp = tempdir().unwrap();
+        let (repo, _home) = make_repo(tmp.path());
+        let changed = repo.work_tree().join("changed.txt");
+        let untouched = repo.work_tree().join("untouched.txt");
+
+        std::fs::write(&changed, b"v1").unwrap();
+        std::fs::write(&untouched, b"stable").unwrap();
+        let id = repo.snapshot("pre-turn:1").expect("snapshot");
+
+        std::fs::write(&changed, b"v2").unwrap();
+
+        let stat = repo
+            .snapshot_diff_stat(&id)
+            .expect("diff stat")
+            .expect("the snapshot differs, so something must be reported");
+        assert!(stat.contains("changed.txt"), "got: {stat}");
+        // The stat describes the restore's effect, not the workspace's whole
+        // uncommitted state — a file the restore will not touch must not appear.
+        assert!(!stat.contains("untouched.txt"), "got: {stat}");
+
+        // After restoring, the two sides agree: nothing left to report. (Which
+        // is why the caller must capture this *before* the restore runs.)
+        repo.restore(&id).expect("restore");
+        assert_eq!(repo.snapshot_diff_stat(&id).expect("diff stat"), None);
     }
 
     #[test]
--- a/crates/tui/src/snapshot/repo.rs
+++ b/crates/tui/src/snapshot/repo.rs
@@ -48,6 +48,41 @@ pub struct Snapshot {
     pub session_id: Option<String>,
 }
 
+/// What a file-scoped restore did to one path, relative to the working tree
+/// it was applied to.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PathRestoreAction {
+    /// Both the snapshot and the working tree had the path; its content came
+    /// back from the snapshot.
+    Modified,
+    /// The snapshot had the path and the working tree no longer did, so the
+    /// restore recreated the file.
+    Recreated,
+    /// The working tree had the path and the snapshot did not, so the restore
+    /// removed the file.
+    Removed,
+}
+
+impl PathRestoreAction {
+    /// Stable wire name, also used by the runtime API response.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Modified => "modified",
+            Self::Recreated => "recreated",
+            Self::Removed => "removed",
+        }
+    }
+}
+
+/// Report of what [`SnapshotRepo::restore_paths`] did to one path.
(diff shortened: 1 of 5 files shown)


crates/tui/src/snapshot/repo.rs
```diff
@@ -591,6 +626,180 @@ impl SnapshotRepo {
...
+            .collect();
+        if !tracked.is_empty() {
+            let mut args: Vec<String> = vec![
+                "checkout".to_string(),
+                id.as_str().to_string(),
+                "--".to_string(),
+            ];
+            args.extend(tracked);
+            let arg_refs: Vec<&str> = args.iter().map(String::as_str).collect();
+            let checkout = run_git(&self.git_dir, &self.work_tree, &arg_refs)?;
```
Hmbown, line 834: [P1] Treat the requested filename as a literal Git pathspec. `--` ends options but does not disable pathspec patterns. On macOS I reproduced `git checkout <snapshot> -- "file[12].txt"` restoring both the literal `file[12].txt` and the unrelated `file1.txt`. Both paths can exist in this snapshot, so the `target_paths.contains(rel)` check does not protect the second file. `run_git` does not set `GIT_LITERAL_PATHSPECS`. Use literal pathspecs for the checkout and comparison (for example the global `--literal-pathspecs` option) and add a regression asserting the second file remains unchanged. This reproduction tested the actual Git command used here; it was not a compiled test of the PR helper.

Hmbown (changes requested):
Reviewed current head 960043a. I am holding this PR for correctness fixes, not a rebase request. The existing review findings about exact snapshot ownership, trust/admission, and error propagation are real blockers. I also reproduced an additional single-file isolation failure in the Git primitive used here (inline). Please include regression coverage for literal metacharacter filenames, exact originating tool/snapshot identity, untrusted and active threads, failed snapshot comparisons, and directories. Windows currently has one failing absolute-path test: the `/tmp/ws` fixture is not an absolute Windows path. Keep the whole-tree and file-scoped actions separate in the API documentation.

Hmbown (approved):
let me take a look at this - think my agent was a bit too reactive

Hmbown:
Thanks @gaord — this is the engine half HengQuWorld/CodeWhale-VSCode#3 needed, and the whole-tree gate and half-undo fixes were the right call. Rather than ask you to chase our churn, I pushed a maintainer commit on top of your two (yours are untouched; you stay the author of the feature).

What the follow-up commit changes, on the confirmed findings:

- **Exact snapshot identity.** The request now carries `snapshot_id` (the `tool:`/`pre-turn:` snapshot from the change the user selected) and `expected_hash` (`sha256:<hex>` of the reviewed bytes, or `absent`). The server no longer picks "the newest snapshot that differs", because an unrelated later snapshot could erase a user's edit while keeping the tool's change. A stale hash is a 409 and nothing is touched.
- **Literal pathspecs.** `file[12].txt` no longer restores `file1.txt`; Git runs with `--literal-pathspecs` for every file-scoped operation, with a regression that keeps the sibling intact.
- **Regular files only.** Directories, symlinks anywhere in the path, and `.git` components are refused before anything is compared; a gitignored file that the safety backup cannot retain is refused instead of removed; removal errors propagate instead of reporting `removed`.
- **No half-undo.** Snapshot open/list/compare failures abort patch-undo and preserve the conversation; an unborn side repo is distinguished from broken history; fork depth/history is validated before any file changes.
- **No racing an active turn.** All three restore routes reserve the runtime's config/checkpoint admission through an owned guard, reject active turns in the same, nested or parent workspace with 409, and move the reservation into the worker so a client disconnect cannot release it mid-Git. Trust and session binding are read under that admission. File restore mirrors the thread trust/Full Access rule you added to patch-undo.
- **Docs and Windows.** `docs/RUNTIME_API.md` now has a "Workspace restore endpoints" section (body, 200/400/404/409/422/500, probe, backup and file-versus-workspace semantics). The `workspace_relative_path` test now uses a real absolute temp path, which is what failed on `windows-latest`.

For the GUI half: the Changes panel needs to keep the snapshot id for each change record (labels are `tool:<call_id>`, listed by `GET /v1/snapshots`) and send the SHA-256 of the bytes it displayed. That is the contract the extension should probe for.

Local evidence: 75 focused tests (snapshot repo, undo/patch-undo/file-revert, restore guard, fork, revert_turn) passed, crate Clippy and fmt clean. Hosted CI on this head decides the merge.

Hmbown:
Follow-up commit from an independent second review of the maintainer change: admission is now awaited on the request (a client that gives up while queued no longer leaves a second undo behind the next one), a fork whose records were saved is returned even if its `thread.forked` event append fails, an unavailable workspace directory is a 409 for both routes instead of "nothing to restore", the server-wide snapshot route validates its id, and a file-scoped removal leaves the emptied parent directory alone. Docs updated accordingly. No change to the request/response contract for the GUI.

PR #6114, opened 2026-09-12
fix(tui): follow symlinked user-level context files

### LINKED ISSUE #6115: bug(tui): a symlinked user-level AGENTS.md is dropped silently
## Description

The user-level instruction layer refuses a symlinked file, and that refusal is collected as a **warning** rather than an error. Pointing a user-level context file at a shared file therefore drops the entire layer silently: the model runs without the global instructions, and nothing in the normal session flow says so.

Affected candidates, in the loader's own priority order (`load_global_agents_context`):

1. `~/.codewhale/AGENTS.md`
2. `~/.agents/AGENTS.md`
3. `~/.deepseek/AGENTS.md`
4. `~/.codewhale/instructions.md`, `~/.agents/instructions.md`, `~/.deepseek/instructions.md`

## Steps to reproduce

1. Put the shared instruction set anywhere, e.g. `~/AGENTS.md`.
2. Point the user-level path at it:
   ```
   mkdir -p ~/.deepseek && ln -sfn ~/AGENTS.md ~/.deepseek/AGENTS.md
   ```
3. Start codewhale in a workspace that has no project-level `AGENTS.md` of its own, and ask for something that only the global file establishes.
4. `/context report` shows the reason under "Project context warnings" — `Refusing symlinked context file /home/<user>/.deepseek/AGENTS.md`.

## Expected behavior

The user-level layer is loaded, following the link to its target. At the very least, an unusable user-level file is surfaced as an error rather than as a warning that scrolling past is enough to miss.

## Actual behavior

`load_context_file` returns `ProjectContextError::Symlink`; `load_global_agents_context` pushes it onto `warnings`, moves on to the next candidate, and with no other candidate present returns no instructions at all. The session proceeds without the user-level layer.

## Impact

Every session, silently, for anyone who keeps one instruction file and shares it across agents — which is the ordinary layout (`~/.config/<agent>/AGENTS.md -> ~/AGENTS.md`). The failure is invisible by construction: nothing in the TUI start-up path shows the warning unless you go looking in the context report, and the model's behaviour just looks slightly off.

The natural workaround is a hardlink, and it fails in exactly the same invisible way: rewriting the canonical file atomically gives it a new inode, and the hardlink silently keeps the old contents. So the substitute trades a visible refusal for a stale copy with no signal at all.

## Provenance

The refusal was introduced by `fa5108f54 fix(security): reject symlinked project context` (2026-06-21). That commit's message and its tests are about **project** instruction files and repo constitution files — "Refuse symlink-backed project instruction and repo constitution files before rendering them into model-facing context" — and its diff does not touch the user-level loader at all. Because both layers share `load_context_file`, the user-level layer inherited the refusal.

The guard itself is right where it was aimed: a link planted inside an untrusted checkout can point the loader outside the repository, and that path stays refused. The user-level layer is the operator's own file in `$HOME`, where that escape does not apply.

## Proposed behavior

Follow the link for the user-level candidates only, resolving to the target and opening the resolved path (`open_context_file` uses `O_NOFOLLOW`, so the resolved path — not the link — is what must be handed to it). Workspace files, rules directories and the repo constitution keep the existing refusal. A fix along those lines is open as #6114.

## Environment

- OS: Arch Linux
- codewhale version: 0.9.13 (source build); tree inspected `2d37e1486`
- Install method: `cargo install --git https://github.com/Hmbown/Codewhale codewhale-cli`
- Model/provider: `command_code` (OpenAI-compatible gateway)
- Terminal app: tmux (`TERM=tmux-256color`)
- Shell: fish
- `codewhale doctor`: not pasted — the relevant path here is file resolution, which the context report shows directly.

## Logs

Warning text as emitted:

```
Refusing symlinked context file /home/<user>/.deepseek/AGENTS.md
```

Reproduces on the current `main` (`2d37e1486`).

7jrxt42BxFZo4iAnN4CX:
## What

The user-level instruction layer (`~/.codewhale/AGENTS.md`, `~/.agents/AGENTS.md`, `~/.deepseek/AGENTS.md`) now follows a symlink to its target instead of refusing it. Files inside a checkout keep the existing refusal.

Closes #6115

## Why

`load_context_file` refuses symlinked files, and `load_global_agents_context` collects that refusal as a **warning**, not an error. So a symlinked user-level file dropped the whole layer silently: the model ran without the global instructions and nothing told the user.

The guard itself is right for its original target. It was introduced in `fa5108f54 fix(security): reject symlinked project context`, whose message and tests cover project instruction files and the repo constitution — a link planted inside an untrusted checkout can point the loader at anything on the machine. That threat does not apply to the user-level layer: the file lives in the operator's own `$HOME`, and pointing it at a shared file is the ordinary way to keep one instruction set across several agents (`~/.config/<agent>/AGENTS.md` -> `~/AGENTS.md`). A hardlink only substitutes by accident, and detaches silently whenever the canonical file is rewritten atomically (new inode) — which is the failure mode this replaces.

## Changes

- `crates/tui/src/project_context.rs`: the loader keeps its exact behaviour but is split so the symlink policy is a parameter; `load_global_context_file` resolves the link and continues with the target. `open_context_file` opens with `O_NOFOLLOW`, so the resolved path must be the one handed to it, never the link.
- Two tests: a symlinked user-level `AGENTS.md` is read, and a symlinked workspace `AGENTS.md` is still refused.

## Verification

- `cargo test -p codewhale-tui --lib project_context` — **77 passed, 0 failed** (12432 filtered out).
- **Negative control** (so the new test is a reproducer rather than an ornament): with the following policy flipped back to `false`, `test_symlinked_global_agents_is_followed` fails with

  ```
  a symlinked user-level AGENTS.md must be read: ["Refusing symlinked context file /tmp/.tmp2eogYp/.deepseek/AGENTS.md"]
  ```

  and with the policy on it passes.
- Existing symlink guards are unaffected: `rules_rejects_symlinked_directory`, `rules_rejects_symlinked_files`, `symlinked_foreign_fragment_does_not_claim_importable_instructions` and `signature_does_not_follow_symlinked_foreign_fragment_directories` all still pass, so the workspace escape stays closed.

## Notes

- A broken symlink or a symlink to a directory still produces a load error that is reported as a warning and falls through to the next candidate — the same shape as before this change.

### DIFF
--- a/crates/tui/src/project_context.rs
+++ b/crates/tui/src/project_context.rs
@@ -977,7 +977,7 @@ fn load_global_agents_context(workspace: &Path, home_dir: Option<&Path>) -> Opti
         let path = join_relative_components(home, candidate);
 
         if context_candidate_exists(&path) {
-            match load_context_file(&path) {
+            match load_global_context_file(&path) {
                 Ok(content) => {
                     let mut ctx = ProjectContext::empty(workspace.to_path_buf());
                     ctx.instructions = Some(content);
@@ -1014,6 +1014,48 @@ fn generate_ephemeral_context(workspace: &Path) -> Option<String> {
 
 /// Load a context file with size checking
 fn load_context_file(path: &Path) -> Result<String, ProjectContextError> {
+    load_context_file_with_symlink_policy(path, false)
+}
+
+/// Load a user-level context file, following a symlink to its target.
+///
+/// The refusal in [`load_context_file`] protects checkouts: a link planted in
+/// an untrusted repository could point the loader at anything on the machine.
+/// The user-level layer is the operator's own file in `$HOME`, where that
+/// escape does not apply and a symlink is the ordinary way to share one
+/// instruction set between agents (`~/.deepseek/AGENTS.md` -> `~/AGENTS.md`).
+/// Refusing it there failed silently: the refusal is collected as a warning,
+/// so the whole user-level layer was dropped without any visible error.
+fn load_global_context_file(path: &Path) -> Result<String, ProjectContextError> {
+    load_context_file_with_symlink_policy(path, true)
+}
+
+/// Resolve a symlinked context file to its target.
+///
+/// `open_context_file` opens with `O_NOFOLLOW`, so the resolved target is what
+/// must be opened, never the link itself.
+fn resolve_symlinked_context_path(path: &Path) -> Result<PathBuf, ProjectContextError> {
+    let metadata = fs::symlink_metadata(path).map_err(|source| ProjectContextError::Metadata {
+        path: path.to_path_buf(),
+        source,
+    })?;
+    if !metadata.file_type().is_symlink() {
+        return Ok(path.to_path_buf());
+    }
+    fs::canonicalize(path).map_err(|source| ProjectContextError::Metadata {
+        path: path.to_path_buf(),
+        source,
+    })
+}
+
+fn load_context_file_with_symlink_policy(
+    path: &Path,
+    follow_symlinks: bool,
+) -> Result<String, ProjectContextError> {
+    let resolved = follow_symlinks
+        .then(|| resolve_symlinked_context_path(path))
+        .transpose()?;
+    let path = resolved.as_deref().unwrap_or(path);
     let metadata = fs::symlink_metadata(path).map_err(|source| ProjectContextError::Metadata {
         path: path.to_path_buf(),
         source,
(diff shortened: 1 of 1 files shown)


Hmbown (approved):
Reviewed the trusted global-file boundary and retained workspace symlink guards. Local context tests: 77 passed, 0 failed; npm tests: 521 passed, 0 failed; check:web and formatting passed. All applicable hosted checks, including macOS, are green at 77d407eca15bea2659824189f5f33eed5e8fd16e. No open review threads.

PR #6170, opened 2026-09-14
fix(weixin-bridge): add Quick Start part to weixin-bridge's doc to make it more easy to be started for potential users

VincentCorleone:
## Summary

The Weixin bridge could not be used as documented. `integrations/weixin-bridge/README.md` pointed at a non-existent `/opt/codewhale/weixin-bot-bridge` path, gave no runtime startup command, and named an env file that nothing reads. Worse, once it did run, the first incoming WeChat message failed with `EACCES` and was dropped silently: the thread map defaulted to `/var/lib/codewhale-weixin-bot-bridge` regardless of `WEIXIN_STATE_DIR`, so every message failed on write inside the `getUpdates` catch. A first-time user saw a scannable QR, a successful login, and then no reply and no actionable error.

This PR defines a Quick Start in the README (modeled on `integrations/wecom-bridge/DEPLOYMENT.md`) and implements the behavior that section describes.

**Fix the write path and make the failures visible**

- Default `WEIXIN_THREAD_MAP_PATH` inside `WEIXIN_STATE_DIR`. Setting only `WEIXIN_STATE_DIR` left the map at the old absolute path, so every incoming message failed on write and the message was silently dropped.
- Create the state dir in `saveSyncBuf`, matching `saveAccount`. The poll cursor can be the first persisted write on a fresh install.
- Probe thread-map writability at startup and exit with a readable message, instead of failing later inside the message loop.
- Log the resolved thread map path next to the state dir.

**Make the login QR scannable in place**

- Add `src/qr.mjs`: a zero-dependency byte-mode QR encoder (versions 1-10, ECC level L) rendered with half-block glyphs, matching the Rust side's `Dense1x2` renderer in `crates/tui/src/runtime_api.rs`. The bridge keeps its no-npm-deps property.
- Keep printing the URL below the QR so a terminal that mangles half-block glyphs still has a path through.

**Add the missing local setup path**

- `scripts/dev-up.sh` plus `npm run bridge` start the runtime and the bridge in one terminal, generate a shared `CODEWHALE_RUNTIME_TOKEN`, wait for `/health`, and clean up the runtime on `Ctrl-C`.
- `.gitignore`: ignore `integrations/*-bridge/.state/` (it holds the live bot token in `account.json`), and re-include `integrations/*/scripts/**`, which the global `*.sh` rule was hiding — without that exception `package.json` would ship pointing at a file that never landed.

**Simplify the Quick Start and add screen captures**

- Restructure around the two stages a user performs: `终端发起` (two-terminal and single-terminal) and the two WeChat-side steps, `微信端扫码接应` and `微信端验证信道效果`.
- Add two captures under `docs/assets/` showing the QR login and a real `/status` exchange. Both are redacted before publishing: the local username is removed from the logged paths, the one-time `qrcode=` login token is masked, and the QR's bottom-right corner is blacked out so the published code can no longer be scanned.
- Correct the `首次配对` step 3 wording: the bridge returns runtime status when the sender is authorized, and returns the refusal carrying the `user_id` only when they are not allowlisted.

## Testing

The Rust gate boxes above are **not applicable** to this change and were not run: no Rust source is touched. The diff is confined to `integrations/weixin-bridge/` and `.gitignore`. I did not run them because they would answer nothing about this diff.

Commands actually run, all from `integrations/weixin-bridge/`:

```
npm run check    # node --check on index.mjs, lib.mjs, qr.mjs + bash -n scripts/dev-up.sh
npm test         # node --test test/*.test.mjs -> tests 12, pass 12, fail 0
```

Direct product evidence, not assertions over internals:

- `npm run bridge` completed QR login against the real iLink endpoint, and the same bridge then answered `/status` from a real WeChat account with `runtime=ok`. That exchange is the second added capture.
- The terminal QR output was decoded back to the exact login URL with OpenCV (`cv2.QRCodeDetector`), 9/9 payloads including UTF-8, so the printed code is machine-readable rather than merely QR-shaped.
- Startup fail-fast verified against an unwritable path: the bridge exits immediately with `Thread map directory is not writable: ...` instead of dropping messages later.

## Checklist

- [ ] This PR adds a new layer/module/abstraction — it names or deletes the layer it replaces

Notes on the unchecked boxes:

- **New layer/module:** `src/qr.mjs` is a new module, but it is not a layer over an existing one. It has no predecessor to name or delete, and the bridge has no prior QR dependency (`package.json` has no `dependencies` field at all). It is documented as matching the existing Rust-side renderer so the two stay consistent.
- **TUI behavior:** no TUI code changed, so there is nothing to verify there. The added screenshots are captures of the bridge's own terminal output, not a UI change.
- **Co-authored-by:** the commit carries `Co-Authored-By: DeepSeek-V41-Flash <service@deepseek.com>`, which is DeepSeek's published `Contact Email` from its Terms of Use. It is not a GitHub numeric noreply address, so I left this box unchecked rather than claim credit registration that will not happen. Maintainers may drop the trailer at merge time if they prefer it gone.

## Notes for the reviewer

- I am not in `.github/APPROVED_CONTRIBUTORS`, so the PR gate may hold the workflow runs at `action_required`. A `/lgtm` on this PR should clear it.
- Scope note: I fixed the bridge to the point where it runs and answers, and stopped there. The bridge is still not registered in `crates/tui/src/remote_setup/registry.rs`, so `/setup bridge` and `codewhale remote-setup --bridge weixin` do not list it. That is a separate change against the Rust registry and I did not bundle it here.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6170" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

No-Issue: fixes a local setup defect in integrations/weixin-bridge found by running it; no tracking issue exists.


Hmbown:
thank you so much for this!!! will get to reviewing it shortly

Hmbown (changes requested):
Thanks for this — the EACCES fail-loud probe and the thread-map-inside-stateDir default are exactly right, and the dev-up health wait closes the pairing race. I verified the Devin findings by hand-tracing; two are real and need fixing before merge:

**1. `placeData` drops data columns 3 and 0 for every version (blocker).** `if (col <= 6) col -= 1` shifts on *every* low pair, not once around the timing column. Trace for v1 (size 21): pairs visited are (20,19)…(8,7), then 6→5 gives (5,4), then 3→2 gives (2,1), then 0 stops the loop — columns 3 and 0 are never written, so trailing codewords are silently truncated and the QR misdecodes. The comment rationalizing `<=` is wrong: only column 6 needs skipping. Fix: `if (col === 6) col -= 1;` (matches Nayuki qrcodegen). Please also add a placement test: encode a payload whose codewords are all `0xFF`, then assert known cells in columns 0–3 are dark — that fails today and passes after the fix.

**2. `invert` destroys half-block information (latent — the caller never passes it, but fix while here).** The replace maps `▀`/`▄` to space; correct inversion is space↔`█`, `▀`↔`▄`.

Considered and declined: duplicate-impl (a Node bridge cannot reuse the Rust crate; the mirror is documented), allow-unlisted default (first-pairing necessity on a localhost dev script), health-check trust (localhost dev-only). Two nits, non-blocking: probe `WEIXIN_STATE_DIR` too when the thread map lives elsewhere (finding 3 is a real micro-gap), and have the two-terminal README path export `WEIXIN_STATE_DIR` up front so non-root users never meet the probe error.

All checks are green and the scope is right — happy to approve on the `=== 6` fix + placement test.

Issue #6174, opened 2026-09-14
ACP: session/new returns a provider-prefixed sessionId that session/load cannot resolve


sequico:
## Description

An ACP client cannot resume a session it created with `session/new`: the id `session/new` returns is not resolvable by `session/load`, while `session/list` reports the bare form that is.

Three observations from one ACP server, all in a throwaway `CODEWHALE_HOME` (no turn sent, no model call, no real session store read or written):

1. `session/new` returns a **provider-prefixed** id.
2. `session/list` returns **bare** uuids.
3. `session/load` accepts **only the bare** form. The prefixed form fails with `-32602`, and it still fails when a session file is named exactly like the prefixed id.

An ACP client has no id for a session it just created other than the one `session/new` returned, so a client that persists and replays that id — the normal thing to do — can never resume the session.

## Steps to reproduce

Everything below runs against a scratch home; nothing outside it is touched.

```bash
export CODEWHALE_HOME=$(mktemp -d)
ac() { printf '%s\n' "$@" | codewhale serve --acp; }
```

**1. What `session/new` advertises:**

```bash
ac '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{"fs":{"readTextFile":true,"writeTextFile":true}}}}' \
   '{"jsonrpc":"2.0","id":2,"method":"session/new","params":{"cwd":"<any-directory>","mcpServers":[]}}'
```

```json
{"jsonrpc":"2.0","id":2,"result":{"sessionId":"codewhale-95bb3fa8-8784-4eed-b34a-59072e9e9823","modes":{...},"models":{...}}}
```

Note the `codewhale-` prefix on the id.

**2. What `session/list` advertises for existing sessions:**

```json
{"jsonrpc":"2.0","id":2,"result":{"sessions":[{"sessionId":"c45e7916-ab6a-4585-8d09-d3649fd92f38","title":"...","cwd":"...","createdAt":"...","updatedAt":"...","messageCount":232}, ...]}}
```

Bare uuid, no prefix.

**3. What `session/load` accepts.** With exactly one persisted session file present, `415488d4-14a0-46ff-8477-7a3989a36de2.json`:

bare id — succeeds:

```json
{"jsonrpc":"2.0","id":2,"result":{"sessionId":"415488d4-14a0-46ff-8477-7a3989a36de2","modes":{...},"models":{...},"configOptions":[...]}}
```

[...]

Issue #6150, opened 2026-09-13
Op::SendMessage god-payload → TurnSpec, and prove the UI never awaits send() on the input path


Hmbown:
From the 0.9.14 refactor backlog.

`Op::SendMessage` (`ops.rs` ~109+) carries ~20 fields incl. `Box<ResolvedRuntimeRoute>`, `Box<CompactionConfig>`, hook executor, provenance. Extract `struct TurnSpec` so new per-turn authority fields stop accreting into the enum arm; every ACP/app-server fix (#6088, app-server client issue) touches this anyway.

Also: `ENGINE_OP_CHANNEL_CAPACITY = 32` (`engine.rs` ~90). If anything on the poll thread `send().await`s into a full op channel, input wedges while the engine is mid-turn. `handle.rs` exposes async `send` — trace the call chain from the event loop. If it is already `try_send` + error toast, close this half with a comment.

Size: S + investigation.

Hmbown:
Triage (v0.9.14 run, partially verified — not in https://github.com/Hmbown/Codewhale/pull/6161): the input-path half is **already correct** for SendMessage — `dispatch.rs:844` reserves capacity off the render thread and hands off with `send_reserved_op` (:887), with the invariant written in the comment. Residual send().await sites to audit/convert: `handlers.rs:2086` (CancelSubAgent), `:2570` (SyncSession). The 20-field god-payload → TurnSpec is the large slice and should follow the audit's conclusion.

Hmbown:
Half landed on v0914-chunk2 (PR #6175): the input path no longer awaits a full op channel — `4d70e050` keeps `send()` off the UI thread's poll path. The `Op::SendMessage` → `TurnSpec` extraction is still open: ~136 construction/match sites, mostly tests.

Hmbown:
[...]

Issue #6035, opened 2026-09-10
Model pins don't propagate: fleet members and agent profiles keep a retired id after the provider moves on


Hmbown:
## Problem

A model id is pinned independently in at least six places, with no shared owner and no migration when a vendor retires an id. On one real machine, after DeepSeek released V4.1 Flash as `deepseek-flash` and dropped `deepseek-v4-flash` from the account roster:

```
~/.codewhale/config.toml:2      default_text_model = "deepseek-v4-flash-vision-exp"
~/.codewhale/config.toml:15     [providers.deepseek]      model = "deepseek-flash"
~/.codewhale/settings.toml:38   default_model             = "deepseek-v4-flash-vision-exp"
~/.codewhale/settings.toml:51   [provider_models] deepseek = "deepseek-v4-flash-vision-exp"
~/.codewhale/agents/scout.toml:4   model = "deepseek-v4-flash"   provider = "deepseek"
~/.codewhale/agents/custom.toml:4  model = "deepseek-v4-flash"   provider = "deepseek"
~/.codewhale/fleets/default.toml:19 model = "deepseek-v4-flash"  provider = "deepseek"
```

**14 of 19 DeepSeek pins named ids the account roster no longer lists.** Changing the provider's model in one place moved the main session to `deepseek-flash` while every subagent kept running `deepseek-v4-flash` — visibly, in the Agents panel, three workers on a retired id while the parent was on the new one.

## The nuance that makes this not a simple "purge old ids"

`deepseek-v4-flash` is **still served** — the subagents worked, producing 277k/142k/270k tokens. It is absent from `/v1/models` but the endpoint still answers. So this is soft-deprecation, not breakage.

It also legitimately still exists in Codewhale's catalog, because third-party hosts (Fireworks, OpenCode Zen, SiliconFlow, Together, Baseten…) serve their own copies and are unaffected by a first-party roster change. A fleet pinning `deepseek-v4-flash` on one of *those* providers is entirely correct.

So the rule cannot be "retire the id globally". It has to be **per route**: this id is no longer offered *on this provider*.

## What is missing

1. **No single owner for "which model does this role use".** Provider default, global default, per-provider setting, agent profile and fleet member each hold their own copy.
2. **No propagation.** Changing a provider's model does not reach agent profiles or fleet members that name the same provider.
3. **No deprecation signal.** Nothing notices that a pinned id vanished from the roster of the provider it is pinned to.
[...]

Issue #5587, opened 2026-08-23
Dead-code sweep phases 2-4: 75 test-only markers, ~242 stale allows, blanket-allow conversions


Hmbown:
Continuation of the audited sweep (phase 1 landed in e5ca0aa86: eight verified-dead items removed, facts-drift contract documented). Full audit of all 379 `allow(dead_code)` sites in crates/tui/src, classified with whole-workspace reference indexes:

- **TRULY DEAD, Tier B/C (18 remaining)**: paste_burst on_plain_char_no_hold / try_append_char_if_active, active_cell forget_tool, world_state render_incremental_text / is_noop / render_world_diff (whole WorldStateDiff surface is unwired), cost_status with_identity, tools/spec with_sandbox_policy, llm_client RetryConfig with_request_timeout / with_total_timeout, TranslationStatus enum, mcp disconnect_all, registry execute_full_with_context, runtime_threads resume_thread_with_agent_rebind (+ its only-caller helper collect_agent_rebind_hints), settings persist_provider_model_selection_as_default, provider_lake all_catalog_providers, ReasoningEffort cycle_next_for_provider, skills/audit AUDIT_MAX_PACKAGE_BYTES / AUDIT_MAX_FILES. Tier C items (mcp/registry/runtime_threads/settings) deserve an intent check before deletion.
- **Dead under blanket allows**: features.rs enabled_features / feature_spec_by_key; sandbox/policy.rs trait SandboxExecutor; fleet/profile.rs load_personal_agent_profiles_tolerant; tui/git_status.rs list_worktrees.
- **TEST-ONLY (75 markers)**: convert to #[cfg(test)] (largest honest win; full list in the audit). llm_client/mock.rs is a whole-module cfg(test) candidate.
- **STALE ALLOWS (~242)**: attribute guards a prod-used item. Authoritative pass: `RUSTFLAGS="--force-warn dead_code" cargo check -p codewhale-tui --all-targets` (overrides #[allow]), then drop every attribute it does not flag. tools/handle.rs both impl-level allows and settings.rs TuiPrefs impl are fully stale.
- **LEAVE ALONE (19)**: cfg_attr-gated platform items; PUBLIC_SANDBOX_BACKENDS (external facts-drift contract, now labeled); OpenRouter wire-capture fields; ToolExecGuard RAII variants; rlm.rs type-shape assertion.
[...]

PR #6175, opened 2026-09-15
v0.9.14 slice run 2: lazy MCP, session recovery + picker UX, launch remedy row

### LINKED ISSUE #5715: Session recovery is invisible to the model: after a force-quit, the agent cannot find what it was doing
## Problem

Force-quit Codewhale mid-task and the next run has no idea what it was doing. The
work is not lost — it is on disk — but the model cannot see it.

Reported by **Gary** in the Codewhale WeChat group: *"If you force-close codewhale
during task execution, next time you run it, it doesn't remember what it was doing
before."*

**AIGC 老炮** added the precedent worth stealing: Cursor has the same problem, but
an agent there can search the local cache records and recover the conversation
content. The fix is not more persistence — it is making the persistence we already
have legible to the model.

## What already exists

- `~/.codewhale/sessions/` — session JSON per session
- `~/.codewhale/sessions/checkpoints/` — crash-recovery checkpoints (#5697 made the
  durable completion commit preserve these)
- `crates/tui/src/session_manager.rs` — the store and checkpoint machinery
- `/resume`, `/session`, `/sessions` — human-facing commands
  (`crates/tui/src/commands/groups/session/`)

## The gap

There is no model-facing surface over any of it. The tool catalog has
`native_memory`, `remember`, `todo` — nothing that reads the session store. So
recovery depends entirely on the *user* knowing to type `/resume`. A fresh session
that follows an interrupted one has no way to notice it was interrupted, let alone
find out what it was mid-way through.

## Proposal

1. **A bounded `session` tool** — search and read prior sessions *for the current
   workspace*. Follow the `native_memory` recall pattern: cap the output and point
   at a bounded getter for one entry rather than dumping a transcript.
2. **A startup signal** — when a session begins in a workspace whose previous
   session ended mid-turn, put that fact in context, with the handle needed to
   recover. The model should be able to say "you were partway through X" without
   being asked.

## Constraints

- Bounded output. A session transcript is unbounded; the tool must never inline one.
- Workspace-scoped. Sessions contain whatever was on screen — never surface one
  workspace's sessions inside another.
- No second prompt authority, and no second `run_turn`.
- Recovery is a *proposal*, never an automatic resume: the model may report that
  prior work exists, and the user decides.

No-Issue: N/A — this is the issue.

### LINKED ISSUE #6014: feat(tui): Session Picker UX — hide empty auto-sessions, highlight current session, page-scroll the list, widen the list pane
## Problem

The Session Picker (`Ctrl-R` / `/sessions`) has four usability gaps that make it hard to navigate and trust:

1. **Empty auto-created sessions clutter the list.** Sessions with 0 messages and the default title ("Session") appear in the picker, even though the same definition (`is_empty_auto_created_session`) is already used to hide them from the launch "recent work" list and from `--continue`/auto-resume. The picker is the odd one out — it shows empty sessions while the other surfaces hide them.
2. **The current session is not highlighted, and its id is not discoverable.** Opening the picker from a live session does not indicate which row is the session you are in. There is no command or surface that prints the current session id, so you can only find it by searching text — and even then the list shows only a truncated 8-char id.
3. **PageUp/PageDown scroll the history pane, not the session list.** The list scrolls only by one row at a time (arrows/`j`/`k`), so navigating a long list is slow; PgUp/PgDn unexpectedly move the preview pane instead.
4. **The list pane is narrow and titles truncate.** The two-pane layout gives the list only ~36% of width (history gets ~64%), so session rows (id + title + count + mode + relative time) are squeezed and titles get cut on the right even on wide terminals.

Affected: anyone who uses `/sessions` / `Ctrl-R` to resume or inspect sessions, especially with many sessions per workspace.

## Proposed solution

1. **Hide empty auto-created sessions in the picker** — apply the existing `is_empty_auto_created_session` filter in the picker's session projection so it matches the launch list and `--continue` behavior. (One filter condition; no new concept.)
2. **Highlight and identify the current session** — mark the row that matches `app.current_session_id` (e.g. a `current` tag / distinct style) when the picker opens from a live session; and expose the current session id somewhere usable (status line hint or a `/session id`-style command), closing the same gap behind #6001.
3. **Make PgUp/PgDn scroll the list when it owns focus** — page the session list by a viewport per press (like the history pane already does for previews), with focus rules for list vs preview (exact behavior is an open question; see below).
4. **Give the list pane a fairer share of width** — evaluate 50/50 or list-first split when there is enough terminal width, and consider dropping the id from the row when a meaningful title exists (id stays as a secondary/status detail).

## Use case

- Open `/sessions` after a long work week: the list is clean (no "Session" rows), the row you are working in is visibly marked, and you can page down through 40+ sessions to find the one you want instead of arrow-keying line by line.
- You need to `codewhale exec --session <id>` or reference the current session: id is shown without digging.
- On a 141-column terminal the list shows actual titles instead of truncated ones.

## Alternatives considered

- **Do nothing** (rely on `/` search): works for finding *other* sessions, but doesn't help visibility of the current one or the empty-row noise.
- **Revert to the pre-#5262 flat list**: no — the session tree (#5262) is better; this issue is about the picker, not the model.
- **Auto-delete empty sessions instead of hiding them**: discussed; hiding is consistent with the existing launch/`--continue` behavior and is non-destructive (sessions may gain messages later via `exec --session`).
- For the width issue: keep history at 64% but make rows denser (drop id when title is meaningful) — a lighter change than re-layout.

## Impact

Every `/sessions` user. Medium-high: the empty-session fix removes visible noise, the current-session highlight answers a common "which one am I in" question, and paging is the difference between a usable and an annoying list past ~20 sessions. All four are picker-only changes; no session-store or data-model changes.

## Additional context

- Related: #2934 (sidebar sessions panel — introduced the preselect-by-id flow), #5262 (session tree / journal — why sessions exist and how titles are derived), #4629 (related "must not create empty Work" session leak), #6001 (open — resume hint omits session id; same "id opacity" root).
- Code-verified behaviors (main @ aedb88b9):
  - Picker opens via `SessionPickerView::new` (no preselection) on the Ctrl-R / launch paths; `new_selecting(session_id)` exists but is used only by callers that already know the target id (sidebar rail, command flows with a preselected session, foreign-session import) — the current-session row is never auto-highlighted on the plain picker path.
  - List keys: arrows/`j`/`k` move selection one row; `PageUp`/`PageDown` call `scroll_history` (preview pane), not list scrolling.
  - Row format: `truncate_id` (8 chars) + title (≤32) + message count + mode + relative time, whole row truncated to pane width.
  - Two-pane split: history 64% / list 36% at widths ≥ 95 columns.
  - `is_empty_auto_created_session` (0 messages + default title) is applied in launch-recent and `--continue`, but not in the picker's `select_sessions` projection.

## Open questions (for maintainers)

- **Focus model for paging**: should PgUp/PgDn always page the list (relegating preview scrolling to arrows/secondary keys), or switch focus list↔preview first (e.g. Tab), or page whichever pane is focused? (Variant A: list-only paging; B: focus-switch + per-pane paging; C: keep preview on PgUp/PgDn, add dedicated list page keys.)
- **Current-session indicator**: inline tag (`current`), distinct color, or a footer/status line showing `session id` for the active session — which is preferred given the TUI visual language?
- **Width**: rebalance to 50/50 vs dense rows (drop id when a real title exists) vs keep layout and only fix truncation math.

## Acceptance criteria

- [ ] Empty auto-created sessions (0 messages, default title) are not listed in `/sessions` / `Ctrl-R` (consistent with launch list and `--continue`)
- [ ] When the picker is opened from a live session, the current session row is visually marked; some surface prints the current session id (status line or command)
- [ ] `PgUp`/`PgDn` page the session list by one viewport per press (per the chosen focus model)
- [ ] Session rows show meaningful titles without truncation on 141-column terminals (list pane widened or row densified)
- [ ] Archived / all-workspace toggles (`e`, `x`, `a`) and `/` search still behave unchanged

### LINKED ISSUE #6033: MCP: connect servers lazily at point of use instead of eagerly at boot
## Problem

Codewhale connects every configured MCP server during session boot. With 23 servers configured — 10 sitting at `auth=not-logged-in`, three that download packages on first run (`npx -y` ×2, `uvx`) — boot does a large amount of work for servers the session will never use, and a single slow one is felt by the user.

0.9.13 bounded the two unbounded waits that made this catastrophic:

- `6d83a3d8d` — `/mcp` no longer waits forever on a cold boot.
- `30bf8543a` — the **turn path** (`build_turn_tool_registry_and_catalog` → `wait_for_explicit_mcp_boot`, `engine.rs:4459`) no longer parks indefinitely. Its inner `tokio::select!` awaited `rx.recv()` with no deadline, so a server that accepts a connection and then goes quiet stopped the session answering.

Those are deadlines on a design that still does the work. The founder's comparison was the right one: other agents tolerate a dozen unconnected servers with no lag, because they never connect them.

## The shape worth adopting

`refs/codex/codex-rs/codex-mcp/`:

- Per-server `startup_timeout`, default 30s, overridable per server (`rmcp_client.rs:102`).
- All servers start concurrently in a `JoinSet` (`connection_manager.rs:252`).
- **`defer_startup`** with a `startup_trigger` watch channel (`:88`, `:131-132`, `:589`) — a server can be registered without being connected.
- The timeout applied **at the point of use** (`:919`), not at boot.

The consequence is the property we want: *a server nobody uses is never connected, so it can never block anything.*

Codewhale already holds the intent — `start_mcp_session_boot` documents that "optional servers stay in the background unless the task explicitly selects their tools". The gap is that they are still connected eagerly.

## Proposed work

1. Register configured servers without connecting them.
2. Resolve a connection lazily when a tool from that server is first selected, under a per-server timeout.
3. Make `startup_timeout` per-server configurable, since `npx`/`uvx` first-run downloads legitimately need longer than an HTTP dial.
4. Keep the failure reporting that landed in 0.9.13 — the launch-screen block naming failed and unauthorized servers (`04a6e41b3`) is how a user finds out, and lazy connection must not make failures invisible instead of merely slow.

## Related

- #6030 — a configured server that authenticates mid-session cannot be attached at all, and `/mcp reload` crashes. Fixing lifecycle properly likely subsumes part of that.
- The founder's setup is the natural test case: 23 servers, 10 unauthenticated, 3 that download on first run.

### LINKED ISSUE #6037: Fleet members over-pin their model, so nothing follows the provider when a vendor retires an id
## The design is already right; the written data is not

`FleetMember` (`crates/tui/src/fleet/store.rs:125`) already supports everything needed:

* `model: Option<String>` — *"Absent with* `provider` *absent = inherit the session route (the operator route when the Fleet has one)."*
* `shortlist: bool` — *"A role-less model choice, not an executable roster member."*

So a member that omits `model` **already follows the provider automatically**. No new mechanism is required.

## What a real machine actually contains

`~/.codewhale/fleets/default.toml`, 27 members:

|  | count | consequence |
| -- | -- | -- |
| explicit `model` pin | **20** | can never follow the provider |
| inherits the session route | 7 | follows correctly |
| `shortlist = true` | **0** | the role-less form is unused |

When DeepSeek released V4.1 Flash as `deepseek-flash` and dropped `deepseek-v4-flash` from the account roster, the main session moved but every pinned member stayed on the retired id — visible in the Agents panel as three workers on `deepseek-v4-flash` while the parent ran `deepseek-flash`.

Note the old id is **still served** (those workers produced 277k/142k/270k tokens), so this is soft-deprecation, not breakage. And `deepseek-v4-flash` legitimately still exists for third-party hosts (Fireworks, OpenCode Zen, SiliconFlow, Together), which are unaffected by a first-party roster change. Any fix must be **per route**, never a global purge of an id.

## Test fixtures are in a real user's shipped fleet

Four members are not roles at all — and they are **executable role members**, not `shortlist` rows:

`orgcustom-router-id`, `model-a`, `model-b`, `legacy-saved-model`

Finding how these reached a user's default fleet matters independently of the rest: something is writing test data into shipped defaults.

## Work

1. **Find what writes the pins.** `/fleet add`, the ⇧F path from `/model`, and whatever generated the default fleet all appear to write an explicit `provider` + `model` even when inheriting would be correct. Default to inheritance; make the pin the deliberate opt-out it is documented to be.
2. **Use** `shortlist: true` for role-less model choices. Twenty pinned members and zero shortlist rows suggests the flag is never set by the code paths users actually take.
3. **Surface drift, do not silently rewrite.** When a pinned `(provider, model)` no longer appears in that provider's live roster, say so in `/status` and the fleet view, and offer a scoped migration. A silent route rewrite is forbidden by C04 and `AGENTS.md`, and a pin is a deliberate choice.
4. **Trace the fixture leak** and remove the residue from shipped defaults.

## Not in scope

Do **not** remove the `fleet` concept or the `agents` profile layer. Both were examined and both are load-bearing — see the sibling issue on that question. This is a data and defaults problem, not a modelling one.

### LINKED ISSUE #6085: feat(tui): launch-screen MCP block — decide copy/paste (and click) affordances; it is read-only by design today
## Context

The launch card prints the MCP status block before the first message. Commit `684bbdc1d` (2026-09-11) consolidated it to summary + at most one problems row (`✕` failures, `⚠` logins, remedy on the tail, names fold into `+N` before the command sheds).

The 0.9.13 remaining-work list still carries: "Launch-screen MCP block: needs copy/paste and click affordances."

## What exists today (source-verified 2026-09-11 at 89621d5a5)

- `crates/tui/src/tui/underwater.rs:1671-1676` — the rows are deliberately informational, not selectable: "A fourth interactive row would have to join `LaunchRowId` and the paint/click/keyboard ordering the card shares, and it would buy nothing … the composer below has focus from the first frame, so the block simply prints the exact command to type."
- `LaunchRowId` is `NewSession` / `Recent` / `SeeAll` only (`underwater.rs:110-123`; click mapping `:226-232`).
- The problems row prints a real remedy string (`/mcp login <name>`, `/mcp`) — but the user retypes it by hand, and there is no way to copy the failing server names.
- Generic `ViewEvent::CopyToClipboard` exists (`tui/ui/handlers.rs:1247`); the block never emits it.

## Decision requested

Either:

- add a copy affordance for the blocked-server names / remedy (keyboard-first, matching the card's grammar), and optionally a click entry that runs `/mcp`; or
- record explicitly that read-only is final, and drop this from the remaining-work list.

## Notes

Not decided here on purpose — the non-interactive choice is recorded in-code and may be right. This issue exists so the remainder is either built or closed.

### LINKED ISSUE #6099: feat(exec): let `codewhale exec` opt into hooks and typed permission rules
## Problem

Hooks and the sibling `permissions.toml` are enforced only in the interactive TUI and its
engine loop. `codewhale exec` — the headless one-shot path, which is exactly where an
unattended run needs a guard most — fires no hooks and evaluates no typed rules.

Both are documented behaviours rather than defects:

`docs/HOOKS.md:22` and the Scope table:

> | `codewhale exec` (headless one-shot) | no |

and the section intro: hooks "are a **TUI runtime feature**. Every firing point lives in
the interactive TUI and in the engine turn loop it drives."

The typed-rule helpers have the same shape — the only non-test call sites are
`crates/tui/src/core/engine/turn_loop.rs:3259` and `crates/tui/src/acp_server.rs:700-708`.

The result is a posture gap rather than an error: the operator's deny rules and hook gates
are silently absent from the unattended run, and nothing says so. In headless mode a call
that would need approval is decided by `auto_approve` alone — approve when set, deny
otherwise (`crates/tui/src/exec_agent.rs:757-764`), with
`approval_mode: if auto_approve { Bypass } else { ... }` at `:484-495`. With Full Access
configured for the session, that means no gate at all.

## Proposed solution

An explicit opt-in on the headless path, defaulting to today's behaviour so no existing
script changes:

- `codewhale exec --hooks` — fire the configured hooks. At minimum the steering
  `tool_call_before` (so a deny can block a tool call) and `shell_env`; observer events
  could be a separate switch if noise is a concern.
- `codewhale exec --permissions` (name is a suggestion) — evaluate the sibling
  `permissions.toml` typed rules before dispatch, at least `deny`.

Open questions worth settling in the same change:

- **Failure policy.** A foreground `tool_call_before` hook that cannot answer is
  fail-closed only when that hook sets `continue_on_error = false`; otherwise the call
  proceeds. An operator arming a guard for an unattended run most likely wants the strict
  form as the `exec` default, or at least a visible warning that it is not strict.
- **`ask` outcomes.** Nothing can be prompted headlessly, so `ask` has to resolve the way
  it does today (fail-closed) — and should say so rather than silently becoming a deny.
- **Naming.** `--hooks` / `--permissions` is a placeholder; any shape that reaches both
  surfaces under conscious operator control is fine.

## Use case

Running the agent unattended — CI, cron, a long batch job, an overnight task — with the
same safety policy that already applies interactively. Today the operator has two options:
carry the policy into the script by hand (different rules, no hook gates, no receipts), or
accept that the unattended run has no guard. Because an unattended run cannot ask, the
gate matters more there, not less.

## Alternatives considered

- **Wrap the headless command externally** (a wrapper script that pre-checks commands).
  Loses hooks entirely and duplicates policy in a second place that drifts.
- **Route the work through the interactive TUI** (drive it under tmux). Keeps hooks but
  requires a human-shaped surface for an unattended job, and breaks on restart.
- **Change the default so `exec` always fires hooks.** Rejected as the wrong direction —
  a config that pages an on-call rotation should not start firing from a CI job merely
  because it exists. Hence opt-in flags.

## Impact

Any user who relies on `permissions.toml` or hooks as a safety layer and also runs
`codewhale exec`. Frequency depends on the user's automation, but the failure mode is
silent: the guard is assumed present and is not. The plumbing already exists, which makes
this a comparatively small change — `run_tool_call_before_hooks` is already called from a
non-TUI surface (`crates/tui/src/acp_server.rs:785`), and
`exec_shell_ask_rule_decision_for_policy` was written as the seam "so headless protocol
adapters ... enforce the same sibling `permissions.toml` policy as the interactive engine"
(`crates/tui/src/core/engine.rs:7568-7576`).

## Additional context

Related: the same policy gap exists for sub-agent tool loops, reported separately — the
headless and child surfaces are both outside the interactive engine loop, so whatever
mechanism fixes one is a candidate for the other.

## Environment

- OS: Arch Linux
- codewhale version: 0.9.13 (dev), source git checkout `8d66b9a238a225007b314f3785d53e5686e96404`
- Install method: `cargo install --git https://github.com/Hmbown/Codewhale codewhale-cli`
- Model/provider: `deepseek/deepseek-v4.1-flash` over a custom `command_code` provider
- Terminal app: tmux (`TERM=tmux-256color`)
- Shell: fish

### LINKED ISSUE #6140: MCP server: retire the DeepSeek-specific tool calls and make the stdio loop async-native
From the 0.9.14 refactor backlog.

1. `mcp_server.rs` ~345-424 `handle_deepseek_call` builds a request and calls `client.create_message` directly with its own thread storage. Founder note 2026-09-11: the MCP tool should not be DeepSeek-exclusive any more — this is deepseek-era residue. Delete it or generalize to a provider-neutral `ask_model` routed through the engine; never a special-cased client call (same second-authority smell as #6088).
2. `run_mcp_server` (`mcp_server.rs` ~84) is a blocking stdio loop invoked via `block_in_place` (`lib.rs` ~2522), constructs its own multi-thread `Runtime::new()` and calls `runtime.block_on(registry.execute_full(..))` per request. Make it async-native: `tokio::io::stdin/stdout`, `AsyncBufRead::lines`, request handling as futures on the caller's runtime, delete the private runtime and every `block_on`. A stdio JSON-RPC server is serialized by definition.

Size: M. Do both together.

### LINKED ISSUE #6141: Finish the execpolicy migration: one consumer of tui/src/execpolicy is left
From the 0.9.14 refactor backlog.

`tui/src/execpolicy/` (matcher.rs, rules.rs) still exists and exactly one file imports it: `tools/shell.rs` ~3389 (`crate::execpolicy::{ExecPolicyDecision, load_default_policy}`), against ~110 imports of `codewhale_execpolicy`. Port matcher/rules into the crate and delete the copy. This is the AGENTS.md corollary made literal: migrate the last consumer or do not start.

Size: M. Acceptance: `tui/src/execpolicy/` is gone, shell.rs uses the crate, permission tests unchanged.

### LINKED ISSUE #6149: Blocking calls in async paths: audit thread::sleep and std::fs in tool paths, pick one convention
From the 0.9.14 refactor backlog.

- `thread::sleep` in production paths: `cloud_dispatch.rs` ~1613 (sandbox-ready poll loop), `lane/src/runtime.rs` ~1401 (10 ms), plus others outside tests. On the runtime they park a worker thread. Convert to `tokio::time::interval` with cancellation, or a notify on the resource itself. Audit first — some are already under `spawn_blocking`.
- ~30 non-test files mix `async fn` with `std::fs::` (`tools/file_search`, `read_media`, `terminal_session`, `revert_turn`, `tool_preparation`, `turn_loop` helpers). Decide one convention and enforce it once: (a) tool I/O uses `tokio::fs`, or (b) `execute_planned_tools` runs handlers on the blocking pool (`spawn_blocking` per batch — 81 sites already use the pattern). (b) is the ponytail answer for tools that read big files.

Deliverable: the convention written down in docs plus the sites moved. Size: M.

Hmbown:
## What / why

Second stacked slice run on the v0.9.14 milestone, stacked on `origin/main` (`433685b2`, post-#6161 merge). Nine issue slices plus one lint fix, one commit per issue, each verified before commit.

## Closes

- Closes #6099 — `codewhale exec --hooks` opts into configured hooks and typed permission rules.
- Closes #6037 — fleet members inherit the operator route instead of pinning it, so a vendor-side model id change propagates.
- Closes #6149 — convention adopted and enforced once: blocking calls run off Tokio workers (`spawn_blocking`/`block_in_place` at the async boundary), the cited `thread::sleep`/`std::fs` sites audited (the `cloud_dispatch` poll loop already runs on a dedicated `cw-dispatch-*` OS thread, not a runtime worker), and `scripts/check-blocking-calls-budget.py` ratchets the remaining inventory.
- Closes #6141 — `tui/src/execpolicy/` deleted; matcher + TOML rules live in `codewhale-execpolicy`; `tools/shell.rs` (the last consumer) uses the crate.
- Closes #6033 — MCP servers connect lazily: boot eagerly connects only `required`/`always_load`/tool-selected servers, the per-turn backstop is removed, `/mcp` and the launch card distinguish idle from in-flight connects.
- Closes #6140 — `serve --mcp` stdio loop is async-native (no `block_in_place`); the DeepSeek-specific `deepseek_chat`/`deepseek_reasoner` tool calls are retired.
- Closes #5715 — prior interrupted sessions are visible to the model: read-only `session_search`/`session_get` tools (workspace-scoped, bounded, untrusted-data-marked) plus a one-line recovery hint injected only when a durable checkpoint marks an interrupted prior session.
- Closes #6014 — session picker: hides empty auto-created sessions via the shared projection (`SessionQuery::without_empty_auto_created`), explicit selection still lands via a filter-widening fallback, the live session's row gets a localized `current` label + `WHALE_ACTION` ink, the preview lists the full session id, PgUp/PgDn page the list (clamped) with Shift+PgUp/PgDn keeping history scroll, wide layout gives the list 44%.
- Closes #6085 — the launch card's MCP problems row joins `LaunchRowId` (the shared paint/click/keyboard ordering): Up/Down lands on it and Enter/click types the printed remedy (`/mcp login <name>` or `/mcp`) into the composer — no clipboard dependency, and the command is visible before a second Enter sends it.

## Refs (partial slices — status comments on the tracker)

- Refs #6150 — the input path no longer awaits a full op channel; the `Op::SendMessage` → `TurnSpec` extraction remains.
- Refs #5529 — budget-death receipts now name surviving workspace changes; the issue's other sub-parts remain.
- Refs #6035 — `doctor` flags fleet/agent model pins missing from the live roster; propagation/migration is not implemented.
- Refs #5587 — one audited cluster of test-only `#[allow(dead_code)]` markers converted to `#[cfg(test)]` and the budget ratcheted to 369; later phases remain.

## Commits

- `6e524d07` feat(exec): opt in to configured hooks with `codewhale exec --hooks` (#6099)
- `ac6a881d` fix(fleet): inherit the operator route instead of pinning it (#6037)
- `4d70e050` fix(tui): stop awaiting the op channel on the input path (#6150)
- `47834cc6` fix(runtime): keep blocking calls off Tokio workers (#6149)
- `290938cc` refactor(execpolicy): port TOML rules into codewhale-execpolicy (#6141)
- `a4c4aa29` fix(subagent): name surviving workspace work in budget-death receipts (#5529)
- `47666265` feat(doctor): flag fleet model pins missing from the live roster (#6035)
- `890b7d03` refactor(tests): gate audited test-only helpers behind cfg(test) (#5587)
- `711c538e` refactor(execpolicy): rename from_str to parse for clippy
- `be214c86` feat(mcp): connect servers lazily instead of all at boot (#6033)
- `75e3a819` refactor(mcp): make serve --mcp async-native and retire the model tools (#6140)
- `eefdcb42` feat(session): make prior interrupted sessions visible to the model (#5715)
- `7192b50a` fix(tui): make the session picker browsable and current-aware (#6014)
- `c5ed2e6c` feat(tui): make the launch card's MCP problems row run its remedy (#6085)

## Gate output (what actually ran on this branch)

- `cargo fmt --all -- --check` → clean
- `cargo check -p codewhale-tui --all-targets --locked` → clean
- Focused suites per slice: session_picker 29 · session_control acceptance 42 · localization parity 51 · session-filtered lib 766 · subagent 783 · doctor 123 · engine MCP 462 · launch_card 16 — all green
- `scripts/check-dead-code-budget.py` → PASS (369, exactly at budget) · `check-blocking-calls-budget` → PASS (616 sites)
- One known environment-sensitive lib failure observed once (`compaction`-count flake); it fails identically on `origin/main` without these changes.
- **Not run locally:** the full workspace nextest suite, `npm test`, `npm run check:web` — hosted CI on this PR is the gate.

Generated with [Devin](https://devin.ai)

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6175" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Changes MCP connection timing, engine turn admission, and daemon async I/O on core runtime paths; wire `send_message` JSON stays compatible but behavior shifts when servers connect.
> 
> **Overview**
> This PR lands a **v0.9.14-style slice** across runtime conventions, MCP boot, sessions UI, and execpolicy structure.
> 
> **Blocking calls (#6149):** Documents the Tokio rule in `AGENTS.md`, adds a CI **blocking-calls budget** ratchet, moves daemon socket setup to **`tokio::fs`**, runs local Whisper transcription on **`spawn_blocking`**, and replaces **`thread::sleep`** with **`yield_now`** on async user-command paths.
> 
> **MCP (#6033):** Session boot connects only **required / always-load / tool-selected** servers; **`mcp_tools`** no longer sweeps every server each turn. Explicit tool selections **start and await their own connects** (with deadline/cancel), and **“connecting”** is tracked from the pool instead of inferred as enabled−connected.
> 
> **Sessions (#5715, #6014):** Injects a **recovery hint** into the frozen system prompt when a prior interrupted checkpoint exists; the picker gets **current-session labeling**, **full ID in preview**, updated **PgUp/PgDn** hints, and passes **`with_current_session`** when opened.
> 
> **Protocol / engine:** Extracts per-turn fields into **`TurnSpec`** (`Op::SendMessage(TurnSpec)` and protocol twin) without changing the tagged JSON shape; refactors **`handle_send_message`** to take one spec struct.
> 
> **Execpolicy (#6141):** Consolidates TOML rules in **`codewhale-execpolicy`** (`matcher`, `toml_rules`, **`RuleDecision`**, **`parse`**); drops default-path loaders from the crate.
> 
> Also documents **`codewhale exec --hooks`**, bumps **base64/shlex**, and updates session-related locale strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 137fb70a921a3f59b1b7a3f05e456e3fc19f53a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hmbown:
Overnight CI triage on this branch. Five of the six red checks are fixed; one is not mine and is described below so it does not get lost.

## Fixed on this branch

| Check | Cause | Fix |
| --- | --- | --- |
| `cargo-deny (advisories)` / `cargo-audit` | RUSTSEC-2026-0285 (rustls, TLS 1.3 handshake across encryption levels) | `61f4bff1a2` — rustls 0.23.45. Note #6105 bumped 0.23.**44** and merged to main, but cargo-deny's own output says "Upgrade to >=0.23.45", so that bump did not clear it |
| `Lint & Type Check` (web) | `facts.generated.ts` stale at toolCount 76, workspace has 78 | `aeabd5a96d` — regenerated via `npm run prebuild` |
| `Lint & Type Check` (web, second failure) | `docs/public-surface-facts.json` holds the same fact by hand, also 76 | `77e24c0b5f` |
| `Lint` (clippy) | 7 errors, 5 of them in test targets that only `--all-targets` reaches — incl. a `println!` probe under `#![deny(clippy::print_stdout)]` | `26b6ecba5b` |
| `Test (windows-latest)` | **Did not compile.** `ShellKind`'s `expect(dead_code)` annotations were written unix-first; on Windows the four Windows variants are constructed (expectations unfulfilled) and `Sh`/`Custom` are not (dead) | `aca2174447` |

Also on the branch: `2e467d64b0` fixes #6174 (ACP `session/new` minting ids `session/load` cannot resolve), and `0d0a1196e3` seals the PTY harness against a live local Ollama — four PTY suites fail on any developer machine running Ollama on :11434, because the startup adoption probe rewrites the launch screen the tests wait for. `seal_home` only sealed the filesystem; that leak arrives over loopback.

## Not mine: `portable`

`portable` (Pet conformance) fails at `./pet/verify.sh --no-swift` on a TypeScript↔Rust **frame-hash parity mismatch**, not on anything in the checks above. It has been red on this branch since `a06d2296` at 08:12, before any of tonight's triage, and Pet conformance has never run on `main` — so this branch is where it is exercised first.

The fix is very likely in the uncommitted `pet_sim.rs` / `pet_widget.rs` / `pet_cameo.rs` sitting in the `codewhale` checkout rather than here. Leaving it to whoever owns that slice: merging with it red would put failing pet parity on `main`, so this should not land until it is resolved or deliberately waived.

## Verification

`./scripts/dev-test.sh tui` — 12707 tests run, 12707 passed, 19 skipped.
`cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or` — exits 0.
PTY: `cargo nextest run -p codewhale-tui --test cucumber --features long-running-tests -E 'test(/_pty::/)'` — 9 passed, from 5 passed / 4 failed on the same machine.

One honest gap: the Windows fix could not be checked locally — `cargo check --target x86_64-pc-windows-msvc` cannot build `ring` without an MSVC-targeting C compiler. CI is its first real test.

Hmbown:
Update, including a retraction.

**Reverted:** `68431ba097` (the #6207 store-adoption fix) is reverted in `bc77a0ff5d`. CI caught a race my local runs did not: the predicate I added samples store emptiness with no liveness check, so a store that is empty only because the holding host has not flushed yet is indistinguishable from an abandoned one. `runtime_store_binding_survives_launch_snapshot_and_resume` failed on macOS and ubuntu at line 284 while passing here in isolation — the classic ordering-dependent signature. It also missed scope-pinned automations, which live outside the runtime store tree. Full reasoning on #6207. That is a data-loss boundary, so it comes out rather than being patched forward at speed.

I also need to correct something I wrote earlier in this thread: I said the seven pre-existing binding tests "pass untouched", offering that as evidence the policy widened by exactly the empty case. A green targeted run is not evidence about a test whose failure mode is ordering-dependent, and I presented it as though it were.

**`Codewhale review` is not a code failure.** It is a spend gate:

```
Complete PR review requires 5 passes at 200000 characters per pass, but max_passes is 1.
Opt in with max_passes/--max-passes of at least 5 only after approving the provider spend and run duration.
```

Raising it needs a human to approve provider spend, so I have not. It is also a size signal — this branch is large enough to need five review passes.

**Still standing from the earlier triage:** the rustls advisory (`61f4bff1a2`), both web fact staleness fixes, the seven clippy errors, the Windows compile fix (`aca2174447` — and `Test (windows-latest)` has since progressed past the compile step that was failing), the #6174 ACP session-id fix, and the PTY Ollama seal.

**Still not mine:** `portable`, the pet TypeScript↔Rust frame-hash parity mismatch, red since `a06d2296` at 08:12 and never exercised on `main`.

Issue #6193, opened 2026-09-15
No runtime performance gate exists: nothing measures or defends the speed of the user path


Hmbown:
## What is missing

`BUILD_PERFORMANCE.md` covers *build* speed. Nothing in the repository measures
*runtime* speed: no `benches/`, no criterion / iai / divan in `Cargo.lock`, no
`.perf` cases, no budget constants, no gate that fails when a user path gets
slower. Verified 2026-09-15.

So every claim about responsiveness in this milestone ("streaming is bursty",
"typing feels laggy", "make it smooth and fast") is currently unfalsifiable. The
stream-pacing change in `9b41aeaac` was reproduced and measured with a scratch
harness that was thrown away; nothing keeps that measurement honest next month.

## What good looks like, learned from deepseek-harness

`refs/dsh/benchmarks/AGENTS.md` runs required repository-level performance gates
whose measured path crosses package ownership. The transferable parts, in the
order I would take them:

1. **Organize by measured user path, not by package tree.** One directory per
   path. DSH has `session-open`, `agent-continuation`, `conversation-fold`,
   `active-stream-reconnect`, `terminal-io`, `long-session-browser`.
2. **Fixed synthetic inputs from reviewed constants.** "Never use recorded
   Sessions, user material, ambient repositories, or network services."
3. **Budgets are reviewed source constants.** "Environment variables must not
   override performance budgets."
4. **Say how the verdict is reached.** "Report enough raw and aggregate
   measurements to explain each verdict, including whether a budget uses a
   median, minimum, absolute value, or ratio."
5. **Separate expectations by machine.** Reference-machine expectations are
   recorded separately from the shared CI time scale and variance headroom, and
   the time scale is never applied to memory or to dimensionless ratios.
6. **Fresh process, private root, bounded child** for any wall-clock or
   retained-memory sample; remove owned roots after failure as well as success.
7. **One named benchmark per real defect class.** `active-stream-reconnect`
   exists because a reconnect carrying a 100,000-delta reasoning prefix has to
   fold without stalling — a specific shape, not a generic "streaming is fast".

## The decision this needs first

A harness means a benchmark dependency, and criterion / iai / divan are all
absent from `Cargo.lock`. Adding one is a dependency decision on a repository
currently working to *reduce* dependency surface (#6151), so it should be a
deliberate choice rather than a drive-by. The alternatives are a dependency-free
`Instant`-based harness in a child process, or adopting one of the above.

[...]

Issue #6155, opened 2026-09-14
Pet: qualify the /pet habitat in a real terminal and accept the shared owner across TUI + desktop


Hmbown:
0.9.13 ships `/pet` (#6154, from #6110 plus the prepared shared-owner/Kitty patch). What it ships on is test evidence and the prepared-archive receipts; these acceptance items were explicitly not done and stay open here (split out of #6109 so the umbrella can track the cross-surface owner work):

1. **Real-terminal visual qualification** of the Kitty pixel habitat: docked→full transitions, continuous motion, resize, Escape image cleanup, a fallback (braille) terminal, tmux/SSH behaviour, reduced motion. Measure delivered frame timing, CPU, RSS, bandwidth and input latency in the terminal; the only numbers today are the raster microbenchmark (17.09 ms mean / 19.55 ms p95, ~150 KB/frame, excludes transport) and a 30 Hz simulation. 60 fps is a target, not a measured result. Computer Use access to Ghostty/Terminal was denied in the preparing session; do not relabel browser or native screenshots as terminal proof.
2. **Shared-owner acceptance**: open the TUI habitat and the desktop/browser view on one live session, compare identity/tick/digest at a retained frame, interact from either side, close and reopen the TUI, confirm continued ownership, one producer, one audio lease, durable appearance. Exercise owner restart, a competing owner, source change, storage failure.
3. **Engine metadata attachment beyond the TUI**: only the TUI producer forwards accepted Engine metadata today; app-server does not (#6139, #6152).
4. **Founder direction "Codewhale by default"** for the appearance: not interpreted in 0.9.13 (the eight prepared presets are unchanged). Decide what the default look is and make it the default.
5. **Completion outcome fidelity**: the pet projection keeps only a derived `failed`; carry the typed `completed|interrupted|failed` terminal status and receipt identity through the metadata projection so a cancelled turn never gets a success gesture (see the animation packet in `artifacts/pet-review-20260912/CORE-PET-GPUI-CONTRACT.md`).

Contract and evidence: `pet/SHARED.md`, `pet/QA.md`, `docs/design/DOT_WHALE_PET.md`, `artifacts/pet-review-20260912/{PREPARED-PET-HANDOFF,RUST-PET-MODE-TAKEOVER,CORE-PET-GPUI-CONTRACT}.md`.

Hmbown:
[...]

PR #6222, opened 2026-09-15
TUI redesign: Shoreline, one palette across both clients

Hmbown:
Re-inks the TUI to the GPUI client's palette and moves it toward the conventions the terminal-agent market converged on. **Draft: the headline item is not done** — see "What is left" below.

Pushed primarily because it was local-only. Five of these ten commits are not the redesign at all: they were stranded on a local `main` that is 129 commits behind `origin/main` and had never been pushed anywhere — a runtime-api settings-schema route, a config `Float` setting kind, the web `/plugins` page, the Computer Use 0.2.2 sync, and the 0.9.13/GPUI-direction doc. Publishing this branch is what rescues them; `main` itself is untouched and still needs reconciling separately.

## The argument

Not aesthetic, structural. `AGENTS.md` asks for "one coherent runtime, one store, one design system", and `TRANSITION.md` makes the GPUI app the product client. A terminal that opens on a saturated navy gradient with decorative sea life while the desktop opens on warm charcoal is not two themes, it is **two design systems**. `SHORELINE_UI_THEME` cites the same slots `codehwhale-gpui/src/workspace/mod.rs:147-164` installs into GPUI's `Theme`, which collapses that.

`underwater` is not deleted. It keeps its water column, ombre and ambient life, and stays selectable — it stops being the ground the product opens on.

## Why it holds rather than half-applies

- `adapt::theme_remap_active` lists both new presets, so all 84 direct `WHALE_*`/`TEXT_*` call sites land on Shoreline slots. Without it the frame repaints while every widget keeps its cold grey, which looks worse than not doing it.
- The mode ramp stays separate from `accent_primary`. `theme_semantic_foreground_role` resolves mode slots before the action lane, so equal values collapse the ANSI-16 role matrix. Caught by `every_selectable_theme_keeps_mode_badges_distinct`, and fixed in the theme rather than the test.
- Every audited pair clears its contrast floor under `contrast::theme_contrast_violations`.

## What is left, and why it matters

**The one-status-row merge is not started.** The market survey in `docs/design/TUI_REDESIGN.md` found exactly one convergent rule we violate — one status row, never two — and it is untouched, so this branch delivers roughly the first half of what the doc argues for. The blocker is real: the metrics line owns the `/model` and context-inspector pointer targets, so re-homing them has to land in the same commit that hides the row or the slice ships a regression.

**Goldens record no colour.** `golden_harness.rs:27-49` dumps `cell.symbol()` only, so this entire re-ink is invisible to all 41 golden buffers — and so is the next one. The harness documents an "ink plane" at `:84-94` that would close the hole and is unimplemented. That should be in scope here, not adjacent to it: shipping a design system whose correctness is unfalsifiable is the same mistake #6193 identifies for performance, and right now both holes are open.

**One product decision:** making `underwater` non-default means the pet and ambient life (#6155, and the in-flight `pet_sim`/`pet_widget` work) ship to zero fresh installs. Either the pet gets a home in Shoreline or it becomes a deliberate opt-in. Not a bug — a call someone has to make.

## Verification

`cargo test -p codewhale-palette --lib --locked` — 81 passed, 0 failed, including the contrast floor, mode-badge distinctness, permission-lane and ANSI-16 audits over both new presets (re-run independently on this tree).

Two failures on this branch are pre-existing and unrelated: `runtime_api::tests::set_config_rejects_unknown_key_with_bad_request` (`runtime_api/tests.rs:8250` expects "unknown config key", `settings.rs:1758` emits "unknown setting") and `tui::views::tests::every_settings_row_reaches_a_store`. Both verified present at the branch point `4a85cb7877`.


Hmbown:
Closing as landed. All 10 commits on this branch are already on `main` via the rebased lane (`feat/shoreline-redesign-rebased-20260915`), e.g. `9d3f6f2a5` (Shoreline palette), `a08d315d9` (brand marks), `542867c88` / `8839b6178` (theme + golden reconciliation), `687e27c6f` (design record). Verified by subject against `origin/main` at `13d08f659`; the branch adds no file that main lacks. Nothing left to merge.

Issue #6178, opened 2026-09-15
v0.9.14: App-server typed command catalog for client slash-command surfaces


Hmbown:
## Why this matters

GPUI Hmbown/codewhale-apps#556 (Linear APPS-28) requires the composer to route slash commands through shared command definitions — "the existing Codewhale command registry, aliases, hidden tiers and user-command shadowing," with host commands like `/settings` or `/compact` never submitted as paid model prompts. Without a served catalog, the desktop must hardcode a second command list that rots against the registry.

## Current behavior

- The command registry, aliases, hidden tiers and user-command shadowing live TUI-side; there is no `/v1/commands` (or equivalent) route — the full `/v1/*` route table in `crates/tui/src/runtime_api.rs` contains no command surface.
- Clients cannot distinguish "Engine-handled host command" from "model prompt" without duplicating the registry.
- #6145 covers the internal `crates/command-contract` adoption question; this issue is the client-facing read surface regardless of where the registry lands.

## Desired behavior

1. A route serving the typed command catalog: name, aliases, summary, hidden/tier flag, argument shape, and whether it is host-handled vs model-bound — enough for a client to complete, validate, and never submit a host command as a prompt.
2. User-command shadowing reflected in the served result (a user command that shadows a builtin must win).
3. Forward-compatible: unknown command kinds must be ignorable by older clients.

## Acceptance criteria

- [ ] `GET` on the route returns the full catalog including aliases and hidden-tier markers
- [ ] A user-defined command shadowing a builtin appears with the winning definition
- [ ] Each entry states host-handled vs model-bound so a client can gate submit
- [ ] `cargo test -p codewhale-tui` route test covers catalog shape and shadowing

Related: Hmbown/codewhale-apps#556, #6145 (command-contract adoption), #6152.

Hmbown:
Landed in `70cda6e43` + `d5291dfb66`.

- `GET /v1/commands` serves the typed catalog: name, aliases, localized summary, usage, parsed literal subcommands, `takes_arguments`, hidden/discovery tier, `kind` (`builtin`/`user`, forward-compatible), `binding` (`host`/`prompt`), `shadowed_by`/`shadowed_aliases` for user-command shadowing.
- Shadowing verified over HTTP: a workspace `.codewhale/commands/model.md` makes the builtin `model` report `shadowed_by: "model"` while the winning user row serves with `binding: "prompt"`.
- Catalog is derived from the existing command registry + user registry — no second command list to rot.

[...]

Issue #6168, opened 2026-09-14
App-server: GET /v1/instructions for GPUI Memory/instructions (or name the real path)


Hmbown:
GPUI Hmbown/codewhale-apps#618 Memory module uses `GET /v1/memory*` when served. Instructions fail closed: Apps probed `GET /v1/instructions` as missing and opens local AGENTS/CLAUDE via the Files module instead.

Do not invent a second instructions store in the desktop. Reconcile `crates/tui/src/runtime_api.rs` and constitution/constitution surfaces against today's tip. If instructions already exist under another path, name it. Otherwise serve read (and supported write) for workspace instruction files with sandbox/permission receipts — or document that Files + AGENTS/CLAUDE is the intended path and close this as wontfix with the real route name.

Related: Hmbown/codewhale-apps#618, Hmbown/Codewhale#6163 (files).

Hmbown:
**Reconciled 2026-09-14** (Claude Fable 5.1 at Hunter's direction, 0.9.14 takeover) against `crates/tui/src/runtime_api.rs` and the prompt assembly at today's tip.

**Real paths today.** `/v1/instructions` was never a route. Instructions reach the model from four Engine-owned sources, none of which is served over HTTP as a list:
1. the workspace's `AGENTS.md`-style project context (`crates/tui/src/project_context.rs::load_project_context`, with parent-directory discovery and `.codewhale/rules/` / `.claude/rules/`),
2. the selected cross-agent instruction fragment (`.cursorrules`, `.clinerules`, `.github/copilot-instructions.md`, … via `codewhale_core::fragments::load_selected_project_instruction_fragment`),
3. the configured `instructions = [...]` file list (`Config::instructions_paths`),
4. the repo constitution `.codewhale/constitution.json` and the user constitution.
`GET /v1/memory` is memory, not instructions.

[...]

PR #6229, opened 2026-09-15
Native-client runtime API routes: files, artifacts, jobs, git, LSP, secrets, targets

### LINKED ISSUE #6163: App-server: GET /v1/artifacts and GET /v1/files for GPUI Preview and Files
GPUI Preview/Files fail closed on missing HTTP. Probed and named:

* `GET /v1/artifacts` — 404
* `GET /v1/files` — 404

GPUI uses local bounded reads against the workspace root instead. Do not invent a second file store in the desktop.

Reconcile `crates/tui/src/runtime_api.rs` against today's tip before adding routes. If these exist under another path (like plugins vs `/v1/plugins`), name it. Otherwise serve list/read (and write+revision for files) with sandbox/permission receipts.

Related: Hmbown/codewhale-apps#564, Hmbown/codewhale-apps#619.

### LINKED ISSUE #6166: App-server: GET /v1/jobs for GPUI long-running job inspect and stop
GPUI Hmbown/codewhale-apps#615 needs Codewhale-owned long-running job inspect/control. Apps probed and named:

* `GET /v1/jobs` — 404 (or not served for GPUI)

GPUI must fail closed and must not invent a second job registry in the desktop. Reconcile `crates/tui/src/runtime_api.rs` and TUI `/jobs` against today's tip before adding routes. If jobs already exist under another path (task/process control plane), name it. Otherwise serve list + bounded output cursor + scoped stop with ownership checks (PID reuse and foreign-session IDs rejected).

Related: Hmbown/codewhale-apps#615, Hmbown/Codewhale#6160 (terminal bytes — separate from jobs).

### LINKED ISSUE #6179: v0.9.14: App-server credential set/clear with source and writable metadata for GPUI providers
## Why this matters

GPUI Hmbown/codewhale-apps#574 (Linear APPS-48) requires first-run provider setup entirely in the desktop: "UI sees configured/source/writable metadata, never raw secret values in catalog/logs. Environment-managed credentials must not appear successfully overwritten." Today the desktop can show that a provider is configured but cannot write a credential — setup still requires the TUI or editing files.

## Current behavior

- `GET /v1/providers` serves `credentialState` — a sanitized structural classification (`crates/tui/src/runtime_api.rs:6109-6113`, "the API never echoes credentials back to callers"). Correct, but read-only.
- No route sets, clears, or verifies a provider credential; there is no `credentials` write surface in the route table.
- `credentials_present` on `/v1/operate` is a readiness flag, not management.

## Desired behavior

1. A write route that accepts a credential for a provider, stores it in the existing credential owner, and returns a receipt — never echoes the secret.
2. Per-provider metadata extended or served alongside: `source` (keychain/env/config/managed) and `writable` so a client can disable the control with a truthful reason ("managed by environment") instead of letting a write fail late.
3. Clear/revoke for client-owned credentials; environment- or managed-owned credentials refuse with a named reason.
4. Optional verify/probe that reports reachability without exposing the credential.

## Acceptance criteria

- [ ] Set persists through the existing credential store and flips `credentialState` on the next `GET /v1/providers`
- [ ] Env-managed credentials refuse overwrite with a named reason; the metadata lets clients show this before submit
- [ ] No secret value appears in any response, log line, or error
- [ ] `cargo test -p codewhale-tui` route tests cover set, clear, refused env-managed write, and no-echo

Related: Hmbown/codewhale-apps#574, #6028 (account link / auth model), #5718 (one worker system / Keychain path).

Hmbown:
Native-client runtime API routes for the GPUI desktop: jobs, workspace files
and session artifacts, git, diagnostics, context, targets/SSH/cloud, LSP, voice,
and a write-only provider credential route. 28 files, ~+5,900 lines, of which
1,758 are tests.

`docs/RUNTIME_API.md` documents the surface.

## Why

The GPUI desktop is the product client (`CURRENT_DECISIONS.md` §14). It is
Engine-complete for reads but could not drive long-running jobs, review a diff,
export diagnostics, switch targets, or complete first-run provider setup without
dropping the user back into the TUI. Each of those needed a Codewhale-owned
route rather than a second registry in the desktop.

## Closes

- Closes #6166 — `GET /v1/jobs` plus `GET|POST /v1/threads/{id}/jobs`,
  `/jobs/{job_id}`, `/output`, `/stdin`, `/kill`. Jobs created here carry an
  `api:{thread_id}` owner scope, so a foreign or reused id is rejected rather
  than controlled. Covered by
  `jobs_api_lists_creates_streams_stdin_and_kills`.
- Closes #6163 — workspace files and session artifacts.
- Closes #6179 — `DELETE /v1/providers/{id}/key`, plus `credentialSource` /
  `credentialWritable` on `GET /v1/providers`. See below.

## It also turns `main` green

`main` is red at `9cdfa92bc`: CI's blocking-calls gate is advisory on pull
requests and **blocking on pushes to main**, and the Lint job failed on
`crates/tui/src/runtime_api/workspace.rs: std_fs sites 1 > budget 0`.

That site is real. `workspace_instructions` called `std::fs::canonicalize` in
its async body, *after* awaiting the `spawn_blocking` it already had — a
blocking syscall on a Tokio worker. It now rides that same closure.

This branch had added four more of the same class, which would have made the
red worse on merge: `list_logs`, `list_crashes` and `process_info` did
directory walks, per-file stats and a `/proc` read inline, and four LSP
handlers called `resolve_workspace_file` — which canonicalizes and stats every
path component — straight from their async bodies. All now use the blocking
pool.

**The gate could not have caught the LSP one.** The scanner is per-file and
lexical: the `std::fs` calls it counts live in `workspace.rs`, while the
handlers calling through to them live in `lsp.rs`, so a helper is invisible at
its call site. Worth knowing before anyone reads a passing budget as proof.

The same lexical rule cuts the other way, which is why eight sites are
budgeted rather than rewritten: a sync `fn` holding `std::fs` is counted even
when every caller is already inside `spawn_blocking`. Every caller of
`confined_directory`, `list_workspace_directory`, `precheck_file_target`,
`list_files`, `read_named_window` and `rss_bytes` was read before budgeting.

## #6179 — clearing a credential, and saying who owns one

The credential route could set a key but never clear one, and served no way to
tell "you have no key" apart from "your key is owned elsewhere and this control
cannot change it".

`DELETE /v1/providers/{id}/key` clears through a *shared* owner rather than a
second implementation: `crates/config/src/credentials.rs` gains
`clear_provider_api_key`, and `codewhale auth clear` now calls it — which
deletes its inline body and a helper. That move fixes a real CLI defect on the
way past: the secret-store delete was `let _ = secrets.delete(...)`, so
`auth clear` printed success while the key could still be in the keyring. Both
callers now report the backend error.

`GET /v1/providers` gains `credentialSource`, `credentialWritable` and a
reason, so a client disables its control with a truthful explanation instead of
letting a write fail late. Both verbs refuse a credential Codewhale does not
own with `409` — notably a literal key in a config file, which still wins at
request time, so writing the secret store would have reported a success the
user could never observe. That is the "must not appear successfully
overwritten" the issue asks for.

The provider-catalog non-secret projection guard correctly failed on the new
fields. Widening an allow-list is an argument, not a formality: the fields are
admissible because `credentialSource` is one of four fixed enum spellings that
can never interpolate a value, a path or an environment variable name. The test
now asserts that closed vocabulary, not just the key set.

## Partially addresses, and says so

**#6179** is *not* closed by this PR. `PUT /v1/providers/{id}/key` satisfies the
set half: it persists through the same transactional path as
`codewhale auth set`, caps the body at 4 KiB, rejects control characters,
returns a redacted receipt (backend + post-write `credentialState`), and has
deliberately no GET — a route that can return a secret can leak one. Still
missing from that issue's acceptance criteria: clear/revoke, `source` and
`writable` metadata, and the named refusal for environment-managed credentials.
Tracked there, not silently folded in here.

## Two defects the merge surfaced, and what was done about them

The merge of `origin/main` (51-commit v0.9.14 slice-2) hit three textual
conflicts, two of which hid problems a textual merge cannot see:

- **`GET /v1/commands` was implemented twice.** Both sides added it and axum
  rejects that at runtime ("Overlapping method route"). They are not duplicates:
  main's carries user-registered commands and shadowing; this branch's carried
  localized descriptions, argument-shape semantics and composer hints. That is a
  public API contract decision, not a merge mechanic, so main's shipped handler
  stays and this branch's route, module and test were removed. The lost fields
  are written up in #6230 with the full comparison and are recoverable from
  `060711b2f1`.
- **`OP_KINDS` never listed `get_context_budget`.** The `GetContextBudget`
  variant and its `kind_str()` arm were added without the const array, which
  `protocol_covers_engine_ops` checks. Pre-existing on this branch rather than a
  merge artifact — it would have failed CI on its own. Added in variant order.

## Evidence

Re-run on this exact head (`0285df1ee`), not inherited from the handoff:

| Gate | Result |
| --- | --- |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --all-targets --all-features --locked` (CI's allow list) | clean |
| `codewhale-config --lib` | `test result: ok. 704 passed; 0 failed; 1 ignored` |
| `codewhale-cli --lib` | `test result: ok. 390 passed; 0 failed; 0 ignored` |
| `runtime_api` | see below |
| Changelog receipt | `sync-changelog.sh` + `derive-changelog.mjs` regenerate with no diff |

The handoff claimed `runtime_api` "250 passed". The first run here reported
`test result: FAILED. 251 passed; 2 failed; 0 ignored; 0 measured; 12494 filtered out`.
The two failures were
`compatibility_stream_closes_losslessly_across_replay_live_handoff` and
`compatibility_stream_exposes_and_resolves_user_input_without_answer_echo`,
both "deadline has elapsed" on a subscription handshake.

They are **not** a regression, and the reason is worth recording: both tests
arrived on `origin/main` in #5717, older than this branch, and both use
`ci_scaled(Duration::from_secs(2))` — 2s locally, 8s only when `CI` is set. That
run happened on a 14-core host at load average 17 with three concurrent cargo
lanes. Re-run isolated on the same head:

```
running 2 tests
test runtime_api::tests::compatibility_stream_closes_losslessly_across_replay_live_handoff ... ok
test runtime_api::tests::compatibility_stream_exposes_and_resolves_user_input_without_answer_echo ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 12745 filtered out
```

And the whole module at `--test-threads=2` on the current head:
`test result: ok. 258 passed; 0 failed; 0 ignored; 0 measured; 12494 filtered out`.

So: green isolated and green at low concurrency, flaky under parallel load. That is a different claim from
"green", and the local gate being load-fragile is worth someone's attention
independently of this PR.

Lint gates: `check-provider-registry.py`,
`test_check_command_crate_boundaries.py`,
`test_check_command_migration_manifest.py`, `check-dead-code-budget.py`,
`check-readme-translations.py`, `check-tui-locale-parity.py`,
`check-persistence-backlog-budget.py`, `check-readme-locales.sh`,
`check-tui-product-vocabulary.sh`, `web/scripts/check-locales.mjs` — all pass.

`check-runtime-contract-budget.py` fails, and **not because of this branch**:

```
[runtime-contract-budget] ERROR: identity changed for Act full tool names
[`tool_catalog.modes.act.full.tool_names`] (added=['session_get', 'session_search'] removed=[])
```

Neither symbol appears anywhere in this branch's diff. Both landed on `main` in
`eefdcb42f5` (#5715) without a matching update to
`scripts/runtime-contract-budget.json`, which still lists 50 Act-full tool names
and neither of these. This is exactly the case `ci.yml:459-462` anticipates —
"a branch can fail it for debt it inherited rather than added" — so the gate is
advisory on pull requests and blocking on pushes to `main`. Flagging rather than
folding a manifest edit into an unrelated PR.

Windows: `git_routes_drive_a_real_workspace_repo` was the only test failing the
required `Test (windows-latest)` check (15280 run, 1 failed). Windows git
defaults to `core.autocrlf=true`, so the checkout that `POST /v1/git/discard`
performs rewrote `v1\n` to `v1\r\n`. That is git behaving as configured, so the
fixture pins `core.autocrlf=false` alongside the `user.email`/`user.name` it
already set. The assertion is unchanged and not weakened. This host cannot build
`x86_64-pc-windows-msvc`, so that half rests on CI.

CI owns the exhaustive run; `cargo nextest --workspace` was not run locally.


Hmbown:
Merged current `main` in (`6e3439f7e8`) and this is now out of draft. The branch was 53 behind; it is now even with the 51-commit slice-2 merge.

## Conflicts, and the two things a textual merge could not see

Three textual conflicts, all additive collisions — kept both sides in each. But two hid real problems:

**The `workspace.rs` conflict cut through a function.** This branch's #6163 files block and main's #6168 instructions block have no symbol collisions (34 vs 5 items, checked), so both belong. But the conflict boundary ran through `write_workspace_file`, whose closing brace sat on main's side. Taking both sides verbatim produced an unclosed delimiter. Restored.

**`GET /v1/commands` was implemented twice**, and axum rejects that at runtime:

```
Overlapping method route. Handler for `GET /v1/commands` already exists
```

They are not duplicates. Main's carries user-registered commands and shadowing; this branch's carries localized `description` + resolved `locale`, `requires_required_argument` (vs a flat `takes_arguments`), `composer_wants_trailing_space`, `palette_runs_directly`, `show_in_empty_discovery` and `unlisted` — a richer composer/palette contract.

Picking between them is a public API decision, not a merge mechanic, so I kept main's shipped handler (#6178 is closed against it) and removed this branch's route, module and test. **The lost capability is written up field-by-field in #6230**, and everything is recoverable from `060711b2f1`. That decision should be made deliberately rather than by whoever resolves a conflict.

## One pre-existing defect on this branch

`OP_KINDS` never listed `get_context_budget`. The branch added the `GetContextBudget` variant and its `kind_str()` arm but not the const array — and `op.rs` has a test asserting *"OP_KINDS must list every variant in order"*. This would have failed CI on its own, independent of the merge. Fixed in variant order. Same for one `assert_eq!(..., false)` clippy error in the voice-route test.

## Verification

- `./scripts/dev-test.sh tui` — **12713 run, 12713 passed**, 19 skipped (it was 12712/1 before the `OP_KINDS` fix)
- `./scripts/dev-test.sh protocol` — 68 passed
- CI clippy gate `--workspace --all-targets --all-features -D warnings` — exits 0

I did not write the original two commits; I brought them up to date and verified them. CI here is still their first hosted check.

Hmbown:
## CodeQL triage — 10 high-severity alerts, all assessed

The CodeQL check reports `10 new alerts including 10 high severity security vulnerabilities`. I read each one rather than waving at the aggregate. **None block, and here is the reasoning for each so a reviewer does not have to redo it.**

### 6 × `rust/cleartext-transmission` — `crates/tui/src/runtime_api/tests.rs:15335-15407`

All six are inside `session_artifacts_list_and_bounded_read`, a test that drives a **loopback** listener: `http://127.0.0.1:<ephemeral>` with `bearer_auth` and fixture tokens that are not credentials. The rule fires on `http://` + a bearer header; there is no transmission off the machine and no real secret. Test-only.

### 1 × `rust/uncontrolled-allocation-size` — `crates/tui/src/runtime_api/diagnostics.rs:152`

The rule sees request-controlled `limit` reach `Vec::with_capacity` and misses the range check two lines above:

```rust
let limit = query.limit.unwrap_or(READ_LIMIT_DEFAULT);
if !(1..=READ_LIMIT_MAX).contains(&limit) {        // READ_LIMIT_MAX = 4 MiB
    return Err(ApiError::bad_request(...));
}
...
let mut window = Vec::with_capacity(limit.min(64 * 1024));   // preallocation capped at 64 KiB
file.take(limit as u64).read_to_end(&mut window)             // read bounded by limit
```

Validated range, preallocation capped independently of the validated value, bounded read. Not reachable as written.

### 2 × `rust/cleartext-logging` — `crates/cli/src/lib.rs:2665, 2672`

These are **mine**, added in `5dd4966f2`, so they got the most scrutiny. What is interpolated is `slot` — a secret-store slot *name* like `deepseek`, never a value — and `error`, a `SecretsError`. That enum is entirely structural: `Keyring(String)` (OS backend status text), `Io`, `Json` (serde position, not content), `InsecurePermissions { path, mode }`, `ReadOnly`. No variant carries key material, and no key is read on this path — `clear_provider_api_key` deletes without fetching.

Worth stating plainly: the alternative to printing that error is what this commit **removes**. The previous code was `let _ = secrets.delete(...)`, so `codewhale auth clear` reported success while the key could still be sitting in the keyring. Staying silent to keep the linter quiet would restore a real security defect to suppress a false one.

(`crates/cli/src/lib.rs` already carries 6 pre-existing alerts for this same rule on `main`.)

### 1 × `rust/path-injection` — `crates/tui/src/session_manager.rs:1360`

**Pre-existing, not introduced here.** `main` already reports exactly one `rust/path-injection` alert for this file, and `git diff --name-only origin/main..HEAD` does not include `session_manager.rs` — this PR does not touch it.

### Also still red, and deliberately

`Codewhale review` fails with `Complete PR review requires 2 passes at 200000 characters per pass, but max_passes is 1`. That is a provider-spend decision, not a code defect, and raising `max_passes` is not mine to authorize.

---

For the record: I earlier read the passing `Analyze (rust)` job as meaning the CodeQL check was stale. It was not — the check aggregates alerts, and the Analyze job succeeding only means the scan ran. Corrected here.

Issue #6207, opened 2026-09-15
bug(tui): session picker refuses saved sessions whose runtime store exists but is not the running host's own


7jrxt42BxFZo4iAnN4CX:
## Description

In the TUI, opening a saved session fails with:

```
Failed to restore session: This session belongs to another Runtime host. Resume it in a new
Codewhale process to reopen its saved store.
```

The session's runtime store directory exists, its ownership validates, and no other Codewhale
process is running — so nothing else holds the store. The refusal sits in the session-switch guard,
which treats "the bound directory exists" as "another live host owns it".

## Steps to reproduce

1. Launch Codewhale in a workspace; sessions created there are bound to the host's store,
   e.g. `~/.codewhale/sessions/<host-session-id>/runtime`.
2. Lose the host session record while its store directory stays on disk. A conversation created in
   that host keeps the durable binding to it.
3. Start a new Codewhale process (new host, its own store) and open the saved session (picker or
   `--resume` from the TUI).
4. The refusal above.

## Expected behavior

Opening the session should succeed, because the bound store both validates and is free:

- `RuntimeStoreBinding::validate_existing_store()` (`crates/tui/src/runtime_threads.rs:2918`) accepts
  it — the owner record parses and the recomputed scope
  `sha256(owner_id ‖ 0x00 ‖ <store>/events.lock)` equals the scope saved in the session;
- the store's process-owner lock (`runtime-process.owner.lock`) is free — no live host holds it;
- `RuntimeThreadManager::open_for_session` (`crates/tui/src/runtime_threads.rs:4636-4668`) already
  takes this exact path when a process opens a session with a binding: existing store + passing
  validation ⇒ adopt it.

So either adopt the store the same way in the switch path, or fall back to transcript recovery,
instead of refusing because the *running* process happens to be bound to a different store. At
minimum, the message should name the real condition and the remedy that works.

## Actual behavior

Fail-closed, in `apply_loaded_session_with_goal` (`crates/tui/src/tui/ui/apply.rs:3561-3573`):

```rust
if let Some(binding) = session.metadata.runtime_store.as_ref()
    && let Some(tasks) = app.runtime_services.task_manager.as_ref()
    && tasks.session_store_binding().as_ref() != Some(binding)
{
    if binding
        .is_missing_session_store()            // apply.rs:3566-3567
        .map_err(|error| error.to_string())?
    {
        recovered_binding = tasks.session_store_binding();
    }
    if recovered_binding.is_none() {
[...]

PR #6233, opened 2026-09-15
Tell the user the resume route that actually works (#6225)

### LINKED ISSUE #6225: Nifty selfdiagnosis; unable to resume from within new process
## Description

Failed to restore session: This session belongs to another Runtime host.

## Steps to reproduce

1. Start fresh `codewhale` session -> /quit `codewhale`
2. Start new `codewhale` session -> /resume -> pick the previous session (from 1.)
3. Failed to restore session: This session belongs to another Runtime host.

## Expected behavior

Resume the session

## Actual behavior

Failed to restore session: This session belongs to another Runtime host.

## Environment

- OS: macOS 26.6.2 (25G83)
- codewhale version: codewhale 0.9.13 (a0b81f619b66)
- Install method: curl
- `codewhale doctor` summary: A-OK
- Model/provider: deepseek-flash
- Terminal app: iTerm2
- Shell: (OhMy)zsh

Image

<br\>

So I let the agent run a trace, what is happening here:

What the message is

It's a Codewhale 0.9.13 guard, not a filesystem error. I found the exact literal in your binary (~/.local/bin/codewhale, 0.9.13 a0b81f619b66), sitting right next to its sibling:

runtime work is active; wait for the current turn, maintenance, and background
tasks to finish, or cancel that specific work before switching sessions
This session belongs to another Runtime host. Resume it in a new Codewhale
process to reopen its saved store.

So the refusal happens on the session-switch / submit path: your running TUI's engine can only serve one runtime store, and the session you're asking for is bound to a different one.

Why it's firing — and it is *not* a live conflict

Your previous session is:

- `ebc1d18a-56fd-4827-b565-165523397618` — title "Find your actual prompt at './The First Monday …'", same workspace (\<redacted\>), 119 messages, JSON parses clean, last written 2026-09-14T22:41:53Z (00:41 local). Nothing is lost.

Each session that 0.9.13 opens records its runtime binding in its own JSON:

- previous session → data_dir: sessions/9cc60efb-…/runtime, execution_scope: 3a94e7f4…
- this session (2b337826-…, started 00:55) → data_dir: sessions/2c9a8a3d-…/runtime, execution_scope: ac988fd6…

I probed the ownership locks directly:

- `lsof -c codewhale` → exactly one process, PID 41035, the TUI on /dev/ttys001, cwd = Planner.
- That process holds 2c9a8a3d-…/runtime/runtime-process.owner.lock (this session) — LOCKED.
- 9cc60efb-…/runtime/runtime-process.owner.lock (the previous session) — FREE, and no lsof entry. Same for its tasks/execution-owners/3a94e7f4….lock.

So no Codewhale process is actually sitting on the old store. The check is against a persisted identity, not liveness. Your current process minted a brand-new store at 00:55 (because it started a new session rather than resuming), and the older session's saved store is stamped with the scope of the earlier process (boot d51a9d03). A process that has already bound itself to one store can't re-home the other in-place — hence the prescribed fix in the message itself: open it in a fresh process, which adopts the saved store at boot.

Corroborating detail: this session's checkpoint carries the previous scope 3a94e7f4… inside its imported work graph (see .work-graph-import-archive/, sessions/checkpoints/2b337826-….json), i.e. the new session inherited the old work graph while the transcript stayed in the old session.

Fix

Quit this TUI, then from the workspace:

cd \<redacted\>
codewhale resume ebc1d18a

(codewhale resume ebc1d18a-56fd-4827-b565-165523397618 or --last also work; codewhale run --continue continues the most recent session for the workspace.) The next time, resume first instead of starting a new session — once a new session exists, that switch is what's refused.

If you hit the other guard instead — This runtime is already active in another process… — its own escape hatch is CODEWHALE_RUNTIME_DIR pointed at a different directory.

Caveats, stated plainly

- I could not execute the resume path to prove the fix: codewhale resume/thread/sessions need to write ~/.codewhale/state.db, and my sandbox is workspace-write only (they failed for me with unable to open database file). Note there is currently no state.db anywhere under ~/.codewhale — if your shell run reports failed to open state db, that's a second, separate issue.
- I also couldn't extract the exact predicate the binary compares (the execution_scope values aren't a hash of any obvious identifier — I tested workspace, hostnames, store paths, boot IDs, with several separators/prefixes). The "can't re-home in-process, use a fresh process" reading is inference from the message, the lock probes, and the binding data — consistent, but not proven end-to-end.
- Fallback if a fresh process still refuses: the stale pointer is the runtime_store block in ~/.codewhale/sessions/ebc1d18a-….json. Backing that file up and clearing that block (or the session_boot_owners.json stamp) should let a new process re-home the session. I have not tested that and won't touch it without your say-so.

One question that would sharpen this: did you hit it via the in-TUI session picker//resume, or on the command line when you launched at 00:55?

---

Pretty nifty; resuming (the exact process) from the CLI works!

Hmbown:
The session-switch refusal told users to do something impossible.

```
This session belongs to another Runtime host.
Resume it in a new Codewhale process to reopen its saved store.
```

#6225 hits this from the plainest path on a clean install — start `codewhale`, `/quit`, start `codewhale`, `/resume`, pick the session — and the advice cannot be followed, because they **were** in a new process.

## Why the advice was circular

Starting a new process and then picking the session from `/resume` returns to this same switch path. The route that actually works is opening the session **at launch**: `TaskManager::start` (`event_loop.rs:849`) passes the saved binding through to `open_for_session`, which calls `validate_existing_store()`, points the runtime at that store and opens it. That is adoption. The picker inside a running session is not — it can rebind the conversation but cannot carry a store's durable work across, so it refuses.

Both halves were true; neither was actionable.

## What this changes

Only the message, which now names the condition and the command, with the session id interpolated so it is copy-pasteable:

> This session's saved Runtime store belongs to a different host. Switching to it from inside a running session cannot carry that store's queued work across, but opening it directly can: run `codewhale resume <id>` from your shell.

**This is not a fix for the refusal itself.** That is #6207, and it is still wrong for the case #6225 reports — a store that exists, is owned by nobody, and holds nothing. But users should not wait on that to reach their sessions, and `codewhale resume <id>` works today on 0.9.13 with no patch.

## Test

The binding test asserted the old literal string. It now asserts the two properties that matter — the refusal points at the direct-open path, and it names the session — rather than pinning wording:

```rust
assert!(error.contains("codewhale resume"), ...);
assert!(error.contains(&loaded.metadata.id), ...);
```

## Checks

- `./scripts/dev-test.sh tui` — **12707 run, 12707 passed**, 19 skipped
- `./scripts/dev-test.sh tui runtime_store_binding` — 7 passed
- CI clippy gate `--workspace --all-targets --all-features -D warnings` — exits 0
- `cargo fmt --check` — clean on both files

No-Issue: message-only remediation. This deliberately does not close #6225 — it makes the refusal's advice followable (`codewhale resume <id>`) without fixing the refusal itself, which is #6207 and stays open, as the body above states. Using `Closes` here would auto-close a bug that is still reproducible.


Hmbown:
I had written the same message fix independently before reading this PR, and have dropped my commit — this one is better (full id, copy-pasteable). One thing mine had that this does not, offered as a note rather than a change:

The advice is only actionable when the store can actually be adopted. `open_for_session` calls `binding.validate_existing_store()`, which fails closed on a store whose `owner.json` is unreadable or whose recomputed scope does not match the saved one. In that case `codewhale resume <id>` refuses too, and the new message has replaced one circle with another — a quieter one, because the user now has a command that looks authoritative.

Guarding it is a couple of lines at the same site:

```rust
let remedy = match binding.validate_existing_store() {
    Ok(()) => format!("…run `codewhale resume {}` from your shell.", session.metadata.id),
    Err(error) => format!("That store cannot be adopted as it stands: {error:#}"),
};
```

Not a blocker — the common case is a store that validates, which is exactly #6225. Take it or leave it.

PR #6240, opened 2026-09-15
Salvage: GitHub App review precision, pinned source context and evidence-based findings

### LINKED ISSUE #6135: Improve GitHub App review precision with pinned source context and evidence-based findings
Founder requested higher-quality coding feedback specifically from the GitHub App that reviews pull requests. The verified published review on PR Hmbown/Codewhale#5842 uses the Actions-backed `codewhale review --pr --post` path; its claim that matching RegistryMutation twice necessarily moves it needs a compiler-backed negative control.

Executing owner: Codex, this GitHub review-quality task. C06 behavioral quality reference; this is not Auto-Review permission policy. Claim exact paths in an isolated Codewhale worktree: crates/tui/src/tools/review.rs, crates/tui/src/tools/review_pr.rs, crates/tui/src/tools/review_hunks.rs, the review prompt assembly hunk in crates/tui/src/lib.rs, and docs/GITHUB_APP.md. Preserve the shared checkout's dirty lanes.

Scope: reuse the shared PR snapshot and hunk authority to supply bounded surrounding source from the exact head, retain the complete original diff and existing input/pass budgets, make missing context explicit, strengthen the reviewer contract around reachable introduced defects, actionable fixes and honest validation. Verify pinned-head/dirty-checkout isolation, source bounds and anchors with local fixtures; run npm test && npm run check:web and focused Rust checks.

Evidence lanes remain separate. Local implementation and commits only; no provider calls, public push, posted review, deployment, or new model budget is authorized by this slice.

Hmbown:
Rescues `14eb885d4e` off `fix/github-review-quality-20260913`, a branch that has had **no PR since 2026-09-13** and that nobody had ever built.

Partially addresses #6135. See "Why this does not close it" below.

## What it does

Gives the GitHub App reviewer pinned source context around each hunk and pushes it toward concrete, evidence-bearing findings instead of generic commentary. 5 files, +406/−25: a new `crates/tui/src/tools/review_pr.rs`, changes to `review.rs` and `review_hunks.rs`, one `lib.rs` wiring line, and `docs/GITHUB_APP.md`.

## Why a cherry-pick and not a branch merge

**Do not merge that branch.** It carries a second, unrelated commit — `4517d294c0`, a Computer Use 0.2.2 sync — which conflicts across ~8 `crates/tui/plugins/computer-use/` paths. An earlier audit reported the branch as "zero conflicts"; that measured the *whole-branch* merge and is misleading as written.

The single sha applies clean, verified before picking:

```
git merge-tree --write-tree --merge-base=14eb885d4e^ origin/main 14eb885d4e   # exit 0
git diff --stat origin/main <resulting-tree>                                   # == the commit's own diffstat
```

Nothing rides along.

## The part that mattered

The commit **compiles but does not pass CI's clippy line**:

```
error: this expression creates a reference which is immediately dereferenced by the compiler
   --> crates/tui/src/tools/review_pr.rs:625:63
    = note: `-D clippy::needless-borrow` implied by `-D warnings`
```

Fixed in the second commit here. A clean cherry-pick tells you the text applied; it tells you nothing about whether the result builds under the gates. This one had sat unbuilt for two days.

## Why this does not close #6135

Two gaps, both worth someone's attention rather than being buried under a merge:

1. **The negative control named in the issue body has no in-repo artifact.** The issue calls for a compiler-backed `RegistryMutation` control; what exists is an assertion in a commit message. Until that control exists as something CI runs, the precision claim is unverified.
2. **`source_context` fails quiet on the cases that need it most.** It returns `None` when `budget < 512`, and on overflow it discards an already-built report *including* its `unavailable_files` counters. So the large PRs with the most surrounding context to lose get none, and emit no receipt saying so. That is a silent-failure shape, not a size limit.

#6135 stays open for both.

## Evidence

```
cargo check -p codewhale-tui --all-features --locked        clean
cargo fmt --all -- --check                                  clean
cargo clippy --workspace --all-targets --all-features
  --locked  (CI's exact allow list)                         clean  (after the fix)
test result: ok. 60 passed; 0 failed   (tools::review, --test-threads=2)
check-blocking-calls-budget.py   624 sites / 180 files, within budget
check-dead-code-budget.py        PASS, 254 at budget
```


Hmbown:
## Windows failure: cause, and a fix that keeps the coverage

`Test (windows-latest)` fails on this branch:

```
FAIL [0.586s] (9569/15288) codewhale-tui tools::review_pr::tests::source_context_is_bounded_line_exact_and_uses_literal_paths
panicked at crates\tui\src\tools\review_pr.rs:692:56:
called `Result::unwrap()` on an `Err` value: Os { code: 123, kind: InvalidFilename,
  message: "The filename, directory name, or volume label syntax is incorrect." }
```

**Cause.** The test creates a file literally named `[literal]*.rs` to prove pathspecs are passed literally rather than globbed. `*` is one of Windows' reserved filename characters (`< > : " / \ | ? *`), so `std::fs::write` cannot create it at all. The test was never runnable on Windows — it is not a regression, it arrived that way in `14eb885d4e`, on a branch that had no PR and that nobody built.

**I see a `#[cfg(unix)]` gate in progress in a shared worktree.** That works, and I have deliberately not touched it — but it drops Windows coverage for the literal-pathspec guard permanently, and that guard exists because a glob interpretation silently reads the *wrong file*. Windows is where a path bug is most likely to differ.

**Alternative that keeps all three platforms.** Change one line:

```rust
let path = "[l]iteral-other.rs";
```

`[` and `]` are legal in NTFS filenames; only `* ? < > : " / \ |` are not. The property under test survives intact, because the decoy is still reachable by a glob interpretation:

- read **literally**, the pattern names this file: `[l]iteral-other.rs`
- read as a **glob**, `[l]` matches the single character `l`, so the pattern resolves to `literal-other.rs` — the existing `"wrong glob match"` decoy the test already creates two lines below

So `assert!(!context.to_string().contains("wrong glob match"))` still fails if anything globs, which is the whole point of the test. The `nested/` relative-path decoy is unaffected.

Happy either way — flagging it because the cfg-gate is the kind of fix that is invisible later, and this guard is worth keeping on the platform where paths behave differently.

For the record, this is the second thing this salvaged commit needed that a clean cherry-pick could not reveal: it also failed CI's clippy on a needless borrow (fixed in `672a54a22`). Both are the cost of a commit sitting unbuilt on a branch with no PR.

Hmbown:
## Windows: restored rather than gated — `1c29c4423`

`55fffb3db` got the diagnosis exactly right, and that diagnosis is what made the fix obvious: `*` is a reserved NTFS filename character, so `std::fs::write(dir.path().join("[literal]*.rs"))` returned `InvalidFilename` (os 123) before a single assertion ran. **That test has never executed on Windows since it was written** — this was not a regression.

I've added a commit on top rather than rewriting that one, because the analysis in it is worth keeping in the history.

What changed: the `#[cfg(unix)]` gate comes off and the fixture is spelled portably.

```rust
let path = "[l]iteral-other.rs";
```

Only NTFS-legal characters — the reserved set is `< > : " / \ | ? *`, and brackets are not in it — and it proves exactly the same property:

- read **literally**, the pathspec names that file;
- read as a **glob**, `[l]` matches the single character `l` and the pattern resolves to `literal-other.rs`, the "wrong glob match" decoy the test already creates two lines down.

So an implementation that globs still reads the decoy and still trips `assert!(!context.to_string().contains("wrong glob match"))`. The `nested/` relative-path decoy is unaffected.

Verified empirically rather than reasoned from the spec — with both files present:

```
$ ls
[l]iteral-other.rs
literal-other.rs
$ for m in [l]iteral-other.rs; do echo "$m"; done
literal-other.rs
```

**Why not just take the gate.** It works, and it is the smaller diff. But it gives up this guard on the one platform whose filename rules caused the failure — and path handling is precisely where Windows differs from the other two. A guard that exists because globbing silently reads the *wrong file* is worth keeping where paths behave differently.

Honest limits: I could not compile or run this locally. Both worktrees that hold this branch are occupied by other sessions — one is mid-release-build, the other holds live work — and I made this commit through git plumbing specifically so I would not write into either. CI runs all three platforms on this PR and is the verification. If Windows still fails, the gate is the right fallback and I will say so.

Thanks to whoever pushed `55fffb3db` — finding `os 123` is a much better root cause than "Windows is flaky", and the remedy was the only part worth changing.

Hmbown:
Merging with one documented red: `Codewhale review` failed on provider spend, not code — pass 1/1 hit provider stop reason `length` after burning the full 65,536-token output budget on reasoning (job 104605859980, usage: 23,591 in / 65,536 out, publication: not_attempted). Every substantive gate is green: Test on macos/ubuntu/windows, Lint, Safety gate, npm smoke, DCO, link. Per the #6229 precedent this class of red does not block.

Issue #6242, opened 2026-09-15
workspace_declares_exactly_one_turn_loop has been green the whole time main has had two turn loops — it matches a name, not a property


Hmbown:
`crates/core/tests/single_turn_loop.rs` is the guard for the product's central
architectural rule: **one turn loop.** It has been passing continuously while
`main` has had two.

## What it actually asserts

```rust
if trimmed.starts_with("async fn run_turn")
    || trimmed.starts_with("pub async fn run_turn")
    || trimmed.starts_with("pub(crate) async fn run_turn")
    || trimmed.starts_with("pub(super) async fn run_turn")
```

It scans for a function **named `run_turn`** and asserts there is exactly one.
Today that finds `crates/tui/src/core/engine/turn_loop.rs:669`, and passes.

## What it cannot see

`crates/tui/src/acp_server.rs`:

```
:69    const MAX_ACP_TOOL_ROUNDS: usize = 50;
:1286  async fn run_agentic_prompt_turn<R, W, F, Fut>(
:1305      for _round in 0..MAX_ACP_TOOL_ROUNDS {
```

That is a turn loop — its own bounded tool-call round loop, its own prompt
assembly, its own conversation store — and the guard is blind to it because it
is not spelled `run_turn`. This is the subject of #6088, filed independently by
reading the code. The guard contributed nothing to finding it and would not
have.

## Why this is worth fixing rather than shrugging at

A guard that can only be satisfied by naming is a guard an author passes by
choosing a different name — and nobody chose a different name maliciously here,
which is the point. `run_agentic_prompt_turn` is a perfectly natural name. The
rule it violates is one of the few this product describes as existential:

> A second turn loop, a second event authority, or a second prompt authority
> anywhere in the stack is the thing this product exists to not be.

A test that reports compliance while the violation is 5,000 lines away in the
same crate is worse than no test, because it is cited as evidence.

## Ask

Assert the property, not the identifier. Some candidates, none of them free:

- Detect the *shape*: a bounded loop over tool-call rounds that calls a
  provider-streaming entry point. More work, catches renames.
- Make the allow-list explicit: enumerate the files permitted to drive a turn
  and fail on any other file that calls the streaming client directly.
  `acp_server.rs` calls `client.create_message_stream` directly today, which is
  the cheap signal.
- At minimum, extend the name match to the known second loop and make the test's
  own doc comment state plainly what it does and does not cover, so nobody reads
  a pass as proof of the rule.

[...]

Issue #6241, opened 2026-09-15
The dead-code ratchet is blind to expect(dead_code), so the budget can fall without the suppression falling


Hmbown:
`scripts/check-dead-code-budget.py` counts one spelling:

```python
PATTERN = re.compile(r"allow\(\s*dead_code\b")
```

`expect(dead_code)` is a suppression too, and the ratchet cannot see it. So
rewriting `#[allow(dead_code)]` as `#[expect(dead_code)]` lowers the budget
number by the full count while removing no dead code at all.

## This already happened, in the sweep the ratchet was built to measure

`137fb70a9` (#5587) removed 119 `allow(dead_code)` and added 92
`expect(dead_code)`. The budget recorded a 115-point improvement. Real
suppression sites went from 384 to 363 — **a reduction of 21**.

On `main` today:

```
allow(dead_code)   244      <- the only thing the ratchet counts
expect(dead_code)  109      <- invisible to it
                   ---
total              353      against a recorded budget of 254
```

The gate passes. It is measuring a spelling, not a property.

## Why `expect` makes this worse rather than better

`#[expect]` is the better attribute — it errors when the lint *stops* firing, so
it cannot rot silently the way `#[allow]` does. That is precisely why a sweep
will keep converting toward it, and precisely why the ratchet will keep reading
those conversions as progress. Left alone, the number drifts to zero while the
suppressed surface stays where it is.

## Ask

Count both spellings. `expect` is still worth preferring, so if the two should
not be weighted equally, track them as two numbers rather than one — but do not
leave a gate whose headline figure improves when nothing changed.

A regression test would also help: the peer budget scripts have
`scripts/test_check_runtime_contract_budget.py` and
`test_check_persistence_backlog_budget.py`; there is no
`test_check_dead_code_budget.py`.

## Evidence

Counts above are from `origin/main` at `bce761083`:

```sh
grep -rEo 'allow\(\s*dead_code'  crates --include='*.rs' | wc -l   # 244
grep -rEo 'expect\(\s*dead_code' crates --include='*.rs' | wc -l   # 109
python3 -c "import json;print(json.load(open('scripts/dead-code-budget.json'))['total'])"   # 254
```

Found while adversarially re-verifying a claim that #5587 was complete. It is
not — but that is #5587's problem. This is the gate's.

PR #6243, opened 2026-09-15
Budget the editor-handoff pause sleep, and note why main looked green without it

### LINKED ISSUE #6165: `/hooks edit` hands the terminal to $EDITOR without pausing the TUI input thread — keystrokes get split between the editor and the composer
## Summary

`/hooks edit` (also `/hooks add`, `/hooks new`) launches `$VISUAL`/`$EDITOR` on
`.codewhale/hooks.toml` **without pausing the TUI terminal-input thread**. The TUI keeps
reading stdin while the editor owns the terminal, so the user's keystrokes are split
between the two readers. The result is that the editor cannot be quit interactively
(`Esc`, `:q!`, `Enter` never arrive intact), the TUI simultaneously swallows the same
keystrokes into the composer, and the session looks completely frozen.

Every other place that hands the terminal to a child does perform the handshake
(`Ctrl+Shift+O` composer editor, interactive shell commands). This one path was missed.

## Environment

- codewhale `0.9.13` (npm wrapper `0.9.13` + native binary built from the `codewhale-cli`
  `0.9.13` crate; `codewhale --version` → `codewhale 0.9.13`)
- Linux, xfce4-terminal, PTY `271x49`, `TERM=xterm-256color`
- `$VISUAL` and `$EDITOR` both unset → the documented `vi` fallback is used
- Editor observed: `vi` (vim), which creates `.hooks.toml.swp` in `.codewhale/`

## Steps to reproduce

1. Start the TUI in a trusted workspace.
2. Type `/hooks edit` and press `Enter`.
3. `vi` opens on `<workspace>/.codewhale/hooks.toml` (template is seeded on first use).
4. Press `Esc`, then `:q!`, then `Enter`.

## Expected behavior

`vi` exits, control returns to the TUI, the Hooks screen reports "Hooks unchanged." — the
normal external-editor round trip.

## Actual behavior

- The editor does **not** exit. Nothing the user types is enough to leave it.
- The keystrokes are *partially* delivered: in a controlled run, `:` and `!` landed in
  vi's command line, while `Esc` and `Enter` were consumed by the TUI. The editor ends up
  in an unusable half-state (command line open, no way to confirm or cancel).
- The TUI consumes the rest, so fragments of the user's editing are submitted as composer
  input. In the reported session, `~/.deepseek/composer_history.txt` contained the entries
  `q`, `!!`, `q` sitting between two genuine prompts — consistent with the pieces of a
  `:q!` quit attempt being captured by the TUI (including its `Enter`) instead of reaching
  the editor.
- The session appears hung. The only reliable recovery is to kill the editor (or the
  session) from a *different* terminal — which the user cannot do if the TUI is their only
  terminal.

## Root cause

The TUI reads stdin on a dedicated pump thread, `TerminalInputPump`:

- `codewhale-tui/src/tui/ui/terminal_input.rs:77` — `spawn_parts()` spawns the
  `codewhale-terminal-input` thread.
- `codewhale-tui/src/tui/ui/terminal_input.rs:90-98` — the loop only stops reading when
  `thread_paused` is set; otherwise it runs `event::poll(...)` and `event::read()`
  continuously, consuming keystrokes that a foreground child should own.
- `codewhale-tui/src/tui/ui/terminal_input.rs:199` — `pause_for_child_terminal()` is the
  only mechanism that makes the pump yield the terminal (it waits for `paused_ack`, with a
  500 ms timeout).

`with_suspended_tui()` (`codewhale-tui/src/tui/external_editor.rs:200`) handles **only**
crossterm state — `disable_raw_mode()` at `:222`, `leave_alt_screen()` at `:224`, and the
mirror calls at `:231`/`:233`. It has no access to the pump and therefore never pauses it.
Suspending raw mode alone is not sufficient: the pump thread keeps calling `event::read()`
on the same tty regardless of the terminal's mode.

The two external-editor call sites differ:

**Correct — composer editor (`Ctrl+Shift+O`), `codewhale-tui/src/tui/ui/event_loop.rs`:**

```rust
// :6505
let editor_result = terminal_input.pause_for_child_terminal().and_then(|()| {
    let result = prepare_terminal_input_handoff(&terminal_input, &mut pending_terminal_events)
        .and_then(|ready| {
            if ready {
                crate::tui::external_editor::spawn_editor_for_input(/* :6512 */ )
            } else { /* refuse handoff on pending Esc / Ctrl+C */ }
        });
    terminal_input.resume_after_child_terminal();   // :6526
    result
});
```

This does all three things: pause the pump, drain buffered input so it cannot leak into the
child (`prepare_terminal_input_handoff`, `codewhale-tui/src/tui/ui/terminal.rs:44`), and
refuse the handoff if a cancellation key is pending. The same pattern is used for
interactive shells (`event_loop.rs:3191`).

**Broken — `/hooks edit`, `codewhale-tui/src/tui/ui/apply.rs`:**

```rust
// :2075
AppAction::EditProjectHooks => {
    edit_project_hooks_from_tui(terminal, app, config);   // no terminal_input available
}

// :2550
fn edit_project_hooks_from_tui(terminal: &mut AppTerminal, app: &mut App, config: &Config) {
    // ...
    // :2565
    let outcome = crate::tui::external_editor::spawn_editor_for_path(
        terminal,
        app.use_alt_screen(),
        app.use_mouse_capture,
        app.use_bracketed_paste,
        &path,
    );
}
```

Neither the action handler nor `edit_project_hooks_from_tui` has a `&TerminalInputPump`.
The signature makes the handshake structurally impossible at this layer, so it can never be
added by accident — it was simply never wired up.

Note the same asymmetry exists between `spawn_editor_for_input` (used by the correct path)
and `spawn_editor_for_path` (used by the broken one): both call `with_suspended_tui`, so
both *look* correct, but the pause lives in the caller and only one caller has it.

## Controlled reproduction

Run on the reporter's machine against `0.9.13`, in an isolated workspace so no real project
files were touched. Full script: `codewhale-issue-experiment.py` (attached alongside this
report). It allocates its own PTY, runs a throwaway `codewhale`, drives the UI, and checks
`ps` for the spawned `vi`.

Observed output (abridged):

```
>>> workspace-trust prompt detected; pressing 1 (trust this workspace)
>>> typing: /hooks edit (char by char)
[check 1] vi processes: ['3984267 3981902 Sl+  pts/34   vi /tmp/.../work/.codewhale/hooks.toml']

>>> injecting exit keys: ESC : q ! ENTER (spread out)
[check 2] vi processes after exit keys: ['3984267 3981902 Sl+  pts/34   vi /tmp/.../work/.codewhale/hooks.toml']

RESULT: vi STILL RUNNING -> keystrokes swallowed (reproduced)
```

The final screen dump of the PTY shows vi's status line ending in `::!` — the `:` and `!`
of the exit sequence reached the editor, but `Esc` and `Enter` did not. That split is the
signature of two readers competing for one tty, and it is reproducible on demand.

## Impact

- `/hooks edit` is unusable with any terminal editor (`vi`, `vim`, `nano`, `emacs -nw`),
  which is exactly the fallback the resolver picks when `$VISUAL`/`$EDITOR` are unset
  (`external_editor.rs:40-50`).
- The failure mode is maximally confusing: the editor *partially* responds, so it does not
  look like a missing handoff — it looks like a hung agent.
- The user's typed characters are captured as composer input, so failed quit attempts are
  submitted as messages to the model.
- Recovery requires a second terminal. On a single-terminal setup the whole session is
  lost.

## Related path with the same defect

`codewhale-tui/src/tui/history.rs:2937` (`try_open_file_at_line`, invoked on mouse click of a
`path:line` in tool output via `codewhale-tui/src/tui/mouse_ui.rs:1680`) spawns `$EDITOR`
with `Command::new(&editor).spawn()` at `:2963` and never suspends the TUI at all — not even
raw mode or the alt screen. Its fallback is also `vim`. Terminal editors launched from that
path compete for the same tty while the TUI keeps rendering.

## Companion defect: the same missing invariant, in the other direction

There is a mirror image of this bug, reported separately in
`codewhale-issue-sigttin-suspend-poisons-terminal.md`: the TUI's *own* process group can end
up in the background, at which point its next read of the tty raises `SIGTTIN` and the kernel
stops the whole group mid-turn. Because the TUI installs no `SIGTSTP`/`SIGTTIN`/`SIGCONT`
handler, the very modes this file's code path also enables — raw mode, mouse capture,
bracketed paste, the alternate screen — are never torn down, and the terminal goes on to
"poison the shell with raw escape reports"
(`codewhale-tui/src/tui/ui/terminal.rs:119-120`).

The two defects are opposite failures of the same thing. Here the TUI keeps competing for a
tty that a child owns; there the TUI loses its claim on the tty entirely and is stopped by
the kernel. In both cases **terminal ownership is implemented per call-site rather than as
one enforced invariant** — every "hand the terminal to a child" and every "take it back"
decision is made by whichever call site remembers to make it.

## Test gap

There are unit tests for the handshake primitives themselves —
`codewhale-tui/src/tui/ui/tests.rs:25635`
(`terminal_input_child_pause_drains_codewhale_events_before_editor_handoff`) and `:25686`
(`terminal_input_handoff_preserves_pending_cancellation_keys`) — but nothing asserts that
each editor entry point uses them. `spawn_editor_for_path` and
`edit_project_hooks_from_tui` have no test coverage at all, which is why the missing pairing
went unnoticed.

## Suggested fix

Items 1–4 each close this bug; item 5 is what keeps it closed.

1. Thread `&TerminalInputPump` down to `edit_project_hooks_from_tui` and wrap the
   `spawn_editor_for_path` call in the same `pause → prepare_terminal_input_handoff →
   resume` sequence used at `event_loop.rs:6505`. Extract that sequence into one helper so
   both call sites share it.
2. Move the pause into `with_suspended_tui` by giving it access to the pump, so *every*
   caller is correct by construction and the "caller must remember to pause" invariant
   disappears.
3. Add a regression test that enumerates external-editor entry points and asserts each one
   pauses the pump before spawning — or a PTY-level integration test in the style of the
   reproduction script above.
4. Whatever the fix, consider documenting that `$EDITOR` must be a GUI editor or one that
   returns immediately (`code --wait`, `subl -w`), since that is currently the only safe
   configuration for the terminal-editor paths.
5. Reduce all of the above to one invariant instead of more call-site fixes: a single
   "terminal ownership" abstraction that owns raw mode, the mouse/paste/alt-screen modes and
   the input pump together, and that every path must pass through — handing the terminal to a
   child, stopping for a job-control signal, resuming, or exiting. Items 1–2 fix this entry
   point; only this generalises to the companion defect above, and to the next call site
   nobody has written yet.

## Reporter notes

- The `.hooks.toml.swp` swap file is left behind when the editor is killed, and the
  template file itself is unchanged; no data was lost in the reported incident.
- Killing the editor from another terminal (`kill -TERM <vi pid>`) does restore the TUI, so
  the TUI side is not itself deadlocked — it is blocked in `wait()` on the child while its
  input pump keeps stealing keystrokes.

Hmbown:
One line of JSON, plus something about the gate that is worth knowing.

No-Issue: ratchet bookkeeping for a site #6165 landed deliberately; the underlying convention is #6149.

## What

#6239 (which closed #6165) added a bounded `thread::sleep` to
`tui/ui/terminal_input.rs::pause_for_child_terminal`. The blocking-calls ratchet
has no entry for that file, so it now reports:

```
crates/tui/src/tui/ui/terminal_input.rs: thread_sleep sites 1 > budget 0
```

This adds the one entry. Nothing else in the file changes.

## Why budgeted and not rewritten

This is the case `check-blocking-calls-budget.py`'s own message names — "raise
the budget with `--update` if the site can only run on synchronous code" — and
its author had already written the justification at the call site:

> a bounded retry, capped by `TERMINAL_INPUT_CHILD_PAUSE_TIMEOUT`, in a
> synchronous API whose caller is about to block this very thread on a
> foreground editor for as long as the user keeps it open. `tokio::time` is not
> reachable from here and would not change what the thread does.

That reads correctly to me. The caller is about to hand the terminal to
`$EDITOR` and block until the user quits it; a few polled milliseconds before
that is not what parks the thread. The sibling sleep at `:192` is already inside
a `thread::Builder::spawn`, which is why the scanner never counted it.

## The part worth reading

**Main is currently green on this, and that is not evidence.**

The budget steps are gated on `if: needs.changes.outputs.heavy == 'true'`. On
main's run for `e8a226552` all four budget steps report `skipped`:

```
skipped  Check dead-code budget
skipped  Check blocking-calls budget
skipped  Check runtime-contract budget
skipped  Check persistence-backlog budget
```

Combined with `continue-on-error` on pull requests, a violation can cross a PR
advisory-only and then land on a main push that never runs the check. That is
exactly the path this took: green PR, green main, real violation, fires on
whatever heavy push comes next — at which point it looks like that unrelated
push's fault.

I am not changing the gating here; narrowing `heavy` or making these steps
unconditional is a CI-cost tradeoff someone should make deliberately. But the
current shape means **a green main does not tell you the ratchets passed**, and
that seemed worth stating where someone will read it.

## Evidence

```
check-blocking-calls-budget.py   625 sites across 181 files, within budget
```

Diff is one added object:

```json
+  "crates/tui/src/tui/ui/terminal_input.rs": {
+    "thread_sleep": 1
+  },
```


Hmbown:
## A second, larger blind spot in this same gate — partly confirmed, partly not

*(Reposted: the first version of this comment lost several filenames to shell backtick expansion.)*

A multi-agent review of the guard surface flagged `scripts/check-blocking-calls-budget.py` for a defect beyond the one this PR patches. I verified the mechanism and could **not** reproduce the headline number, so both are below.

### Confirmed: the pattern anchors on the literal `std::fs::`

```python
"std_fs": re.compile(
    r"\bstd::fs::(?:read|read_to_string|write|create_dir|…)\b"
),
```

A file that does `use std::fs;` and then calls `fs::read_to_string(...)` — the more idiomatic of the two spellings — is invisible to this gate. Same call, same blocking syscall, not counted.

Measured on `origin/main` (raw occurrences, no scope filtering):

```
std::fs:: sites the gate can see : 3253
bare  fs:: sites it cannot see   :  476  across 390 files
```

Largest uncounted: `crates/tui/src/runtime_api/tests.rs` 298, `crates/tui/src/config/tests.rs` 296, `crates/tui/src/project_context.rs` 184, `crates/tui/src/session_manager.rs` 154, `crates/tui/src/plugins/tests.rs` 121, `crates/config/src/tests.rs` 120.

### Not confirmed: "the gate sees 45% of sites"

The review reported 624 recorded sites against 1403 under a widened pattern. I could not reproduce that ratio. The two figures measure different things — the budget's 624 is **after** scope filtering (tests, `spawn_blocking` and dedicated-thread scopes are excluded), while a raw pattern count is not. My raw numbers put the blind spot nearer 13% of occurrences, and a large share of those sit in `tests.rs` files the gate deliberately ignores anyway.

**So: the blind spot is real, its size is unconfirmed, and the honest production-only number is the one nobody has measured yet.** Widening the regex and re-running `--update` would produce it in one command; I did not do that here because it would bury a one-line fix under a re-baseline.

### Why it belongs in its own issue, not this PR

Widening the pattern moves the budget number for reasons unrelated to any code change. That deserves its own commit whose message says the number rose because the measurement got honest, not because the tree got worse. Folding it in here would make this PR's one-line intent unreviewable.

It is also another instance of the class in **#6242** — a gate that asserts a spelling rather than a property. That issue is the archetype; this is the third confirmed instance, after the dead-code ratchet (**#6241**) and the skipped-steps problem described in this PR's own description.

One fact for whoever picks it up: there are **zero** `use tokio::fs` sites in the tree, so a widened pattern needs no import-gating to avoid false positives.

Issue #6213, opened 2026-09-15
v0.9.14: Eliminate per-frame/per-delta recomputation and clones in the TUI hot path


Hmbown:
From `codewhale-ops/PERF-OPPORTUNITIES-20260915.md` §3 (T1, T4–T7) — per-frame and per-delta work in the TUI that scales with output size / conversation length / catalog size. Each item independently shippable; confirm the call path on [S] items.

- [ ] **T1 [V]** — `tui/src/tui/output_rows_cache.rs:107`: on a cache **hit** (the common case during streaming at up to 120 FPS) it returns `entry.rows.clone()` — every `String` + styled span per row. Caller `history/tool_output.rs:372` also runs `hash_str(output)` over the full output bytes per frame per cell to build the key. Fix: store `Arc<Vec<OutputRow>>` and clone the Arc (the "caller iterates without a lock" property the doc comment wants is preserved); key the cache externally by `(tool-result id, width)` so the per-frame full-output hash disappears.
- [ ] **T4 [V]** — `core/engine/turn_loop.rs:5234–5262`: every `InputJsonDelta` pushes `partial_json`, then `parse_tool_input` runs the full repair ladder (`arg_repair::repair` + up to 3 `serde_json::from_str` fallbacks) over the **entire** accumulated buffer and `parsed.value.clone()`s the result. A 100KB `apply_patch` in ~2KB deltas = ~50 full parses + 50 deep clones per tool call, on the per-token path. Fix: mirror into `tool_state.input` only at `ContentBlockStop` (already the authoritative parse point), or every N bytes / on a debounce; `swap` the parsed value in instead of cloning. (The sibling O(n²) verbose-log copy at this site was already fixed — this is the surviving half.)
- [ ] **T5 [S]** — `core/engine/tool_catalog.rs:737–744,853–866`: `tool_search` runs `tool.input_schema.to_string().to_lowercase()` per deferred tool per query, plus `tool.name.to_lowercase()` inside the per-term scoring loop. The catalog is static between queries — build `haystack` + `name_lower` once at catalog construction (side map keyed by tool name).
- [ ] **T6 [S]** — `tui/views/mod.rs:2696–2729` `row_matches_filter`: ~11 `to_lowercase` allocations per settings row, several passes per interaction (`visible_items()` + `matching_row_indices()`). Compute lowered filter terms once per `update_filter`; cache a lowered search blob per `ConfigRow` at load.
[...]

PR #6267, opened 2026-09-16
perf(tui): lower the settings filter once per pass, not once per row (#6213 T6)

Hmbown:
## Summary

Part of #6213 (item **T6**).

No-Issue: partial work on #6213 — the filter-term hoist only; T4, T5 and the per-row half of T6 remain, so that issue must stay open.

## What changed

`row_matches_filter` lowercased and split `self.filter` **on every row**, and the settings view calls it twice per interaction — once from `visible_items`, once from `matching_row_indices`. For a filter like `auto approve`, that is the same `trim().to_lowercase()` and the same `split_whitespace()` repeated `2 × rows` times per keystroke.

The term list moves into `filter_terms()`, computed once per pass and passed in as `&[String]`. Empty terms still mean "everything matches", and the matching expression is untouched, so what the filter accepts is identical.

## What deliberately did not change

The larger half of T6 — lowercasing each row's own label, value, scope and hint (about 10 allocations per row per pass) — is **not** here. Caching those needs the `row_display_value` dependency settled first: unlike the rest, it is not a pure function of the row, so a cached blob keyed on the row alone could go stale on screen. I would rather leave that visible than ship a cache with an unsound invalidation rule.

## Verification

```
cargo check -p codewhale-tui --all-targets --all-features --locked   (clean)
test result: ok. 329 passed; 0 failed; 0 ignored; 0 measured; 12479 filtered out
    (tui::views::)
```

Local only, macOS/aarch64. No hosted CI claim; no benchmark, so the saving is argued from the removed work rather than measured.


Hmbown:
Re-ran `Test (ubuntu-latest)`; it passed. All five required checks are green, including `Test (macos-latest)`.

The original failure was **not** the SHA-6380 env-var class the handoff note guessed. Under cargo-nextest each test runs in its own process, and this test spawns the real binary with `env_clear()` and a fresh `TempDir` HOME — an intra-process env race is structurally impossible. It was a 250 ms `CLI_PERSIST_TIMEOUT` race that lost the dry-run receipt under Ubuntu load, and it had happened once before on PR #6104 (`c810bc0458`, 2026-09-12) at the same `:180`.

Tracked as #6269, with the test-side half in #6270.

PR #6171, opened 2026-09-14
feat(providers): add AICraft OpenAI-compatible provider template

BX166:
## Summary

Adds **AICraft** (`aicraftapi.com`) as an OpenAI-compatible provider setup template, following the same descriptor-row pattern as SenseNova, Baseten, Groq, Cerebras, and Command Code.

- `AICRAFT_TEMPLATE_ID = "aicraft"`
- `AICRAFT_BASE_URL = "https://aicraftapi.com/v1"`
- `AICRAFT_DEFAULT_MODEL = "claude-4.6-sonnet"`
- `AICRAFT_API_KEY_ENV = "AICRAFT_API_KEY"`

## Changes

- Four new consts plus one `ProviderSetupApply::Compatible` descriptor row, inserted before the unpublished `agnes` entry.
- Updated `settings_value_names_fillable_then_unpublished` to include `aicraft` in the fillable list.
- Added an `aicraft` entry to `hosted_openai_compat_hosts_are_templates_not_enum_variants`.
- Updated the module header comment.

## Test plan

- The two touched unit tests (`settings_value_names_fillable_then_unpublished`, `hosted_openai_compat_hosts_are_templates_not_enum_variants`) are expected to pass under the existing `crates/config` test suite.

---

<a href="https://app.devin.ai/review/hmbown/codewhale/pull/6171" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

No-Issue: provider setup template addition; no tracking issue was filed for AICraft. (Line added by a maintainer to satisfy the required `link` check — the contribution itself is unchanged.)

PR #6288, opened 2026-09-16
feat(providers): add the AICraft OpenAI-compatible provider template (#6171)

Hmbown:
Lands @BX166's contribution from #6171. Their two commits are here verbatim,
with their authorship intact — this branch exists only because the fix below
could not be pushed to their fork.

## What this adds

An `AICraft` (`aicraftapi.com`) OpenAI-compatible provider setup template,
following the descriptor-row pattern already used by SenseNova, Baseten, Groq,
Cerebras and Command Code: four consts plus one
`ProviderSetupApply::Compatible` row, inserted before the unpublished `agnes`
entry, with the two touched unit tests updated and the row documented in
`docs/PROVIDERS.md`.

No new enum variant and no compiled model roster — live `GET /v1/models` and the
Codewhale catalog remain the offering list.

## Why an integration branch

Their PR's required `Lint` check was red on `Check formatting`: once `aicraft`
joins the tuple in `hosted_openai_compat_hosts_are_templates_not_enum_variants`
the line exceeds the width, so `cargo fmt --all -- --check` wants it wrapped.
Six lines, entirely mechanical.

`maintainerCanModify` is true on #6171, but pushing the fix to
`BX166/Codewhale` returned 404, so per `AGENTS.md` ("when the push is refused,
land the resolved merge on `integration/<topic>-<pr>-<date>` and land from
there") it lands here instead. Their branch is untouched.

Their CI had also never run at all: as a first-time contributor, every run sat
at `action_required` from Sept 14 until approved today, and the approval has to
be repeated on each new commit. That is a repo setting, not anything they did.

## Verification

- `rustfmt --edition 2024 --check` clean on the merged
  `crates/config/src/provider_templates.rs`
- Their branch was updated from current `main` first, so this carries today's
  release work rather than a five-day-old base

Closes #6171


Hmbown:
Superseded by #6289. Closing the integration branch too — see the note on #6171.

PR #6321, opened 2026-09-18
feat(tui): record the mode each turn ran in on the turn

gaord:
## Summary

A turn's mode was only recoverable from the thread, which answers how the thread is set up *now*: switch mode while a turn is running and the completed turn is reported under a mode it never ran in. A client that needs "how did *this* run" — the GUI's plan-approval action is one — had to guess, or keep its own bookkeeping, which only covers turns it started itself.

`TurnRecord.mode` receipts the mode from the same policy projection that already produces `permission_posture` (`start_turn_with_source`), so the fact travels with the run: `GET /v1/threads/{id}`, the start-turn response, and the `turn.started` / `turn.completed` payloads all serialize the record directly. The manual compaction turn now computes that projection once instead of twice.

Records this runtime did not run — an imported conversation, a settlement for an unaccepted turn — carry none, as do pre-existing records. The field is additive (`#[serde(default, skip_serializing_if = "Option::is_none")]`, the convention `permission_posture` already follows), so no schema bump.

**What this does not do:** it is still "the mode the turn ran in", not "the turn produced a plan". Nothing in the runtime marks a message as a plan — Plan mode shares the one base prompt by design, and `update_plan` is a `model_visible() == false` replay tool — so a plan is ordinary prose. Distinguishing a plan from an answer needs a new signal on the turn, and this PR deliberately does not invent one.

No-Issue: no tracking issue — a small client-facing field, opened from the consumer side (a GUI client that currently guesses the mode) rather than from an upstream report.

## Testing

- [x] `cargo fmt --all -- --check` — run scoped (`cargo fmt -p codewhale-tui -- --check`), clean

**CI is this PR's only verification of compilation and tests.** The toolchain installed here is rustc 1.96.0 (2026-05-25), older than current stable, and `codewhale-tui` does not compile with it at all: 17 `#[expect(dead_code)]` sites do not fire on that version, so `-D warnings` rejects the crate (`unfulfilled_lint_expectations`) — in both the test and non-test cfg. None of those sites are touched by this PR, and the identical failure reproduces with this change stashed, so it is not caused here. `cargo fmt` does not depend on codegen or the lint, and passes. The test the lanes should exercise is `runtime_api::tests::turn_record_reports_the_mode_it_ran_in`, which starts a real turn with `mode: "plan"` and asserts the mode on both the start response and the stored thread detail.

## Checklist

- [x] This PR adds a new layer/module/abstraction — it names or deletes the layer it replaces — no new layer: one field on the existing turn record, populated from the projection that record already uses
- [x] Updated docs or comments as needed — `docs/RUNTIME_API.md` SDK contract table, plus the field's doc comment
- [x] Added or updated tests where relevant — new HTTP test, and 12 existing turn fixtures updated for the new field
- [ ] Verified TUI behavior manually if UI changes — no UI change: nothing in the TUI reads the field
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address — nothing harvested; authored here


Hmbown (approved):
Code review: clean, well-scoped slice. Additive `Option` field following the `permission_posture` convention (no schema bump), honest known-limitation note (mode-ran-in vs produced-a-plan), per-case `None` rationale comments, and a live-server test proving the record carries the mode. The compaction projection-dedup is a strict improvement.

The red checks are not this PR: changelog-parity and version-drift fail on main content (commits ff695506e0/95bc5c008f/6286136e6b, tui CHANGELOG slice), none from these 4 files. Needs a rebase once main heals, then merge on green.

PR #6336, opened 2026-09-18
fix(tui): /config theme editor no longer exits on an arrow key

Serendo:
The theme editor live-previews the highlighted choice on every arrow key, and each preview emits `ConfigUpdated { persist: false }`. The host applies that through `refresh_config_view_if_open`, which popped the modal and rebuilt it with `ConfigView::new_for_app` — dropping the transient `editing` state.

The first arrow key therefore closed the editor. The next one fell through to the non-editing key map, where Left/Right switch category, and the highlight snapped back to row 0. Through a real terminal this reads as "the panel closes as soon as I press a direction key", and the theme never actually moves.

`refresh_config_view_if_open` now rebuilds through `ConfigView::rebuild_preserving`, which refreshes the rows from the new snapshot (a persisted commit still has to show up) while carrying `editing` over, re-locating it by row key so a reordered or filtered snapshot cannot attach it to the wrong row. A vanished row drops the editor rather than resurrecting it, and the highlight is clamped into the refreshed choice list.

Enter opens the editor without emitting, so the browser unit test drives the refresh explicitly; the PTY test in `crates/tui/tests/cucumber/` checks the whole path against the shipped binary, since this only reproduces in the host glue. Both fail against v0.9.13 without the change.

Fixes #theme-nav-exit

## Testing

## Checklist

- [ ] This PR adds a new layer/module/abstraction — it names or deletes the layer it replaces


Hmbown (approved):
Verified locally: the new unit test passes (`config_refresh_keeps_the_open_theme_editor`, 1 passed), `cargo fmt --check` is clean after I pushed a rustfmt-only commit to the branch, and clippy reports nothing in the touched files (the crate-wide too-many-arguments errors are pre-existing). One non-blocking note: `rebuild_preserving` carries the previous `choices` list, so if a refresh ever changes a key's available choices, the editor shows the stale list until reopened — fine for the static theme list, worth a re-derive if this path ever serves dynamic choice sets. Thanks @Serendo — clean diagnosis, host-glue PTY coverage included.

Issue #6346, opened 2026-09-18
test(tui): run_tests_cwd_scopes_cargo_to_subdir fails on Windows verbatim path prefix


Hmbown:
## Why this matters

`tools::test_runner::tests::run_tests_cwd_scopes_cargo_to_subdir` asserts the rendered cargo command contains the project dir, but on Windows the two sides render different path forms: the command carries the `\\?\` verbatim prefix while the expected dir does not (or short/long-name variance). Observed failing on a PR run while main passed the same test twice — either a runner-image-dependent path normalization or an order-dependent fixture shape. Windows-only red that blocks PRs with zero relation to the diff.

## Current behavior

```
cargo must run in the scoped dir, ran: (cd \\?\C:\Users\runneradmin\AppData\Local\Temp\.tmpHjf9bL\project && cargo test)
```

- `crates/tui/src/tools/test_runner.rs:352` — the assertion
- `crates/tui/src/tools/test_runner.rs:83-90` — workdir via `context.resolve_existing_dir`
- `crates/tui/src/tools/test_runner.rs:155-160` — `format_command` renders `workdir.display()`

## Desired behavior

1. The assertion compares the same normalization on both sides (canonicalize-then-compare, or strip the verbatim prefix before comparing).
2. The test passes deterministically across Windows runner images.

## Repro or evidence

- Failing run: PR #6333 CI, `Test (windows-latest)`, which otherwise compiled and ran the suite.
- Main runs `c0906628` / `91f99e16` passed the same test — runner- or order-dependent.

## Acceptance criteria

- [ ] `cargo nextest run -p codewhale-tui -E 'test(run_tests_cwd_scopes_cargo_to_subdir)'` passes on a Windows runner regardless of tempdir short/verbatim form.
- [ ] No change to product behavior: test-only or normalization-only fix.

## Related

- Observed on: #6333.

PR #6347, opened 2026-09-18
fix(tui): double codewhale-main stack to 32 MiB for debug poll chains

Hmbown:
## Problem

Every PR's ubuntu test leg has failed since Sep 16: the `plugin_toml_binary_lifecycle_skill_and_stdio_mcp_acceptance` cucumber test hangs ~83 s and fails, showing a blank frame where the `trusted.` receipt should be.

The TUI process dies right after the plugin trust confirm: `thread 'codewhale-main' has overflowed its stack` (SIGABRT). The test's liveness check was fooled because the crashed process lingers as a zombie.

## Diagnosis (gdb at the guard page, Linux aarch64 repro)

47 frames, ~16.5 MiB consumed on the 16 MiB `codewhale-main` stack — **not recursion** (the chain unwinds cleanly to `thread_start`). The weight is debug-build poll frames of the giant top-level futures:

| frame | stack |
|---|---|
| `run_async_main_inner` | ~5.5 MiB |
| `run_tui` | ~3.2 MiB |
| poll `Box<handle_view_events>` | ~2.1 MiB |
| `execute_command_input` | ~1.2 MiB |
| `run_async_main` + dispatch + interactive | ~1.9 MiB |
| trust-confirm sync chain (`spawn_tui_engine` → `Engine::new` → `DeepSeekClient::new` → `resolve_runtime_route` → `Config::clone`) | ~0.5 MiB |

The trust-confirm engine respawn is simply the deepest command path; the margin was zero, so any small in-window addition tipped it over. Any future addition to these futures would re-tip a re-baselined 16 MiB.

## Fix

`CODEWHALE_MAIN_STACK_BYTES` 16 → 32 MiB. Cost is address space only (thread stacks commit lazily). Test-thread `RUST_MIN_STACK` deliberately stays 16 MiB: engine-only chains on test threads measured ~2.5 MiB, and the full UI-loop chain lives in spawned binaries that size their own stacks — the ci.yml comment now records that split.

## Verification

Linux repro container (`rust:latest`, aarch64, hermetic test home):
- before: `plugin_toml_binary_...` FAILED in 83.47 s (SIGABRT after trust confirm)
- after: **`test result: ok. 1 passed; 0 failed`** in 5.37 s

No-Issue: CI-leg regression fix; the previously-failing ubuntu leg on this PR is itself the acceptance record.


Hmbown:
Gate judgment record (per repo merge-under-gate rules):

- **Test (ubuntu-latest): PASS** — the acceptance record for this PR. The plugin_toml_binary acceptance that has failed every PR since Sep 16 passes with the 32 MiB stack.
- **Test (windows-latest): FAIL — pre-existing, not this PR.** The only failure is `tools::test_runner::tests::run_tests_cwd_scopes_cargo_to_subdir` (test_runner.rs:352), the known verbatim-path assertion tracked as #6346 and fixed in #6348. This PR touches `crates/tui/src/lib.rs` (a stack-size constant) and a ci.yml comment; it cannot affect a Windows-only path-string assertion.
- **Lint, Safety, Version drift, Integrations, npm wrapper smoke, Workflow lint: PASS.**

Intent: admin-merge once macOS reports, or earlier if the self-hosted mac runner stays queued — this unblocks every open PR's ubuntu leg.

Hmbown:
Landed via #6348's merge (77f758298) — the stacked branch carried this PR's commit 9f13b1f71. The full green gate arrived on #6348's run: ubuntu PASS (the plugin_toml_binary acceptance that had failed every PR since Sep 16), macOS PASS, windows PASS (via the #6346 fix). This PR's own ubuntu-leg acceptance record: https://github.com/Hmbown/CodeWhale/actions/runs/35401589363

Issue #76, opened 2026-04-26
Streaming reply silently halts mid-response — no SSE idle-timeout


Hmbown:
## Symptom

Long model replies sometimes stop mid-stream. The TUI shows the partial response and the cost line, then the active cell never closes. No error surfaces; the user sees \"Request cancelled\" only after they hit Esc, or just empty space.

Reproduced today during a long shadow-math research session under heavy concurrent load (parallel cargo builds saturating the host). The active cell drew partial output (tool calls, several paragraphs of model text), then went silent at \`\$0.15\` shown in the footer with no completion event.

## Root cause (likely)

\`crates/tui/src/client/chat.rs:181\`

\`\`\`rust
while let Some(chunk_result) = byte_stream.next().await {
    ...
}
\`\`\`

This awaits the next chunk indefinitely. If the upstream connection is silently stalled (TCP keep-alive succeeds but no bytes are being produced — common under provider/proxy hiccups, or when the local host is under heavy load and the connection's read side gets starved), \`stream.next()\` never resolves and never returns \`None\`. The agent loop is parked. Nothing prints, nothing errors.

## Reference: how Codex handles this

\`/Volumes/VIXinSSD/codex-main/codex-rs/codex-api/src/sse/responses.rs\` wraps the stream read in a tokio timeout:

\`\`\`rust
// codex-rs/codex-api/src/sse/responses.rs:430
let response = timeout(idle_timeout, stream.next()).await;
\`\`\`

…with \`idle_timeout: Duration\` carried on the \`Provider\` (\`codex-rs/codex-api/src/provider.rs:49\`: \`pub stream_idle_timeout: Duration\`). When the timer fires, the SSE driver returns a structured error that the agent loop turns into a recoverable failure (retry / surface to user).

## Proposed fix

1. Add a \`stream_idle_timeout\` field on the deepseek client (default \`30s\`, configurable via \`DEEPSEEK_STREAM_IDLE_TIMEOUT_SECS\` env / config).
2. Wrap the chunk read in \`crates/tui/src/client/chat.rs\` in \`tokio::time::timeout(idle, byte_stream.next())\`. On timeout: drop the connection, return an \`LlmError::StreamIdleTimeout { elapsed_ms }\`.
3. Same treatment in \`crates/tui/src/client/responses.rs\` for the Responses path.
[...]

Issue #34, opened 2026-04-25
`read_file` should auto-extract text from PDFs (and other binary docs)


Hmbown:
## Motivation

Recent friction: \`docs/DeepSeek_V4.pdf\` was right there in the repo and the agent couldn't read it without shelling out to \`pdftotext\`. The user reasonably expected \`read_file(\"docs/DeepSeek_V4.pdf\")\` to Just Work.

Same friction would apply to .docx, .epub, .html-as-rendered, and (eventually) image OCR.

## Proposal

In the existing \`read_file\` tool (find in \`crates/tui/src/tools/\`), detect file extension and content type:

| Extension | Behavior |
|---|---|
| \`.pdf\` | Shell out to \`pdftotext -layout - -\` (stdin/stdout). If unavailable, return structured \`{type: \"binary_unavailable\", reason: \"pdftotext not installed\", hint: \"brew install poppler\"}\`. |
| \`.docx\` | Use \`pandoc\` if available; otherwise binary_unavailable. |
| \`.html\` | Strip tags via the existing helpers (re-use logic from \`crates/tui/src/tools/web_search.rs:434\`). |
| Anything else binary | Return \`{type: \"binary\", size_bytes: N, sha256: \"...\"}\` instead of garbled UTF-8. |

For PDFs, optionally support a \`pages\` parameter (\`pages: \"1-5\"\` or \`pages: \"10\"\`) so the model can page through large docs without blowing context.

## Implementation

- Detect extension first; fall back to content-type sniffing (read first 4 bytes; \`%PDF\` etc.) for files without extensions.
- Subprocess timeout (10s default, configurable).
- Cap output at the existing read_file size limit; mark \`truncated: true\`.
- Add \`tool_uses[\"read_file\"][\"pdf_pages_extracted\"]\` to the metrics so we can see how often it's hit.

## Acceptance criteria

- \`read_file(\"docs/DeepSeek_V4.pdf\")\` returns plain text.
- Without \`pdftotext\` installed, returns a clear \"install poppler\" hint instead of crashing or returning binary garbage.
- A \`pages\` param works to read just a slice.

## Target

v0.5.1 — small win, removes a recurring \"can you find the paper\" / \"can you read the PDF\" round trip.

Issue #12, opened 2026-04-11
Roadmap: simplify DeepSeek API key setup and remove auth/provider drift


Hmbown:
## Summary
`deepseek-tui` should feel clearly **DeepSeek-only** and **API-key-first**. Right now the codebase and docs still carry broader credential/provider language that makes setup feel more complicated than it should.

This issue is about simplifying first-run setup around the one thing users actually need: a DeepSeek API key.

## Scope
- Make first-run onboarding clearly DeepSeek API key driven
- Simplify `login`, `logout`, `doctor`, and model discovery around the DeepSeek flow
- Reduce or remove user-facing provider/auth drift that suggests multiple auth systems when the intended product is just DeepSeek + API keys
- Tighten docs so new users can get from install to first prompt quickly
- Audit config/help text for leftover terminology that implies a broader auth system than we want

## Acceptance Criteria
- Fresh install can be configured quickly with a DeepSeek API key and minimal confusion
- User-facing CLI/docs do not imply extra auth flows unless they are intentionally supported
- Setup/help text consistently presents the product as DeepSeek-first
- Relevant config and onboarding paths are tested

## Starting Points
- `crates/tui/src/main.rs`
- `crates/cli/src/main.rs`
- `crates/config/src/lib.rs`
- `README.md`
- `docs/CONFIGURATION.md`

Hmbown:
Closing as stale — these April 11 roadmap items predate the v0.4.x / v0.5.0 work and are being replaced by concrete, actionable issues (#28-#35). Will reopen specific items if/when they're picked up.

PR #6361, opened 2026-09-19
feat(runtime-api): terminal byte stream (#34), stream resume + idempotent submit (#76), pet agent-count pin (#12)

Hmbown:
## Summary

Two Core unblocks from the tracker (`codewhale-app#61`), on one branch.

### #34 — the Engine's terminal byte stream

The app-side ticket asks Core for "authenticated byte input/output, resize, exit and bounded replay" over the stateful terminal path, and says to link real Core work before claiming those paths. Two commits:

1. **The owner gained the primitives** (`crates/tui/src/tools/terminal_session.rs`). The pty master is now retained (it was dropped after the reader/writer clones were taken), so `resize_session` can reach the kernel's window; `OutputBuffer::read_since` reads from an **absolute cursor without consuming** and sets `gap` when the 512 KiB ring has moved past it (report, not repair); `session_exit_status` and `kill_session` make exit observable. `take_output` now reuses that same cursor arithmetic instead of its own copy.
2. **The routes that consume them** (`crates/tui/src/runtime_api/terminal.rs`), which is also what makes them live code — `cargo clippy --all-targets` failed the lib build on the primitives alone, correctly, and this PR is the consumer.

- `GET /v1/terminal/{name}/output` — wire-shaped like the jobs stream on purpose (`cursor`/`max_bytes`/`format` → `offset`/`next_cursor`/`total`/`dropped`). Reads never consume, so several clients can hold independent cursors.
- `POST /v1/terminal/{name}/input` — `text` or `base64`, bounded per frame. Input stays attributable by route: this is the client's writer, `terminal_send` is the agent's.
- `POST /v1/terminal/{name}/resize` and `POST /v1/terminal/{name}/kill`.
- Routes **attach to shells the Engine already owns and never create one** (404 otherwise) — an HTTP request must not conjure a shell the Engine does not know about.
- `RuntimeCapabilities` gains `terminal_stream`, `terminal_input`, `terminal_resize`, `terminal_kill`, set from `cfg!(unix)`, so a client gates its pane on the flag rather than discovering the gap from a failed request. Windows answers `501` for all four (the owner is Unix-only), which keeps "this build cannot do terminals" distinguishable from "that session is gone".
- `docs/RUNTIME_API.md` documents the family in the GPUI section with its four stated limitations: no `wait_ms` long poll, no scrollback recovery, live sessions only (a restarted Engine reports no session rather than pretending to reattach), no runtime-sdk wrapper yet.

### #12 — the pet's agent count

The ticket's premise ("no producer populates `Activity.parallel`") turned out to be stale: Core already emits it end to end (engine → `event_loop.rs:1946` → `metadata()`'s allowlist → `pet-native.js:1796` `agent:${id}` spans → the `· ×N` caption), and the JS half is already covered by `pet/tests/pet-engine.test.mjs`. So there is no producer to write — a second one would be a second authority for a count that already flows. What was missing is any test of the Rust half, which is the half that fails *silently*. This pins it, and the pin is proven to work: deleting `Event::AgentSpawned` from the allowlist fails it with `agent spawns are observed`. The app-side rendering half stays in codewhale-app.

### #76 — resumable streaming and idempotent submission

Two commits, both Core-side.

**Resume from `Last-Event-ID`.** The durable half already existed (per-thread `seq`, a JSONL event log, `since_seq` replay with a bounded tail), but nothing on the wire could use it: every SSE frame was written without an `id:`, so a browser `EventSource` had nothing to resume from. Journal frames now carry their durable `seq` as the SSE id — journal frames only, since the `stream.progress` frames are transport progress and an id there would invite a client to resume from a point it never received. `stream_thread_events` reads `Last-Event-ID` as the cursor when no explicit `since_seq` was asked for; an explicit query cursor wins, so a deliberate replay-from-zero is never silently overridden by a stale header. An opaque id (proxy, older client) starts from the durable head instead of failing to open the stream. `RuntimeCapabilities` gains `event_stream_resume` so a client can gate its reconnect controls on it.

**A replay is an acknowledgement, not an admission.** `POST /v1/threads/{id}/turns` answered `201` whether it admitted a new turn or handed back the original for a repeated `operation_key` — so a client retrying after a dropped response could not tell whether it had created a second turn. The admission path now reports the disposition (`start_turn_with_reporting_replay`; the two replay returns say `true`, the tail says `false`), and the route answers `200 { ..., idempotent_replay: true }` for a replay, keeping `201` for a fresh admission. The flag is omitted on admission, so existing responses stay byte-identical, and the 94 existing `start_turn` callers are untouched.

The existing `turn_endpoint_operation_key_returns_original_and_conflicts_on_mismatch` test asserted the old `201`; it now pins the stronger contract, including `409` on a changed request with the same key, exactly one `SendMessage`, and exactly one turn.

**Not in this PR:** app-server minting an `operation_key` for its own submissions (client half, app lane), and the Windows ConPTY port for the terminal family.

Additional verification for these two commits:

- `... tests.rs thread_event_frames_carry` → `1 test run: 1 passed` (first frame's `id:` equals its payload `seq`; reconnect with only `Last-Event-ID` lands on the next durable event; `?since_seq=0` with a stale header still replays from zero) · `last_event_id_accepts` → `1 passed` · `... tests.rs event` → `170 passed`
- `... tests.rs turn_endpoint_operation_key` → `1 test run: 1 passed`, plus the paths the signature change touches: `turn_operation` 5 passed · `start_turn` 5 passed · `agent_mail` 6 passed · `thread_goal` 5 passed · `steer` 33 passed
- clippy `--all-targets -D warnings` exit 0 · fmt exit 0 · both ratchets PASS

### Unreding main (same branch, no extra PR)

`main` was red on two jobs, and neither is visible in a PR rollup — `check-runtime-contract-budget` is advisory on pull requests and fatal on push, and `Test (macos-latest)` is not one of the three required checks.

**`Lint` — an unrecorded identity change.** `execute_tools` now enters every catalog outside Plan (`tool_catalog.rs:350`, code-mode Phase 1 `e23ce514c`), moving the Act and Operate full tool names and identity digests; and the `agent` tool advertises its `cwd` parameter, growing the shared active surface by 256 schema bytes / 64 estimated tokens everywhere that tool appears (including Plan full, which is why it grew without gaining a tool). The checker refuses identity drift and `--update` cannot bypass it, so this is the explicit maintainer edit the file's header asks for: identities re-pinned, 14 ceilings raised to the measured values, and both facts recorded in the file's `_comment` history. `check-runtime-contract-budget.py` → `PASS: all 55 metrics are exactly at budget.`

**`Test (macos-latest)` — a real race, proven fixed.** `threads_running_lists_active_turns_and_clears_on_settle` forced a synthetic settle into the durable store, but the engine still owns that record and can persist afterwards; the listing is read from the store, so a later engine write puts the turn back in flight and the single read fails. The assertion now polls until the settle wins (with a `ci_scaled(5s)` deadline, so a genuinely stuck turn still fails).

Proven rather than argued: with the record flipped back to `InProgress` after the test's write and settled 300ms later, the old single read fails with the CI shape verbatim (`left: [... "status": "in_progress"]`, `right: Array []`) while the polling version passes the identical injection (0.549s — it waited for the settle). The injection was reverted; the commit holds only the fix.

## Testing

- `scripts/dev-test.sh crates/tui/src/runtime_api/tests.rs terminal_routes_serve` → `Summary [0.208s] 1 test run: 1 passed` — **the full seam over HTTP against a real PTY**: input through the route executes in the Engine's own session, output comes back through the route at an absolute cursor, a repeat read at the same cursor returns identical bytes, resize is confirmed by the shell's own `stty size` reporting `40 100` (a handler that merely stored the numbers would fail this), and kill is observed as `running: false`.
- `... terminal_output_for_an_unknown` → `1 test run: 1 passed`; `... terminal_capabilities` → `1 test run: 1 passed`.
- `... crates/tui/src/tools/terminal_session.rs terminal` → `360 tests run: 360 passed, 12543 skipped`.
- `... crates/protocol/src/runtime/mod.rs runtime::` → `10 tests run: 10 passed`.
- `cargo clippy -p codewhale-tui -p codewhale-protocol --all-targets --locked -- -D warnings -A clippy::uninlined_format_args -A clippy::too_many_arguments -A clippy::unnecessary_map_or` exit 0 · `cargo fmt --all -- --check` exit 0
- `check-blocking-calls-budget.py` → `601 sites across 177 files, within budget` · `check-dead-code-budget.py` → `PASS: 185 attributes, exactly at budget.`

**Not verified: Windows.** The owner is `#[cfg(unix)]` end to end and every test here is `cfg(all(test, unix))`; ConPTY qualification is its own slice, as the ticket says.

## Checklist

- [x] No new layer replacing an existing one: this is the existing terminal owner plus routes over it
- [x] Docs updated (`docs/RUNTIME_API.md`, and the limitations are stated in the module)
- [x] Tests added where they protect behaviour that fails silently
- [ ] TUI manual check: n/a (no TUI surface changed)

No-Issue: Core half of tracker codewhale-app#61; #34 and #12 live in the app repo, so there is no issue in this repo to close.


Hmbown:
This branch is being carried into the 0.10.0 readiness PR [#6370](https://github.com/Hmbown/Codewhale/pull/6370) as a merge of `work/0.9.14-core-unblocks` at `041e628` (all ten commits, authorship intact), rather than re-implemented.

The three conflicts against current main were all generated or measured files and resolved to main's newer versions: `docs/public-surface-facts.json` and `web/lib/facts.generated.ts` (the 0.10.0 capture facts) and `scripts/runtime-contract-budget.json` (main's 2026-09-19 Linux lock-in already carries the agent `cwd` / `execute_tools` ceilings this branch measured; every metric value is identical, only the `_comment` history differed).

On the merged tree (rustc 1.98.1, Linux): `terminal_routes_serve_a_live_engine_session_over_http`, `terminal_output_for_an_unknown_session…`, `runtime_info_advertises_terminal_capabilities`, `thread_event_frames_carry_the_id…`, `last_event_id_accepts_only_decimal_cursors`, `turn_endpoint_operation_key_returns_original…`, `threads_running_lists_active_turns_and_clears_on_settle`, `a_parked_approval_announces…`, `agent_events_forward…` each 1 passed; `sleep_guard` 2 passed; `terminal_session` 15 passed; `runtime_api::tests::` 249 passed; `turn_loop` 67 passed. If #6370 lands by squash instead of merge, this PR will need closing by hand with that credit.

---

PR #6388, opened 2026-09-21
fix(tasks): refuse a posture the thread would reject at task admission

Hmbown:
## Summary

Follow-up to #6386 (Devin finding on the admission path). `TaskManager::add_task_with_id` stored `permission_posture` verbatim and left its validation to the worker, which runs `RuntimePolicyProjection::from_request` only when it opens the task's thread. An unknown spelling, or `never`, therefore passed `POST /v1/tasks` with a 2xx and a queued `TaskRecord`, sat in the durable queue, and failed only once a worker claimed it.

Admission now runs the same projection, with the mode and auto-approve bits the record will carry (`req.mode` or the manager's default mode, `req.auto_approve`), so the refusal happens at the boundary the request crossed and nothing is queued. There is no second parser: the worker's own projection is the validator, so admission and execution cannot disagree about what is accepted. Accepted values are stored as before (the pin test from #6386 is unchanged); only the refusal moved earlier.

No-Issue: closes the review finding left on #6386, thread linked there.

## Testing

Test build: `cargo rustc -p codewhale-tui --lib --profile test --all-features --locked` (rustc 1.98.1), run under `scripts/with-hermetic-test-home.sh`.

- `task_manager::tests::add_task_refuses_a_posture_the_thread_would_reject` → `test result: ok. 1 passed; 0 failed` ("sideways" and "never" both refused, queue stays empty).
- `task_manager::tests::` → `test result: ok. 57 passed; 0 failed`.
- `runtime_policy::` → `test result: ok. 4 passed; 0 failed`.
- Proof the test judges the fix: with the admission call forced off (`if false { … }` around it, rebuilt), the same test → `test result: FAILED. 0 passed; 1 failed`: the "sideways" task was queued as a `TaskRecord` instead of being refused. The fix was restored afterwards; the committed file is the fixed one.
- `cargo fmt --all -- --check`: clean.
- Non-test `cargo check -p codewhale-tui --lib --locked`: exit 0, 0 warnings.

- [x] Targeted tests and proof run locally (above)
- [ ] Workflow lint, Lint, Test ubuntu / macos / windows — CI on this head

## Checklist

- [x] This PR adds a new layer/module/abstraction — it names or deletes the layer it replaces (no new layer; reuses `RuntimePolicyProjection::from_request`)
- [x] Verified TUI behavior manually if UI changes — no UI change
- [x] Harvested/co-authored credit uses a GitHub numeric noreply address (n/a)

https://claude.ai/code/session_0134iUMxmGuXiG1LzPgfZVnv

---
_Generated by [Claude Code](https://claude.ai/code/session_0134iUMxmGuXiG1LzPgfZVnv)_


Hmbown:
Thanks @claude — both findings verified against the branch head (`b18747302`) rather than taken on trust. Both are accurate.

**1. Mode coverage gap — confirmed, and it is coverage only, not a defect.**

The guard does validate mode. `task_manager.rs:1693` passes the resolved mode as `from_request`'s first argument:

```rust
RuntimePolicyProjection::from_request(
    req.mode.as_deref().filter(|m| !m.trim().is_empty()).unwrap_or(&self.cfg.default_mode),
    req.permission_posture.as_deref(),
    req.auto_approve,
)?;
```

So an unknown mode spelling *is* refused at admission today; the new test simply does not exercise that half. Rather than respin this PR's CI for a test-only addition — and push another ~30 minute Windows/macOS lane in front of a release candidate — I will land the mode case in the issue-sweep PR that is already queued for CI. Flagging that explicitly so it is a commitment with a place to land, not a deferral.

**2. Two default-mode literals and the blank-mode quirk — confirmed, pre-existing, not introduced here.**

Verified both halves:

- Admission falls back to `self.cfg.default_mode` (`task_manager.rs:1696`); `create_thread` falls back to the hardcoded `"agent"` (`runtime_threads.rs:7331`). They agree today only because every production `TaskManagerConfig` sets `default_mode: "agent"`.
- The blank case is real: admission filters `Some("   ")` to the default before validating, but `task.mode` is persisted from `req.mode.unwrap_or_else(...)` at `task_manager.rs:1724` **without** that filter, so the record keeps `"   "` verbatim. It is masked because `create_thread` re-filters blank modes to the same default.

You are right that this means the "admission and execution cannot disagree" invariant this PR establishes for posture does not yet fully hold for blank mode. That is a pre-existing seam and out of scope for a guard-clause addition, but it is the kind of thing that stops being harmless the moment a non-`"agent"` `default_mode` ships. Tracking it rather than silently widening this PR.

No changes to this PR; merging on the verified behavior.

Hmbown:
@codewhale-agent — your four evidence gaps, settled against `runtime_policy.rs` and `task_manager.rs` at head `b18747302`. You were right to raise them as gaps rather than findings; none is a defect, and one is a real (pre-existing) seam.

**(1) Value identity between admission and execution.** Holds for posture and auto-approve: the record stores `permission_posture: req.permission_posture` and `auto_approve: req.auto_approve.unwrap_or(false)` — exactly what was validated. Mode has one seam: admission filters blank before defaulting, but `task_manager.rs:1724` stores `req.mode.unwrap_or_else(...)` *without* that filter, so `Some("   ")` is validated as the default yet persisted verbatim. It is masked because `create_thread` re-filters blank modes to the same default. Pre-existing, not introduced here, and already noted on @claude's thread.

Your sharper sub-point — a user-configurable `default_mode` that the projection rejects would refuse every mode-less task at admission — is correct, but that is the desirable direction: it now fails at the boundary instead of after the task has sat in the durable queue and a worker has claimed it. That is the whole point of the PR.

**(2) Untrimmed mode forwarded to the projection.** Not a disagreement risk. Both sites use the identical shape — filter on `!m.trim().is_empty()`, then forward the *raw* string. Admission: `task_manager.rs:1693`. Execution: `runtime_threads.rs:7329`. Whatever `parse_runtime_mode` does with padding, it does the same thing in both places, so the two cannot diverge.

**(3) Internal callers passing a persisted or legacy posture.** No such caller. `add_task_with_id` is the only path that stages a *new* record (`admit_task_record(task, false)`). Crash recovery replays an already-staged record via `recover_task_admission` → `admit_task_record(staged, true)`, which does not re-run the projection — so an older persisted spelling is not retroactively refused. Existing queued work is unaffected.

**(4) Purity, cost, and error-text stability.** `from_request` (`runtime_policy.rs:62-86`) is pure: two parses, a match, returns a `Copy` struct. No I/O, no allocation beyond the error paths. Both rejection paths contain the literal the test asserts on:

- `"unsupported permission posture {value:?}; expected ask, auto-review, or full-access"` — covers `"sideways"`
- `"permission posture 'never' is not part of the Runtime product contract"` — covers `"never"`

So `contains("permission posture")` holds for both inputs the test uses.

Merging on this. Thanks for scoping these as gaps rather than asserting them — that framing made them fast to close.

